PR #2, opened 2021-03-16
add test so i can see if coverage works

J-Thompson12:
Apparently with the right coveralls permission you can just add the check in there. So I have removed my custom check.


josecordaz (approved):
LGTM

PR #12, opened 2021-03-22
Custom coverage check to show failure in actions history

J-Thompson12:
@jrryjcksn @palemtnrider I am leaving it up to you guys if this is something that we want. Circle ci was worked the same way as github actions where the coveralls check didnt show up in the history. It only shows up on the PR. If we want this custom step then it will show in the history if it failed.


J-Thompson12:
not a clean branch

PR #16, opened 2021-03-24
[WKP-1915] Create "wego version" command and wego binary

jrryjcksn:
- Sets up make
- Creates `wego` binary
- Adds `wego` version command that checks if a newer version is available and informs the user if o


josecordaz (approved):
LGTM

PR #18, opened 2021-03-24
basic version of docker and s3

J-Thompson12:
I am using my own aws keys because i couldnt get the ones in 1password to work


josecordaz (approved):
LGTM

PR #20, opened 2021-03-24
fix docker push

J-Thompson12:


.github/workflows/deploy.yml
```diff
@@ -24,12 +26,16 @@ jobs:
...
-        uses: actions/checkout@v2
-      - name: Push to Docker Hub
+      - name: Login to GitHub Packages Docker Registry
+        uses: docker/login-action@v1
+        with:
+          username: ${{ secrets.DOCKER_HUB_USERNAME }}
+          password: ${{ secrets.DOCKER_HUB_PASSWORD }}
+      - name: Checkout
+        uses: actions/checkout@master
+      - name: Build and push
```
josecordaz, line 36: I think this step is missing right before the Build and Push action
```
      - name: Set up Docker Buildx
        uses: docker/setup-buildx-action@v1
```

josecordaz (approved):

PR #24, opened 2021-03-26
add bin check

J-Thompson12:
https://weaveworks.atlassian.net/browse/WKP-1916


cmd/wego/main.go
```diff
@@ -33,6 +35,15 @@ func configureLogger() {
 func main() {
 	rootCmd.PersistentFlags().BoolVarP(&options.verbose, "verbose", "v", false, "Enable verbose output")
 	rootCmd.AddCommand(version.Cmd)
+
+	if checkResponse, err := checkpoint.Check(&checkpoint.CheckParams{
```
josecordaz, line 39: @palemtnrider @jrryjcksn  Would it be a good idea to use github api to check latest releases instead? In order to `go-checkpoint` to work we need to manually add a record in the DB throught a web page which is easily forgotten and easier to make a mistake in any of the forms fields. Mark and I looked at the web page a few days ago and the DB was outdated, the last release for wksctl  was `0.1.0-alpha.1`. Not sure if the info generated by this is still helpfull. I think if we can have less manuall steps like this it would be better. Or if we can create that record automatically after making a release, it would be cool too. 🙂

J-Thompson12: Makes sense i can look into it

J-Thompson12: So turns out i never looked too much at version.go That file already does the check by calling the releases endpoint.

cmd/wego/version/cmd.go
```diff
@@ -2,42 +2,28 @@ package version
...
 	"fmt"
-	"os"
 
+	log "github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
+	"github.com/weaveworks/go-checkpoint"
 )
 
+// The current wego version
+var Version = "v0.0.0"
```
josecordaz, line 12: I would rather prefer to have this dynamic. Do can do something like this https://github.com/weaveworks/wksctl/pull/331/files#diff-9099d16d1f5039d08a9c52239830f5355aa0ad958d4f8f18bd5457581acbf546R354

josecordaz: @J-Thompson12  I just saw that you had this already. What did we change it ?

J-Thompson12: So we had a conversation at standup yesterday and we need to use checkpoint because it does metrics as well as the version check

josecordaz (approved):

PR #30, opened 2021-03-31
Added darwin binary build

josecordaz:


.github/workflows/deploy.yml
```diff
@@ -7,17 +7,21 @@ on:
 name: Deploy
 jobs:
   publish-s3-artifact:
-    runs-on: ubuntu-latest
+    runs-on: ${{matrix.os}}
+    strategy:
+      matrix:
+        os: [ubuntu-latest, macOS-latest]
```
jrryjcksn, line 13: Just wondering whether this will take longer than doing both makes in one go by setting GOOS and GOARCH? :thinking:

jrryjcksn (approved):
LGTM

.github/workflows/deploy.yml
josecordaz: good question, I'll test that out 🙂

PR #41, opened 2021-04-02
add flux binary to wego

J-Thompson12:
I looked at a few different options but from what I saw this was the best way to do it. If there are any other ideas out there i would be glad to hear it


cmd/wego/flux/cmd.go
```diff
@@ -0,0 +1,34 @@
...
+	_ "embed"
+	"fmt"
+	"os"
+	"os/exec"
+
+	"github.com/spf13/cobra"
+)
+
+//go:embed flux
+var flux []byte
```
luizbafilho, line 13: omg, so simple. That's awesome.

One thing tho, since the dependencies script downloads them to `tools/bin`, do you mind changing to pointing to there?

take a look at https://github.com/weaveworks/weave-gitops/pull/43 it should be merged shortly.

cmd/wego/flux/cmd.go
```diff
@@ -0,0 +1,34 @@
...
+
+var Cmd = &cobra.Command{
+	Use:   "flux",
+	Short: "Use flux commands",
+	Run:   runCmd,
+}
+
+func init() {
+	_ = os.WriteFile("bin/flux", flux, 0755)
```
luizbafilho, line 22: I think we should use the virtual file system we use in `wks` so we don't have to save anything into the user's file system since we can call it from anywhere.

cmd/wego/flux/cmd.go
```diff
@@ -0,0 +1,34 @@
...
+}
+
+func runCmd(cmd *cobra.Command, args []string) {
+	c := exec.Command("./bin/flux", args...)
+
+	// run command
+	if output, err := c.CombinedOutput(); err != nil {
+		fmt.Println("Error:", err)
+	} else {
+		fmt.Printf("Output: %s\n", output)
```
luizbafilho, line 32: after printing the error, we should also exit with an error. `os.Exit(1)`

luizbafilho (changes requested):

cmd/wego/flux/cmd.go
jrryjcksn, line 22: Not sure where this is supposed to write the binary... We don't have a bin directory at install time (that I"m aware of). Maybe this should go in a temp directory?

jrryjcksn (changes requested):
I think we need to put the flux binary in a different location at run time.

cmd/wego/flux/cmd.go
luizbafilho: can't we run it from memory?

cmd/wego/flux/cmd.go
J-Thompson12: Okay so thats going to be an interesting thing to do. With go embed i am not allowed to use a file whose path has ../ I am unable to go outside the package. So I can move all of this to the root but that may not be what we want.

cmd/wego/flux/cmd.go
```diff
@@ -0,0 +1,63 @@
...
+	Short: "Use flux commands",
+	Run:   runCmd,
+}
+var path = "temp/bin"
+var exePath string
+
+func init() {
+	var dep dependencies
+	if _, err := toml.DecodeFile("tools/dependencies.toml", &dep); err != nil {
+		fmt.Fprintf(os.Stderr, "Error: %v\n", err)
```
jrryjcksn, line 34: This is really a nit, but we have exactly the same error handling in a bunch of places. Since we're exiting when we get an error, we could wrap that in a function. Something like:

```
func checkError(err error) {
   if err != nil {
      fmt.Fprintf(os.Stderr, "Error: %v\n", err)
      os.Exit(1)
   }
}

and then:

checkError(toml.DecodeFile("tools/dependencies.toml", &dep))
exePath = path + "/flux-" + dep.Flux.Version

if _, err := os.Stat(path); os.IsNotExist(err) {
   checkError(os.MkdirAll(path, 0755))
   checkError(os.WriteFile(exePath, fluxExe, 0755))
}
```

cmd/wego/flux/cmd.go
```diff
@@ -0,0 +1,63 @@
...
+	Use:   "flux",
+	Short: "Use flux commands",
+	Run:   runCmd,
+}
+var path = "temp/bin"
+var exePath string
+
+func init() {
+	var dep dependencies
+	if _, err := toml.DecodeFile("tools/dependencies.toml", &dep); err != nil {
```
luizbafilho, line 33: This won't work. This runs every time you run your binary, not at build time, which means if you run the `wego` binary outside the wego repo it breaks. 

```
weave-gitops git:(flux-bin) cd
➜  ~ /Users/luiz/Projects/weaveworks/weave-gitops/bin/wego flux 
Error: open tools/dependencies.toml: no such file or directory
```

you should use build flags instead, you can take a look at the version command. You can then use the `stoml` tool to read the version from the deps file, look at how to use that in the `tools/download-deps.sh` file.

Regarding the path you save the file, we should save it under `$HOME/.wego/bin` and use it from there, if you keep this way saving in `temp/bin` it will save the file in the user's current folder wherever they are running the command.

I got a bug, the way you passed the rest of the arguments to the flux bin don't work.

```
weave-gitops git:(flux-bin) bin/wego flux install -h
Use flux commands

Usage:
  wego flux [flags]

Flags:
  -h, --help   help for flux

Global Flags:
  -v, --verbose   Enable verbose output
```
when running that, we should see flux output, something like:

```
install -h
The install command deploys Flux in the specified namespace.
If a previous version is installed, then an in-place upgrade will be performed.

Usage:
  flux install [flags]

Examples:
  # Install the latest version in the flux-system namespace
  flux install --version=latest --namespace=flux-system
....
```

It looks like cobra takes control over everything and thinks those arguments are to it. I'm not sure how to solve that one, you'll have to google that one, let me know if you get stuck with that, I can help you.

luizbafilho (changes requested):

cmd/wego/flux/cmd.go
J-Thompson12: So i gotta ask.  At what point does embedding this not make sense anymore. I mean if it's listed as a dependency and getting installed can we not just use the installed flux bin instead of the embedded one.

luizbafilho: Those dependencies are build dependencies. They are gonna be installed only on our machines and in the ci. and that installing process is done via make file, so most of those are only there to assist us during development. That's why we embed flux and copy to user's file system, otherwise, we would need to ask them to manually install flux before using wego.

jrryjcksn: The only down side I see to that is that the main reason for being for wego is to make the process of using flux2 simple and intuitive. Perhaps the wego binary could do a download and install if it doesn't find flux locally?

josecordaz: maybe use `toml.Decode` instead of `toml.DecodeFile` so you can avoid file path location issue 🙂

josecordaz: I was able to avoid those issues. Using the change above and running the command like this `./weave-gitops/bin/wego flux -- install -h` . Not 100% we want that. 

image

J-Thompson12: I mean its not great but its better than what i had.

luizbafilho: since this is only for debugging, I'm good with that. Great discovery.

luizbafilho (approved):
It looks great.

jrryjcksn (approved):
LGTM

PR #43, opened 2021-04-05
Make target to download binaries dependencies

luizbafilho:
I mostly brought the scripts from the wks repo.

### DIFF
--- a/tools/download-deps.sh
+++ b/tools/download-deps.sh
@@ -44,12 +44,12 @@ instantiate_url() {
 # and allow overriding with a local tool by setting the environment variable "$LOCAL_<tool>" (e.g. $LOCAL_wk)
 # (useful for testing and particularly useful on darwin since we don't publish a "wk" version for darwin)
 download_dependency() {
-  local tool="${1}"
-  local bin_dir="${2}"
-  local dependencies_toml="${DEP_FILE}"
-  local localToolVar="\$LOCAL_"${tool}
-  local localTool
-  localTool=$(eval "echo ${localToolVar}")
+    local tool="${1}"
+    local bin_dir="${2}"
+    local dependencies_toml="${DEP_FILE}"
+    local localToolVar="\$LOCAL_"${tool}
+    local localTool
+    localTool=$(eval "echo ${localToolVar}")
 
     if [ -n "${localTool}" ]; then
         cp "${localTool}" "${bin_dir}"
@@ -71,6 +71,8 @@ download_dependency() {
     special_tarpath_url=(${special_tarpath//;/ }) # split out special paths which contain <url>;<path in tarball>
     local tarpath
     tarpath=$(instantiate_url "$(run_stoml tarpath)")
+    local custom_bindir
+    custom_bindir=$(run_stoml bindir)
     echo $tarpath
     if check_url "${binarypath}"; then
         url_and_path="${binarypath}"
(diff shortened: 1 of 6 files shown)


tools/download-deps.sh
```diff
@@ -0,0 +1,97 @@
...
+    run_stoml() {
+        local property="${1}"
+        "${bin_dir}"/stoml "${dependencies_toml}" "${tool}"."${property}"
+    }
+
+    local url_and_path
+    local fetch
+    local binarypath
+	local special_tarpath
+	local special_tarpath_url
```
palemtnrider, line 68: Nit formatting

tools/download-deps.sh
```diff
@@ -0,0 +1,97 @@
...
+    binarypath=$(instantiate_url "$(run_stoml binarypath)")
+	special_tarpath=$(instantiate_url "$(run_stoml special_tarpath)")
+	special_tarpath_url=(${special_tarpath//;/ }) # split out special paths which contain <url>;<path in tarball>
+    local tarpath
+    tarpath=$(instantiate_url "$(run_stoml tarpath)")
+    echo $tarpath
```
palemtnrider, line 74: Would it be helpful to print the other paths/urls?  Maybe print them inside the if statements below?

tools/download-deps.sh
```diff
@@ -0,0 +1,97 @@
...
+        url_and_path="${binarypath}"
+        fetch=do_curl_binary
+    elif check_url "${special_tarpath_url}"; then
+        url_and_path="${special_tarpath}"
+        fetch=do_curl_tarball_with_path
+    elif check_url "${tarpath}"; then
+        url_and_path="${tarpath}"
+        fetch=do_curl_tarball
+    else
+        echo "No valid path for tool:" "${tool}"
```
palemtnrider, line 85: Seems like we should exit here instead of trying to continue?

tools/download-deps.sh
```diff
@@ -0,0 +1,97 @@
...
+    fi
+    "${fetch}" "${tool}" "${url_and_path}" "${bin_dir}"
+}
+
+# Don't use $RELEASE_GOOS here, should be whatever is running the script.
+STOML_URL="https://github.com/freshautomations/stoml/releases/download/v0.4.0/stoml_$(goos)_amd64"
+do_curl_binary "stoml" "${STOML_URL}" "${BIN_DIR}"
+tools=(flux)
```
palemtnrider, line 93: Is this temporary for testing just the download of flux? I was thinking the list of tools to download would come directly from the dependencies.toml file. i.e., everything listed in there is downloaded.

tools/functions.sh
```diff
@@ -0,0 +1,124 @@
...
+        echo "386"
+        ;;
+    *)
+        error "uknown arch: ${arch}"
+        ;;
+    esac
+}
+
+mktempdir() {
+    mktemp -d 2>/dev/null || mktemp -d -t 'wk-quickstart'
```
palemtnrider, line 78: Please replace wk-quickstart with something related to wego

tools/functions.sh
```diff
@@ -0,0 +1,124 @@
...
+    local url="${2}"
+
+    log "Downloading ${url}"
+    curl --progress-bar -fLo "${path}" "${url}"
+}
+
+do_curl_binary() {
+    local cmd="${1}"
+    local url="${2}"
+    local default_path="${HOME}/.wks/bin"
```
palemtnrider, line 92: replace .wks with wego (and in the other places below)

palemtnrider (changes requested):
Great to have the binaries downloaded dynamically.  I made a few minor comments.  Looks like the functions file came from WKP - I'd suggest removing the functions we don't need.

J-Thompson12 (approved):
LGTM

PR #45, opened 2021-04-05
add cluster status

jrryjcksn:


pkg/status/status.go
```diff
@@ -0,0 +1,43 @@
...
+
+	if lookupHandler("deployment coredns -n kube-system") == nil {
+		return Unmodified
+	}
+
+	return Unknown
+}
+
+func kubectlHandler(args string) error {
+	return exec.Command(fmt.Sprintf("kubectl get %s", args)).Run()
```
luizbafilho, line 42: are we relying on `kubectl` to be available for our calls? 

Not just this one, but as a general practice, how are we planning on interacting with the kube API?

jrryjcksn: That's a good point. We've always installed `kubectl` in the past (there are a lot of "resources" that use it). Honestly, though it seems "cleaner" to use the go client, I'm not sure it's worth the extra credential manipulation and setup when we're doing simple things like this. I would vote to ensure kubectl is installed and use it when it makes sense (maybe all the time? -- depends on what we need to do later).

josecordaz: where would `kubectl` command be called from ? inside one of the nodes? or from the customer machine ?

jrryjcksn: In this case, we're talking about using `kubectl` from the wego binary on the customer machine. Normally, we would use the go client for anything in a controller within the cluster.

luizbafilho: It feels that this can/should be useful within the controller. And having a single way of talking to kube API is probably ideal. These days with `client-go` talking directly to kube API programmatically is not that hard because it uses the same mechanism as `kubectl`. Here are couple of examples:

out of cluster: https://github.com/kubernetes/client-go/blob/master/examples/out-of-cluster-client-configuration/main.go
in cluster: https://github.com/kubernetes/client-go/blob/master/examples/in-cluster-client-configuration/main.go

PR #48, opened 2021-04-08
fix push s3

J-Thompson12:
fixes https://app.zenhub.com/workspaces/weaveworks-foss-603cb6b1d4dff9000e242dc4/issues/weaveworks/weave-gitops/47

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -37,9 +37,7 @@ fmt:
 	go fmt ./...
 # Run go vet against code
 vet:
-	echo vet ran first	------------------------------------------------------
 	go vet ./...
 
 dependencies:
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -33,7 +37,9 @@ fmt:
 	go fmt ./...
 # Run go vet against code
 vet:
+	echo vet ran first	------------------------------------------------------
```
jrryjcksn, line 40: These look like debugging statements... are they left over?

jrryjcksn (approved):
Curious about the "ran first" output but otherwise LGTM

Makefile
J-Thompson12: Yeah whoops

PR #52, opened 2021-04-13
fix coverage test

J-Thompson12:
I removed the delete fake flux because that gets taken care of when we get a new setup for each step


jrryjcksn (approved):
\o/

PR #78, opened 2021-04-24
Refactor test script

rokshana-b:
Changed the test script to make code more reusable

### DIFF
--- a/test/acceptance/test/smoke_tests.go
+++ b/test/acceptance/test/smoke_tests.go
@@ -8,7 +8,6 @@
 package acceptance
 
 import (
-	"os"
 	"os/exec"
 
 	. "github.com/onsi/ginkgo"
@@ -16,15 +15,6 @@ import (
 	"github.com/onsi/gomega/gbytes"
 	"github.com/onsi/gomega/gexec"
 )
-
-func FileExists(name string) bool {
-	if _, err := os.Stat(name); err != nil {
-		if os.IsNotExist(err) {
-			return false
-		}
-	}
-	return true
-}
 
 var _ = Describe("WEGO Acceptance Tests", func() {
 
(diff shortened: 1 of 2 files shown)


test/acceptance/test/smoke_tests.go
```diff
@@ -26,10 +26,16 @@ func FileExists(name string) bool {
...
+var _ = Describe("WEGO Acceptance Tests", func() {
+
+	var session *gexec.Session
+	var err error
 
 	BeforeEach(func() {
 
+		Context("Given I have a wego binary installed on my local machine", func() {
+			Expect(FileExists(WEGO_BIN_PATH)).To(BeTrue())
+		})
```
iahmad9, line 38: I think Context container is not recommended to be used in BeforeEach or It blocks, instead it is preferred to keep them outside. Any particular reason for using it in this way?

test/acceptance/test/smoke_tests.go
```diff
@@ -26,10 +26,16 @@ func FileExists(name string) bool {
 	return true
 }
 
-var _ = Describe("WEGO Acceptance Tests ", func() {
+var _ = Describe("WEGO Acceptance Tests", func() {
+
+	var session *gexec.Session
+	var err error
```
iahmad9, line 32: Good one 👍

test/acceptance/test/smoke_tests.go
rokshana-b: It was the only way I could make the script run... I wanted to run the 'Given' condition before each test case but the hierarchy of 'Describe', 'Context', 'It', 'By' was giving me problems

rokshana-b: I've changed `Context` to `By` — seems to work just fine.

iahmad9 (approved):

test/acceptance/test/smoke_tests.go
```diff
@@ -104,5 +90,43 @@ var _ = Describe("WEGO Acceptance Tests ", func() {
...
+			Expect(err).ShouldNot(HaveOccurred())
+		})
+
+		VerifyUsageText()
+
+	})
+
+	It("Verify that wego flux can print out version information", func() {
+
+		By("When I run 'wego flux -- -version", func() {
```
iahmad9, line 122: would be good if this is changed to `wego flux -- -v` as that's what the test case is testing.

PR #105, opened 2021-05-04
release workflow

J-Thompson12:
Honestly not really sure how to test this so hopefully it works.


luizbafilho:
Is it possible for one workflow to depend on the other? it would important to only allow a release if the unit tests and the acceptance tests pass.

On testing, you can fork the project and do that in your fork, that way you create releases without messing up the official repo.

jrryjcksn (approved):
Looks okay to me. I guess we'll see...

J-Thompson12:
@luizbafilho you reminded me what i was looking into before vacation. I know there isnt a way to depend on jobs from another workflow. I figured though that the tests would be run on the pr then by the time it got to this we would know that the tests had passed.

PR #106, opened 2021-05-04
36/wego add command

jrryjcksn:
Provides a first cut of `wego add`.

- Creates wego repo if not present
- Installs Application CRD, source controller, kustomize controller, and GitHub secret for wego repo into cluster
- Installs GitHub secret plus source and kustomize controllers into wego repo for newly added repo


cmd/wego/add/cmd.go
```diff
@@ -0,0 +1,290 @@
...
+kind: Application
+metadata:
+  name: {{ .AppName }}
+spec:
+  path: {{ .AppPath }}
+  url: {{ .AppURL }}
+`
+
+// Will move into filesystem when we store wego infrastructure in git
+const appCRD = `apiVersion: apiextensions.k8s.io/v1beta1
```
luizbafilho, line 35: Since you are creating a crd here, should we then create the story to start implementing the wego controller?

jrryjcksn: We can, but I'm still not sure what to put in it yet. I had the choice to either tell flux to ignore `app.yaml` or turn it into a CRD so that flux wouldn't fail on it. I figured I'd go ahead and make the CRD now...

luizbafilho: The way I see could be like: `wego add` only drops the `app.yaml` there and nothing more, it would be the controller's responsibility to create the source and kustomization|helm custom resources, as a default behavior. That way we kinda abstract away flux entirely. In case the user decides they want more control, we would allow them to override the flux crs, by dropping them into the repo.

for now, we can keep the command adding those flux crs so we have a working command, but as soon we have the controller in place we could let the controller manage them.

cc @palemtnrider

jrryjcksn: That makes sense to me. In particular, we can consult the app.yaml to determine which controllers a particular app needs

josecordaz (approved):
LGTM

PR #116, opened 2021-05-06
Fix issues with flux-system namespace and adding when wego repo exists

jrryjcksn:


jrryjcksn:
More coverage coming shortly...

josecordaz (approved):
LGTM

PR #120, opened 2021-05-06
increase coverage on integration tests

J-Thompson12:
I looked it over and this looks like it is accurate on its coverage.


luizbafilho:
Looks like its trying to get coverage from the fakes: https://coveralls.io/builds/39429875/source?filename=pkg%2Ffluxops%2Ffluxopsfakes%2Ffake_flux_handler.go

I've seen there is a ignore option in `go-acc` has a `--ignore` flag, could try out calling it with `fakes`?

J-Thompson12:
@luizbafilho go-acc can use any flags from go test. I have it set to test every go file in the dir. I should add something to skip those though. Good catch.

jrryjcksn (approved):
\o/

test/acceptance/test/utils_test.go
```diff
@@ -0,0 +1,12 @@
+package acceptance
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestFileExists(t *testing.T) {
+	require.True(t, FileExists("utils.go"))
```
josecordaz, line 10: kind suggestion. It would be great to have a file used specifically for this test. In case the rare case utils.go changes name. There are some methods to create temp files like this:
```
tempFile, err := ioutil.TempFile("", "")
os.Remove(tempFile.Name())
```
Then we don't need to rely any external resources to test this function  🙂

josecordaz (approved):
LGTM

test/acceptance/test/utils_test.go
J-Thompson12: that is a very valid point

PR #135, opened 2021-05-11
Added ns wego-system when applying source and kustomize to flux

josecordaz:


luizbafilho (changes requested):
that should be configurable via command flag, the `install` defaults to `wego-system` but the user can pick whatever namespace he wants.

josecordaz:
> that should be configurable via command flag, the `install` defaults to `wego-system` but the user can pick whatever namespace he wants.

@luizbafilho  How can I take the namespace used in the install command for the add command ?

luizbafilho (approved):
LGTM

PR #136, opened 2021-05-11
Adding helm deployment option

luizbafilho:


pkg/cmdimpl/add.go
```diff
@@ -269,16 +283,23 @@ func Add(args []string, allParams AddParamSet) {
 
 	// Create controllers for new repo being added
 	source := generateSourceManifest()
-	kust := generateKustomizeManifest()
+
```
jrryjcksn, line 286: When we switch over to profiles, it will definitely be possible for a profile to include both helm charts and other artifacts that get kustomized. Arguably, even today a user _might_ want to add an app/repo that includes both. Do we know it's okay to just pick one or the other?

jrryjcksn (changes requested):

pkg/cmdimpl/add.go
luizbafilho: In the profiles world, we would be adding the profile and letting it handle the helm and kustomize.

Although I've never seen that kind of mixed-use, generally people chooses one for an given application, it's definitely possible. How do you suggest we could expose that? how would the command look like?

jrryjcksn: Fair enough about profiles handling this; it'll be interesting to see what we have to rip out when they come around. 

I don't know what a good UX is for that yet. I asked Cornelia a while ago if it would be okay in the early phases to just install both the controllers by default and she thought that would be all right (and a lot simpler).

luizbafilho: I don't think that's going to be possible because it breaks flux expectations. For instance, when creating a Kustomization pointing to a helm chart path it breaks with the following error: 

```
kustomize create failed: failed to decode Kubernetes YAML from /tmp/podinfo046500499/charts/podinfo/Chart.yaml: error unmarshaling JSON: while decoding JSON: Object 'Kind' is missing in '{"apiVersion":"v1","appVersion":"5.2.0","description":"Podinfo Helm chart for Kubernetes","engine":"gotpl","home":"https://github.com/stefanprodan/podinfo","maintainers":[{"email":"stefanprodan@users.noreply.github.com","name":"stefanprodan"}],"name":"podinfo","sources":["https://github.com/stefanprodan/podinfo"],"version":"5.2.0"}'
```

Or when creating a helm pointing to the a kustomization folder you'll see:

```
message: HelmChart 'flux-system/flux-system-podinfo' is not ready
reason: ArtifactFailed
status: "False"
type: Ready
```

jrryjcksn: I wasn't actually referring to mixing them in that sense; only that an "app" might include multiple artifacts of different types. Someone may want to use external code that is already packaged in multiple formats. Imagine wanting to create an observability app that pulled in both Prometheus and Grafana but one of them was packaged as a helm chart and the other was just manifests w/ kustomization.

luizbafilho: do you think that is a phase 1 concern?

jrryjcksn: I guess I'm not sure why we would deliberately disallow it in phase 1 since it's no extra work to support it -- it's less work, in fact, since we don't have to expose the choice to the user. Selecting one or the other seems like an optimization specifically to reduce the set of controller pods

luizbafilho: Ahh now I see the problem, the comment you added is misleading `// Create controllers for new repo being added` that doesn't add the controller, it added the custom resources for the app. I'll change it to make it clear.

jrryjcksn: It adds a custom resource but also adds the manifests for the source and kustomize controllers for the new repo (which will cause the controllers to be created after a sync)

luizbafilho: the source and kustomize custom resources for the new repo are added before that. At that point they were applied already.

jrryjcksn: It adds **an application CR** into the wego repo along with **the manifests for the two controllers**. If you don't agree with this statement, can we just have a live conversation? It sure seems like we're talking past each other.

jrryjcksn (approved):
LGTM

PR #141, opened 2021-05-12
add access test

J-Thompson12:
Okay I will mention I am not a fan of all the err returns but I couldnt really figure out another way using the ginko tests.

### DIFF
--- a/cmd/wego/add/cmd.go
+++ b/cmd/wego/add/cmd.go
@@ -31,6 +31,7 @@ func init() {
 	Cmd.Flags().StringVar(&params.Path, "path", "./", "Path of files within git repository")
 	Cmd.Flags().StringVar(&params.Branch, "branch", "main", "Branch to watch within git repository")
 	Cmd.Flags().StringVar(&params.PrivateKey, "private-key", filepath.Join(os.Getenv("HOME"), ".ssh", "id_rsa"), "Private key that provides access to git repository")
+	Cmd.Flags().BoolVar(&params.IsPrivate, "private", true, "Set access control on the repo")
 }
 
 func runCmd(cmd *cobra.Command, args []string) {
--- a/pkg/cmdimpl/add.go
+++ b/pkg/cmdimpl/add.go
@@ -80,6 +80,7 @@ type AddParamSet struct {
 	Branch         string
 	PrivateKey     string
 	DeploymentType string
+	IsPrivate      bool
 }
 
 var (
@@ -246,7 +247,12 @@ func Add(args []string, allParams AddParamSet) {
 
 		orgRef := cgitprovider.NewOrgRepositoryRef(cgitprovider.GITHUB_DOMAIN, owner, fluxRepoName)
 
-		repoInfo := cgitprovider.NewRepositoryInfo("wego repo", gitprovider.RepositoryVisibilityPrivate)
+		access := gitprovider.RepositoryVisibilityPrivate
+		if !params.IsPrivate {
+			access = gitprovider.RepositoryVisibilityPublic
+		}
+
+		repoInfo := cgitprovider.NewRepositoryInfo("wego repo", access)
 
 		repoCreateOpts := &gitprovider.RepositoryCreateOptions{
 			AutoInit:        gitprovider.BoolVar(true),
--- a/test/acceptance/test/core_operations_test.go
+++ b/test/acceptance/test/core_operations_test.go
@@ -2,7 +2,6 @@
 // +build acceptance
 // +build !smoke
 
-// ^^ temporarily disabling this test until we have it in ginkgo format.
 package acceptance
 
 // Runs basic WeGO operations against a kind cluster.
@@ -12,16 +11,18 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"strings"
-	"testing"
 	"time"
 
 	"github.com/fluxcd/go-git-providers/github"
 	"github.com/fluxcd/go-git-providers/gitprovider"
-	log "github.com/sirupsen/logrus"
-	"github.com/stretchr/testify/require"
-	"github.com/weaveworks/weave-gitops/pkg/cmdimpl"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	"github.com/onsi/gomega/gbytes"
+	"github.com/onsi/gomega/gexec"
+	"github.com/prometheus/common/log"
 	"github.com/weaveworks/weave-gitops/pkg/flux"
 	"github.com/weaveworks/weave-gitops/pkg/fluxops"
 	"github.com/weaveworks/weave-gitops/pkg/status"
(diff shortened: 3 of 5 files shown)


jrryjcksn:
Instead of:
```
err = foo()
```
you can say:
```
Expect(foo()).Should(Succeed())
```

test/acceptance/test/core_operations_test.go
```diff
@@ -64,176 +65,314 @@ var (
...
+			Expect(setUpTestRepo()).Should(Succeed())
+		})
+
+	})
+
+	It("Verify add private repo when repo does not already exist", func() {
+
+		By("When i run 'wego add .'", func() {
+			dir, err := os.Getwd()
+			Expect(err).Should(Succeed())
```
jrryjcksn, line 99: This one should be the other way :-)

test/acceptance/test/core_operations_test.go
```diff
@@ -64,176 +65,314 @@ var (
...
+			Expect(os.Chdir(tmpDir)).Should(Succeed())
+			Expect(os.Chdir(dir)).Should(Succeed())
```
jrryjcksn, line 101: Why are we changing to do different directories in a row? Is it to test if `tmpDir` is set up correctly?

test/acceptance/test/core_operations_test.go
```diff
@@ -64,176 +65,314 @@ var (
...
+			command := exec.Command(WEGO_BIN_PATH, "add", ".")
+			session, err = gexec.Start(command, GinkgoWriter, GinkgoWriter)
+			Expect(err).Should(Succeed())
```
jrryjcksn, line 104: This one too. I think it read weird to say "expect error should succeed"

test/acceptance/test/core_operations_test.go
```diff
@@ -64,176 +65,314 @@ var (
...
+		})
+
+		By("Then a private repo with name foo-cluster-wego is created on the remote git", func() {
+			Expect(ensureWegoRepoExists()).Should(Succeed())
+		})
+
+		By("kubectl get pods -n wego-system should list the source and kustomize controllers", func() {
+			Expect(waitForNginxDeployment()).Should(Succeed())
+			command := exec.Command("sh", "-c", utils.Escape("kubectl get pods -n wego-system"))
```
jrryjcksn, line 113: We do this same sort of command and session setup several times. It might be nice to pull it out so we could say something like: `Expect(runCommandForGinkgo(<cmd>)).Should(Succeed())`

test/acceptance/test/core_operations_test.go
```diff
@@ -64,176 +65,314 @@ var (
...
+			Eventually(session).Should(gbytes.Say("kustomize-controller"))
+			Eventually(session).Should(gbytes.Say("source-controller"))
+		})
+	})
+
+	It("Verify add public repo when repo does not already exist", func() {
+
+		By("When i run 'wego add . --private=false'", func() {
+			dir, err := os.Getwd()
+			Expect(err).Should(Succeed())
```
jrryjcksn, line 125: One more. Sorry to be so pedantic.

test/acceptance/test/core_operations_test.go
```diff
@@ -64,176 +65,314 @@ var (
...
+			Expect(os.Chdir(tmpDir)).Should(Succeed())
```
jrryjcksn, line 126: Same question about changing dirs from above

test/acceptance/test/core_operations_test.go
```diff
@@ -64,176 +65,314 @@ var (
...
+		return err
+	}
+
+	url := fmt.Sprintf("https://github.com/%s/%s", os.Getenv("GITHUB_ORG"), name)
+	ref, err := gitprovider.ParseOrgRepositoryURL(url)
+	if err != nil {
+		return err
+	}
+	_, err = client.OrgRepositories().Get(ctx, *ref)
+	if err != nil {
```
jrryjcksn, line 205: This could just be: `return err`; no need for the "if"

test/acceptance/test/core_operations_test.go
```diff
@@ -64,176 +65,314 @@ var (
...
-	require.NoError(t, err)
-	require.NoError(t, utils.CallCommandForEffectWithInputPipeAndDebug("kubectl apply -f -", string(manifests)))
+	if err != nil {
+		return err
+	}
+	return utils.CallCommandForEffectWithInputPipeAndDebug("kubectl apply -f -", string(manifests))
 }
 
-func getWegoRepoName(t *testing.T) string {
+func getWegoRepoName() (string, error) {
```
jrryjcksn, line 251: Could either get rid of this entirely or just have it return `fluxops.GetRepoName()`

test/acceptance/test/core_operations_test.go
```diff
@@ -64,176 +65,314 @@ var (
...
 
-	require.NoError(t, err)
-	originurl := fmt.Sprintf("ssh://git@github.com/%s/%s", os.Getenv("GITHUB_ORG"), getRepoName(t))
-	cmd := fmt.Sprintf(`git remote add origin %s && \
-		git pull --rebase origin main && \
-		git push --set-upstream origin main`,
-		originurl)
-	err = utils.CallCommandForEffectWithDebug(cmd)
-	require.NoError(t, err)
+	err = os.Chdir(dir)
```
jrryjcksn, line 319: can just return `os.Chdir(dir)`

jrryjcksn (changes requested):
A few nits...

test/acceptance/test/core_operations_test.go
J-Thompson12: I actually just copied that over from yours. I wasnt really sure.

test/acceptance/test/core_operations_test.go
J-Thompson12: yeah that makes sense.

test/acceptance/test/core_operations_test.go
jrryjcksn: Do you mean this?
```
	require.NoError(t, os.Chdir(tmpDir))
	defer func() {
		require.NoError(t, os.Chdir(dir))
	}()
```
That's different... it's changing back at the end of the function due to the defer

J-Thompson12: Yeah i do see that

test/acceptance/test/core_operations_test.go
```diff
@@ -60,180 +61,310 @@ spec:
...
+		By("Then a private repo with name foo-cluster-wego is created on the remote git", func() {
+			Expect(ensureWegoRepoExists()).Should(Succeed())
+		})
 
-	cmdimpl.Add([]string{"."}, cmdimpl.AddParamSet{Name: "", Url: "", Path: "./", Branch: "main", PrivateKey: keyFilePath, Namespace: "wego-system", DeploymentType: "kustomize"})
+		By("kubectl get pods -n wego-system should list the source and kustomize controllers", func() {
+			Expect(waitForNginxDeployment()).Should(Succeed())
+			Expect(runCommandForGinkgo("kubectl get pods -n wego-system")).Should(Succeed())
+			Eventually(session).Should(gbytes.Say("kustomize-controller"))
+			Eventually(session).Should(gbytes.Say("source-controller"))
```
iahmad9, line 113: When you use `gbytes.Say`, it implicitly also verify the order of those controllers in the output which is not important in this case. I would suggest using the `ContainsSubstring` or `MatchRegexp` instead.

test/acceptance/test/core_operations_test.go
```diff
@@ -60,180 +61,310 @@ spec:
...
 	clusterName, err := status.GetClusterName()
 	if err == nil {
 		ctx := context.Background()
-		url := fmt.Sprintf("https://github.com/%s/%s", os.Getenv("GITHUB_ORG"), getRepoName(t))
+		name, err := getRepoName()
+		if err != nil {
+			return err
+		}
+
+		url := fmt.Sprintf("https://github.com/%s/%s", os.Getenv("GITHUB_ORG"), name)
```
iahmad9, line 326: SUGGESTION: As it may be GitLab or some privately hosted git in the future, I would suggest not to hard code the URLs.

pkg/cmdimpl/add.go
```diff
@@ -280,7 +281,12 @@ func Add(args []string, allParams AddParamSet) {
 
 		orgRef := cgitprovider.NewOrgRepositoryRef(cgitprovider.GITHUB_DOMAIN, owner, fluxRepoName)
 
-		repoInfo := cgitprovider.NewRepositoryInfo("wego repo", gitprovider.RepositoryVisibilityPrivate)
+		access := gitprovider.RepositoryVisibilityPrivate
+		if !params.IsPrivate {
+			access = gitprovider.RepositoryVisibilityPublic
+		}
+
+		repoInfo := cgitprovider.NewRepositoryInfo("wego repo", access)
```
iahmad9, line 290: Please also add unit tests

cmd/wego/add/cmd.go
```diff
@@ -31,6 +31,7 @@ func init() {
 	Cmd.Flags().StringVar(&params.Path, "path", "./", "Path of files within git repository")
 	Cmd.Flags().StringVar(&params.Branch, "branch", "main", "Branch to watch within git repository")
 	Cmd.Flags().StringVar(&params.PrivateKey, "private-key", filepath.Join(os.Getenv("HOME"), ".ssh", "id_rsa"), "Private key that provides access to git repository")
+	Cmd.Flags().BoolVar(&params.IsPrivate, "private", true, "Set access control on the repo")
```
iahmad9, line 34: please add unit tests.

test/acceptance/test/testsuite_common_test.go
```diff
@@ -40,4 +42,12 @@ var _ = BeforeSuite(func() {
 		WEGO_BIN_PATH = "/usr/local/bin/wego"
 	}
 	log.Infof("WEGO Binary Path: %s", WEGO_BIN_PATH)
+	Expect(checkInitialStatus()).Should(Succeed())
 })
+
+func checkInitialStatus() error {
```
josecordaz, line 48: Just a minor thing. As this function seems to be related to the testing setup rather than the logic itself. It would be cool to have it in some setup purpose environment, not sure what would that be for Ginko/Gomego yet. But this is the link for the setup function when doing normal testing. https://golang.org/pkg/testing/#hdr-Main . I added an implementation here as well just in case. [link](https://github.com/weaveworks/weave-gitops/blob/main/pkg/gitproviders/common_test.go#L137). Probably something we needed to think about for future work. As as a function like that could be reusable for other suites doing testing for other functions. 🙂

josecordaz (approved):
👍🏼 LGTM

test/acceptance/test/testsuite_common_test.go
```diff
@@ -3,13 +3,15 @@
...
 import (
+	"fmt"
 	"os"
 	"testing"
 
 	. "github.com/onsi/ginkgo"
 	ginkgo "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 	log "github.com/sirupsen/logrus"
+	"github.com/weaveworks/weave-gitops/pkg/status"
```
iahmad9, line 14: Could this be avoided, given it’s a black box test, should avoid using internal functions/components.

jrryjcksn (approved):
\o/ LGTM

PR #152, opened 2021-05-13
Moving app crd to install command

### LINKED ISSUE #139: Move app crd to install command
**Short description**
The App CRD should be applied at install phase together with all the other CRDs

https://github.com/weaveworks/weave-gitops/blob/main/pkg/cmdimpl/add.go#L34

luizbafilho:
closes #139


josecordaz (approved):
LGTM

PR #154, opened 2021-05-13
Disabling argument parsing on flux command

luizbafilho:
We no longer need to use `--` after a flux command.

### DIFF
--- a/cmd/wego/flux/cmd.go
+++ b/cmd/wego/flux/cmd.go
@@ -11,10 +11,11 @@ import (
 )
 
 var Cmd = &cobra.Command{
-	Use:     "flux -- [flux commands or flags]",
-	Short:   "Use flux commands",
-	Example: "wego flux -- install -h",
-	Run:     runCmd,
+	Use:                "flux -- [flux commands or flags]",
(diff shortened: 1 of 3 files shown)


cmd/wego/flux/cmd.go
```diff
@@ -11,10 +11,11 @@ import (
 )
 
 var Cmd = &cobra.Command{
-	Use:     "flux -- [flux commands or flags]",
-	Short:   "Use flux commands",
-	Example: "wego flux -- install -h",
-	Run:     runCmd,
+	Use:                "flux [flux commands or flags]",
+	Short:              "Use flux commands",
+	DisableFlagParsing: true,
```
josecordaz, line 16: Nice finding

josecordaz (approved):
LGTM

luizbafilho:
@J-Thompson12 look what got it working. 🎉

cmd/wego/flux/cmd.go
luizbafilho: hahah yep, was looking for something else and I looked to that and had aha 💡 moment.

PR #162, opened 2021-05-17
fix release workflow

J-Thompson12:
I checked out the flux workflow and these are the differences that i noticed that may possibly make a difference.


josecordaz (approved):
LGTM

PR #163, opened 2021-05-17
fix release try 2

J-Thompson12:
There was an error that unshallow didnt make sense with a fetch depth of 0. According to goreleaser docs fetch depth needs to be set so lets hope this works.


josecordaz (approved):
LGTM

PR #175, opened 2021-05-19
house keeping

iahmad9:
* Change the timeout to avoid flakiness when resetting the cluster, going forward we have to find some better solution for this.
* always upload the test results to the testspace even when they fail.

### DIFF
--- a/test/acceptance/test/testsuite_common_test.go
+++ b/test/acceptance/test/testsuite_common_test.go
@@ -16,7 +16,7 @@ func GomegaFail(message string, callerSkip ...int) {
 
 	//Show pods
 	err := showItems("pods")
-	if err == nil {
+	if err != nil {
 		log.Infof("Failed to print the pods")
 	}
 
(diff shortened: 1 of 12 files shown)


test/acceptance/test/testsuite_common_test.go
```diff
@@ -16,7 +16,7 @@ func GomegaFail(message string, callerSkip ...int) {
 
 	//Show pods
 	err := showItems("pods")
-	if err == nil {
+	if err != nil {
 		log.Infof("Failed to print the pods")
 	}
```
J-Thompson12, line 21: This is super minor but wouldnt this function be more helpful if it returned an error instead of just logging it.

J-Thompson12 (approved):

test/acceptance/test/testsuite_common_test.go
iahmad9: Sure thing but couldn't think of a use case where the caller would interested to do much with the error, so thought just log the error and done with it.

PR #194, opened 2021-05-21
Switch to standard "example.com" example emails

jrryjcksn:
It's good practice to not use real email domains in tests/docs.


josecordaz (approved):
LGTM

PR #201, opened 2021-05-24
Attempt to increase coverage

josecordaz:


jpellizzari (approved):
Looks great! 

A PR description will really help whomever comes across this in the future understand why it was added. This one is pretty self-explanatory, but adding any additional context would be helpful for understanding "why" things changed.

I think of the PR description as notes to my future self.

PR #202, opened 2021-05-24
Add Makefile incantation to ensure the static assets directory always exists

jpellizzari:
Ensure that unit-tests will not trip on a syntax error for embedding of the UI

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -79,7 +79,7 @@ jobs:
     - name: Install UI Deps
       run: npm ci
     - name: Build UI Assets
-      run: make cmd/ui/dist/index.html
+      run: make cmd/ui/dist/main.js
     - name: Fake Install flux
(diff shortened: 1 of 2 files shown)


iahmad9 (commented):
After pulling your branch, I tried running `go test -v ./...` from the project root but still seeing the same issue. I am wondering if ui should have its own sub-folder under the project root (like wkp) ?

jpellizzari:
@iahmad9 The nature of `go` embedding means that we will have to let the `Makefile` run the tests, as it ensures that the requisite files will be in place. Luckily, it only has to run once on someone local environment, then `go test` should work.

> I am wondering if ui should have its own sub-folder under the project root (like wkp) ?

It does have its own subfolder. The runtime `main.go` file will still need to be parsed for `go test ./..`

iahmad9 (approved):

PR #210, opened 2021-05-25
Update snyk job to handle npm deps

jpellizzari:
Fixes the broken build on main. From slack:

> the issue is that snyk is looking at the package-lock.json file during a golang job, which is kinda weird. There are "forced" reconciliations that happen on install to avoid vulnerabilities that Snyk doesn't pick up because it doesn't do the install.

This PR removes the offending UI stuff from a job that scans golang files.

### DIFF
--- a/.github/workflows/docker_scan.yaml
+++ b/.github/workflows/docker_scan.yaml
@@ -50,6 +50,8 @@ jobs:
         uses: actions/checkout@v2
       - name: Fake Install flux
         run: mkdir -p pkg/flux/bin && touch pkg/flux/bin/flux
+      - name: Remove UI deps from Scan
+        run: rm package-lock.json && rm package.json && make cmd/ui/dist/index.html
       - name: Run Snyk to check for vulnerabilities
         uses: snyk/actions/golang@master
         env:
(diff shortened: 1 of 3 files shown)


jpellizzari:
Use Github/NPM scanning for UI code

.github/workflows/docker_scan.yaml
```diff
@@ -50,6 +50,8 @@ jobs:
         uses: actions/checkout@v2
       - name: Fake Install flux
         run: mkdir -p pkg/flux/bin && touch pkg/flux/bin/flux
+      - name: Remove UI deps from Scan
+        run: rm package-lock.json && rm package.json && make cmd/ui/dist/index.html
```
iahmad9, line 54: I am not clear, will it still scan the deps in package.json or that's not needed?

iahmad9 (approved):

.github/workflows/docker_scan.yaml
jpellizzari: Snyk will no longer scan `package.json`. I have added a step to the `test` job to do a vulnerability check using `npm audit`. Its a bit easier to use and run locally.

PR #221, opened 2021-05-27
Create repo only when DryRun is false

josecordaz:
I changed the repoCreation call when DryRun is false.
I updated the createRepo tests so they could work when the DryRun flag is false.


jrryjcksn (approved):
\o/ LGTM

PR #231, opened 2021-06-01
Add version check on every command

J-Thompson12:
Added a version check that runs after every command and adds flux version to version command.  https://app.zenhub.com/workspaces/weaveworks-foss-603cb6b1d4dff9000e242dc4/issues/weaveworks/weave-gitops/189

### DIFF
--- a/cmd/wego/main.go
+++ b/cmd/wego/main.go
@@ -48,4 +48,12 @@ func main() {
 		fmt.Fprintf(os.Stderr, "Error: %v\n", err)
 		os.Exit(1)
 	}
+
+	//Run on all commands that are not version
+	if os.Args[1] != "version" {
+		versionInfo, _ := version.CheckVersion()
+		if versionInfo != "" {
+			log.Infof(versionInfo)
+		}
+	}
 }
--- a/cmd/wego/main.go
+++ b/cmd/wego/main.go
@@ -49,11 +49,8 @@ func main() {
 		os.Exit(1)
 	}
 
-	//Run on all commands that are not version
-	if os.Args[1] != "version" {
-		versionInfo, _ := version.CheckVersion()
-		if versionInfo != "" {
-			log.Infof(versionInfo)
-		}
+	versionInfo, _ := version.CheckVersion()
+	if versionInfo != "" {
+		log.Info(versionInfo)
 	}
 }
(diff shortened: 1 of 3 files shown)


cmd/wego/main.go
```diff
@@ -48,4 +48,12 @@ func main() {
 		fmt.Fprintf(os.Stderr, "Error: %v\n", err)
 		os.Exit(1)
 	}
+
+	//Run on all commands that are not version
+	if os.Args[1] != "version" {
```
josecordaz, line 53: I wonder if we can simplify this by removing this `if` and remove the call to CheckVersion in the `version` command 🤔.  So the check can be done in only one place.

cmd/wego/version/cmd.go
```diff
@@ -22,16 +23,36 @@ var Cmd = &cobra.Command{
...
 
 func runCmd(cmd *cobra.Command, args []string) {
-	if checkResponse, err := checkpoint.Check(&checkpoint.CheckParams{
+	logStr, err := CheckVersion()
+	if err != nil {
+		log.Error(err)
+	}
+
+	if logStr != "" {
+		log.Infof(logStr)
```
josecordaz, line 32: `Infof` is usually used when arguments need to be passed in, you might want to try `log.Info`.

cmd/wego/main.go
J-Thompson12: I see what you mean. Yeah that is a very valid point.

cmd/wego/version/cmd.go
```diff
@@ -22,16 +22,27 @@ var Cmd = &cobra.Command{
...
-		fmt.Println("GitCommit:", GitCommit)
-		fmt.Println("BuildTime:", BuildTime)
-		fmt.Println("Branch:", Branch)
+	})
+
+	if err != nil {
+		return "", fmt.Errorf("Unable to retrieve latest version: %v", err)
 	}
+
+	if err == nil && checkResponse.Outdated {
```
josecordaz, line 42: Not need to check `err==nil` here as it is inferred for the previous `if` statement.

josecordaz (approved):
LGTM

cmd/wego/version/cmd.go
J-Thompson12: Whoops forgot about that after changing things up. Good catch.

PR #232, opened 2021-06-01
Added wait logic to make sure the repo exists before continuing the execution

josecordaz:
A wait logic was added to make sure the repository exists before getting any info from it.

There was an error saying the repository doesn't exists.

By adding unit tests.


pkg/gitproviders/common.go
```diff
@@ -71,14 +94,64 @@ func getOwnerType(provider gitprovider.Client, owner string) (string, error) {
 	})
 
 	if err != nil {
-		if errors.Is(err, gitprovider.ErrNotFound) {
-			return OwnerTypeUser, nil
+		if errors.Is(err, gitprovider.ErrNotFound) ||
+			strings.Contains(err.Error(), badCredentialsError) ||
+			strings.Contains(err.Error(), notFoundError) {
```
luizbafilho, line 99: you are checking again the not found error, and bad credentials is an actual error, so it should return as such.

pkg/gitproviders/common.go
```diff
@@ -4,15 +4,22 @@ import (
...
 	"github.com/fluxcd/go-git-providers/gitprovider"
 	"github.com/weaveworks/weave-gitops/pkg/override"
 )
 
+type ProviderAccountType string
+
 const (
-	OwnerTypeUser = "user"
-	OwnerTypeOrg  = "organization"
+	UserAccountType ProviderAccountType = "user"
```
luizbafilho, line 21: One small change here, it's common in go to prefix the constants that are "kinda enums", so instead of `UserAccountType` and `OrgAccountType`, it would be `AccountTypeUser` and `AccountTypeOrg`.

you can see something like that in the HTTP package: https://golang.org/pkg/net/http/#pkg-constants

luizbafilho (changes requested):

pkg/gitproviders/common.go
josecordaz: If you are trying to get a org using a user account that doesn't exist it responds with bad credentials. 
image

For the Not found issue, I guess there is a bug on gogitprovider because it is not caching the error with this condition `errors.Is(err, gitprovider.ErrNotFound)`

pkg/gitproviders/common.go
josecordaz: It seemed easier to have the User and Org as prefix for IDE autocomplete 🤔

pkg/gitproviders/common.go
luizbafilho: I see, so we might need to find a different way, or another endpoint, imagine if the person actually uses the wrong token to query the org, that would return the wrong type, creating another sort of bugs.

luizbafilho: btw, I've just tested and returned as expected the Not Found.

image

josecordaz: did you use your personal token for this test?

josecordaz: I used the same org as you here:
image

luizbafilho (approved):

PR #233, opened 2021-06-01
go-git implementation

luizbafilho:
**What changed?**
Moving all git cli calls to `go-git` library.

**Why?**
It'll make it easier to mock and maintain the code

**How did you test it?**

**Release notes**

**Documentation Changes**


cmd/wego/add/cmd.go
```diff
@@ -42,7 +44,19 @@ func init() {
 
 func runCmd(cmd *cobra.Command, args []string) {
 	params.Namespace, _ = cmd.Parent().Flags().GetString("namespace")
-	if err := cmdimpl.Add(args, params); err != nil {
+
+	authMethod, err := ssh.NewPublicKeysFromFile("git", params.PrivateKey, params.PrivateKeyPass)
```
josecordaz, line 48: does this means that it is strictly necessary to pass in these two flags? If passing them empty means that it will take the default ssh key then ignore my comment. Otherwise I would add as the default option to look for the default ssh key in `~/.ssh/`

pkg/cmdimpl/add.go
```diff
@@ -80,14 +87,20 @@ func updateParametersIfNecessary() error {
 	}
 
 	if params.Url == "" {
-		urlout, err := exec.Command("git", "remote", "get-url", "origin").CombinedOutput()
+		gitClient := git.New(nil)
+		repo, err := gitClient.Open(params.Dir)
+		if err != nil {
+			return err
```
josecordaz, line 93: it would be helpful for debbuging if we have e more detailed error description. Maybe something like fmt.Errof("error accessing %s directory %s`, params.Dir,err) or similar.

pkg/cmdimpl/add.go
```diff
@@ -80,14 +87,20 @@ func updateParametersIfNecessary() error {
...
-			return wrapError(err, "could not get remote origin url")
+			return err
 		}
 
-		url := strings.TrimRight(string(urlout), "\n")
-		fmt.Printf("URL not specified; ")
-		params.Url = url
+		urls := remote.Config().URLs
+
+		params.Url = urls[0]
```
josecordaz, line 103: We need to handle the case of empty urls here so the app does not panic but instead show a descriptive error message unless it is not possible to have empty urls after the err check on repo.Remote.

pkg/cmdimpl/add.go
```diff
@@ -432,36 +428,39 @@ func Add(args []string, allParams AddParamSet) error {
...
-
 	if err != nil {
 		return wrapError(err, "error generating manifest")
 	}
 
-	sourceName := filepath.Join(appSubdir, "source-"+params.Name+".yaml")
-	manifestsName := filepath.Join(appSubdir, fmt.Sprintf("%s-%s.yaml", params.DeploymentType, params.Name))
-	appYamlName := filepath.Join(appSubdir, "app.yaml")
+	appSubdir := filepath.Join("apps", params.Name)
+	if err := os.MkdirAll(appSubdir, 0755); err != nil {
```
josecordaz, line 436: It seems we need to have this under DryRun check

pkg/cmdimpl/status.go
```diff
@@ -15,22 +13,6 @@ import (
 
 // Status provides the implementation for the wego status application command
 func Status(allParams AddParamSet) error {
-	// verify the app is in the wego apps folder
```
josecordaz, line 18: This removes the ability to let the user know the possibility of an unexcited app.

pkg/git/gogit.go
```diff
@@ -0,0 +1,240 @@
...
+}
+
+func New(auth transport.AuthMethod) *GoGit {
+	return &GoGit{
+		auth: auth,
+	}
+}
+
+func (g *GoGit) Open(path string) (*gogit.Repository, error) {
+	g.path = path
```
josecordaz, line 49: It might help to mitigate duplicate assignments if we move this to the New function instead.

pkg/git/gogit.go
```diff
@@ -0,0 +1,240 @@
...
+}
+
+func (g *GoGit) Write(path string, content []byte) error {
+	if g.repository == nil {
+		return ErrNoGitRepository
+	}
+
+	wt, err := g.repository.Worktree()
+	if err != nil {
+		return err
```
josecordaz, line 128: it might help for debugging to have descriptive error messages

josecordaz (changes requested):

josecordaz:
maybe a bit more coverage ? 😬

cmd/wego/add/cmd.go
luizbafilho: no change in current behavior, currently `params.PrivateKey` defaults to the user's private key.

jrryjcksn (commented):
VERY much cleaner overall!

cmd/wego/add/cmd.go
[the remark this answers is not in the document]
luizbafilho: I see. Just took the same approach flux does, but we can change it. Do you mind if I do it in another PR? I'd like to get this one merged asap.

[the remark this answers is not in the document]
jrryjcksn: Sure. No problem

josecordaz (approved):

PR #243, opened 2021-06-03
automate test for running wego add without controllers installed

rokshana-b:
**What changed?**
I added a test case where the automated test will verify that `wego add` cannot run without having wego controllers installed.

**Why?**
It is part of the regression test suite.

**How did you test it?**
This is a test.

**Release notes**
N/A

**Documentation Changes**
N/A


iahmad9 (approved):
LGTM, please remove new test from SmokeTest Suite

PR #245, opened 2021-06-04
exclude main branch

iahmad9:
**What changed?**

**Why?**

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,8 +1,7 @@
-
 on:
   push:
     branches:
-      - main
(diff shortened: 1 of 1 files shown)


markramm (commented):
I can't tell by reading just the commit message and the diff what the purpose of excluding the main branch is.   Could you provide a commit message which explains they why as well?   I mean I can figure it out if I dig, but it should be obvious from the message/comments/diff.

iahmad9:
@markramm this branch is not for merging, it's for testing certain workflows. going to delete it anyway.

PR #250, opened 2021-06-05
Added subcommand app and fixes on giving the app location

josecordaz:
- change on wego add and wego status to be:
  wego app add
  wego app status
- Fixes to support two ways of defining an application location
  1.- using url flag
  2.- using the location argument

Easier for the user and it avoid sthe need of having a repo cloned so it can be added.

Tests were manually with the different options involved. For `wego app add ` in the first scenario I used only the URL and a second one using just the location argument. All results were as expected.

Changes in the use of add and status:
`wego add` --> `wego app add...`
`wego status` -- > `wego app status...`

### DIFF
--- a/cmd/wego/main.go
+++ b/cmd/wego/main.go
@@ -37,15 +37,23 @@ func configureLogger() {
 	}
 }
 
+var ApplicationCmd = &cobra.Command{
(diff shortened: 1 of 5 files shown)


jrryjcksn (approved):
LGTM \o/

PR #253, opened 2021-06-06
Support all three config modes

jrryjcksn:
**What changed?**

- We no longer create a wego repository
  - no more "private" flag
- We have three automation config options:
  - none (only place automation in cluster; no persistent storage)
  - URL (place automation in external repository -- like our old default but no auto-creation), 
  - "" (use the user's app repository)
- Dry run test correctly handles new git client

**Why?**
This is the design favored by the CTO that she will demo at GitOps days

**How did you test it?**
- Tested locally
- Added unit tests

**Release notes**
N/A

**Documentation Changes**
Yes, the getting started guide will need to be updated

### DIFF
--- a/pkg/cmdimpl/add.go
+++ b/pkg/cmdimpl/add.go
@@ -593,7 +593,7 @@ func getUserRepoName() string {
 }
 
 func getGoatRepoName() string {
-	return urlToRepoName(params.AppConfigUrl)
+	return urlToRepoName(strings.TrimSuffix(params.AppConfigUrl, ".git"))
 }
 
 func urlToRepoName(url string) string {
--- a/pkg/cmdimpl/add.go
+++ b/pkg/cmdimpl/add.go
@@ -7,22 +7,20 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"io/ioutil"
-	"os"
 	"path/filepath"
 	"strings"
 	"text/template"
 
-	"github.com/fluxcd/go-git-providers/gitprovider"
 	"github.com/weaveworks/weave-gitops/pkg/fluxops"
 	"github.com/weaveworks/weave-gitops/pkg/git"
-	"github.com/weaveworks/weave-gitops/pkg/gitproviders"
 	"github.com/weaveworks/weave-gitops/pkg/shims"
 	"github.com/weaveworks/weave-gitops/pkg/status"
 	"github.com/weaveworks/weave-gitops/pkg/utils"
 )
 
 type DeploymentType string
+type SourceType string
+type ConfigType string
 
 const appYamlTemplate = `apiVersion: wego.weave.works/v1alpha1
 kind: Application
@@ -36,6 +34,12 @@ spec:
 const (
 	DeployTypeKustomize DeploymentType = "kustomize"
 	DeployTypeHelm      DeploymentType = "helm"
+
+	SourceTypeGit  SourceType = "git"
+	SourceTypeHelm SourceType = "helm"
+
+	ConfigTypeUserRepo ConfigType = ""
+	ConfigTypeNone     ConfigType = "NONE"
 )
 
 type AddParamSet struct {
@@ -49,6 +53,8 @@ type AddParamSet struct {
 	PrivateKeyPass string
 	DeploymentType string
 	Chart          string
+	SourceType     string
+	AppConfigUrl   string
 	Namespace      string
 	DryRun         bool
 	IsPrivate      bool
@@ -199,9 +205,9 @@ func generateSourceManifestGit() ([]byte, error) {
 	secretName := params.Name
 
 	cmd := fmt.Sprintf(`create secret git "%s" \
-			--url="%s" \
-			--private-key-file="%s" \
-			--namespace=%s`,
+            --url="%s" \
+            --private-key-file="%s" \
+            --namespace=%s`,
 		secretName,
 		params.Url,
 		params.PrivateKey,
@@ -218,12 +224,12 @@ func generateSourceManifestGit() ([]byte, error) {
 	}
 
 	cmd = fmt.Sprintf(`create source git "%s" \
-			--url="%s" \
-			--branch="%s" \
-			--secret-ref="%s" \
-			--interval=30s \
-			--export \
-			--namespace=%s `,
+            --url="%s" \
+            --branch="%s" \
+            --secret-ref="%s" \
+            --interval=30s \
+            --export \
+            --namespace=%s `,
 		params.Name,
 		params.Url,
 		params.Branch,
@@ -238,10 +244,10 @@ func generateSourceManifestGit() ([]byte, error) {
 
 func generateSourceManifestHelm() ([]byte, error) {
 	cmd := fmt.Sprintf(`create source helm %s \
-			--url="%s" \
-			--interval=30s \
-			--export \
-			--namespace=%s `,
+            --url="%s" \
+            --interval=30s \
+            --export \
+            --namespace=%s `,
 		params.Name,
 		params.Url,
 		params.Namespace)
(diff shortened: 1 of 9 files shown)


pzfreo:
I have concerns about the --app-config-url name.

I would prefer --gitops-config-location or --gitops-automation-location 

@markramm thoughts?

pkg/cmdimpl/add_test.go
```diff
@@ -171,17 +171,75 @@ func handleGitLsRemote(arglist ...interface{}) ([]byte, []byte, error) {
...
+	},
+	WriteStub: func(a string, b []byte) error {
+		fmt.Println("failing write")
+		shims.Exit(1)
+		return nil
+	},
+}
+
+var ignoreGitClient = gitfakes.FakeGit{
+	CloneStub: func(ctx context.Context, a, b, c string) (bool, error) {
```
luizbafilho, line 208: Unless you explicitly want to see the print, there is no set all methods, it defaults to noop

pkg/cmdimpl/add_test.go
```diff
@@ -171,17 +171,75 @@ func handleGitLsRemote(arglist ...interface{}) ([]byte, []byte, error) {
 	return nil, nil, fmt.Errorf("NO!")
 }
 
-var fakeGitClient = gitfakes.FakeGit{}
+var failGitClient = gitfakes.FakeGit{
+	CloneStub: func(ctx context.Context, a, b, c string) (bool, error) {
```
luizbafilho, line 175: a failure would probably be returning an error instead of exiting the process.

pkg/cmdimpl/add_test.go
```diff
@@ -216,7 +319,7 @@ var _ = Describe("Dry Run Add Test", func() {
 			_ = override.WithOverrides(
 				func() override.Result {
 					deps := &AddDependencies{
-						GitClient: &fakeGitClient,
+						GitClient: &failGitClient,
```
luizbafilho, line 322: Just to exemplify the previous comment, this could be just `GitClient: &gitfakes.FakeGit{}`

pkg/cmdimpl/add.go
```diff
@@ -128,114 +121,26 @@ func updateParametersIfNecessary(gitClient git.Git) error {
...
-		fluxRepoName,
-		params.PrivateKey,
-		params.Namespace)
-	if params.DryRun {
-		fmt.Printf(cmd + "\n")
-	} else {
-
-		_, err = fluxops.CallFlux(cmd)
+func getUrls(client git.Git, remote string) ([]string, error) {
+	if _, ok := client.(*git.GoGit); ok {
```
luizbafilho, line 125: If you do this you won't be able to mock it in the test. It will ignore it every time.

With this, you should be able to mock that

```
GitClient: &gitfakes.FakeGit{
	OpenStub: func(s string) (*gogit.Repository, error) {
		r, err := gogit.Init(memory.NewStorage(), memfs.New())

		_, err = r.CreateRemote(&config.RemoteConfig{
			Name: "origin",
			URLs: []string{"http://foo/foo.git"},
		})
		return r, nil
	},
},
```

pkg/cmdimpl/add_test.go
jrryjcksn: This is so that any use of the code anywhere downstream will fail the test w/o trying to propagate the failure to ensure dry run is working. Returning an error wouldn't work.

luizbafilho: the use of invalid/non-existent repo addresses would probably accomplish the same, without requiring you to implement those, but I'm fine with it. It's up to you.

pkg/cmdimpl/add.go
jrryjcksn: Without a remote, everything panicked. I had no way to fake the remote at the time so I upleveled it.  I'd be very happy to do something better later

pkg/cmdimpl/add_test.go
jrryjcksn: Not for dry run

pkg/cmdimpl/add.go
luizbafilho: sounds good

pkg/cmdimpl/add_test.go
jrryjcksn: That makes sense; I was just trying to be consistent across all the overrides and that was the simplest way. Not wedded to it :-)

luizbafilho (approved):

pkg/cmdimpl/add.go
```diff
@@ -349,134 +270,276 @@ func Add(args []string, allParams AddParamSet, deps *AddDependencies) error {
...
+	}
+	appYaml, err := generateAppYaml()
+	if err != nil {
+		return wrapError(err, fmt.Sprintf("could not create app.yaml for '%s'", params.Name))
+	}
+	// kustomize or helm referencing single user repo source
+	applicationGOAT, err := generateApplicationGoat(userSourceName)
 	if err != nil {
-		return wrapError(err, "could not get flux repo name")
+		return wrapError(err, fmt.Sprintf("could not create GitOps automation for '%s'", params.Name))
```
palemtnrider, line 300: error message is using a different value than what is passed to generateApplicationGoat

pkg/cmdimpl/add.go
```diff
@@ -349,134 +270,276 @@ func Add(args []string, allParams AddParamSet, deps *AddDependencies) error {
...
+	if err := applyToCluster(userRepoSource, targetKustomize, appKustomize); err != nil {
+		return err
+	}
+	appYaml, err := generateAppYaml()
+	if err != nil {
+		return wrapError(err, fmt.Sprintf("could not create app.yaml for '%s'", params.Name))
+	}
+	applicationGoat, err := generateApplicationGoat(userSourceName)
+	if err != nil {
+		return wrapError(err, fmt.Sprintf("could not create GitOps automation for '%s'", params.Name))
```
palemtnrider, line 329: error message is using a different value than what is passed to generateApplicationGoat

pkg/cmdimpl/add.go
```diff
@@ -349,134 +270,276 @@ func Add(args []string, allParams AddParamSet, deps *AddDependencies) error {
...
+	}
+	if err := writeAppYaml(gitClient, appYaml, ".wego"); err != nil {
+		return err
+	}
+	if err := writeGoats(gitClient, ".wego", applicationGoat); err != nil {
```
palemtnrider, line 334: NIt: Since writeAppYaml and writeGoats are performing similar operations, I'd prefer to have the arguments in the same order.  i.e. base path then the manifests.

pkg/cmdimpl/add.go
```diff
@@ -349,134 +270,276 @@ func Add(args []string, allParams AddParamSet, deps *AddDependencies) error {
...
+            --secret-ref="%s" \
+            --interval=30s \
+            --export \
+            --namespace="%s"`,
+		repoName,
+		repoUrl,
+		params.Branch,
+		secretName,
+		params.Namespace)
+	sourceManifest, err := fluxops.CallFlux(cmd)
```
palemtnrider, line 428: Missing the params.DryRun check - so in dry-run mode we will try and create the source but the secret won't be there.

pkg/cmdimpl/add.go
```diff
@@ -349,134 +270,276 @@ func Add(args []string, allParams AddParamSet, deps *AddDependencies) error {
...
+}
 
-	fmt.Fprintf(shims.Stdout(), "Applying wego platform resources...\n")
-	if !params.DryRun {
-		// Install Source and Kustomize controllers, and CRD for application (may already be present)
-		wegoSource, err := generateWegoSourceManifest()
-		if err != nil {
-			return wrapError(err, "could not generate wego source manifest")
-		}
+func generateSource(repoName, repoUrl string, sourceType SourceType) ([]byte, error) {
```
palemtnrider, line 395: We aren't using sourceType in the function.  (I suspect it's because we have a different PR that will add the helm support)

palemtnrider (approved):
LGTM - Just a handful of nits you can choose to make or not.

pkg/cmdimpl/add.go
jrryjcksn: Yes, but that's correct. The message is saying we failed to set up the app; the source is just where the automation for the target should point.

pkg/cmdimpl/add.go
jrryjcksn: We don't actually create the source; we just create the manifest and return it.

PR #261, opened 2021-06-07
Using flux install without export

luizbafilho:
**What changed?**

Using install as-is from flux.

the `dry-run` option is coming in another pr.

**Why?**

**How did you test it?**

**Release notes**

**Documentation Changes**


josecordaz (approved):
LGTM

PR #263, opened 2021-06-07
helm repo source

J-Thompson12:
Added ability to add helm manifest through helm repo. 

Unit tests and tested myself

Command looks like `wego app add . --url=https://grafana.github.io/helm-charts --name=graf1  --chart=loki`

### DIFF
--- a/cmd/wego/add/cmd.go
+++ b/cmd/wego/add/cmd.go
@@ -37,7 +37,7 @@ func init() {
 	Cmd.Flags().StringVar(&params.Path, "path", "./", "Path of files within git repository")
 	Cmd.Flags().StringVar(&params.Branch, "branch", "main", "Branch to watch within git repository")
 	Cmd.Flags().StringVar(&params.DeploymentType, "deployment-type", "kustomize", "deployment type [kustomize, helm]")
-	Cmd.Flags().StringVar(&params.Chart, "chart", "", "Specify chart for --helm-repository source")
+	Cmd.Flags().StringVar(&params.Chart, "chart", "", "Specify chart for helm source")
 	Cmd.Flags().StringVar(&params.PrivateKey, "private-key", filepath.Join(os.Getenv("HOME"), ".ssh", "id_rsa"), "Private key that provides access to git repository")
 	Cmd.Flags().BoolVar(&params.DryRun, "dry-run", false, "If set, 'wego add' will not make any changes to the system; it will just display the actions that would have been taken")
 	Cmd.Flags().BoolVar(&params.IsPrivate, "private", true, "Set access control on the repo")
--- a/cmd/wego/add/cmd.go
+++ b/cmd/wego/add/cmd.go
@@ -37,6 +37,8 @@ func init() {
 	Cmd.Flags().StringVar(&params.Path, "path", "./", "Path of files within git repository")
 	Cmd.Flags().StringVar(&params.Branch, "branch", "main", "Branch to watch within git repository")
 	Cmd.Flags().StringVar(&params.DeploymentType, "deployment-type", "kustomize", "deployment type [kustomize, helm]")
+	Cmd.Flags().StringVar(&params.HelmRepository, "helm-repository", "", "url of remote helm repository")
+	Cmd.Flags().StringVar(&params.Chart, "chart", "", "Specify chart for --helm-repository source")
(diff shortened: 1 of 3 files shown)


pkg/cmdimpl/add.go
```diff
@@ -235,6 +237,26 @@ func generateSourceManifest() ([]byte, error) {
 	return sourceManifest, nil
 }
 
+func generateSourceManifestHelm() ([]byte, error) {
+	if params.Chart == "" {
+		return nil, fmt.Errorf("chart needs to be set")
+	}
```
J-Thompson12, line 243: I added this check in the source because i figured we didnt want to make the source only to have the manifest fail. Pretty sure there might be a better way to check if the chart is valid but I figured i would just get it done for now.

pkg/cmdimpl/add.go
```diff
@@ -433,15 +471,27 @@ func Add(args []string, allParams AddParamSet, deps *AddDependencies) error {
...
-		if err != nil {
-			return wrapError(err, "could not generate source manifest")
+		var source []byte
+		if params.HelmRepository != "" {
+			source, err = generateSourceManifestHelm()
+			if err != nil {
+				return wrapError(err, "could not generate source manifest")
+			}
+		} else {
+			source, err = generateSourceManifestGit()
```
josecordaz, line 481: huge favor, could you please help me make this function to use the new Chart field instead of Path. When I created it, I used `Path` but it seems more accurate to use `Chart` now.

pkg/cmdimpl/add.go
```diff
@@ -433,15 +471,27 @@ func Add(args []string, allParams AddParamSet, deps *AddDependencies) error {
...
 
 		// Create flux custom resources for new repo being added
-		source, err := generateSourceManifest()
-		if err != nil {
-			return wrapError(err, "could not generate source manifest")
+		var source []byte
+		if params.HelmRepository != "" {
+			source, err = generateSourceManifestHelm()
+			if err != nil {
+				return wrapError(err, "could not generate source manifest")
```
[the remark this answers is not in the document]
J-Thompson12: I mean the error that happens depends on which path you choose. If you are adding a helm repo then you know the error was in helm.

pkg/cmdimpl/add.go
J-Thompson12: I talked with @jrryjcksn  at least on whether or not i should use path. We agreed for helm-repository chart made sense but I didnt really ask about using chart for github.

jrryjcksn: Looks like Flux asks the user for "chart" in either case so we should probably follow that lead.

cmd/wego/add/cmd.go
```diff
@@ -33,10 +33,11 @@ var Cmd = &cobra.Command{
 func init() {
 	Cmd.Flags().StringVar(&params.Owner, "owner", "", "Owner of remote git repository")
 	Cmd.Flags().StringVar(&params.Name, "name", "", "Name of remote git repository")
-	Cmd.Flags().StringVar(&params.Url, "url", "", "URL of remote git repository")
+	Cmd.Flags().StringVar(&params.Url, "url", "", "URL of remote repository")
 	Cmd.Flags().StringVar(&params.Path, "path", "./", "Path of files within git repository")
 	Cmd.Flags().StringVar(&params.Branch, "branch", "main", "Branch to watch within git repository")
 	Cmd.Flags().StringVar(&params.DeploymentType, "deployment-type", "kustomize", "deployment type [kustomize, helm]")
+	Cmd.Flags().StringVar(&params.Chart, "chart", "", "Specify chart for --helm-repository source")
```
jrryjcksn, line 40: Looks like this description is no longer accurate (at least the "--" part)

pkg/cmdimpl/add_test.go
```diff
@@ -192,15 +192,70 @@ var _ = Describe("Test helm manifest", func() {
...
+			--namespace=wego-system `
+
+		fakeHandler := &fluxopsfakes.FakeFluxHandler{
+			HandleStub: func(args string) ([]byte, error) {
+				Expect(args).Should(Equal(expected))
+				return []byte("foo"), nil
+			},
+		}
+
+		fluxops.SetFluxHandler(fakeHandler)
```
jrryjcksn, line 247: I made a change in the other helm test in my stuff to not set the fake handler directly (using `WithOverrides` instead) because it leaves the flux handler broken for other tests.

jrryjcksn (approved):
A couple of nits, otherwise LGTM

PR #269, opened 2021-06-08
Cluster reset

iahmad9:
**What changed?**
Try repeatable way of cleaning up the resources on wego flux uninstall.

**Why?**
To make it possible to run tests against a standing cluster without recreating.

### DIFF
--- a/test/acceptance/test/add_tests.go
+++ b/test/acceptance/test/add_tests.go
@@ -28,12 +28,15 @@ var _ = Describe("Weave GitOps Add Tests", func() {
 		var repoAbsolutePath string
 		private := true
 		appManifestFilePath := "./data/nginx.yaml"
+		workloadName := "nginx"
+		workloadNamespace := "my-nginx"
 		defaultSshKeyPath := os.Getenv("HOME") + "/.ssh/id_rsa"
 		addCommand := "app add . "
 		appRepoName := "wego-test-app-" + RandString(8)
 		appName := appRepoName
 
 		defer deleteRepo(appRepoName)
+		defer deleteWorkload(workloadName, workloadNamespace)
 
 		By("And application repo does not already exist", func() {
 			deleteRepo(appRepoName)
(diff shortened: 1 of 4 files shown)


test/acceptance/test/add_tests.go
```diff
@@ -260,7 +273,7 @@ var _ = Describe("Weave GitOps Add Tests", func() {
 		By("Then I should see relevant message in the console", func() {
 
 			Eventually(addCommandOutput).Should(MatchRegexp(`Checking cluster status[.?]+ (Unknown|Unmodified)`))
-			Eventually(addCommandErr).Should(ContainSubstring("WeGO not installed... exiting"))
+			Eventually(addCommandErr).Should(MatchRegexp(`WeGO.*... exiting`))
 		})
 	})
```
rokshana-b, line 279: Shouldn't we verify that wego is exiting _because_ it's not installed?

test/acceptance/test/utils.go
```diff
@@ -203,6 +240,19 @@ func deleteRepo(appRepoName string) {
...
+	_ = waitForNamespaceToTerminate(workloadNamespace, INSTALL_RESET_TIMEOUT)
+}
+
+func runCommandAndReturnOutput(commandToRun string) (stdOut string, stdErr string) {
+	command := exec.Command("sh", "-c", commandToRun)
+	session, _ := gexec.Start(command, GinkgoWriter, GinkgoWriter)
+	Eventually(session).Should(gexec.Exit())
+	return string(session.Wait().Out.Contents()), string(session.Wait().Err.Contents())
+}
+
```
rokshana-b, line 255: I had added a similar function in the utils file. I'll remove mine. 👍

rokshana-b (approved):

test/acceptance/test/add_tests.go
iahmad9: You are right but at the moment it's not very clear what Unknown status means and the difference between Unmodified & Uknown. Could you please verify that with @jrryjcksn

PR #273, opened 2021-06-08
expand ~ using UserHomeDir

jrryjcksn:
**What changed?**
We expand `~/` at the front of a key path to the user's home directory

**Why?**
User feedback

**How did you test it?**
Tested locally


josecordaz (approved):
LGTM

PR #274, opened 2021-06-08
Upload deploy key

josecordaz:
It now uploads the deployment key to the provider

This was done to fix a security issue

To verify I went to the repo settings in github.com and looked the deployment key under Deploy keys tab


jrryjcksn (approved):
LGTM but I'd prefer we didn't have a bunch of blank lines in the imports (other than separating the go builtins from the github packages)

josecordaz:
> LGTM but I'd prefer we didn't have a bunch of blank lines in the imports (other than separating the go builtins from the github packages)

Goland IDE does that automaticaly, I need to fix the configuration.

jrryjcksn (approved):
\o/ :tada:

Issue #267, opened 2021-06-08
wego app add committed all my repo files


jpellizzari:
**Describe the bug**
I would expect the `wego app add` to commit everything in the `.wego` directory, but not the rest of my repo. 

In this particular case, I had run `ssh-keygen` in the wego directly (incorrectly) to get a fresh ssh key. After running the `wego app add` command successfully, GitHub emailed me to let me know I had added an SSH key to an open repository. 

I don't expect `wego` to save me from my own stupidity in this case (regarding ssh keys), but should it commit my entire repo, or just the `.wego` folder?

**Environment**
 - wego: 0.0.4-rc0

**To Reproduce**
1. Following this Getting Started: https://docs.gitops.weave.works/docs/getting-started/
2. `touch somefile.txt`
2. `wego app add . --path=deploy/webapp --branch=master --private-key=<some key>`

**Expected behavior**
The `.wego` folder is the only thing that gets committed.

jpellizzari:
@jrryjcksn is this already fixed?

jpellizzari:
Closed by #277

PR #277, opened 2021-06-08
Don't add files in user repo to commit when committing .wego contents

jrryjcksn:
**What changed?**
We no longer gather up all files in a user's repo when committing .wego contents. This isn't the final solution. It wasn't clear how to avoid committing files already in the index. With this change, though, only files in the index will be picked up.

**Why?**
We shouldn't update the user's repository behind her back

**How did you test it?**
Tested locally and new gogit unit test

**Release notes**

**Documentation Changes**


luizbafilho (approved):

jpellizzari:
closes #267

PR #278, opened 2021-06-08
Add UI layout and nav with fake data

jpellizzari:
Adds the first UI elements to Weave GitOps with some fake data:
Screenshot from 2021-06-08 13-56-42

Screenshot from 2021-06-08 13-56-49

Also creates a new workflow for frontend only stuff cc @luizbafilho 

Adds some linting and testing to CI as well.


package.json
```diff
@@ -5,7 +5,10 @@
   "scripts": {
     "preinstall": "npx npm-force-resolutions",
     "build": "parcel build --no-source-maps ui/index.html --dist-dir cmd/ui/dist",
-    "start": "parcel serve --port 4567 ui/index.html"
+    "start": "parcel serve --port 4567 ui/index.html",
```
AlinaGoaga, line 8: Liking parcel, `Built in 3.76s`!

ui/components/ErrorBoundary.tsx
```diff
@@ -0,0 +1,37 @@
+import * as React from "react";
+import Page from "./Page";
+
+export default class ErrorBoundary extends React.Component<
```
AlinaGoaga, line 4: Oh cool! I've not used this before, just familiarised myself now with the docs on the react website. Perhaps I'd rewrite to follow the latest syntax - not a big deal though

jpellizzari: As I understand it, this is the latest syntax.

https://reactjs.org/docs/error-boundaries.html

I don't think we can do a function component for this, if that's what you mean.

ui/components/Page.tsx
```diff
@@ -0,0 +1,184 @@
...
+                    to={formatURL(n.value)}
+                  />
+                ))}
+              </Tabs>
+            </NavContent>
+          </NavContainer>
+          <ContentContainer>
+            <Content>
+              <TitleBar>
+                <Breadcrumbs>
```
AlinaGoaga, line 163: Interesting, I like the look and feel of having this in the Content section! I think @foot suggested something similar when we were having a look at Figma. I had to come up with a workaround to have it up in the appbar as I don't have access to that sort of information at that point (the appbar is outside the Switch). So I am "pretending" it's the app bar but in fact it's part of the main page (I divided the right side of the Drawer in the section header - containing the breadcrumbs - and the page content).

ui/components/Flex.tsx
```diff
@@ -0,0 +1,32 @@
...
+  flex-direction: ${(props) => (props.column ? "column" : "row")};
+  align-items: ${(props) => (props.align ? "center" : "start")};
+  ${({ between }) => between && "justify-content: space-between"};
+  ${({ center }) => center && "justify-content: center"};
+  ${({ wide }) => wide && "width: 100%"};
+  ${({ wrap }) => wrap && "flex-wrap: wrap"};
+  ${({ end }) => end && "justify-content: flex-end"};
+`;
+
+class Flex extends React.PureComponent<Props> {
```
AlinaGoaga, line 25: This is great, I forgot to give it a go after you sent it to me *face palm*! Could we have it added to the `ui-components` library do you think?

ui/lib/types.ts
```diff
@@ -0,0 +1,4 @@
+export enum PageRoute {
```
AlinaGoaga, line 1: My preference is for "RESTy" stuff :p but again no big deal, we can adopt this pattern if you prefer it

ui/lib/test-utils.tsx
```diff
@@ -0,0 +1,24 @@
...
+    <MuiThemeProvider theme={muiTheme}>
+      <ThemeProvider theme={theme}>{element}</ThemeProvider>
+    </MuiThemeProvider>
+  );
+}
+
+export function withContext(TestComponent, url: string) {
+  const history = createMemoryHistory();
+  history.push(url);
+  return (
```
AlinaGoaga, line 19: Oh cool, could we perhaps use Browser Router here as we do in the main app? I am thinking if we did that we wouldn't need to create the history etc.

ui/components/ErrorBoundary.tsx
AlinaGoaga: Yup exactly that's what I was thinking. Oki np, good to know

ui/lib/test-utils.tsx
AlinaGoaga: Yeah I see your point. This helped me https://testing-library.com/docs/example-react-router/#reducing-boilerplate (second point). I used to use React Testing Library also. Perfectly fine as is, I was just curious :)

AlinaGoaga (approved):
Looking great!

PR #279, opened 2021-06-08
fix upstream helm support

jrryjcksn:
**What changed?**
We now generate the correct source type for upstream helm and we disallow using it in the user repo mode.

**Why?**
Upstream helm did not work due to an oversight when merging in change.

**How did you test it?**
Tested locally

**Release notes**

**Documentation Changes**

### DIFF
--- a/pkg/cmdimpl/add.go
+++ b/pkg/cmdimpl/add.go
@@ -88,6 +88,11 @@ func updateParametersIfNecessary(gitClient git.Git) error {
 		params.DeploymentType = string(DeployTypeHelm)
 	}
 
+	if params.AppConfigUrl == "" && params.SourceType != string(SourceTypeGit) {
+		return fmt.Errorf("Cannot create .wego directory in helm repository:\n" +
+			"  you must either use --app-config-url=none or --appconfig-url=<url of external git repo>")
+	}
+
 	if params.Name == "" {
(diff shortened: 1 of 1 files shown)


pkg/cmdimpl/add.go
```diff
@@ -88,6 +88,11 @@ func updateParametersIfNecessary(gitClient git.Git) error {
 		params.DeploymentType = string(DeployTypeHelm)
 	}
 
+	if params.AppConfigUrl == "" && params.SourceType != string(SourceTypeGit) {
```
palemtnrider, line 91: We should use `string(ConfigTypeUserRepo)` instead of the empty string.

pkg/cmdimpl/add.go
```diff
@@ -88,6 +88,11 @@ func updateParametersIfNecessary(gitClient git.Git) error {
...
+		return fmt.Errorf("Cannot create .wego directory in helm repository:\n" +
```
palemtnrider, line 92: Nit: error message should start with a lower case letter (go style guide)

palemtnrider (approved):
Other than a minor change and a nit, LGTM

pkg/cmdimpl/add.go
jrryjcksn: Good catch

PR #281, opened 2021-06-09
Look in well known locations for ssh key file or allow the user to specify it's path on the command line

jrryjcksn:
**What changed?**
We now use a generated deploy key for flux but allow specification of an ssh key to talk to git repositories over ssh. 

**Why?**
By default, darwin does not store the user's default ssh key in the ssh-agent so we need to allow it to be specified.

**How did you test it?**
It's tested every time we run `wego app add` (particularly on darwin)

### DIFF
--- a/cmd/wego/add/cmd.go
+++ b/cmd/wego/add/cmd.go
@@ -6,14 +6,16 @@ package add
 import (
 	"fmt"
 	"os"
+	"path/filepath"
 	"strings"
 
-	"github.com/go-git/go-git/v5/plumbing/transport/http"
+	"github.com/go-git/go-git/v5/plumbing/transport/ssh"
 	"github.com/lithammer/dedent"
 	"github.com/spf13/cobra"
 	"github.com/weaveworks/weave-gitops/pkg/cmdimpl"
 	"github.com/weaveworks/weave-gitops/pkg/git"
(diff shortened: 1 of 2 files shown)


cmd/wego/add/cmd.go
```diff
@@ -36,17 +38,28 @@ func init() {
...
-		Password: os.Getenv("GITHUB_TOKEN"),
-	})
+	if strings.HasPrefix(params.PrivateKey, "~/") {
+		dir := getHomeDir()
+		params.PrivateKey = filepath.Join(dir, params.PrivateKey[2:])
+	} else if params.PrivateKey == "" {
+		params.PrivateKey = findPrivateKeyFile()
+	}
+
+	authMethod, err := ssh.NewPublicKeysFromFile("git", params.PrivateKey, params.PrivateKeyPass)
```
palemtnrider, line 56: We should add the private-key-pass to the CLI flags.

palemtnrider (approved):
LGTM.

cmd/wego/add/cmd.go
jrryjcksn: If we decide passing the password on the command line is okay, we can expose it. It has long been standard practice to avoid that since it shows up in /proc. If the assumptions around that have changed we can revisit this.

PR #284, opened 2021-06-09
correctly untar flux archive and stop overwriting flux binary in make

jrryjcksn:
**What changed?**
Flux binary was being overwritten

**Why?**
To fix other architectures

**How did you test it?**
Building a release

**Release notes**

**Documentation Changes**


palemtnrider (approved):
LGTM.  Thanks for fixing this.

PR #301, opened 2021-06-09
Automate test case for 'wego add --url' flag

rokshana-b:
**What changed?**
- [ ] Added test case for `wego add app --url`
- [ ] Added `--branch` flag to `wego add app --dry-run` test case

**Why?**
To add test coverage for additional WeGO features

### DIFF
--- a/test/acceptance/test/add_tests.go
+++ b/test/acceptance/test/add_tests.go
@@ -283,13 +283,13 @@ var _ = Describe("Weave GitOps Add Tests", func() {
 		var repoAbsolutePath string
 		var session *gexec.Session
 		private := true
-		branchName := "ci-test-branch"
+		branchName := "test-branch"
 		appManifestFilePath := "./data/nginx.yaml"
 		defaultSshKeyPath := os.Getenv("HOME") + "/.ssh/id_rsa"
 		appRepoName := "wego-test-app-" + RandString(8)
 		workloadName := "nginx"
 		workloadNamespace := "my-nginx"
-		url := os.Getenv("GITHUB_ORG") + "/" + appRepoName + ".git"
+		url := url := "ssh://git@github.com/" + os.Getenv("GITHUB_ORG") + "/" + appRepoName + ".git"
 		addCommand := "app add . --url=" + url + " --branch=" + branchName + " --dry-run"
 		appName := appRepoName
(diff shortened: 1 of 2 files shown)


josecordaz (approved):
LGTM

jrryjcksn (approved):
LGTM. There's also a new app add mode (upstream helm repository) that we can only test with `--url`. So, this will be a good starting place for that too.

test/acceptance/test/add_tests.go
```diff
@@ -279,18 +279,18 @@ var _ = Describe("Weave GitOps Add Tests", func() {
...
 		var session *gexec.Session
 		private := true
-		branchName := "main"
+		branchName := "test-branch"
 		appManifestFilePath := "./data/nginx.yaml"
 		defaultSshKeyPath := os.Getenv("HOME") + "/.ssh/id_rsa"
 		appRepoName := "wego-test-app-" + RandString(8)
 		workloadName := "nginx"
 		workloadNamespace := "my-nginx"
 		url := "ssh://git@github.com/weaveworks-gitops-test/" + appRepoName + ".git"
```
iahmad9, line 292: Should not hardcode the org part, instead should fetch from os.Getenv("GITHUB_ORG") environment variable.

test/acceptance/test/add_tests.go
```diff
@@ -330,4 +334,52 @@ var _ = Describe("Weave GitOps Add Tests", func() {
 			verifyWegoAddCommandWithDryRun(appRepoName, WEGO_DEFAULT_NAMESPACE)
 		})
 	})
+
+	// Eventually this test run will include all the remaining un-automated `wego app add` flags.
+	It("SmokeTest - Verify 'wego app add' works when --url flag is specified", func() {
+		var repoAbsolutePath string
+		private := true
+		appRepoName := "wego-test-app-" + RandString(8)
+		url := "ssh://git@github.com/weaveworks-gitops-test/" + appRepoName + ".git"
```
iahmad9, line 343: Same as above => Should not hardcode the org part, instead should fetch from os.Getenv("GITHUB_ORG") environment variable.

iahmad9 (changes requested):
Please change the hardcode GitHub org to env variable.

PR #302, opened 2021-06-09
Adding git status check before doing add

luizbafilho:
**What changed?**
Adding git status check before doing proceeding with the app add

**Why?**
Avoid committing user's changes.

**How did you test it?**
change something in the app repo and try to add it.

**Release notes**

**Documentation Changes**


jrryjcksn (changes requested):
This looks okay to me but I'd really like to not have the coverage numbers decrease. We're not trying to rush anything out just now.

josecordaz (approved):
LGTM

luizbafilho:
The code I added was covered unless you want us to start testing the errors branches.

jrryjcksn:
Well, the whole point of this is to disallow the add if the repo isn't clean and you set the status stub to always return true so we're not actually testing the intended behavior. The error path is the whole point.

In any case, I'm not trying to be overly pedantic but the coverage is already out of range and making it worse is something I think we can/should avoid; otherwise there's no point to having a coverage threshold at all.

luizbafilho:
I see so the point is not bumping the coverage is covering a case that was not covered, and I ignored it; with that, I agree with you. 

But since you've mentioned, yeah, I now agree with Jordan, this coverage threshold at this phase of the project where things change all the time is contra-productive; we lose quite often a lot of time trying to add coverage to things that will very likely to change in the near future.

We already have good coverage and some guarantee that a good part of the code already works with acceptance tests. I see no need, at the moment, to aim to keep the coverage so high.

iahmad9:
> I see so the point is not bumping the coverage is covering a case that was not covered, and I ignored it; with that, I agree with you.
> 
> But since you've mentioned, yeah, I now agree with Jordan, this coverage threshold at this phase of the project where things change all the time is contra-productive; we lose quite often a lot of time trying to add coverage to things that will very likely to change in the near future.
> 
> We already have good coverage and some guarantee that a good part of the code already works with acceptance tests. I see no need, at the moment, to aim to keep the coverage so high.

My take on this is that we should consider the unit testing part of the implementation which means if the code changes, the unit test will change with it. So in that sense, we should avoid missing the unit tests otherwise if we wait for the code to be matured before writing the unit tests then it's already too late and these will become technical debt. Which in my experience is not easy to fix retrospectively for the number of obvious reasons.

jrryjcksn (approved):
LGTM

pkg/cmdimpl/add_test.go
```diff
@@ -502,4 +500,58 @@ var _ = Describe("Wet Run Add Test", func() {
 				status.Override(shandler))
 		})
 	})
+
+	It("fails when the repository is in a dirty state", func() {
```
bigkevmcd, line 504: To me this feels a bit like http://xunitpatterns.com/Obscure%20Test.html

It's really not clear what it's doing?

Could we maybe extract some functions to clarify what's going on?

luizbafilho: this is not relevant anymore, I'll move this PR out to the new packages that were created after the refactoring.

palemtnrider:
@luizbafilho is this PR still relevant?  If so, can you update with main and address the comments so we can look at merging and closing it?  If it is no longer relevant, please close.

PR #303, opened 2021-06-10
Make install should put binaries in "${GOPATH}/bin"

D3nn:
**What changed?**
Makefile target "install" sends wego binaries to ${GOPATH}/bin instead of .../weave-gitops/bin/

**Why?**
Original binary output wasn't in most developers PATH

**How did you test it?**
Ran `make install` and verified that binaries are correctly being installed in ${GOPATH}

**Release notes**
Users who've previously used `make install` should migrate binaries from old path to ${GOPATH)

**Documentation Changes**
There are currently no documents covering this.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -34,6 +34,10 @@ bin:
 # Build wego binary
 wego: dependencies bin
 
+# Install binaries to GOPATH
+install: bin bin/$(BINARY_NAME)_ui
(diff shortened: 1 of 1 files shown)


D3nn:
This is entirely Makefile updates, so no code coverage changes expected.

josecordaz (approved):
LGTM

PR #309, opened 2021-06-11
Add applications API server

jpellizzari:
Add the initial API code generation and server implementation. A fake `ListApplications` call is implemented for now. There is an example test to demonstrate how testing could work.

Also adds frontend code to the coverage metrics, as well as ignores coverage on generated files. Some of the github actions were converted to Makefile targets to make it easier to troubleshoot locally. @J-Thompson12 would you mind taking a look at that portion?


jpellizzari:
Closes #321

cmd/wego-server/app/app.go
```diff
@@ -0,0 +1,61 @@
...
+
+	"github.com/grpc-ecosystem/grpc-gateway/v2/runtime"
+	log "github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+	pb "github.com/weaveworks/weave-gitops/pkg/api/applications"
+	"github.com/weaveworks/weave-gitops/pkg/server"
+)
+
+func NewAPIServerCommand() *cobra.Command {
+	cmd := &cobra.Command{
```
josecordaz, line 16: Tiny suggestion would be to move this to the cmd folder for consistency and have all commands in the same place.

josecordaz (approved):
LGTM

api/applications/applications.proto
```diff
@@ -0,0 +1,28 @@
...
+package wego_server.v1;
+
+option go_package = "github.com/weaveworks/weave-gitops/pkg/protos/applications";
+
+import "google/api/annotations.proto";
+
+service Applications {
+    rpc ListApplications(ListApplicationsReq) returns (ListApplicationsRes) {
+        option (google.api.http) = {
+            get : "/v1/applications"
```
bigkevmcd, line 13: Following the [Google naming conventions](https://cloud.google.com/apis/design/naming_convention#request_and_response_messages) how about we change these to `ListApplicationsRequest` and `ListApplicationsResponse` ?

api/applications/applications.proto
```diff
@@ -0,0 +1,28 @@
+
+syntax = "proto3";
+
+package wego_server.v1;
+
+option go_package = "github.com/weaveworks/weave-gitops/pkg/protos/applications";
+
+import "google/api/annotations.proto";
```
bigkevmcd, line 8: It'd be nice to bring in the openapi annotations...how about something like...

```protobuf
import "protoc-gen-openapiv2/options/annotations.proto";

option (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_swagger) = {
  info: {
    title: "WeGo Applications API",
    version: "0.1";
    description: "<insert something here>";
  };
  consumes: "application/json";
  produces: "application/json";
};
```

This gets output in the resulting swagger.json

api/applications/applications.proto
bigkevmcd: Also, let's put comments above the methods?

Something like...

```protobuf
service Applications {
    /**
     * ListApplications returns the list of WeGo applications that the authenticated user has access to.
     */
    rpc ListApplications(ListApplicationsRequest) returns (ListApplicationsResponse) {
        option (google.api.http) = {
            get : "/v1/applications"
          };
    }
}
```

These are output in the Swagger, and in the generated documentation.

api/applications/applications.proto
```diff
@@ -0,0 +1,28 @@
...
+message Application {
+    string name = 1;
+}
+
+message ListApplicationsReq {
+
+}
+
+message ListApplicationsRes {
+   repeated Application applications = 1;
```
bigkevmcd, line 27: Comments on fields are added to the Swagger _and_ to the generated HTML documentation.

cmd/ui/main.go
```diff
@@ -2,49 +2,51 @@ package main
...
 
+var log = logrus.New()
+
 func main() {
-	log := logrus.New()
+	apiUrlFlag := flag.String("api-url", "http://localhost:8000", "The URL for the applications API server")
+
 	mux := http.NewServeMux()
 
 	mux.Handle("/health/", http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
bigkevmcd, line 23: Should this perhaps `HEAD` the `api-url` to check that the api-url is healthy?

Or do we consider that if the proxy is available at all, then the system is healthy, irrespective of it's primary dependency?

cmd/ui/main.go
```diff
@@ -2,49 +2,51 @@ package main
...
 		w.WriteHeader(http.StatusOK)
```
bigkevmcd, line 24: You don't need to do this, it's the default value for responses if there's no other status.

https://golang.org/src/net/http/server.go#L1600

cmd/ui/main.go
```diff
@@ -2,49 +2,51 @@ package main
...
 	}))
 
 	assetFS := getAssets()
 	assetHandler := http.FileServer(http.FS(assetFS))
 	redirector := createRedirector(assetFS, log)
 
+	apiUrl, err := url.Parse(*apiUrlFlag)
+
+	if err != nil {
+		log.Errorf("could not parse proxy url")
```
bigkevmcd, line 39: You're using `Errorf` with no formatting strings, but, how about including the actual error message in the output, which might help a user diagnose the issue?

Something like `log.Errorf("could not parse proxy url: %s", err)`

cmd/ui/main.go
```diff
@@ -2,49 +2,51 @@ package main
...
+		os.Exit(1)
+		return
```
bigkevmcd, line 41: You don't need to `return` here, `os.Exit()` will terminate the process...

cmd/ui/main.go
```diff
@@ -116,3 +118,14 @@ func createRedirector(fsys fs.FS, log logrus.FieldLogger) http.HandlerFunc {
 		}
 	}
 }
+
+func proxy(u *url.URL, w http.ResponseWriter, r *http.Request) {
+	proxy := httputil.NewSingleHostReverseProxy(u)
```
bigkevmcd, line 123: You should be creating the proxy once, and reusing it, rather than creating a new one for every request.

pkg/server/server_test.go
```diff
@@ -0,0 +1,19 @@
...
+	"github.com/weaveworks/weave-gitops/pkg/api/applications"
+)
+
+var _ = Describe("ApplicationsServer", func() {
+	It("AddApplication", func() {
+		res, err := client.ListApplications(context.Background(), &applications.ListApplicationsReq{})
+
+		Expect(err).NotTo(HaveOccurred())
+
+		Expect(len(res.Applications)).To(Equal(3))
```
bigkevmcd, line 17: nitpick: you could change this to...

```go
Expect(res.Applications).To(HaveLen(3))
```

cmd/ui/main.go
jpellizzari: Strange. So what is the second argument to `mux.Handle`?

cmd/ui/main.go
jpellizzari: In my mind, the health check controls routing of traffic in k8s (once healthy). I think the `ui` server is healthy even if the api-url is not. It can return the UI that may render errors if the API is not healthy.

Returning a UI with an error message seems better than not routing traffic to this service.

cmd/ui/main.go
bigkevmcd: Sorry, you don't need to `w.WriteHeader(http.StatusOK)`

cmd/ui/main.go
```diff
@@ -2,49 +2,51 @@ package main
...
+		log.Errorf("could not parse proxy url")
+		os.Exit(1)
+		return
+	}
+
+	log.Infof("api proxy url set to %s", apiUrl.String())
+
+	mux.Handle("/v1/", http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		proxy(apiUrl, w, r)
+	}))
```
foot, line 48: Can we call
```golang
if err := pb.RegisterApplicationsHandlerServer(ctx, mux, server.NewApplicationsServer()); err != nil {
  return fmt.Errorf("could not register application: %w", err)
}
```
here to mount all the `/v1` routes here onto `mux` and then completely remove `cmd/wego-server/app/app.go`?

I'm not 100% sure if it would work but would be cool if so.

jpellizzari: @foot How would we specify that only `/v1` routes go to the `server.NewApplicationServer()`? There are static assets that need to be served by this runtime on `/`, ie `index.html` and `main.js`. 

This was my original complaint around `gRPC`. I can mux the API server into the `ui` runtime easily with `twirp`, but its not obvious how to do it with `gRPC`.

jpellizzari: Also it should be noted that `wego-server` is the API (standalone) runtime, the file you commented in is the UI server.

cmd/ui/main.go
jpellizzari: Just an empty handler func then?

cmd/ui/main.go
foot: > @foot How would we specify that only /v1 routes go to the server.NewApplicationServer()?

I think `RegisterApplicationsHandlerServer` only adds the routes in the protobuf file.
https://github.com/weaveworks/weave-gitops/pull/309/files#diff-4f1ecbad04de7fc879a574bf0f86a2d8b2558607c54c9fe2a938c103d020238bR56 on

> Also it should be noted that wego-server is the API (standalone) runtime, the file you commented in is the UI server.

Ah hah! Good point, 2 servers. 👌

api/applications/applications.proto
jpellizzari: Added

api/applications/applications.proto
jpellizzari: Added a field comment

api/applications/applications.proto
jpellizzari: Added comments and changed the req/res type names

cmd/ui/main.go
jpellizzari: Changed to create the proxy in the `main` block

cmd/ui/main.go
jpellizzari: > I think RegisterApplicationsHandlerServer only adds the routes in the protobuf file.

It only accepts a `runtime.ServeMux`, which I cannot append to. I need it to accept a `http.ServeMux` to add `/metrics` or `/health` (for example).

jpellizzari:
@foot @bigkevmcd Updated to embed the API server in the UI runtime, rather than proxying. I assume we need a combined (ui + api) and a standalone runtime (api only)?

cmd/ui/main.go
jpellizzari: No longer relevant, using an "embedded" API server approach instead of proxying

luizbafilho (approved):
looks great

.github/workflows/test.yml
```diff
@@ -201,4 +199,5 @@ jobs:
         export PATH=${PATH}:`go env GOPATH`/bin
         export WEGO_BIN_PATH=$(pwd)/bin/wego
         go get -u github.com/onsi/ginkgo/ginkgo
```
jrryjcksn, line 201: Curious why you replaced `go get -u github.com/onsi/ginkgo/ginkgo` with `go get github.com/onsi/ginkgo` above but not here :thinking:

Makefile
```diff
@@ -81,3 +78,47 @@ ui-test:
 
 ui-audit:
 	npm audit
+
+# Directory for coverage data
+coverage:
```
jrryjcksn, line 83: I think having the `coverage` target do nothing other than create the dir will be confusing. `make coverage` seems like a high-level user command.

Makefile
```diff
@@ -81,3 +78,47 @@ ui-test:
...
+
+proto-deps:
+	@go get \
+		github.com/grpc-ecosystem/grpc-gateway/v2/protoc-gen-grpc-gateway \
+		google.golang.org/protobuf/cmd/protoc-gen-go \
+		google.golang.org/grpc/cmd/protoc-gen-go-grpc \
+		github.com/grpc-ecosystem/grpc-gateway/v2/protoc-gen-openapiv2 \
+		github.com/grpc-ecosystem/protoc-gen-grpc-gateway-ts \
+		github.com/deepmap/oapi-codegen/cmd/oapi-codegen
+	buf beta mod update
```
jrryjcksn, line 114: Where does the `buf` command come from?

cmd/wego-server/main.go
```diff
@@ -0,0 +1,83 @@
...
+	"github.com/grpc-ecosystem/grpc-gateway/v2/runtime"
+	log "github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+	pb "github.com/weaveworks/weave-gitops/pkg/api/applications"
+	"github.com/weaveworks/weave-gitops/pkg/server"
+)
+
+func init() {
+	if os.Getenv("LOG_LEVEL") == "DEBUG" {
+		// Only log the warning severity or above.
```
jrryjcksn, line 20: NIT: "warning" => "debug" in comment

jrryjcksn (approved):
LGTM modulo a few nits

Makefile
jpellizzari: Yeah I thought about that. Mixing "classic" file targets and jobs isn't great. I will see if I can make that clearer or get rid of the target.

Makefile
jpellizzari: Added to README for dev setup

oapi-codegen.config.yaml
```diff
@@ -0,0 +1,5 @@
+output: pkg/openapi-client/wego_server.go
+package: wego_client
+generate:
+  - types
+  - client
```
foot, line 5: Neat! I only had a very quick play w/ this and didn't work immediately, any luck? swagger-codegen was the other I've seen around in grpc-gateway examples etc.

cmd/ui/main.go
```diff
@@ -1,50 +1,43 @@
...
 	assetFS := getAssets()
 	assetHandler := http.FileServer(http.FS(assetFS))
 	redirector := createRedirector(assetFS, log)
 
+	gMux := runtime.NewServeMux()
+	mux.Handle("/v1/", gMux)
+
+	if err := pb.RegisterApplicationsHandlerServer(context.Background(), gMux, server.NewApplicationsServer()); err != nil {
+		log.Fatalf("could not register application: %s", err)
+	}
```
foot, line 39: ⭐ . Sorry I didn't understand the use case properly first. This is neat though being able to mux in the gateway here.

For things like metrics collection etc the proxy model might work better though? Then all api usage reporting is coming from the same container etc.

foot (approved):
lgtm 👌 👏 

The only other things is you could add the `tools.go` trick (https://github.com/weaveworks/wks/blob/master/cmd/capi-server/tools.go) and then `go mod tidy` and all the versions of the protobuf plugins will be versioned in go.mod too.

cmd/ui/main.go
jpellizzari: Hmm yeah good point. The first use case I was optimizing for was the stand-alone binary running outside the cluster. Running inside the cluster will require some thought.

PR #336, opened 2021-06-15
Rearchitecting Add code

luizbafilho:
**What changed?**
This is a major refactor of the `app add` command. It creates a new App Service package, that'll contain the implementation for `app` commands. 

It also creates interfaces for all dependencies the app command needs.

**Note:**
In order to speed up the approval process of this PR, I decided to roll back any UX change, so, now it's behaving as is today. There should be no changes at all, besides possible bugs, introduced in the process.

I can create a new PR only with those UX changes if needed, I'm leaving the description below for historical reasons, but that's not valid anymore.

~As much as possible I haven't change the current behavior, all the 3 modes stay the same and work as current implementation, but as I was doing the development I felt the need to add new flags and make some changes to simplify the "mode selection".~

<details>
<summary>Details</summary>

These were the changes:

**Add in the user repo**
Stays exactly the same, with no changes at all as this is the default behavior, by running the command below the manifests will be generated, applied to the cluster, and committed to the user repo.
```
wego app add .
``` 

**Add without committing manifests**
For this mode, I added a `--commit-manifests` flag, it does the same as the default mode but it only applies the manifests to the cluster and doesn't commit them to any repo.
```
wego app add . --commit-manifests=false
```

**Add with external repo**
This mode had the biggest change, for it I added 3 new flags:
```
--automation-repo string          Repository that will hold the Gitops Automation manifests
--automation-repo-branch string   Repository branch that will hold the Gitops Automation manifests (default "main")
--automation-repo-path string     Repository path that will hold the Gitops Automation manifests (default "./wego")
```

To add an app using an external repo the user can do:

```
wego app add . --automation-repo https://github.com/luizbafilho/wego-automation
```
The above command will read the current directory looking for the app repo URL and save the manifests in `./wego` directory in the `wego-automation` repo.

```
wego app add --url https://github.com/stefanprodan/podinfo --automation-repo https://github.com/luizbafilho/wego-automation
```

This one will create all the manifests save them in `wego-automation` repo, but point them to the repo specified in the `--url`

I did those changes because it felt a bit easier to understand, although I'm not attached to them or any naming I've used, so feel free to question them.

</details>

**Why?**
Testing in the current code is hard, this PR makes it quite easy to inject and define the behavior of the dependencies making it super easy to test our code.

**How did you test it?**

Just use it, try to add apps using the current three modes and check if something breaks.

**Release notes**

**Documentation Changes**


cmd/wego/add/cmd.go
```diff
@@ -58,12 +64,29 @@ func runCmd(cmd *cobra.Command, args []string) {
...
+	}
+
 	gitClient := git.New(authMethod)
+	fluxClient := flux.New()
+	kubeClient := kube.New()
+	gitProviders := gitproviders.New()
 
-	deps := &cmdimpl.AddDependencies{
-		GitClient: gitClient,
+	deps := &app.Dependencies{
```
jpellizzari, line 81: :+1: I like this. It does make me wonder if the `cmdimpl` package is necessary, since it just wraps the library.

pkg/services/app/add.go
```diff
@@ -0,0 +1,503 @@
...
+
+	fmt.Print("done\n\n")
+	fmt.Print("Checking cluster status... ")
+
+	clusterStatus := a.kube.GetClusterStatus()
+	fmt.Printf("%s\n\n", clusterStatus)
+
+	switch clusterStatus {
+	case kube.Unmodified:
+		return errors.New("WeGO not installed... exiting")
```
jpellizzari, line 66: Minor log message improvement: not necessarily exiting in this case.

pkg/services/app/add.go
```diff
@@ -0,0 +1,503 @@
...
+}
+
+func (a *App) Add(params AddParams) error {
+	fmt.Print("Updating parameters from environment... ")
+	params, err := a.updateParametersIfNecessary(params)
+	if err != nil {
+		return errors.Wrap(err, "could not update parameters")
+	}
+
+	fmt.Print("done\n\n")
```
jpellizzari, line 58: We will need to come up with a plan for how to do these `fmt` calls to `stdout`. Maybe we need to add a `log.Writer` to the service deps to capture it as part of a single request.

Does not need to be solved in this PR necessarily.

pkg/services/app/add.go
```diff
@@ -0,0 +1,503 @@
...
+
+	if trimmed != "" {
+		return "ssh://git@github.com/" + trimmed
+	}
+
+	return url
+}
+
+// NOTE: ready to save the targets automation in phase 2
+// func (a *App) writeTargetGoats(basePath string, name string, manifests ...[]byte) error {
```
jpellizzari, line 498: I know its not used yet, but can we avoid adding obfuscation with the term "Goats"?

pkg/services/app/add_test.go
```diff
@@ -0,0 +1,209 @@
...
+	gitProviders *gitprovidersfakes.FakeGitProviderHandler
+
+	appSrv        app.AppService
+	defaultParams app.AddParams
+)
+
+var _ = BeforeEach(func() {
+	gitClient = &gitfakes.FakeGit{}
+	fluxClient = &fluxfakes.FakeFlux{}
+	kubeClient = &kubefakes.FakeKube{
```
jpellizzari, line 27: :partying_face: I dig this. Making the k8s client return what you want is a pain.

pkg/services/app/install.go
```diff
@@ -0,0 +1,9 @@
+package app
+
+import "fmt"
+
+type InstallParams struct{}
+
+func (a *App) Install(params InstallParams) error {
```
jpellizzari, line 7: I sometimes get file fatigue with all my editor tabs. I would propose we keep these in the same file since they share module scope.

jpellizzari (commented):
Structure looks good! Very organized. I cannot speak to the `automation-repo` changes, but I like the overall direction.

pkg/services/app/add.go
luizbafilho: I agree, not a super fan, but struggled to find a good name that means something is `kustomization` or a `helmrelease`

pkg/services/app/add.go
luizbafilho: Agree, I was thinking of adding something similar to flux so we have those nice little icons in the messages, and when calling from the API we could have a more structured format for it.

pkg/services/app/install.go
luizbafilho: sorry I didn't get what you mean.

jpellizzari: I would like to combine everything in `pkg/services/app/` into a single file (except for tests)

luizbafilho:
hi @cdavisafc and @bia tagging you because but I've made some changes to the UX and I'd love your feedback on it. those are mere suggestions since there was a need to input the same path and branch to the external repo, which is different from the path and branch of the app, so I've made a call and picked a name but I'm good with any name.

pkg/services/app/install.go
luizbafilho: ahh, if you open the entire file you'll see is huge, and that's the implementation of a single action. I'm afraid as we add more and more code will be super hard to navigate that.

pkg/services/app/add.go
jpellizzari: The Flux maintainers might have a word to summarize this class of "reconcilable" objects

pkg/services/app/install.go
jpellizzari: Now I am confused. This file (and the others like it) appear to be only 40 lines

jpellizzari: Nevermind I think I understand now

pkg/services/app/add_test.go
```diff
@@ -0,0 +1,209 @@
...
+
+	It("gets the cluster name", func() {
+		err := appSrv.Add(defaultParams)
+		Expect(err).ShouldNot(HaveOccurred())
+
+		Expect(kubeClient.GetClusterNameCallCount()).To(Equal(1))
+	})
+
+	It("creates and deploys a git secret", func() {
+		fluxClient.CreateSecretGitStub = func(s1, s2, s3 string) ([]byte, error) {
```
josecordaz, line 72: I'm curious this seems an overwrite. I would suggest  something like `appSrv.SetFluxClient`or `appSrv.GetFluxClient().CreateSecretGitStub` so the intention is more explicit and easier to tell what the options are. Otherwise I need to lookup and the relation between fluxClient and appSrv. I wonder if there is a way to  go back to the previous logic of `CreateSecretGitStub` if needed in the same `It` 🤔 .

pkg/services/app/add_test.go
```diff
@@ -0,0 +1,209 @@
...
+
+				name, url, namespace := fluxClient.CreateSourceHelmArgsForCall(0)
+				Expect(name).To(Equal("loki"))
+				Expect(url).To(Equal("https://charts.kube-ops.io"))
+				Expect(namespace).To(Equal("wego-system"))
+			})
+		})
+
+		Describe("generates application goat", func() {
+			It("creates a kustomization if deployment type kustomize", func() {
```
josecordaz, line 132: how could we verify that the output for CreateKustomization is right? Is there an easy way to do it with `counterfeiter`

josecordaz (commented):
Overall what I like the most about this object oriented approach is that it would be easier to maintain and looks easier to unit test.

pkg/services/app/app.go
```diff
@@ -0,0 +1,41 @@
...
+	Git          git.Git
+	GitProviders gitproviders.GitProviderHandler
+	Flux         flux.Flux
+	Kube         kube.Kube
+}
+
+type AppService interface {
+	Add(params AddParams) error
+	Status(params StatusParams) error
+	Install(params InstallParams) error
```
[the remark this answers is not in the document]
luizbafilho: 🤦 haha indeed that doesn't make any sense, that should go into another place.

pkg/services/app/add_test.go
luizbafilho: That verification goes in the Flux package tests. I didn't add yet.

pkg/services/app/add_test.go
luizbafilho: I think the effect is the same, you need to understand their relationships in either way.

To go back to the previous logic or change the behavior, just set that again, it's just a pointer.

pkg/services/app/app.go
```diff
@@ -0,0 +1,41 @@
...
+}
+
+type App struct {
+	git          git.Git
+	flux         flux.Flux
+	kube         kube.Kube
+	gitProviders gitproviders.GitProviderHandler
+}
+
+func New(deps *Dependencies) *App {
```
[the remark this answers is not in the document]
luizbafilho: good point

cmd/wego/add/cmd.go
luizbafilho: no, after the migration we can delete that.

pkg/services/app/add.go
luizbafilho: Actually is, I was asked to add that at some point, and just exit there if we can't talk to the cluster.

cmd/wego/add/cmd.go
```diff
@@ -54,26 +57,38 @@ func runCmd(cmd *cobra.Command, args []string) {
...
 	if err != nil {
-		fmt.Fprintf(shims.Stderr(), "failed reading ssh keys: %s\n", err)
-		shims.Exit(1)
+		fmt.Fprintf(os.Stderr, "failed reading ssh keys: %s\n", err)
+		os.Exit(1)
 	}
 
+	if params.Url == "" {
+		if len(args) == 0 {
+			fmt.Fprint(os.Stderr, "no app --url or app location specified")
```
D3nn, line 66: I've noticed with most wego commands (including subcommands), that leaving off parameters/etc at the end spits out an error message with usage output, but not `wego app add`.  For example, the command `wego app status` gives this output:

```
$ wego app status
Error: requires at least 1 arg(s), only received 0
Usage:
  wego app status <app-name> [flags]

Examples:
wego app status podinfo

Flags:
  -h, --help   help for status

Global Flags:
      --namespace string   gitops runtime namespace (default "wego-system")
  -v, --verbose            Enable verbose output

Error: requires at least 1 arg(s), only received 0
```
Should we have `wego app add` follow this convention as well?

D3nn (commented):
Not a blocker, but adding usage output might be worthwhile here as most of the commands I've tried without the final params seemed to work that way.

cmd/wego/add/cmd.go
luizbafilho: good point, I'll check out how to show that.

pkg/services/app/app.go
```diff
@@ -0,0 +1,31 @@
...
+	return &App{
+		git:          git,
+		flux:         flux,
+		kube:         kube,
+		gitProviders: gitProviders,
+	}
+}
+
+// Make sure App implements all the required methods.
+var _ AppService = &App{}
```
jpellizzari, line 31: Seems a bit odd. Something else will complain when you try to use the `App` as an `AppService` if it doesn't implement the interface.

jpellizzari (approved):
Excited about this. LGTM other than one question.

pkg/services/app/app.go
luizbafilho: haha I know, it's a kinda trick the community has, but it will only complain if that is injected or attributed using explicitly somewhere. Like in https://github.com/weaveworks/weave-gitops/pull/336/files#diff-bc17604d5b75b57fd65e69e1996c5c225005be132e891dd100816f4b66a74421R79, unless I set like `var appService AppService = app.New(gitClient, fluxClient, kubeClient, gitProviders)` there won't be complaints about. 

And since using implicit types is the norm, people usually do that to have to guarantee implementation.

pkg/flux/flux.go
```diff
@@ -0,0 +1,169 @@
...
+		return out, fmt.Errorf("failed to create kustomization: %w", err)
+	}
+
+	return out, nil
+}
+
+func (f *FluxClient) CreateHelmReleaseGitRepository(name string, source string, chartPath string, namespace string) ([]byte, error) {
+	args := []string{
+		"create", "helmrelease", name,
+		"--source", "GitRepository/" + source,
```
jrryjcksn, line 94: In the "previous implementation", we needed to post-process this as flux was replacing ".wego" with "./wego". Perhaps that is no longer a problem? It's difficult to tell with such widespread change.

jrryjcksn (approved):
This looks good to me but, honestly, it's a bit hard to tell with this massive a change.

pkg/flux/flux.go
luizbafilho: thank you for catching that up, I remember seeing it but forgot to add it back.

cmd/wego/add/cmd.go
```diff
@@ -54,26 +57,38 @@ func runCmd(cmd *cobra.Command, args []string) {
 
 	authMethod, err := ssh.NewPublicKeysFromFile("git", params.PrivateKey, params.PrivateKeyPass)
 	if err != nil {
-		fmt.Fprintf(shims.Stderr(), "failed reading ssh keys: %s\n", err)
-		shims.Exit(1)
+		fmt.Fprintf(os.Stderr, "failed reading ssh keys: %s\n", err)
```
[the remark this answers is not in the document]
luizbafilho: Initially, that was added to help with testing, but that's not necessary anymore, I've just made a small change to use `RunE` instead, so we can just return the error, and avoid having to `os.Exit` manually.

luizbafilho:
skipped some acceptances tests after discussion with rokshana

PR #344, opened 2021-06-15
Force resolve normalize-url dep

jpellizzari:
Closes https://github.com/weaveworks/weave-gitops/security/dependabot/package-lock.json/normalize-url/open

From slack:

> As I understand it, regex DoS issues are not applicable to us since we only run on the client side. The bad-actor would only be able to lock up their own browser.


mnowster (approved):

mnowster:
Upgrade :allthethings:

PR #345, opened 2021-06-15
Upload deploy key just once

josecordaz:
`wego app add` will only upload one deploy key.

The previous behavior was uploading a new deploy key when running `wego app add` for the same repository every time.

There is just one single deploy keys after running `wego app add` many times. Besides the message `deploy key weave-gitops-deploy-key already exists for repo MY-REPO` properly shown.


pkg/utils/common.go
```diff
@@ -27,3 +28,17 @@ func WaitUntil(out io.Writer, poll, timeout time.Duration, checkDone func() erro
...
+type callback func()
+
+func CaptureStdout(c callback) string {
+	r, w, _ := os.Pipe()
+	tmp := os.Stdout
+	os.Stdout = w
+	c()
+	w.Close()
+	stdout, _ := ioutil.ReadAll(r)
+	os.Stdout = tmp
```
jrryjcksn, line 41: Seems like resetting `os.Stdout` to its original value should be deferred in case the callback errors out.

jrryjcksn (approved):
LGTM w/ one minor nit

PR #350, opened 2021-06-16
flux version from binary and return wego version

J-Thompson12:
Return flux version from the binary

running the command

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,5 @@
 .PHONY: ui-dev
-VERSION=$(shell git describe --always --match "v*")
+VERSION=$(shell git describe --tags --always --match "v*")
 GOOS=$(shell go env GOOS)
 GOARCH=$(shell go env GOARCH)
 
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -8,7 +8,7 @@ BRANCH=$(shell git rev-parse --abbrev-ref HEAD)
 GIT_COMMIT=$(shell git log -n1 --pretty='%h')
 CURRENT_DIR=$(shell pwd)
 FLUX_VERSION=$(shell $(CURRENT_DIR)/tools/bin/stoml $(CURRENT_DIR)/tools/dependencies.toml flux.version)
-LDFLAGS = "-X github.com/weaveworks/weave-gitops/cmd/wego/version.BuildTime=$(BUILD_TIME) -X github.com/weaveworks/weave-gitops/cmd/wego/version.Branch=$(BRANCH) -X github.com/weaveworks/weave-gitops/cmd/wego/version.GitCommit=$(GIT_COMMIT) -X github.com/weaveworks/weave-gitops/pkg/version.FluxVersion=$(FLUX_VERSION)"
+LDFLAGS = "-X github.com/weaveworks/weave-gitops/cmd/wego/version.BuildTime=$(BUILD_TIME) -X github.com/weaveworks/weave-gitops/cmd/wego/version.Branch=$(BRANCH) -X github.com/weaveworks/weave-gitops/cmd/wego/version.GitCommit=$(GIT_COMMIT) -X github.com/weaveworks/weave-gitops/pkg/version.FluxVersion=$(FLUX_VERSION) -X github.com/weaveworks/weave-gitops/cmd/wego/version.Version=$(VERSION)"
```
josecordaz, line 11: The Version field is already filled out in the gorelease job. Are we planning in running a make task to generate it instead?

josecordaz: or maybe we want them in both places right? 🤔

J-Thompson12: Ohhh i just didnt see it doing anything on my build so I figured I would fix it. I havent tried it on a release build

J-Thompson12: So for everything else we do it in both places. I had to change the command for the makefile to actually print out the version.

J-Thompson12: it was just doing my commit before the change to add --tags. not sure why though. Wks has that command and it prints the version

josecordaz: cool man, no problem, it might worth to spent some time later thinking about how to convert this to do the build in a single place so we don't have to maintain two places 🙂

josecordaz (approved):
LGTM

josecordaz (approved):
LGTM

PR #354, opened 2021-06-17
switch to weaveworks-gitops-test repo from wkp-example-org

iahmad9:
**What changed?**
Switching the repo to gitops test repo

**Why?**
To not pollute the wkp test repo.


josecordaz (approved):
LGTM

PR #356, opened 2021-06-17
Parse app config url on ssh url

josecordaz:
It was throwing `could not create git secret: exit status 1` when adding an app using `app-config-url` flag because the url needs to change the format for cloning.


pkg/cmdimpl/add.go
```diff
@@ -128,6 +128,13 @@ func updateParametersIfNecessary(gitClient git.Git) error {
 		params.Url = "ssh://git@github.com/" + trimmed
 	}
 
+	if params.AppConfigUrl != "" {
+		if strings.HasPrefix(params.AppConfigUrl, sshPrefix) {
```
jrryjcksn, line 132: Could we pull this logic out into a function called in both places?

jrryjcksn (changes requested):

jrryjcksn (approved):
LGTM

PR #359, opened 2021-06-17
Avoid getting ginkgo and gomega latest

luizbafilho:
**What changed?**

Avoid using `-u` otherwise it picks up the latest version breaking the execution because that version is different from what is pinned in `go.mod`

**Why?**

**How did you test it?**

**Release notes**

**Documentation Changes**


jpellizzari:
My bad on this one. I had it fixed then chose the wrong method when trying to make both jobs consistent.

jpellizzari (approved):

PR #367, opened 2021-06-21
Fixed url sanitization

josecordaz:
URL sanitation was miss behaving leading to URLs with suffix `.git` that go-git doesn't accept causing errors when attempting to clone the repo. It was tested by adding unit tests  for all possible combinations and making sure smoke and accept tests passed.

### DIFF
--- a/pkg/services/app/add_test.go
+++ b/pkg/services/app/add_test.go
@@ -437,7 +437,7 @@ var _ = Describe("Add", func() {
 
 				name, url, branch, secretRef, namespace = fluxClient.CreateSourceGitArgsForCall(1)
 				Expect(name).To(Equal("bar"))
-				Expect(url).To(Equal("ssh://git@github.com/foo/bar.git"))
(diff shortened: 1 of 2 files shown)


luizbafilho:
Hey Jose, have you manually tested the whole flow and checked the status of the GitRepository? I remember that breaking without the `.git`.

pkg/services/app/add_test.go
```diff
@@ -608,3 +607,57 @@ var _ = Describe("Add", func() {
 		})
 	})
 })
+
+var _ = Describe("sanitizeRepoUrl", func() {
+
```
luizbafilho, line 612: I'm not sure if this is some config on your editor, but we should avoid adding a blank line at the beginning and the end of functions.

luizbafilho (changes requested):
Looks good, just need some manual verification if flux breaks without the `.git` at the end of the URL.

Issue #257, opened 2021-06-07
View App Status in UI


markramm:

markramm:
As an Application Operator I would like togo to the Weave Gitops UI and see the current status of all my applications so that I can determine if there are any problems in the deployment that require my attention. 

@jpellizzari had a visualization from the flux UI, which needs to be adapted to show an list of apps and their status pages.

markramm:
WIreframe: https://share.balsamiq.com/c/iyah4imNHohbKunRZYEbKk.png

markramm:
Container images tag can be separated out to a separate ticket.

PR #368, opened 2021-06-21
Expose application status to the API and UI

jpellizzari:
Part of #257 

Exposes some real data from a WeGO application:
Screenshot from 2021-06-21 13-24-28

Anyone have any ideas on what other data we might want to surface here?

This currently uses the `kubectl` binary like an HTTP interface to the cluster. In a subsequent PR, I will "cut out the middle-man" and query the cluster HTTP API directly, but for now, this is how the `App` service is built, so I wanted to consume it the same way as everything else.


pkg/kube/kube.go
```diff
@@ -99,6 +103,23 @@ func (k *KubeClient) GetClusterStatus() ClusterStatus {
 	return Unknown
 }
 
+func (k *KubeClient) GetApplication(name string) (*wego.Application, error) {
+	cmd := []string{"get", "app", name, "-o", "json"}
+	o, err := k.runKubectlCmd(cmd)
+
+	if err != nil {
+		return nil, fmt.Errorf("could not run kubectl command: %s", err)
```
josecordaz, line 111: Could we add the name of the app somewhere in the error message, it would help for debugging afterwards 🙂

pkg/kube/kube.go
```diff
@@ -99,6 +103,23 @@ func (k *KubeClient) GetClusterStatus() ClusterStatus {
...
+	}
+
+	a := wego.Application{}
+
+	if err := json.Unmarshal(o, &a); err != nil {
+		return nil, fmt.Errorf("could not unmarshal json: %s", err)
```
josecordaz, line 117: probably here too

josecordaz (approved):
LGTM

pkg/kube/kube.go
jpellizzari: Based on this feedback, I tried an error message format like this: 

`return nil, fmt.Errorf("could not run kubectl command for %s: %s", name, err)`

To me, the error message gets a little more confusing:

`could not run kubectl command for my-app: ...`

^^ If I am the user, this makes me think that `my-app` is the problem, when really there is probably another issue.

Since the consumer has to pass in the app name as an argument already, they already have the context for the error message.

pkg/kube/kube.go
jpellizzari: Same as my previous comment; the `Unmarshal` call doesn't really have anything to do with the `app` name. It might fail no matter which `app` name got passed in.

jrryjcksn (approved):
LGTM w/ one question about `branch` field

api/applications/applications.proto
```diff
@@ -26,10 +28,20 @@ service Applications {
...
+    rpc GetApplication(GetApplicationRequest) returns (GetApplicationResponse) {
+        option (google.api.http) = {
+            get : "/v1/applications/{application_name}"
+        };
+    }
 }
 
 message Application {
-    string name = 1; // The name of the application
+    string name = 1;  // The name of the application
```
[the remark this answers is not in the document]
jpellizzari: The only thing in the CRD (currently), are the `path` and `url` fields. I plan on addressing this in a follow-up PR.

PR #370, opened 2021-06-22
Generate secrets with unique name

josecordaz:
**What changed?**
We are creating git secrets per repository.

**Why?**
We were overwriting git secrets causing authentication problems when flux tries to clone the repository.

**How did you test it?**
This was tested manually.

**Release notes**

**Documentation Changes**

### DIFF
--- a/pkg/services/app/add.go
+++ b/pkg/services/app/add.go
@@ -70,7 +70,7 @@ func (a *App) Add(params AddParams) error {
 		return err
 	}
 
-	fmt.Println("Generating deploy key...")
+	fmt.Printf("Generating deploy key for repo %s ...\n", params.Url)
 	secretRef, err := a.createAndUploadDeployKey(params.Url, clusterName, params.Namespace, params.DryRun)
 	if err != nil {
 		return errors.Wrap(err, "could not generate deploy key")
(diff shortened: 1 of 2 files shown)


pkg/services/app/add.go
```diff
@@ -317,36 +323,36 @@ func (a *App) commitAndPush(params AddParams, filters ...func(string) bool) erro
 	return nil
 }
 
-func (a *App) createAndUploadDeployKey(reposUrls []string, clusterName string, namespace string, dryRun bool) (string, error) {
-	secretRef := fmt.Sprintf("weave-gitops-%s", clusterName)
-	if dryRun {
-		return secretRef, nil
+func (a *App) createAndUploadDeployKey(repoUrl string, clusterName string, namespace string, dryRun bool) (string, error) {
+
+	if repoUrl == "" {
```
jrryjcksn, line 328: Is this a valid case? Should we return an error if `repoUrl` is empty?

jrryjcksn (approved):
One small question, but LGTM

pkg/services/app/add.go
josecordaz: I'm with you, I think there is a proper way to do this. I was trying to fix a failure in the unit tests. But that would involve a bit more changes. I prefer to have this merged first as I need to talk with @luizbafilho to clarify some doubts.

luizbafilho: not a valid case, that's left over the UX change, feel free to remove it.

PR #376, opened 2021-06-22
Use kubebuilder CRD workflow

jpellizzari:
Relies on #368 
Adds a new `Makefile` job(s) to create the crd yaml from a Golang struct via `kubebuilder`.

The allows us to build out our CRD from a `go` struct and use the CR as an object throughout the code base.


luizbafilho (approved):

jrryjcksn (approved):
LGTM

PR #384, opened 2021-06-23
Re-arrange routing and add error handling

jpellizzari:
Re-arranges the UI routing to wrap the page content with the navigation elements, and adds some async error message handling:

Screenshot from 2021-06-23 13-08-39

Note that the `<ErrorBoundary />` component handles "synchronous" errors in component render blocks

### DIFF
--- a/ui/App.tsx
+++ b/ui/App.tsx
@@ -6,8 +6,10 @@ import {
   Route,
   Switch,
 } from "react-router-dom";
+import { ToastContainer } from "react-toastify";
 import { ThemeProvider } from "styled-components";
 import ErrorBoundary from "./components/ErrorBoundary";
+import Layout from "./components/Layout";
 import AppContextProvider from "./contexts/AppContext";
 import { Applications as appsClient } from "./lib/api/applications/applications.pb";
 import theme, { GlobalStyle, muiTheme } from "./lib/theme";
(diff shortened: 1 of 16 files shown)


ui/lib/utils.ts
```diff
@@ -16,3 +17,11 @@ export const getNavValue = (currentPage: any): PageRoute | boolean => {
       return false;
   }
 };
+
+export function notifySuccess(message: string) {
+  toast["success"](message);
+}
+
+export function notifyError(message: string) {
+  toast["error"](`Error: ${message}`);
```
foot, line 26: 🔥 🍞 .

I think we are using Snack not Toast cc @AlinaGoaga

ui/App.tsx
```diff
@@ -22,24 +24,31 @@ export default function App() {
...
+                    component={ApplicationDetail}
+                  />
+                  <Redirect exact from="/" to={PageRoute.Applications} />
+                  <Route exact path="*" component={Error} />
+                </Switch>
+              </ErrorBoundary>
+              <ToastContainer
+                position="top-center"
+                autoClose={10000}
+                newestOnTop={false}
```
foot, line 48: Interesting! To avoid newer things coming in and hiding messages? Might get a slightly staler error message here though, but 10s sounds unlikely.

ui/contexts/AppContext.tsx
```diff
@@ -1,15 +1,45 @@
...
+    error: null,
+  });
+
+  React.useEffect(() => {
+    // clear the error state on navigation
+    setAppState({
+      ...appState,
+      error: null,
+    });
+  }, [window.location]);
```
foot, line 29: Nifty. Make sense.

ui/App.tsx
```diff
@@ -6,8 +6,10 @@ import {
   Route,
   Switch,
 } from "react-router-dom";
+import { ToastContainer } from "react-toastify";
```
foot, line 9: Ah hah more toast. Cool. Well its all in one place if we need to reconcile Snackbar and Toast 👌

foot (approved):
Yeah, screenshot LGTM 👌 , error handling. ⭐ .

foot:
btw whats the easiest way to quickly test out the UI and poke around?

ui/lib/utils.ts
jpellizzari: Not married to the lib. It should be easy enough to swap out.

ui/App.tsx
jpellizzari: :thinking: I can't remember why I set that prop. I'll see what happens with multiple errors.

jpellizzari:
> btw whats the easiest way to quickly test out the UI and poke around?

@foot There isn't much to see at the moment. The List of applications is still fake data from the backend. I'll ping you guys when there is real data.

PR #385, opened 2021-06-23
187 add branch

J-Thompson12:
Add a new flag named auto-merge. If set to true add will not create a pr. Default is false

Creates a PR with a branch name that is a hash of url + path + branch to be merged to the branch specified in --branch. 
If a branch with that name already exists add will fail.

Adds a label with the same hash to the app in the cluster. 
If that label exists the add will fail.

removed extra manifests being applied

Added namespace to app manifest so app was created under the correct namespace

Added app goat to external repo apply

### DIFF
--- a/pkg/kube/kube.go
+++ b/pkg/kube/kube.go
@@ -180,12 +180,12 @@ func (k *KubeClient) GetApplications(ctx context.Context, ns string) ([]wego.App
 	return a.Items, nil
 }
 func (k *KubeClient) LabelExistsInCluster(label string) error {
-	cmd := []string{"get", "pods", "--namespace=wego-system", "--show-labels"}
+	cmd := []string{"get", "app", "-l", fmt.Sprintf("weave-gitops.weave.works/app-identifier=%s", label)}
 	o, err := k.runKubectlCmd(cmd)
 	if err != nil {
 		return fmt.Errorf("could not run kubectl command: %s", err)
 	}
-	if strings.Contains(string(o), label) {
+	if !strings.Contains(string(o), "No resources found") {
 		return fmt.Errorf("unable to create resource, resource already exists in cluster")
 	}
 	return nil
--- a/pkg/kube/kube.go
+++ b/pkg/kube/kube.go
@@ -48,7 +48,7 @@ type Kube interface {
 	GetClusterName(ctx context.Context) (string, error)
 	GetClusterStatus(ctx context.Context) ClusterStatus
 	GetApplication(ctx context.Context, name string) (*wego.Application, error)
-	LabelExistsInCluster(label string) error
+	LabelExistsInCluster(ctx context.Context, label string) error
 }
 
 type KubeClient struct {
@@ -179,17 +179,6 @@ func (k *KubeClient) GetApplications(ctx context.Context, ns string) ([]wego.App
 
 	return a.Items, nil
 }
-func (k *KubeClient) LabelExistsInCluster(label string) error {
-	cmd := []string{"get", "app", "-l", fmt.Sprintf("weave-gitops.weave.works/app-identifier=%s", label)}
-	o, err := k.runKubectlCmd(cmd)
-	if err != nil {
-		return fmt.Errorf("could not run kubectl command: %s", err)
-	}
-	if !strings.Contains(string(o), "No resources found") {
-		return fmt.Errorf("unable to create resource, resource already exists in cluster")
-	}
-	return nil
-}
 
 func (k *KubeClient) resourceLookup(args string) error {
 	_, err := k.runKubectlCmd(strings.Split(args, " "))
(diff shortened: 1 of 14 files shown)


cmd/wego/add/cmd.go
```diff
@@ -94,6 +95,8 @@ func runCmd(cmd *cobra.Command, args []string) error {
 
 	appService := app.New(gitClient, fluxClient, kubeClient, gitProviders)
 
+	utils.SetCommmitMessageFromArgs("wego app add", params.Url, params.Path, params.Name)
```
J-Thompson12, line 98: The plan is to set this to the be the exact command that was run. I was unware exactly how to do this and this PR needed done faster so I am pushing this to the backburner.

cmd/wego/add/cmd.go
```diff
@@ -45,6 +45,7 @@ func init() {
 	Cmd.Flags().StringVar(&params.PrivateKey, "private-key", "", "Private key to access git repository over ssh")
 	Cmd.Flags().StringVar(&params.AppConfigUrl, "app-config-url", "", "URL of external repository (if any) which will hold automation manifests; NONE to store only in the cluster")
 	Cmd.Flags().BoolVar(&params.DryRun, "dry-run", false, "If set, 'wego add' will not make any changes to the system; it will just display the actions that would have been taken")
+	Cmd.Flags().BoolVar(&params.AutoMerge, "auto-merge", false, "If set, 'wego add' will merge automatically into the set --branch")
```
luizbafilho, line 48: minor thing, I see we used the `wego add will...` probably because of the dry run message, but I think that is redundant, you are already calling the `wego add` is implied that we are talking about it.

pkg/kube/kube.go
```diff
@@ -149,6 +150,18 @@ func (k *KubeClient) GetApplication(name string) (*wego.Application, error) {
 	return &a, nil
 }
 
+func (k *KubeClient) LabelExistsInCluster(label string) error {
+	cmd := []string{"get", "pods", "--namespace=wego-system", "--show-labels"}
```
luizbafilho, line 154: comment for refactoring later, but this will be quite inefficient, as a cluster grows, querying for all pods takes quite some time, several seconds to minutes.

So, ideally, we should query for the apps with the label on them, something like:

```
 kubectl get app -l hash=my-hash
```

luizbafilho (approved):
Looks good to me, with minor nits that can be handle later.

cmd/wego/add/cmd.go
J-Thompson12: It was because of dry run. I definitely copied it.

pkg/kube/kube.go
jrryjcksn, line 154: We shouldn't assume this is all in `wego-system`. Also, it would be nice to filter the output to only contain the labels.

pkg/services/app/add.go
```diff
@@ -444,6 +527,8 @@ apiVersion: wego.weave.works/v1alpha1
 kind: Application
 metadata:
   name: {{ .AppName }}
+  labels:
+    name: {{ .AppHash }}
```
jrryjcksn, line 531: We shouldn't use `name` as the label name. It should be something qualified like:
`weave-gitops.weave.works/app-identifier`

jrryjcksn (changes requested):

cmd/wego/add/cmd.go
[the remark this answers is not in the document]
J-Thompson12: auto merge seems right to me because we are skipping the PR stage and just merging right to the set branch. Auto push would make sense if this was for the PR. So if auto-merge is false then a PR is created.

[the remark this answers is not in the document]
jrryjcksn: I don't feel strongly about changing this right now (particularly since we'll likely move to the config file stuff later). However, we "commit and push" in the non-PR path (any merging would have had to have already happened) :man_shrugging:

pkg/kube/kube.go
```diff
@@ -149,6 +150,18 @@ func (k *KubeClient) GetApplication(name string) (*wego.Application, error) {
 	return &a, nil
 }
 
+func (k *KubeClient) LabelExistsInCluster(label string) error {
+	cmd := []string{"get", "app", "-l", fmt.Sprintf("weave-gitops.weave.works/app-identifier=%s", label)}
```
jrryjcksn, line 154: Now I think we're only looking in the `default` namespace. I believe we should be using the namespace argument to `wego app add` which _defaults_ to `wego-system`

jrryjcksn (approved):
LGTM

pkg/services/app/add.go
```diff
@@ -351,18 +351,25 @@ func (a *App) generateExternalRepoManifests(params AddParams, secretRef string,
 		return nil, nil, errors.Wrap(err, "could not generate target source manifests")
 	}
 
+	AppGoat, err := a.flux.CreateKustomization(params.Name, repoName, filepath.Join(".", "apps", params.Name), params.Namespace)
```
jrryjcksn, line 354: I feel kind of dumb mentioning this, but this goes against naming conventions (capitalized local variable) :man_shrugging:

pkg/services/app/add.go
```diff
@@ -351,18 +351,25 @@ func (a *App) generateExternalRepoManifests(params AddParams, secretRef string,
...
+	if err != nil {
+		return nil, nil, errors.Wrap(err, fmt.Sprintf("could not create kustomization for '%s' .wego/apps", params.Name))
```
jrryjcksn, line 356: No `.wego` here

pkg/services/app/add.go
J-Thompson12: that was an accident. It was named something else and i deleted the first part. thanks for catching that.

PR #387, opened 2021-06-24
Initial ADR for API definitions with gRPC.

bigkevmcd:
**What changed?**
Start of ADRs on API standardisation

**Why?**
Recommendation to make it easier to build standard, interoperable clients.


palemtnrider:
Thanks, Kevin for capturing this information.  I'd like to see the tooling separated from the use of protocol buffers.  I can envision us changing tooling before moving away from protobufs and having 2 ADRs will make that easier.

palemtnrider (approved):
Approving to get them merged into the code base.

PR #390, opened 2021-06-25
Do not attempt to create deploy key for helm repositories

jrryjcksn:
Also, check if deploy key exists before creating flux secret...

**What changed?**
- We were attempting to create a deploy key at the top of `Add`, before we determined the source type. Deploy keys are not meaningful for helm repositories, so the creation failed
- We created a flux secret for the deploy key before determining if the key already existed; this overwrote the secret containing the key in the cluster but we did not associate the new key with the repository

**Why?**
To fix the use of helm repositories and ensure we did not corrupt the flux secret for the repo if we added another app.

**How did you test it?**
Tested locally

**Release notes**
No

**Documentation Changes**
No

### DIFF
--- a/pkg/services/app/add_test.go
+++ b/pkg/services/app/add_test.go
@@ -41,20 +41,11 @@ var _ = BeforeEach(func() {
 	}
 	gitProviders = &gitprovidersfakes.FakeGitProviderHandler{}
 
-	appSrv = app.New(gitClient, fluxClient, kubeClient, gitProviders)
-
-	deployKeyLookups = 0
-	deployKeyUploads = 0
-
-	gitProviders.UploadDeployKeyStub = func(s1, s2 string, key []byte) error {
-		deployKeyUploads += 1
-		return nil
-	}
-
 	gitProviders.DeployKeyExistsStub = func(s1, s2 string) (bool, error) {
-		deployKeyLookups += 1
 		return false, nil
 	}
+
+	appSrv = app.New(gitClient, fluxClient, kubeClient, gitProviders)
 
 	defaultParams = app.AddParams{
 		Url:            "https://github.com/foo/bar",
@@ -120,14 +111,13 @@ var _ = Describe("Add", func() {
 			defaultParams.SourceType = string(app.SourceTypeGit)
 
 			gitProviders.DeployKeyExistsStub = func(s1, s2 string) (bool, error) {
-				deployKeyLookups += 1
 				return true, nil
 			}
 
 			err := appSrv.Add(defaultParams)
 			Expect(err).ShouldNot(HaveOccurred())
-			Expect(deployKeyUploads).To(Equal(0))
-			Expect(deployKeyLookups).To(Equal(1))
+			Expect(gitProviders.UploadDeployKeyCallCount()).To(Equal(0))
+			Expect(gitProviders.DeployKeyExistsCallCount()).To(Equal(1))
 		})
 
 		It("looks up deploy key and creates secret if not found", func() {
@@ -135,8 +125,8 @@ var _ = Describe("Add", func() {
 
 			err := appSrv.Add(defaultParams)
 			Expect(err).ShouldNot(HaveOccurred())
-			Expect(deployKeyUploads).To(Equal(1))
-			Expect(deployKeyLookups).To(Equal(1))
+			Expect(gitProviders.UploadDeployKeyCallCount()).To(Equal(1))
+			Expect(gitProviders.DeployKeyExistsCallCount()).To(Equal(1))
 		})
 	})
 
@@ -148,8 +138,8 @@ var _ = Describe("Add", func() {
 
 				err := appSrv.Add(defaultParams)
 				Expect(err).ShouldNot(HaveOccurred())
-				Expect(deployKeyUploads).To(Equal(0))
-				Expect(deployKeyLookups).To(Equal(0))
+				Expect(gitProviders.UploadDeployKeyCallCount()).To(Equal(0))
+				Expect(gitProviders.DeployKeyExistsCallCount()).To(Equal(0))
 			})
 		})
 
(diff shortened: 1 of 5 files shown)


pkg/services/app/add_test.go
```diff
@@ -100,7 +115,44 @@ var _ = Describe("Add", func() {
...
+			defaultParams.SourceType = string(app.SourceTypeGit)
+
+			gitProviders.DeployKeyExistsStub = func(s1, s2 string) (bool, error) {
+				deployKeyLookups += 1
+				return true, nil
+			}
+
+			err := appSrv.Add(defaultParams)
+			Expect(err).ShouldNot(HaveOccurred())
+			Expect(deployKeyUploads).To(Equal(0))
```
luizbafilho, line 129: To keep consistent with the rest of the tests you can use the `CallCount` methods, so you can get rid of `deployKeyLookups` and `deployKeyLookups` and only implement the custom Stub when it returns something different from the default values.

In this case, it could be:

```go
err := appSrv.Add(defaultParams)

Expect(gitProviders.DeployKeyExistsCallCount()).To(Equal(1))

Expect(fluxClient.CreateSecretGitCallCount()).To(Equal(0))
Expect(gitProviders.UploadDeployKeyCallCount()).To(Equal(0))
```

luizbafilho (changes requested):
Looks pretty good but minor comments on the tests to keep consistency. 

Although we have to discuss the `gitproviders` mocking approach, I don't like that it requires some special steps to get it done. @josecordaz is the only one who can get that mocked properly.

pkg/services/app/add_test.go
jrryjcksn: I looked at that. That isn't what "CallCount" does. I wish it were. Here's the definition from the gitops fake:

```
func (fake *FakeGitProviderHandler) UploadDeployKeyCallCount() int {
	fake.uploadDeployKeyMutex.RLock()
	defer fake.uploadDeployKeyMutex.RUnlock()
	return len(fake.uploadDeployKeyArgsForCall)
}
```

It actually is just counting the args passed to the call. Justin and I ran into this last week.

pkg/services/app/add_test.go
```diff
@@ -41,6 +43,19 @@ var _ = BeforeEach(func() {
...
+	gitProviders.UploadDeployKeyStub = func(s1, s2 string, key []byte) error {
+		deployKeyUploads += 1
+		return nil
+	}
+
+	gitProviders.DeployKeyExistsStub = func(s1, s2 string) (bool, error) {
+		deployKeyLookups += 1
+		return false, nil
+	}
+
```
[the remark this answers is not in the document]
jrryjcksn: Yes, Justin and I got confused by this last week (and I now think I know why). I expected it to work the way you indicate but it _seemed_ like it was actually counting the args. I now think that might have been due to not regenerating the fakes. I'll definitely fix this. That's what I wanted to do in the first place. :-)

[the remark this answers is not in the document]
luizbafilho: Sounds good, let me know if you need any help with that.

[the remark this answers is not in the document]
jrryjcksn: Agreed on the recorder testing.

pkg/services/app/add_test.go
```diff
@@ -22,8 +22,10 @@ var (
 	kubeClient   *kubefakes.FakeKube
 	gitProviders *gitprovidersfakes.FakeGitProviderHandler
 
-	appSrv        app.AppService
-	defaultParams app.AddParams
+	appSrv           app.AppService
+	defaultParams    app.AddParams
+	deployKeyLookups int
```
luizbafilho, line 27: leftover

pkg/services/app/add_test.go
```diff
@@ -39,6 +41,10 @@ var _ = BeforeEach(func() {
 	}
 	gitProviders = &gitprovidersfakes.FakeGitProviderHandler{}
 
+	gitProviders.DeployKeyExistsStub = func(s1, s2 string) (bool, error) {
```
luizbafilho, line 44: This is the default implementation of the stub method you can just remove it

pkg/services/app/add_test.go
```diff
@@ -100,7 +106,43 @@ var _ = Describe("Add", func() {
...
+		It("looks up deploy key and skips creating secret if found", func() {
+			defaultParams.SourceType = string(app.SourceTypeGit)
+
+			gitProviders.DeployKeyExistsStub = func(s1, s2 string) (bool, error) {
+				return true, nil
+			}
+
+			err := appSrv.Add(defaultParams)
+			Expect(err).ShouldNot(HaveOccurred())
+			Expect(gitProviders.UploadDeployKeyCallCount()).To(Equal(0))
```
luizbafilho, line 119: it would be nice to check the flux calls since that's what actually breaks the code, so adding `Expect(fluxClient.CreateSecretGitCallCount()).To(Equal(0|1))` where is due

pkg/services/app/add_test.go
```diff
@@ -231,6 +272,18 @@ var _ = Describe("Add", func() {
...
 		})
 
+		Describe("avoids deploy key for helm", func() {
+			It("skips secret creation and lookup when source type is helm", func() {
+				defaultParams.Url = "https://charts.kube-ops.io"
+				defaultParams.Chart = "loki"
+
+				err := appSrv.Add(defaultParams)
+				Expect(err).ShouldNot(HaveOccurred())
+				Expect(deployKeyUploads).To(Equal(0))
```
luizbafilho, line 282: leftover

luizbafilho (changes requested):

pkg/services/app/add_test.go
jrryjcksn: Thanks for catching that!

pkg/services/app/add_test.go
jrryjcksn: Makes sense.

pkg/services/app/add_test.go
jrryjcksn: Sorry, missed the fluxClient reference here.

luizbafilho (approved):

PR #392, opened 2021-06-25
Reading private key password

### LINKED ISSUE #357: Providing a `--url` with an encrypted ssh key fails on public repo
Providing a --url with an encrypted ssh key fails on a public repo.

**Environment**
 - wego: `main`
 
**To Reproduce**
1. Run this command on `main`:
`$ wego app add --branch main --path kustomize --name my-app --url git@github.com:stefanprodan/podinfo.git`

**Expected behavior**
The repo should clone and `wego app add` should continue

**Actual Behavior**
The command fails with: `failed reading ssh keys: x509: decryption password incorrect`

**Screenshots**
Screenshot from 2021-06-17 13-47-01

**Additional context**
I think this will be a potential blocker for people trying out `wego`. We don't explicitly let the user know we are going to use their keys, so it may be confusing to hit this. 

I would expect an error when trying to do something with a private repo, but a public repo should work?

luizbafilho:
**What changed?**
- Fixes #357 
- When the user's private key is encrypted, it now prompts a message asking for the password

**Why?**
- wego failed to proceed without the private key password

**How did you test it?**
- `wego app add --branch main --path kustomize --name my-app --url ssh://git@github.com/stefanprodan/podinfo.git --dry-run`

**Release notes**

**Documentation Changes**

### DIFF
--- a/cmd/wego/add/cmd.go
+++ b/cmd/wego/add/cmd.go
@@ -20,6 +20,7 @@ import (
 	"github.com/weaveworks/weave-gitops/pkg/runner"
 	"github.com/weaveworks/weave-gitops/pkg/services/app"
 	"github.com/weaveworks/weave-gitops/pkg/utils"
+	"golang.org/x/term"
 )
 
 var params app.AddParams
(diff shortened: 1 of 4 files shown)


jrryjcksn (approved):
LGTM (and works for me, too)

PR #409, opened 2021-06-29
Add k8s client HTTP implementation

jpellizzari:
Adds a new implementation of the `kube.Kube` interface. Related to #368 .

Previously, the `GetApplication` method used `kubectl` as an HTTP client. This PR cuts out the CLI call to `kubectl` and goes directly to the K8s HTTP API.

Due to the way the `kube.Kube` passes `[]byte` slices around, some functions are not implementable. I cannot for-see a situation where we would allow a user to pass a raw byte buffer and have it applied to the cluster (for example).

This feels like it doesn't really fit under the `kube.Kube` interface, so this is more of an interim solution. I believe we will need to phase out the command-line implementation to avoid duplicating logic between the two paths.


pkg/kube/kubehttp.go
```diff
@@ -0,0 +1,180 @@
...
+	_ = helmv2.AddToScheme(scheme)
+	_ = wego.AddToScheme(scheme)
+	_ = corev1.AddToScheme(scheme)
+
+	return scheme
+}
+
+var WeGONamespace = "wego-system"
+var WeGOCRDName = "apps.wego.weave.works"
+var FluxNamespace = "flux-system"
```
josecordaz, line 36: Could we change this vars to constants and use them in other placers like this https://github.com/weaveworks/weave-gitops/pull/409/files#diff-49dcd96d19df2344f270f32a5b138783f8e8c3ddf2063ec83f998cbf2a29121eL126 ?

pkg/kube/kubehttp.go
```diff
@@ -0,0 +1,180 @@
...
+
+	v1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	"k8s.io/client-go/tools/clientcmd"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	helmv2 "github.com/fluxcd/helm-controller/api/v2beta1"
+	kustomizev1 "github.com/fluxcd/kustomize-controller/api/v1beta1"
+	sourcev1 "github.com/fluxcd/source-controller/api/v1beta1"
+	"github.com/pkg/errors"
+	wego "github.com/weaveworks/weave-gitops/api/v1alpha1"
```
josecordaz, line 15: 💯

pkg/kube/kubehttp_test.go
```diff
@@ -0,0 +1,129 @@
...
+var _ = BeforeSuite(func() {
+	testenv = &envtest.Environment{CRDDirectoryPaths: []string{"../../manifests/crds"}}
+
+	scheme = kube.CreateScheme()
+	cfg, err = testenv.Start()
+
+	Expect(err).NotTo(HaveOccurred())
+
+	ns = corev1.Namespace{
+		ObjectMeta: metav1.ObjectMeta{Name: "wego-system"},
```
josecordaz, line 36: good place for the global constant

pkg/kube/kubehttp_test.go
```diff
@@ -0,0 +1,129 @@
...
+
+	testclient, err = client.New(cfg, client.Options{Scheme: scheme})
+	Expect(err).NotTo(HaveOccurred())
+
+	err = testclient.Create(context.Background(), &ns)
+	Expect(err).NotTo(HaveOccurred())
+})
+
+var _ = BeforeEach(func() {
+	k = &kube.KubeHTTP{
```
josecordaz, line 47: wondering if it would be ok to move this the first line of the next `Describe` 🤔

pkg/kube/kubehttp_test.go
```diff
@@ -0,0 +1,129 @@
...
+
+var _ = Describe("KubeHTTP", func() {
+	It("GetClusterName", func() {
+		name, err := k.GetClusterName(context.Background())
+		Expect(err).NotTo(HaveOccurred())
+
+		Expect(name).To(Equal(testClustername))
+	})
+	It("GetClusterStatus", func() {
+		ctx := context.Background()
```
josecordaz, line 61: `context` could be moved to the beginning of the `Describe` to use it everywhere and avoid duplication

pkg/services/app/add.go
```diff
@@ -45,6 +45,7 @@ type AddParams struct {
 }
 
 func (a *App) Add(params AddParams) error {
+	ctx := context.Background()
```
josecordaz, line 48: tiny improvement I think it would be to add the context as a field in App to avoid creating new contexts in each method.

josecordaz (approved):
LGTM

josecordaz:
I see you figured out how to add the app schema using `kubebuilder` nice!!! 💯 👏🏽

pkg/services/app/add.go
luizbafilho: That's not recommended you can take a look why here: https://blog.golang.org/context-and-structs

luizbafilho (approved):
Awesome it looks great

pkg/services/app/add.go
josecordaz: You are right, it seems this is not the right approach. The actual goal I was trying to achieve here is to have the same context object instance instead of creating one inside each function like this, to take the most advantage of it. From the post `...of its extensibility in order to build a powerful tree of cancellation, deadline, and metadata information through the call stack...`. So one way that might solve this would be to add a field for the context in Add and follow the same pattern everywhere like in the `git` pkg (where we are creating isolated context for each function) for instance and it would help for testing as well.

luizbafilho: yes, ideally this should be injected in the method call.

pkg/kube/kubehttp.go
jpellizzari: >  change this vars to constants
We can, but changing to `const` just prevents runtime re-assignment. If someone added a change to re-assign, I would say "don't do that".

I guess `const` signals intent better in this case.

> use them in other placers like this

Good catch.

pkg/kube/kubehttp_test.go
jpellizzari: I had originally wanted a clean environment on each test, but the `testenv` library doesn't work that way. This object is currently stateless, so not much value in creating a new one every time, but if it has state in the future, I want a clean state on each test.

pkg/services/app/add.go
jpellizzari: The common use case for `context.Context` is to trace requests that cascade into other requests. For us, this would might mean:

```
browser client -> API Server -> K8s API
```
Cancelling the browser request could be made to cancel the K8s API request, or tie then together with a trace, or pass auth information throughout the stack.

Since the command-line invocations do not produce/require a context, its OK to create it at the top-level of this service entry point (IMO). `context.Context` has to be created somewhere.

jpellizzari: > yes, ideally this should be injected in the method call.

It didn't seem appropriate to change the signature in this PR. It would have made the scope much larger. We can do that as we go.

luizbafilho: you are totally right, but I had the impression you'd call the App service methods at some point from the API, if that's the case it would make sense to have the context in the method params.

luizbafilho: I'm good with leaving like that until you need to call it from the API.

pkg/kube/kubehttp_test.go
jpellizzari: See the comments below. A clean context on each test is the desired behavior.

pkg/kube/kubehttp.go
josecordaz: cool

pkg/kube/kubehttp_test.go
josecordaz: that makes sense to me 👍🏼

pkg/kube/kubehttp_test.go
```diff
@@ -124,6 +124,18 @@ var _ = Describe("KubeHTTP", func() {
...
+	It("GetApplications", func() {
+		ctx := context.Background()
+		name := "my-app"
+		// TODO: this currently relies on the previous GetApplication test case.
+		// This is very bad and I intend on fixing the
+		// entire test environment isolation issue in a later PR.
+
+		list, err := k.GetApplications(ctx, name)
+		Expect(err).NotTo(HaveOccurred())
+		Expect(len(list)).To(Equal(1))
```
bigkevmcd, line 137: This could be `Expect(list).To(HaveLen(1))`

PR #410, opened 2021-06-29
Fix a fresh cluster with a repo that already has a deploy key

jrryjcksn:
**What changed?**
We now recreate a deploy key if the cluster does not have a matching secret

**Why?**
If a user recreates her cluster, it will fail if the app repo has already been used and has an existing deploy key

**How did you test it?**
Unit tests

**Release notes**
No

**Documentation Changes**
No


josecordaz (approved):
LGTM

PR #412, opened 2021-06-29
Added command `app list`

josecordaz:
**Describe what has changed in this PR**
New command `app list`.

**Tell your future self why have you made these changes**
We needed an option to list apps already added to the cluster.

**How have you verified this change? Tested locally? Added unit test/integration/acceptance test(s)?**
Manually using the cli and also using the UI

**Release notes**
Added support to list apps with `wego app list`

**Documentation Changes**
Added support to list apps with `wego app list`


pkg/kube/kube_test.go
```diff
@@ -178,4 +178,41 @@ var _ = Describe("GetApplication", func() {
...
+		runner.RunStub = func(cmd string, args ...string) ([]byte, error) {
+			return res, nil
+		}
+
+		apps, err := kubeClient.GetApplications("wego-system")
+		Expect(err).ShouldNot(HaveOccurred())
+		Expect(2).To(Equal(len(*apps)))
+
+		for i, a := range appsList.Items {
+			apps := *apps
```
jrryjcksn, line 211: Another nit, but this seems like an unnecessarily obscure usage. I feel like we should avoid shadowing a variable with a different type

jrryjcksn (approved):
LGTM

pkg/kube/kube_test.go
josecordaz: it didn't seem right to me either 🙂. I forgot to look up for a different approach. This way allowed me to fix an issue with the compiler when I tried to use `*apps[i].name`. I would deeply appreciate a suggestion for the right way. For this approach in particular I guess I also missed to have this assignation outside of the `for` 🤔

josecordaz: found it 🙂

jrryjcksn: :+1:

PR #414, opened 2021-06-30
Update add test

iahmad9:
**What changed?**
* Pulling repo before checking the presence/absence of .wego folder
* @jpellizzari suggested moving the test setup-related initializations to a function to avoid repition.
* Update the kind and k8s version.
* 

**Why?**
Housekeeping as well as the current kind version seems to be incompatible with CI infra.

### DIFF
--- a/test/acceptance/test/add_tests.go
+++ b/test/acceptance/test/add_tests.go
@@ -829,7 +829,7 @@ var _ = Describe("Weave GitOps Add Tests", func() {
 		})
 	})
 
-	FIt("Verify wego can add kustomize-based app with 'app-config-url=NONE'", func() {
+	It("Verify wego can add kustomize-based app with 'app-config-url=NONE'", func() {
 		var repoAbsolutePath string
 		private := true
 		appRepoName := "wego-test-app-" + RandString(8)
--- a/test/acceptance/test/add_tests.go
+++ b/test/acceptance/test/add_tests.go
@@ -829,7 +829,7 @@ var _ = Describe("Weave GitOps Add Tests", func() {
 		})
 	})
 
-	It("Verify wego can add kustomize-based app with 'app-config-url=NONE'", func() {
+	FIt("Verify wego can add kustomize-based app with 'app-config-url=NONE'", func() {
 		var repoAbsolutePath string
 		private := true
 		appRepoName := "wego-test-app-" + RandString(8)
@@ -875,10 +875,11 @@ var _ = Describe("Weave GitOps Add Tests", func() {
 		})
 
 		By("And I should not see wego components in the remote git repo", func() {
+			pullGitRepo(repoAbsolutePath)
 			folderOutput, _ := runCommandAndReturnOutput(fmt.Sprintf("cd %s && ls -al", repoAbsolutePath))
-			Eventually(folderOutput).ShouldNot(ContainSubstring(".wego"))
-			Eventually(folderOutput).ShouldNot(ContainSubstring("apps"))
-			Eventually(folderOutput).ShouldNot(ContainSubstring("targets"))
+			Expect(folderOutput).ShouldNot(ContainSubstring(".wego"))
+			Expect(folderOutput).ShouldNot(ContainSubstring("apps"))
+			Expect(folderOutput).ShouldNot(ContainSubstring("targets"))
 		})
 	})
 })
--- a/test/acceptance/test/add_tests.go
+++ b/test/acceptance/test/add_tests.go
@@ -72,7 +72,7 @@ var _ = Describe("Weave GitOps Add Tests", func() {
 		})
 
 		By("And repos created have private visibility", func() {
-			Expect(getRepoVisibility(os.Getenv("GITHUB_ORG"), tip.appRepoName)).Should(ContainSubstring("true"))
+			Expect(getRepoVisibility(GITHUB_ORG, tip.appRepoName)).Should(ContainSubstring("true"))
 		})
 	})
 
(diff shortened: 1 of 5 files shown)


test/acceptance/test/utils.go
```diff
@@ -268,6 +277,16 @@ func runCommandAndReturnOutput(commandToRun string) (stdOut string, stdErr strin
 	return string(session.Wait().Out.Contents()), string(session.Wait().Err.Contents())
 }
 
+func generateTestInputs() TestInputs {
```
jpellizzari, line 280: I like this better. One thing to keep in mind is that multiple return values might make things a little more expressive, although 4 return values is probably pushing the limit.

test/acceptance/test/add_tests.go
```diff
@@ -360,53 +339,49 @@ var _ = Describe("Weave GitOps Add Tests", func() {
...
 		branchName := "test-branch-02"
-		uniqueSuffix := RandString(6)
-		appManifestFilePath := getUniqueWorkload("xxyyzz", uniqueSuffix)
-		workloadName := "nginx-" + uniqueSuffix
-		workloadNamespace := "my-nginx-" + uniqueSuffix
-		defaultSshKeyPath := os.Getenv("HOME") + "/.ssh/id_rsa"
+
 		wegoNamespace := "my-space"
-		url := "ssh://git@github.com/" + os.Getenv("GITHUB_ORG") + "/" + appRepoName + ".git"
+		url := "ssh://git@github.com/" + os.Getenv("GITHUB_ORG") + "/" + tip.appRepoName + ".git"
```
jpellizzari, line 354: I like to use `fmt.SprintF()` for long string concatenations: https://golang.org/pkg/fmt/#Sprintf

Purely stylistic, but it make things easier to read IMO.

test/acceptance/test/add_tests.go
```diff
@@ -209,24 +194,24 @@ var _ = Describe("Weave GitOps Add Tests", func() {
...
 		By("And I should see workload for app2 is deployed to the cluster", func() {
-			verifyWorkloadIsDeployed(workloadName2, workloadNamespace2)
+			verifyWorkloadIsDeployed(tip2.workloadName, tip2.workloadNamespace)
 		})
 
 		By("And repos created have private visibility", func() {
-			Expect(getRepoVisibility(os.Getenv("GITHUB_ORG"), appRepoName1)).Should(ContainSubstring("true"))
-			Expect(getRepoVisibility(os.Getenv("GITHUB_ORG"), appRepoName2)).Should(ContainSubstring("false"))
+			Expect(getRepoVisibility(os.Getenv("GITHUB_ORG"), tip1.appRepoName)).Should(ContainSubstring("true"))
+			Expect(getRepoVisibility(os.Getenv("GITHUB_ORG"), tip2.appRepoName)).Should(ContainSubstring("false"))
```
jpellizzari, line 206: Lots of repeated `os.Getenv` calls. I would read it once from the `env` and store it in a top level variable. That way, changing the variable or logic only needs to be changed once.

test/acceptance/test/utils.go
```diff
@@ -298,6 +317,7 @@ func gitAddCommitPush(repoAbsolutePath string, appManifestFilePath string) {
 	command := exec.Command("sh", "-c", fmt.Sprintf(`
                             cp -r %s %s &&
                             cd %s &&
+							sleep 10s &&
```
jpellizzari, line 320: :worried: Why do we need to `sleep` after changing directories?

jpellizzari (commented):
My only real concern is the `sleep` command in the shell call in `gitAddCommitPush`,

Everything else is just thoughts.

test/acceptance/test/utils.go
iahmad9: You are right, actually this sleep should be after hub create, as it takes few secs after the call to create the repo on GitHub.

test/acceptance/test/utils.go
iahmad9: I tried that first but then it felt like too many return values and there is good chance that we need more return values in the future. Moreover, it means for every single extra return for new tests, we need to update all tests.

test/acceptance/test/add_tests.go
iahmad9: Yeah, I am not happy with it either, let me do it next pass.

test/acceptance/test/utils.go
iahmad9: Moved to sleep to the right function.

test/acceptance/test/add_tests.go
iahmad9: Fixed.

jpellizzari (approved):
:+1: 

I do want to point out that placing arbitrary timeouts is a code smell for non-deterministic tests.

test/acceptance/test/utils.go
```diff
@@ -278,7 +297,8 @@ func initAndCreateEmptyRepo(appRepoName string, IsPrivateRepo bool) string {
                             mkdir %s &&
                             cd %s &&
                             git init &&
-                            hub create %s %s`, repoAbsolutePath, repoAbsolutePath, os.Getenv("GITHUB_ORG")+"/"+appRepoName, privateRepo))
+                            hub create %s %s &&
+							sleep 10s`, repoAbsolutePath, repoAbsolutePath, GITHUB_ORG+"/"+appRepoName, privateRepo))
```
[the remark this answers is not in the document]
iahmad9: I ll look into implementing polling around this instead of hardcoded timeout.

PR #416, opened 2021-06-30
Updated link in README.md to point to the latest release

jrryjcksn:
**What changed?**
The release link now points to the current release

**Why?**
We want users to use the new release


palemtnrider (approved):
LGTM

PR #426, opened 2021-07-01
add test for helm chart from git repo with app-config-url=NONE

jrryjcksn:
**What changed?**
Test added for adding a helm chart from a git repo with `--app-config-url=NONE`

**Why?**
This use case was not covered

**How did you test it?**
This is the test


josecordaz (approved):
LGTM

PR #429, opened 2021-07-02
Add k8s test isolation (via namespace)

jpellizzari:
Adds test isolation for `controller-runtime` tests. This is based on the `kustomization-controller` tests: https://github.com/fluxcd/kustomize-controller/blob/main/controllers/kustomization_controller_test.go

Thanks to @stefanprodan and @stealthybox for the help ^^

Also requires a namespace for fetching applications.


api/applications/applications.proto
```diff
@@ -53,7 +53,8 @@ message ListApplicationsResponse {
 }
 
 message GetApplicationRequest {
-    string application_name = 1; // The name of an application
```
bigkevmcd, line 56: I suspect these would be better as just `name` and `namespace` as you're asking for the `Application`

pkg/kube/kube.go
```diff
@@ -47,8 +47,8 @@ type Kube interface {
 	FluxPresent(ctx context.Context) (bool, error)
 	GetClusterName(ctx context.Context) (string, error)
 	GetClusterStatus(ctx context.Context) ClusterStatus
-	GetApplication(ctx context.Context, name string) (*wego.Application, error)
 	LabelExistsInCluster(ctx context.Context, label string) error
+	GetApplication(ctx context.Context, name, namespace string) (*wego.Application, error)
```
bigkevmcd, line 51: How about using [`types.NamespacedName`](https://pkg.go.dev/k8s.io/apimachinery/pkg/types#NamespacedName) which is more strongly typed version of this?

pkg/kube/kube_suite_test.go
```diff
@@ -1,13 +1,83 @@
...
+	}()
+
+	k8sClient = k8sManager.GetClient()
+	Expect(k8sClient).ToNot(BeNil())
+	close(done)
+}, 60)
+
+var _ = AfterSuite(func() {
+	By("tearing down the test environment")
+	err := testEnv.Stop()
```
bigkevmcd, line 67: You could simplify this to...

```go
Expect(testEnv.Stop()).To(Succeed())
```

pkg/kube/kube_suite_test.go
```diff
@@ -1,13 +1,83 @@
...
+	Expect(err).ToNot(HaveOccurred())
+})
+
+func init() {
+	rand.Seed(time.Now().UnixNano())
+}
+
+var letterRunes = []rune("abcdefghijklmnopqrstuvwxyz1234567890")
+
+func randStringRunes(n int) string {
```
bigkevmcd, line 77: how about just `randomString` instead of including runes in the name?

But, you could also just delegate to https://pkg.go.dev/k8s.io/apimachinery/pkg/util/rand#String ?

pkg/kube/kubehttp.go
```diff
@@ -168,7 +170,7 @@ func (c *KubeHTTP) SecretPresent(ctx context.Context, secretName string, namespa
 func (c *KubeHTTP) GetApplications(ctx context.Context, namespace string) ([]wego.Application, error) {
 	result := wego.ApplicationList{}
 
-	if err := c.Client.List(ctx, &result); err != nil {
+	if err := c.Client.List(ctx, &result, namespaceOpts(namespace)); err != nil {
```
bigkevmcd, line 173: I think you could simplify this to https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/client#InNamespace and remove the `namespaceOpts` function below.

```go
if err := c.Client.List(ctx, &result, client.InNamespace(namespace)); err != nil { }
```

pkg/kube/kubehttp_test.go
```diff
@@ -9,48 +9,26 @@ import (
...
-
 var _ = Describe("KubeHTTP", func() {
+	var (
+		namespace *corev1.Namespace
+		err       error
+	)
+
+	BeforeEach(func() {
+		namespace = &corev1.Namespace{}
+		namespace.Name = "kube-test-" + randStringRunes(5)
```
bigkevmcd, line 22: This could be shortened to...

```go
namespace = &corev1.Namespace{Name: "kube-test-" + randStringRunes(5)}
```

Or, if you were getting rid of `randStringrunes`...

```go
import "k8s.io/apimachinery/pkg/util/rand"

namespace = &corev1.Namespace{Name: "kube-test-" + rand.String(5)}
```

bigkevmcd (commented):
Some nice work in here...mostly minor suggestios.

api/applications/applications.proto
jpellizzari: Agreed! Changed

pkg/kube/kube.go
jpellizzari: Much better :+1:

luizbafilho (approved):

J-Thompson12 (approved):

PR #430, opened 2021-07-02
add pr tests

J-Thompson12:
Added acceptance tests for creating a pr. Going to add more in another PR

**Why?**

**How did you test it?**

**Release notes**

**Documentation Changes**

### DIFF
--- a/pkg/kube/kube.go
+++ b/pkg/kube/kube.go
@@ -207,7 +207,7 @@ func (k *KubeClient) runKubectlCmdWithInput(args []string, input []byte) ([]byte
 	return out, nil
 }
 func (k *KubeClient) LabelExistsInCluster(ctx context.Context, label string) error {
-	cmd := []string{"get", "app", "-l", fmt.Sprintf("weave-gitops.weave.works/app-identifier=%s", label)}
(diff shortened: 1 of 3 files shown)


pkg/kube/kube.go
```diff
@@ -207,7 +207,7 @@ func (k *KubeClient) runKubectlCmdWithInput(args []string, input []byte) ([]byte
 	return out, nil
 }
 func (k *KubeClient) LabelExistsInCluster(ctx context.Context, label string) error {
-	cmd := []string{"get", "app", "-l", fmt.Sprintf("weave-gitops.weave.works/app-identifier=%s", label)}
+	cmd := []string{"get", "app", "-l", fmt.Sprintf("weave-gitops.weave.works/app-identifier=%s", label), "--all-namespaces"}
```
josecordaz, line 210: I wonder if it would be ok to make the name of the function and the label parameter a bit more descriptive 🙂 . It looks like we are looking for a very specific label value rather than a label itself.

josecordaz (approved):
LGTM

pkg/kube/kube.go
J-Thompson12: I think if we end using more labels we will for sure need to do that. I am not sure if we ever plan on doing that though

PR #434, opened 2021-07-06
Add test case for wego app add --chart

rokshana-b:
**Why?**
Adding test cases for:
 - `wego app status`
 - `wego app add -h`
 - `wego app status -h`
 - `wego app list -h`
 - `wego flux -v`
 -  wego quits if 'flux-system' namespace is present


test/acceptance/test/add_tests.go
```diff
@@ -962,11 +959,17 @@ var _ = Describe("Weave GitOps Add Tests", func() {
...
+			folderOutput, _ := runCommandAndReturnStringOutput(fmt.Sprintf("cd %s && ls -al", repoAbsolutePath))
 			Expect(folderOutput).ShouldNot(ContainSubstring(".wego"))
 			Expect(folderOutput).ShouldNot(ContainSubstring("apps"))
 			Expect(folderOutput).ShouldNot(ContainSubstring("targets"))
 		})
+
+		By("And I can see wego app status", func() {
+			statusOutput := runCommandAndReturnSessionOutput(fmt.Sprintf("%s app status %s", WEGO_BIN_PATH, appName))
+			Eventually(statusOutput).Should(gbytes.Say(
+				`Latest successful deployment time: 202[0-9]-[0-9][0-9]-[0-9][0-9]T[0-9][0-9]:[0-9][0-9]:[0-9][0-9]Z\n.*\n.*gitrepository\/` + appName + `\s*True\s*Fetched revision: main\/.{40}\smain\/.{40}\s*False\s*\n.*\n.*kustomization\/` + appName + `\s*True\s*Applied revision: main\/.{40}\smain\/.{40}\s*False`))
```
iahmad9, line 971: Nit: For the readability of code, I would expand it to multiple lines.

rokshana-b: Made requested changes

jrryjcksn (approved):
This all looks okay to me but I'm concerned that the help tests are really brittle and are only helpful when someone inadvertently changes the output. Any legitimate change will break tests. Maybe we should consider storing the help strings in a separate file that we can pull in to both the command code and the tests. That way deliberate changes to the help text wouldn't break any tests. :thinking:  Let's discuss this at the standup.

jpellizzari:
> This all looks okay to me but I'm concerned that the help tests are really brittle and are only helpful when someone inadvertently changes the output. Any legitimate change will break tests. Maybe we should consider storing the help strings in a separate file that we can pull in to both the command code and the tests. That way deliberate changes to the help text wouldn't break any tests.  Let's discuss this at the standup.

I was going to leave similar feedback, but I realized that this PR is following suit with way the prior tests were written. I have created this issue to track the conversation: https://github.com/weaveworks/weave-gitops/issues/454

PR #437, opened 2021-07-06
The external config repository is always SourceTypeGit -- we shouldn't use the source type parameter

jrryjcksn:
**What changed?**
We no longer consult the source type parameter when deciding whether or not to create a deploy key in the external repo case. For a helm repository, we were failing to create a deploy key, even though the config repository is necessarily a git repository.

**Why?**
To correct the behavior when adding an upstream helm repository with an external configuration repository

**How did you test it?**
I tested it locally and there will be a new acceptance test going back shortly (which exposed the issue).

**Release notes**
No

**Documentation Changes**
No

### DIFF
--- a/pkg/services/app/add.go
+++ b/pkg/services/app/add.go
@@ -112,7 +112,6 @@ func (a *App) Add(params AddParams) error {
 		return err
 	}
 
-	fmt.Printf("Generating deploy key for repo %s ...\n", params.Url)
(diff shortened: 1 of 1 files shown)


J-Thompson12 (approved):

J-Thompson12:
Minor but if you are not passing in the type anymore then we dont need the source check inside createAndUploadDeployKey

jrryjcksn:
Actually, we do, because it is also used at the top of Add where we need to distinguish the two cases

PR #438, opened 2021-07-07
Refactor git providers

josecordaz:
**What changed?**
- For go git integration tests we now can set an id to the cache used in a specific Describe section.
- All git providers tests migrated to use ginkgo framework.

**Why?**
- It was getting hard to maintain.
- There were 2 bugs fixed when running this tests locally.
 -- The URL filter was not replacing values properly.
 -- When generating cache sometimes it was throwing Bad Credentials errors due to conflicts with other tests modifying GITHUB_TOKEN env variable.

**How did you test it?**
- By rerunning all the tests locally and through Github actions.


jrryjcksn (approved):
LGTM. 

Would it be possible at some point to add a comment explaining how to generate the cache files?

luizbafilho:
Do we still need to manually edit the cache files after generate them?

pkg/gitproviders/common.go
```diff
@@ -483,3 +473,12 @@ func NewUserRepositoryRef(domain, user, repoName string) gitprovider.UserReposit
 		},
 	}
 }
+
+func waitUntilRepoCreated(ownerType ProviderAccountType,owner, name string) error{
```
jpellizzari, line 477: Looks like a formatting miss here. I would advise configuring your editor to run `gofmt` on save. That ensures everyone's code looks roughly the same.

josecordaz:
> Do we still need to manually edit the cache files after generate them?

Sorry man, I'm confused. Why would we need to do something like that?. I probably know where the confusion is, though. Even though this runs as part of the unit tests job. I would like to see this as some sort of mix between unit tests and integration tests. Let's have a further discussion about this when I add documentation to clarify how this works in a follow up PR.

PR #445, opened 2021-07-08
Add readme and template for ADRs

palemtnrider:
**What changed?**
Adding a readme describing ADRs and a template

**Why?**
We need to start capturing architectural decisions and this is a lightweight which is commonly used.

**How did you test it?**
N/A

**Release notes**
N/A

**Documentation Changes**
N/A


bigkevmcd (commented):
Can I propose using https://github.com/npryce/adr-tools instead?

This reduces the cost of creating and maintaining the ADRs.

palemtnrider:
Let me give adr-tools a try.  I don't mind filling out a template but if that makes it even easier to capture the info I'm all for it.

palemtnrider:
@bigkevmcd I used adr-tools to generate #458 and it does simplify the process a bit.  I've updated the readme to reference adr-tools and updated the template to match the adr-tools one for those that don't want to install adr-tools

bigkevmcd (approved):

PR #450, opened 2021-07-08
Add check to download deps only once

jpellizzari:
Adds some Makefile magic to ensure the `dependencies` target only runs once.


luizbafilho (approved):
aha, nice, simple but effective.

josecordaz (approved):
LGTM

PR #452, opened 2021-07-08
Use go struct and yaml marshalling for adding applications

jpellizzari:
Changes our templating implementation to use the struct. This will ensure that any changes to the CRD will be reflected in the file that gets applied/saved.

The scope of this PR has expanded to move the `app_test` package into the `app` package to allow for testing private functions. This was necessary to add test coverage for things like the app hash label.

Shout-out to @bigkevmcd for the `cmp` package for struct comparison

### DIFF
--- a/pkg/services/app/add.go
+++ b/pkg/services/app/add.go
@@ -33,8 +33,9 @@ const (
 	SourceTypeGit  SourceType = "git"
 	SourceTypeHelm SourceType = "helm"
 
-	ConfigTypeUserRepo ConfigType = ""
-	ConfigTypeNone     ConfigType = "NONE"
+	ConfigTypeUserRepo        ConfigType = ""
+	ConfigTypeNone            ConfigType = "NONE"
+	WeGOAppIdentifierLabelKey            = "weave-gitops.weave.works/app-identifier"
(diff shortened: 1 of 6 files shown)


pkg/services/app/add.go
```diff
@@ -547,41 +550,28 @@ func (a *App) writeAppGoats(basePath string, name string, clusterName string, ma
...
-		return nil, errors.Wrap(err, "could not parse app yaml template")
+	gvk := wego.GroupVersion.WithKind(wego.ApplicationKind)
+	app := wego.Application{
+		TypeMeta: metav1.TypeMeta{
+			Kind:       gvk.Kind,
+			APIVersion: gvk.GroupVersion().String(),
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      params.Name,
+			Namespace: params.Namespace,
```
J-Thompson12, line 561: We still need to have the appHash for the label so we dont add two of the same applications

jpellizzari:
@J-Thompson12 nice catch! Fixed with expanded test coverage, PTAL.

J-Thompson12 (approved):

PR #457, opened 2021-07-09
add prlink to cli output

J-Thompson12:
addresses https://app.zenhub.com/workspaces/weaveworks-foss-603cb6b1d4dff9000e242dc4/issues/weaveworks/weave-gitops/374 outputs the pr link to the cli output.

use the cli output to merge the pr in the test

### DIFF
--- a/pkg/gitproviders/common.go
+++ b/pkg/gitproviders/common.go
@@ -31,8 +31,8 @@ type GitProviderHandler interface {
 	RepositoryExists(name string, owner string) (bool, error)
 	DeployKeyExists(owner, repoName string) (bool, error)
 	UploadDeployKey(owner, repoName string, deployKey []byte) error
-	CreatePullRequestToUserRepo(userRepRef gitprovider.UserRepositoryRef, targetBranch string, newBranch string, files []gitprovider.CommitFile, commitMessage string, prTitle string, prDescription string) error
-	CreatePullRequestToOrgRepo(orgRepRef gitprovider.OrgRepositoryRef, targetBranch string, newBranch string, files []gitprovider.CommitFile, commitMessage string, prTitle string, prDescription string) error
+	CreatePullRequestToUserRepo(userRepRef gitprovider.UserRepositoryRef, targetBranch string, newBranch string, files []gitprovider.CommitFile, commitMessage string, prTitle string, prDescription string) (string, error)
+	CreatePullRequestToOrgRepo(orgRepRef gitprovider.OrgRepositoryRef, targetBranch string, newBranch string, files []gitprovider.CommitFile, commitMessage string, prTitle string, prDescription string) (string, error)
 	GetAccountType(owner string) (ProviderAccountType, error)
 }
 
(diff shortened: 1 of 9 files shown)


pkg/gitproviders/common.go
```diff
@@ -260,11 +260,11 @@ func UploadDeployKey(owner, repoName string, deployKey []byte) error {
 func DeployKeyExists(owner, repoName string) (bool, error) {
 	return gitProviderHandler.(GitProviderHandler).DeployKeyExists(owner, repoName)
 }
-func CreatePullRequestToUserRepo(userRepRef gitprovider.UserRepositoryRef, targetBranch string, newBranch string, files []gitprovider.CommitFile, commitMessage string, prTitle string, prDescription string) error {
+func CreatePullRequestToUserRepo(userRepRef gitprovider.UserRepositoryRef, targetBranch string, newBranch string, files []gitprovider.CommitFile, commitMessage string, prTitle string, prDescription string) (string, error) {
```
jpellizzari, line 263: There are way too many arguments to this function. Anything more than 4 tends to get unwieldy. I realize it was set up that way before this PR, just wanted to comment.

jpellizzari: Given that this func returns a `string`, I don't know what the string represents. I assume a URL in this case? Can we return a golang `url.Url` type? Or possibly the pull request object that is being created?

pkg/services/app/add.go
```diff
@@ -690,11 +690,21 @@ func (a *App) createPullRequestToRepo(params AddParams, basePath string, repo st
 		}
 
 		orgRepoRef := gitproviders.NewOrgRepositoryRef(github.DefaultDomain, org, repoName)
-		return a.gitProviders.CreatePullRequestToOrgRepo(orgRepoRef, params.Branch, appHash, files, utils.GetCommitMessage(), fmt.Sprintf("wego add %s", params.Name), fmt.Sprintf("Added yamls for %s", params.Name))
+		prLink, err := a.gitProviders.CreatePullRequestToOrgRepo(orgRepoRef, params.Branch, appHash, files, utils.GetCommitMessage(), fmt.Sprintf("wego add %s", params.Name), fmt.Sprintf("Added yamls for %s", params.Name))
+		if err != nil {
+			return err
```
jpellizzari, line 695: Can we return a wrapped error here indicating what went wrong?

test/acceptance/test/add_tests.go
```diff
@@ -1075,15 +1077,17 @@ var _ = Describe("Weave GitOps Add Tests", func() {
 		})
 
 		By("And I run wego app add command for app", func() {
-			runWegoAddCommand(repoAbsolutePath, addCommand, WEGO_DEFAULT_NAMESPACE)
+			output, _ := runWegoAddCommandWithOutput(repoAbsolutePath, addCommand, WEGO_DEFAULT_NAMESPACE)
+			re := regexp.MustCompile(`(http|ftp|https):\/\/([\w\-_]+(?:(?:\.[\w\-_]+)+))([\w\-\.,@?^=%&amp;:/~\+#]*[\w\-\@?^=%&amp;/~\+#])?`)
```
jpellizzari, line 1081: What is this checking for? A valid URL? Should `ftp` be allowed?

J-Thompson12: no it probably shouldnt.

pkg/gitproviders/common.go
J-Thompson12: I can absolutely return the pr object and get the url string in add. Unfortunately I dont think I can lower the number of arguments. Gitproviders needs all of that info and I could take some of that info and generate it in common but that could limit future functionality.

pkg/services/app/add_test.go
```diff
@@ -472,22 +472,6 @@ var _ = Describe("Add", func() {
...
-			_, _, newBranch, files, _, _, _ := gitProviders.CreatePullRequestToUserRepoArgsForCall(0)
-			Expect(newBranch).To(Equal("wego-bf22e886ea99b1891c16bba5529d7f0b"))
-			Expect(files).To(Not(BeEmpty()))
-
-			Expect(gitClient.WriteCallCount()).To(Equal(0))
-			Expect(kubeClient.ApplyCallCount()).To(Equal(2))
-			Expect(fluxClient.CreateSecretGitCallCount()).To(Equal(1))
-			Expect(gitClient.CloneCallCount()).To(Equal(0))
-			Expect(gitProviders.CreatePullRequestToUserRepoCallCount()).To(Equal(1))
-		})
```
J-Thompson12, line 490: These tests cant be run now that createrepo is returning the url object. These tests are still covered in common_test.go

pkg/gitproviders/common.go
jpellizzari: Yeah I would say that would be a separate PR to make that change.

I would use a `struct` as an argument like we do on `AddParams` to make it a little more understandable.

jpellizzari (approved):

pkg/services/app/add_test.go
J-Thompson12: So the test cant get the url from the url object because it was never set. Before it never did anything with the url.  Now it errors out in add.go because that stuff doesnt exist. Anyway i realized these were duplicate tests anyway and are checked in acceptance and common_test.

PR #471, opened 2021-07-14
Use the app repo url to generate the app hash

### LINKED ISSUE #486: The app hash for the app CR is generated using the url of the config repo when using external repo config option.
**Describe the bug**

**Environment**
 - wego: [e.g. v0.1.0]
 - kubernetes:  [e.g. 1.20.4]
    - [ ] KinD - _version_]
    - [ ] k3s - _version_
    - [ ] cloud [e.g., EKS, AKS]  _version_
 - Browser + version: [e.g. chrome 74, safari 12, firefox 87]
 - 
**Affects versions**

**To Reproduce**

**Expected behavior**

**Config and Logs**
If applicable, add logs to help explain your problem. _please compress the output before attaching_ 
- [ ] wego config
- [ ] `wego` cli call output (and parameters)
- [ ] wego-controller logs
- [ ] Events from wego-* namespaces
- [ ] `kubectl cluster-info dump` 
- [ ] Prometheus alerts

**Screenshots**

**Additional context**

josecordaz:
Fixes #486 

**What changed?**
- Fixed app hash 
- Refactor to have one single point to generate the app hash

**Why?**
- The app hash generator was using the config repo url instead of the app repo url.
- Having a single point to generate the app has leads to easier maintenance.

**How did you test it?**
Updating unit tests

**Release notes**

**Documentation Changes**


J-Thompson12:
We use both the url or the app config url. It depends on which path its going down.

pkg/services/app/add.go
```diff
@@ -581,10 +581,10 @@ func makeWegoApplication(params AddParams) wego.Application {
 	return app
 }
 
-func generateAppYaml(params AddParams, repo string) ([]byte, error) {
+func generateAppYaml(params AddParams) ([]byte, error) {
 	app := makeWegoApplication(params)
 
-	appHash, err := utils.GetAppHash(repo, params.Path, params.Branch)
+	appHash, err := utils.GetAppHash(params.Url, params.Path, params.Branch)
```
J-Thompson12, line 587: If you change it here you need to also change where we are checking if the appHash exists already when adding a new app.

pkg/services/app/add.go
```diff
@@ -113,15 +113,36 @@ func (a *App) Add(params AddParams) error {
 		return errors.Wrap(err, "could not generate deploy key")
 	}
 
+	if err = a.setAppHash(params); err != nil {
```
jpellizzari, line 116: Adding this as state on the service will not work long term. The hash would persist between HTTP requests and there would be race conditions between `go` routines.

josecordaz: The intention is not to save the Add progress state here, we would need something else like a SyncMap where the app hash could be used as a key(or something similar). The actual intention is just to be a bit more aligned to POO as it helps with understanding. For instance `generateAppManifests` could be a behavior of the App struct (`(a *App) generateAppYaml(params AddParams)`) as it creates this YAML content for the App itself.

jpellizzari (commented):
@josecordaz I think this looks better, but can you add some detail to the PR description about why this change was necessary? I'm not sure I fully understand the context.

jpellizzari (approved):

Issue #473, opened 2021-07-14
Error on acceptance tests


josecordaz:
**Describe the bug**

```
Verify multiple apps dir can be added to the cluster using single repo for wego config [It]
  /home/runner/work/weave-gitops/weave-gitops/test/acceptance/test/add_tests.go:807

  Expected success, but got an error:
      <*errors.errorString | 0xc000028be0>: {
          s: "Error: Failed to find the resource nginx-r2li0n of type deploy, timeout reached",
      }
      Error: Failed to find the resource nginx-r2li0n of type deploy, timeout reached

  /home/runner/go/pkg/mod/github.com/onsi/gomega@v1.14.0/internal/assertion/assertion.go:38
```

The root of the problem seems to be this
![image.png](https://images.zenhubusercontent.com/606389311ead0cc4788a9a72/6c75f430-0f93-4592-872b-056cc6625ca1)

It seems that sometimes `hub create` throws this, not completely sure why.

**Environment**
 - wego: [e.g. v0.1.0]
 - kubernetes:  [e.g. 1.20.4]
    - [ ] KinD - _version_]
    - [ ] k3s - _version_
    - [ ] cloud [e.g., EKS, AKS]  _version_
    - [ ] other - _name_ _version_
 - Browser + version: [e.g. chrome 74, safari 12, firefox 87]
 - 
**Affects versions**

**To Reproduce**

**Expected behavior**

**Config and Logs**
If applicable, add logs to help explain your problem. _please compress the output before attaching_ 
- [ ] wego config
- [ ] `wego` cli call output (and parameters)
- [ ] wego-controller logs
- [ ] Events from wego-* namespaces
- [ ] `kubectl cluster-info dump` 
- [ ] Prometheus alerts

**Screenshots**

**Additional context**

josecordaz:
The actual problem seems to be flaky results due to corrupt binary by other builds

PR #474, opened 2021-07-14
Fix acceptance tests

josecordaz:
Fixes: #473

**What changed?**
Using a unique name to backup Wego binaries for smoke and integration tests to use.
Added logs to print out the commit used by the build and the one used by the tests.
Using random folder  as destination directory for hub clone.

**Why?**
It seems to be flaky tests results where the wego binary might get corrupted by other builds.

**How did you test it?**
Probably not a final fix but it might help.

**Release notes**

**Documentation Changes**


josecordaz:
Reminder: Remove repo field from `generateAppManifests` function

josecordaz:
Reminder: Return app hash instead

PR #480, opened 2021-07-15
fix_indentation_issue

D3nn:
**What changed?**
SilenceUsage and SilenceError were both incorrectly indent in the `cmd/wego/app/status/cmd.go`

**Why?**
General hygiene 

**How did you test it?**
Compiled and validated use of the `wego app status` command (which worked previously despite the indentation issue)

**Release notes**
NO

**Documentation Changes**
NO


jrryjcksn (approved):
LGTM

jpellizzari (approved):
LGTM. Maybe we can add a check to CI that if `go fmt` edits any files, it fails CI.

The issue with only some folks running formatters is that it dirties up the PRs when the next person comes along.

PR #481, opened 2021-07-15
Add basic git provider Oauth flow

### LINKED ISSUE #461: Implement git provider Oauth flow
Users will need to authenticate with a git provider to add an application. With the current CLI experience, it is assumed they have a Github token specified in their `env`.

Eventually, we will store the users oauth + refresh token in a secret on the cluster. This will be correlated with an OIDC ID token that we can use to look up the oauth + refresh token for making requests on the users behalf. The Secret will be RBAC'd to a specific user.

jpellizzari:
Closes #461 

Adds a first draft of an Oauth flow. This is NOT production ready and will require more work. Please note the README change to include the`.env` file. I am the only one with the `CLIENT_ID` and `CLIENT_SECRET` at the moment; working with IT on figuring out how to safely store that info.

Users will be presented with a login button for GH:
Screenshot from 2021-07-16 12-21-35

They will be taken to GH to authorize the app:
Screenshot from 2021-07-16 12-19-32

Upon return, they will be able to see their email address and have a simple logout action:

Screenshot from 2021-07-16 12-20-11

Notes:
* The authentication logic is currently baked into the `applications` server. I can invision a scenario where it is its own separate service.
* This PR currently forces a user to login to see Applications. That isn't required, but was used as an example.
* The auth token is returned to the UI and stored in `localStorage`. My understanding is that this is OK from a security perspective, but the future implementation will not necessarily return the token to the UI.
* Refresh token is not currently stored
* The OAuth2 callback is handled by the UI, rather than the backend. It is possible to do this on the backend and never return the "challenge" code to the frontend (which I have done as a defense-in-depth measure in other projects), but its weird on an SPA + API architecture.
* There is no CSRF protection at the moment in the form of the `state` parameter. That will be added later.
* Tests are very basic, as this code will change drastically in the near future.


api/applications/applications.proto
```diff
@@ -36,6 +36,25 @@ service Applications {
...
+    }
+
+    rpc Authenticate(AuthenticateRequest) returns (AuthenticateResponse) {
+        option (google.api.http) = {
+            post : "/v1/authenticate/{provider_name}"
+            body: "*"
+        };
+    }
+
+    rpc GetUser(GetUserRequest) returns (GetUserResponse) {
```
jpellizzari, line 53: Document these

api/applications/applications.proto
```diff
@@ -60,3 +79,39 @@ message GetApplicationRequest {
 message GetApplicationResponse {
     Application application = 1;
 }
+
+message GetAuthenticationProvidersRequest {
+
+}
+
+message OauthProvider {
```
jpellizzari, line 87: Document fields

cmd/wego-server/main.go
```diff
@@ -28,6 +29,7 @@ func init() {
 }
 
 func main() {
+	godotenv.Load()
```
jpellizzari, line 32: Document this in the readme

pkg/middleware/middleware.go
```diff
@@ -0,0 +1,75 @@
...
+	http.ResponseWriter
+	Status int
+}
+
+func (r *statusRecorder) WriteHeader(status int) {
+	r.Status = status
+	r.ResponseWriter.WriteHeader(status)
+}
+
+func WithLogging(h *runtime.ServeMux) http.Handler {
```
jpellizzari, line 25: Comments

pkg/server/server.go
```diff
@@ -55,3 +62,66 @@ func (s *server) GetApplication(ctx context.Context, msg *pb.GetApplicationReque
...
+	}
+
+	return &pb.AuthenticateResponse{
+		Token: token.AccessToken,
+	}, nil
+}
+
+func (s *server) GetUser(ctx context.Context, msg *pb.GetUserRequest) (*pb.GetUserResponse, error) {
+	token, err := middleware.ExtractToken(ctx)
+	if err != nil {
```
jpellizzari, line 102: Check the error here to make sure its really an unauthorized situation

ui/contexts/AppContext.tsx
```diff
@@ -1,45 +1,75 @@
...
+// Due to the way the grpc-gateway typescript client is generated,
+// we need to wrap each individual call to ensure the auth header gets
+// injected into the underlying `fetch` requests.
+// This saves us from having to rememeber to pass it as an arg in every request.
+function wrapClient<T>(client: any): T {
+  const wrapped = {};
+
+  _.each(client, (func, name) => {
+    wrapped[name] = (payload, options: RequestInit = {}) => {
+      const token = getToken();
```
jpellizzari, line 17: Check for no token here?

ui/contexts/AppContext.tsx
```diff
@@ -1,45 +1,75 @@
...
-  };
+    // Fetch the user once at app startup
+    appsClient
+      .GetUser({})
+      .then(({ user }) => {
+        setUser(user);
+      })
+      .catch((err) => {
+        console.error(err);
+        console.log(location.pathname);
```
jpellizzari, line 62: remove

ui/lib/storage.ts
```diff
@@ -0,0 +1,21 @@
...
+    localStorage.setItem(TOKEN_KEY, token);
+  }
+}
+
+export function getToken() {
+  if (window && window.localStorage) {
+    return localStorage.getItem(TOKEN_KEY);
+  }
+
+  return null;
```
jpellizzari, line 14: ?

ui/pages/Auth.tsx
```diff
@@ -0,0 +1,105 @@
...
+    padding-left: 65px;
+  }
+
+  &.gitlab {
+    background-color: orange;
+    color: white;
+  }
+`;
+
+// const LoginIcon = styled.i`
```
jpellizzari, line 64: Comment or remove

ui/pages/OAuthCallback.tsx
```diff
@@ -0,0 +1,26 @@
...
+};
+
+function OauthCallback({ className }: Props) {
+  const { authenticate, error } = useAuth();
+  const { query, navigate } =
+    useNavigation<{ code: string; provider: string; next?: string }>();
+
+  React.useEffect(() => {
+    authenticate(query.code, query.provider).then(() => {
+      navigate(query.next ? `/${query.next}` : PageRoute.Applications);
```
jpellizzari, line 19: Type safety on `query.next`?

jpellizzari: more trouble than its worth for now

ui/contexts/AppContext.tsx
jpellizzari: Passing through empty string/null to the `Authorization` header is probably fine for now.

josecordaz:
I tried to run this but got an issue at the time of opening the page:
image
I created my own test github app.
I updated the values and rename the file to be `.env`. Any chance I need to change something else before running this?

jpellizzari:
@josecordaz This is an issue with macOS. There is a file descriptor limit that you can increase:

https://superuser.com/questions/433746/is-there-a-fix-for-the-too-many-open-files-in-system-error-on-os-x-10-7-1

I need to tune that watch job to only look in certain directories. It is too optimistic right now.

pkg/gitproviders/common.go
```diff
@@ -34,17 +42,41 @@ type GitProviderHandler interface {
...
+}
+
+// TODO: currently, this always returns the github implementation.
+// This eventually needs to return the different git provider implementations.
+func New(providerName ProviderName) (defaultGitProviderHandler, error) {
+	switch providerName {
+	case ProviderNameGithub:
+		return defaultGitProviderHandler{}, nil
+	}
+	return defaultGitProviderHandler{}, fmt.Errorf("provider name %s is not supported", providerName)
```
palemtnrider, line 71: Since we are returning an error, we should return nil as the defaultGitProviderHander

pkg/gitproviders/common.go
```diff
@@ -34,17 +42,41 @@ type GitProviderHandler interface {
...
 
-func New() defaultGitProviderHandler {
-	return defaultGitProviderHandler{}
+type User struct {
+	Email string
+}
+
+// TODO: currently, this always returns the github implementation.
+// This eventually needs to return the different git provider implementations.
+func New(providerName ProviderName) (defaultGitProviderHandler, error) {
```
palemtnrider, line 66: Since this is exported and in a common file we should have a go doc comment. Same comment for the other exported functions

pkg/gitproviders/common.go
```diff
@@ -449,6 +481,23 @@ func (h defaultGitProviderHandler) CreatePullRequestToOrgRepo(orgRepRef gitprovi
 	return pr, nil
 }
 
+func (h defaultGitProviderHandler) OauthConfig() OauthProviderConfig {
```
palemtnrider, line 484: Can you add a bit of clarification on these two functions?  They are off an object, defaultGitProviderHander, but don't use the object in the implementation.  Having comments will help us when we return to look at this code.

pkg/gitproviders/github.go
```diff
@@ -61,3 +66,47 @@ func WithGithubProviderHandler(handler GithubProviderHandler, fun func() error)
 	}()
 	return fun()
 }
+
+func (h DefaultGithubProviderHandler) OauthConfig() OauthProviderConfig {
+	return &oauth2.Config{
```
palemtnrider, line 71: I'd prefer this function return error if either of the environment variables are unset instead of returning an OauthProviderConfig that won't work.

pkg/gitproviders/gitprovidersfakes/fake_git_provider_handler.go
```diff
@@ -460,6 +486,124 @@ func (fake *FakeGitProviderHandler) GetAccountTypeReturnsOnCall(i int, result1 g
...
+	fake.getUserMutex.Lock()
+	ret, specificReturn := fake.getUserReturnsOnCall[len(fake.getUserArgsForCall)]
+	fake.getUserArgsForCall = append(fake.getUserArgsForCall, struct {
+		arg1 context.Context
+		arg2 *oauth2.Token
+	}{arg1, arg2})
+	stub := fake.GetUserStub
+	fakeReturns := fake.getUserReturns
+	fake.recordInvocation("GetUser", []interface{}{arg1, arg2})
+	fake.getUserMutex.Unlock()
```
palemtnrider, line 499: Can you elaborate on unlocking here instead of a defer Unlock() right after taking the lock?

palemtnrider (commented):
Partial review comments so far...

pkg/gitproviders/common.go
jpellizzari: We are returning a struct by value, so we cannot return `nil` here.

pkg/gitproviders/common.go
jpellizzari: The `defaultGitProviderHandler` needs to match the `GitProviderHandler` interface, so that is why the function users a receiver.

pkg/gitproviders/gitprovidersfakes/fake_git_provider_handler.go
jpellizzari: This is generated code.

josecordaz:
> @josecordaz This is an issue with macOS. There is a file descriptor limit that you can increase:
> 
> https://superuser.com/questions/433746/is-there-a-fix-for-the-too-many-open-files-in-system-error-on-os-x-10-7-1
> 
> I need to tune that watch job to only look in certain directories. It is too optimistic right now.

I'm going to try to tune the watch job. I already tried to modify the values of maxfiles and maxfilesperproc a few times but it still doen't work. I wonder if I need to set them to a higher number 🤔. This was my last attempt:
image

jpellizzari:
> > @josecordaz This is an issue with macOS. There is a file descriptor limit that you can increase:
> > https://superuser.com/questions/433746/is-there-a-fix-for-the-too-many-open-files-in-system-error-on-os-x-10-7-1
> > I need to tune that watch job to only look in certain directories. It is too optimistic right now.
> 
> I'm going to try to tune the watch job. I already tried to modify the values of maxfiles and maxfilesperproc a few times but it still doen't work. I wonder if I need to set them to a higher number . This was my last attempt:

@josecordaz The `npm run watch` job is not necessary for running the server. It just runs tests on file changes

jpellizzari:
Closing in favor of this approach instead: https://github.com/weaveworks/weave-gitops/issues/498

josecordaz:
I was able to fix the `too many open files` issue with this https://wilsonmar.github.io/maximum-limits/ 🙂

PR #488, opened 2021-07-17
Use path for kustomize apps and chart for helm apps to generate app hash

josecordaz:
**What changed?**
App hash generator will use Chart parameter for helm type apps

**Why?**
The hash generator was using Path parameter all the time, which had the same value when adding more than 1 helm app, leading to  duplicate app hashes, resulting in failures when trying to add two different charts from the same helm repository.

**How did you test it?**
Manually

**Release notes**

**Documentation Changes**


pkg/services/app/add.go
```diff
@@ -132,6 +132,23 @@ func (a *App) Add(params AddParams) error {
 	}
 }
 
+func getAppHash(params AddParams) (string, error) {
+	var appHash string
+	var err error
+	if DeploymentType(params.DeploymentType) == DeployTypeHelm {
+		appHash, err = utils.GetAppHash(params.Url, params.Chart, params.Branch)
```
jpellizzari, line 139: Since this `utils.GetAppHash` function accepts three strings, it doesn't really do anything that is application-specific. I can use any three strings and get a valid return value.

Can we change this hashing function to accept `AddParams` and return a hash? That way, to the user of this function it reads as "if you give me the same `AddParams`, I will give you the same content hash every time` (with the caveat that we only care about certain fields).

josecordaz: I like that

josecordaz: I changed the function `utils.GetAppHash(url, path, branch string)` to be `utils.GetHash(inputs ...string)`(A bit more open as it will now accept N amount of strings). BTW there is a function `getAppHash` already that receives AddParams which I believe fits with the quoated statements you mention.

pkg/services/app/add.go
```diff
@@ -132,6 +132,23 @@ func (a *App) Add(params AddParams) error {
 	}
 }
 
+func getAppHash(params AddParams) (string, error) {
+	var appHash string
+	var err error
+	if DeploymentType(params.DeploymentType) == DeployTypeHelm {
+		appHash, err = utils.GetHash(params.Url, params.Chart, params.Branch)
```
jpellizzari, line 139: I would argue that this `GetHash` function isn't providing any abstraction anymore; it is just wrapping the native `go` libraries to create a content hash. 

It also doesn't get called anywhere but this package (as far as I can tell), so I would move that code here to avoid unnecessary indirection.

jpellizzari (approved):
:+1: with a note about reducing some unnecessary indirection

pkg/services/app/add.go
josecordaz: yes, I agree, let's fix that as well

PR #490, opened 2021-07-19
Return errors rather than dropping them during PR creation

jrryjcksn:
**What changed?**
- We now return wrapped errors during PR creation where we were previously ignoring them and returning nil (leaving PRs uncreated)
- Removed `fluxops` package as it was no longer needed

**Why?**
So that PR creation will fail appropriately if we can't retrieve github credentials.

**How did you test it?**
Unit tests

**Release notes**
No

**Documentation Changes**
None


pkg/flux/logs_test.go
```diff
@@ -110,48 +103,26 @@ func TestSetup(t *testing.T) {
...
-	binPath := fmt.Sprintf("%v/flux-%v", fluxPath, version.FluxVersion)
-	require.FileExists(t, binPath)
-
-	version.FluxVersion = "0.12.0"
-	SetupFluxBin()
-	require.NoFileExists(t, binPath)
-	binPath = fmt.Sprintf("%v/flux-%v", fluxPath, version.FluxVersion)
-	require.FileExists(t, binPath)
+	defer os.RemoveAll(dir)
+	override.WithOverrides(func() override.Result {
```
jrryjcksn, line 109: I added this because I noticed the test was running against the user's actual home dir. Possibly we should pull out all "os" operations (UserHomeDir, Exit, Stat, etc.) into a separate client object for mocking :thinking:

pkg/services/app/add.go
```diff
@@ -673,24 +672,19 @@ func (a *App) createPullRequestToRepo(params AddParams, basePath string, repo st
 
 	owner, err := getOwnerFromUrl(repo)
 	if err != nil {
-		return nil
```
jpellizzari, line 676: :cold_sweat: 

I'm trying to think of a way to catch this with linting...

pkg/services/app/add.go
```diff
@@ -673,24 +672,19 @@ func (a *App) createPullRequestToRepo(params AddParams, basePath string, repo st
...
-		if err != nil {
-			return nil
-		}
-
-		orgRepoRef := gitproviders.NewOrgRepositoryRef(github.DefaultDomain, org, repoName)
+		orgRepoRef := gitproviders.NewOrgRepositoryRef(github.DefaultDomain, owner, repoName)
 		prLink, err := a.gitProviders.CreatePullRequestToOrgRepo(orgRepoRef, params.Branch, appHash, files, utils.GetCommitMessage(), fmt.Sprintf("wego add %s", params.Name), fmt.Sprintf("Added yamls for %s", params.Name))
 		if err != nil {
-			return fmt.Errorf("unable to create pull request: %s", err)
+			return fmt.Errorf("unable to create pull request: %w", err)
```
jpellizzari, line 687: :+1:

jpellizzari (approved):

pkg/services/app/add.go
jrryjcksn: Boy, that would be nice!

PR #495, opened 2021-07-19
Using git-providers to create test repos instead of hub as it seems to be unstable to create repositories

josecordaz:
**What changed?**
Using git-providers to create test repositories.

**Why?**
`hub` was used previously for the same purpose. It seems unstable to create them. Sometimes throws this particular error:
image

**How did you test it?**
Integrations tests should run more stable now.

**Release notes**

**Documentation Changes**


J-Thompson12:
I think this looks fine but if we are going to remove hub for create the repo we might as well stay consistent and use git providers for all the other repo actions like clone and delete.

josecordaz:
> I think this looks fine but if we are going to remove hub for create the repo we might as well stay consistent and use git providers for all the other repo actions like clone and delete.

Right, that would be great. Just that git-providers doesn't work for cloning. Probably we can use `git` directly instead. 🤔

J-Thompson12 (approved):
this works for me then

Issue #470, opened 2021-07-13
Re-work GitProviderHandler interface


jpellizzari:
Currently, the `GitProviderHandler` interface returns the `githubProviderHandler` struct by default. This needs to be changed to return different `GitProvider` implementations (ie Gitlab, Bitbucket). The `.New()` function should return the different git provider implementations, each of which matches the `GitProviderHandler` interface.

jpellizzari:
Closing, as this is probably no longer relevant as of our Auth ADR and new #548 epic

PR #496, opened 2021-07-19
Passing git provider token in each add call

### LINKED ISSUE #467: Re-work Add() function to use key from parameters
Currently, the App service uses a single git client with a key passed in when the service is instantiated. 

This will not work for a multi-user API server, and we need to pass in the auth credentials when the `.Add()` command is invoked.

The `AddParams` might be a good place to add the git auth info.

luizbafilho:
**What changed?**
- closes #467 
- Making Add use git provider token passed in each call

**Why?**
- Making it compatible with a long-running app service.

**How did you test it?**

**Release notes**

**Documentation Changes**


jpellizzari (approved):
Nice work! I also created this issue: https://github.com/weaveworks/weave-gitops/issues/470

I think these changes address the major issues raised in this issue. @luizbafilho, what do you think?

luizbafilho:
It does address, we just need to set up the `gitprovider.Config` with the correct provider name when initializing it. That could be done by parsing the repo url.

PR #497, opened 2021-07-20
Fetch source and automation data for a given application

### LINKED ISSUE #491: Store references to Kustomization and Source k8s objects
When we `add` an app, we will need to save a reference to the flux Kustomization/HelmRelease and Source objects that are related. This will allow us to draw the reconciliation graph later.

After chatting with Flux maintainers, the best way to express the relationship between `Application` and `Kustomization | HelmRelease` and `Source` is via an `ownerReference`.

Note that this has garbage collection implications. K8s documentation on `ownerReference`:
https://kubernetes.io/docs/concepts/workloads/controllers/garbage-collection/

https://weaveworks.slack.com/archives/C0CG4EP6F/p1626710106321700

Implementation notes:
* I would imagine we would bake in an owner reference to the `KustomizationSpec |SourceSpec` when we create them

jpellizzari:
Closes #491 

In order to retrieve the sources and automatons for a given `Application`, we need to know some things about the related objects. We assume that a source and automation are named the same as the application.

A temporary UI to demonstrate:
Screenshot from 2021-07-21 07-58-20


josecordaz (approved):
LGTM

palemtnrider (commented):
Mainly minors changes to increase clarity.

pkg/server/server.go
```diff
@@ -44,14 +51,110 @@ func (s *server) ListApplications(ctx context.Context, msg *pb.ListApplicationsR
...
+	}
+
+	name := types.NamespacedName{Name: app.Name, Namespace: app.Namespace}
+
+	if err := s.kube.GetResource(ctx, name, src); err != nil {
+		return nil, fmt.Errorf("could not get source: %w", err)
+	}
+
+	if err := s.kube.GetResource(ctx, name, automation); err != nil {
+		return nil, fmt.Errorf("could not get automation: %w", err)
```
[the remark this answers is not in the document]
jpellizzari: @palemtnrider `Deployment` is already a word used in Kubernetes. Do we not already call these things "GOAT" for "GitOps Automation" (or at least that is what I thought it was called)?

luizbafilho (approved):

api/applications/applications.proto
```diff
@@ -38,10 +38,22 @@ service Applications {
...
 }
 
+// This object represents a single condition for a Kubernetes object.
+// It roughly matches the Kubernetes type defined here: https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1#Condition
+message Condition {
+    string type      = 1;
+    string status    = 2;
+    string reason    = 3;
+    string message   = 4;
+    string timestamp = 5;
```
palemtnrider, line 48: Can you elaborate on using string as the type instead of timestamp?  Having timestamp instead of strings makes it easier for non-human interfaces to consume. Also, a UI will have an easier time converting to local time when displaying to the user.

api/applications/applications.proto
```diff
@@ -38,10 +38,22 @@ service Applications {
...
+
 message Application {
-    string name = 1;  // The name of the application
-    string path = 2;  // The file path where the k8s yaml files for this application are stored.
-    string url  = 3;  // The git repository URL for this application
+    string   name                            = 1;  // The name of the application
+    string   path                            = 2;  // The file path where the k8s yaml files for this application are stored.
+    string   url                             = 3;  // The git repository URL for this application
+    repeated Condition source_conditions     = 4;  // A list of conditions for the Source related to this Application
+    repeated Condition automation_conditions = 5;  // A list of conditions for the Kustomization or HelmRelease for this application
```
palemtnrider, line 56: the CRD uses the term `deployment` instead of `automation`

pkg/server/server.go
```diff
@@ -44,14 +51,110 @@ func (s *server) ListApplications(ctx context.Context, msg *pb.ListApplicationsR
...
 func (s *server) GetApplication(ctx context.Context, msg *pb.GetApplicationRequest) (*pb.GetApplicationResponse, error) {
 	app, err := s.kube.GetApplication(ctx, types.NamespacedName{Name: msg.Name, Namespace: msg.Namespace})
-
 	if err != nil {
 		return nil, fmt.Errorf("could not get application: %s", err)
 	}
 
+	src, automation, err := findFluxObjects(app)
+	if err != nil {
+		return nil, fmt.Errorf("could not get flux objects for application: %w", err)
```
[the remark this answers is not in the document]
palemtnrider: I don't see the app name being added to the error here or in findFluxObject?

pkg/server/server.go
```diff
@@ -44,14 +50,110 @@ func (s *server) ListApplications(ctx context.Context, msg *pb.ListApplicationsR
...
+// Returns k8s objects that can be used to find the cluster objects.
+// The first return argument is the source, the second is the automation
+func findFluxObjects(app *wego.Application) (client.Object, client.Object, error) {
+	st := app.Spec.SourceType
+	if st == "" {
+		// Apps that were created before the SourceType field exists will not have a SourceType defined.
+		// Assume git, since thats what the CLI defaults to.
+		st = wego.SourceTypeGit
+	}
+
```
palemtnrider, line 111: It's possible the only one of the objects exists when findFluxObjects is called and I think that should be OK.  If only the source object is present, we should still return it and vice versa for the deployment type.  In fact, neither may exist in the cluster yet and that should be OK.

pkg/services/app/app.go
```diff
@@ -16,11 +16,13 @@ type DeploymentType string
 type SourceType string
 
 const (
-	DeployTypeKustomize DeploymentType = "kustomize"
-	DeployTypeHelm      DeploymentType = "helm"
+	// TODO: use wego.DeployType and wego.SourceType everywhere
```
palemtnrider, line 19: we should use the ones from application_types here instead of defining them again.

palemtnrider (changes requested):
Looking go so far.  I suggested a few items I'd like to see cleaned up before we merge.

api/applications/applications.proto
jpellizzari: Looks like I used an `int32` in the Flux UI:
https://github.com/fluxcd/webui/blob/f48d23e5095a04949ad276fc6da1349f380df457/pkg/clustersserver/clustersserver.go#L712

Which comes back in seconds. Hopefully enough resolution?:
https://github.com/fluxcd/webui/blob/f48d23e5095a04949ad276fc6da1349f380df457/ui/pages/Events.tsx#L34

Will fix!

pkg/services/app/app.go
jpellizzari: See the comment on the line below. It would muddy up the intent of this PR. Follow up issue created.

api/applications/applications.proto
jpellizzari: From another comment that may have shown as "outdated":

> `Deployment` is already a word used in Kubernetes. Do we not already call these things "GOAT" for "GitOps Automation" (or at least that is what I thought it was called)?

So I think we should discuss changing that command-line flag for this reason ^^. That would be outside of the scope of this PR, so I will create a follow up and change this to be consistent for now.

pkg/server/server.go
```diff
@@ -44,14 +51,110 @@ func (s *server) ListApplications(ctx context.Context, msg *pb.ListApplicationsR
...
+	if src == nil {
+		return nil, nil, errors.New("invalid source type")
+	}
+
+	at := app.Spec.DeploymentType
+	if at == "" {
+		// Same as above, default to kustomize to match CLI default.
+		at = wego.DeploymentTypeKustomize
+	}
+	var automation client.Object
```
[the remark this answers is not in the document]
jpellizzari: See above re `Deployment`

api/applications/applications.proto
jpellizzari: Changed to `int32`

pkg/server/server.go
jpellizzari: Changed to return `nil` with no error in the event of a "NotFound" error from kubernetes

api/applications/applications.proto
jpellizzari: Changed to `deployment_conditions`

pkg/server/server.go
```diff
@@ -44,14 +51,110 @@ func (s *server) ListApplications(ctx context.Context, msg *pb.ListApplicationsR
...
+	switch at {
+	case wego.DeploymentTypeHelm:
+		automation = &helmv2.HelmRelease{}
+	case wego.DeploymentTypeKustomize:
+		automation = &kustomizev1.Kustomization{}
+	}
+
+	if automation == nil {
+		return nil, nil, errors.New("invalid automation type")
```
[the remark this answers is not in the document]
jpellizzari: See the second review in this PR for further discussion

palemtnrider (approved):

Issue #322, opened 2021-06-14
Pause GitOps automation for an application


palemtnrider:
**Short description**
As an Application operator, I want to pause the gitops automation for application foo so that I can manipulate the application manifests without impacting my running application.

**Acceptance criteria**
**given** cluster with wego installed and application foo added via `wego app add foo`
**when** issuing `wego app pause foo`
**then** gitops runtime doesn't pull updates from my repo when I commit and push to the branch and/or path I used in `wego app add`
**when** issuing `wego app pause foo` a second time
**then** an error isn't displayed

**given** cluster with wego installed and application foo paused via `wego app pause foo`
**when** issuing `wego app unpause foo`
**then** gitops runtime resumes  pull updates from my repo when I commit and push to the branch and/or path I used in `wego app add`
**when** issuing `wego app unpause foo` a second time
**then** an error is not displayed (but an info message saying already reconciling is displayed)

Must work on app added with and without branch and path values

**Work breakdown**
(optional)Task breakdown of the work
- [ ] _create X_
- [ ] _design X_
- [ ] _..._

J-Thompson12:
I need to clarify some things on this task. Basically what the assumption is about where this command will be run.

1. The command is always run from inside the repository so the code has access to the target yaml and can add suspend to it.

2. The command can be run from anywhere and the use the app name to get the url from the app describe and clone the repo to a tmp dir then add suspend to it and push. 

3. A possible third option that is the correct one that I havent thought of.

jpellizzari:
Re-opening to assign to QA since this is user facing

jpellizzari:
Closing in favor of #528

PR #501, opened 2021-07-20
App pause

J-Thompson12:
**What changed?**
added pause and unpause features.

**How it was Tested**
Tested locally. Going to create an issue to add acceptance tests for it.

**Docs?**
https://github.com/weaveworks/weave-gitops-docs/pull/59

### DIFF
--- a/pkg/services/app/unpause.go
+++ b/pkg/services/app/unpause.go
@@ -1,6 +1,6 @@
 package app
 
-import "github.com/weaveworks/weave-gitops/pkg/flux"
+import wego "github.com/weaveworks/weave-gitops/api/v1alpha1"
 
 type UnpauseParams struct {
 	Name      string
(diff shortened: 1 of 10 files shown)


pkg/services/app/unpause.go
```diff
@@ -0,0 +1,45 @@
...
+	"context"
+	"fmt"
+)
+
+type UnpauseParams struct {
+	Name      string
+	Namespace string
+}
+
+func (a *App) Unpause(params UnpauseParams) error {
```
jpellizzari, line 13: The `Pause` and `Unpause` implementations are almost identical. Would it be possible to combine the logic into one function, and pass in the desired operation (`pause` or `unpause`) as an argument?

You could still keep the exported function separate, but they would be thin wrappers around a single `pauseOrUnpause` function that decides which flux operation to do.

jpellizzari:
Closes #322

jpellizzari (approved):
LGTM, nice work. Only nits, merge at your discretion.

PR #508, opened 2021-07-22
Sanitize cluster name

D3nn:
**What changed?**
Sanitized the cluster names, which comply now with RFC 1053 naming used by K8s

**Why?**
Creating clusters with invalid cluster names would cause failures

**How did you test it?**
Create unit-test case for this with a number of examples.

**Release notes**

**Documentation Changes**
Probably should add language about the validation possibly changing the cluster name in weave-gitops-docs

### DIFF
--- a/pkg/kube/kube_test.go
+++ b/pkg/kube/kube_test.go
@@ -9,8 +9,10 @@ import (
 	wego "github.com/weaveworks/weave-gitops/api/v1alpha1"
 	"k8s.io/apimachinery/pkg/types"
 
-	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/ginkgo/extensions/table"
+
 	"github.com/weaveworks/weave-gitops/pkg/kube"
 	"github.com/weaveworks/weave-gitops/pkg/runner/runnerfakes"
 )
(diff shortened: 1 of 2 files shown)


pkg/kube/kube_test.go
```diff
@@ -129,6 +129,39 @@ var _ = Describe("GetClusterName", func() {
 	})
 })
 
+var _ = Describe("FixInvalidClusterName", func() {
```
jpellizzari, line 132: There is a ginkgo extension that can help make these independent test cases: https://pkg.go.dev/github.com/onsi/ginkgo/extensions/table

jpellizzari (approved):

jpellizzari (commented):
@D3nn I just remembered we have the KubeHTTP implementation as well. Could you add this to the corresponding method?

jrryjcksn (approved):
LGTM

pkg/kube/kube_test.go
D3nn: Just pushed a change to use the extension.

D3nn:
@jpellizzari:
> @D3nn I just remembered we have the KubeHTTP implementation as well. Could you add this to the corresponding method?

From looking at the kubehttp.go code, it looks like it's purpose is to query for existing objects already deployed (including of course the creation of the cluster itself.  Given that the objects are already deployed in the cluster, it's safe to presume that the cluster name for the cluster would already be valid.  NO?

jpellizzari:
> @jpellizzari:
> 
> > @D3nn I just remembered we have the KubeHTTP implementation as well. Could you add this to the corresponding method?
> 
> From looking at the kubehttp.go code, it looks like it's purpose is to query for existing objects already deployed (including of course the creation of the cluster itself. Given that the objects are already deployed in the cluster, it's safe to presume that the cluster name for the cluster would already be valid. NO?

:+1: I think I understand. I'm wondering if both `kubectl` and the `go-client` parse the same config. I guess we can wait and find out.

PR #511, opened 2021-07-23
256/remove application - refactor code to collect resource information

jrryjcksn:
**What changed?**
As I started to consider how to find all the resources to remove, I noticed that the necessary information was spread out through the code and based on AddParams. This collects all information about created resources in an AppResourceInfo structure (derived from a wego.Application) so that it will stay consistent with applications and be available to the remove code (which doesn't have AddParams).

**How did you test it?**
Other than the first part of remove (which is covered by unit tests), this didn't change behavior and is tested by the existing unit tests.

**Release notes**
No

**Documentation Changes**
No


pkg/services/app/add.go
```diff
@@ -692,14 +717,136 @@ func (a *App) createPullRequestToRepo(params AddParams, gitProvider gitproviders
...
-	prLink, err := gitProvider.CreatePullRequestToUserRepo(userRepoRef, params.Branch, appHash, files, utils.GetCommitMessage(), fmt.Sprintf("wego add %s", params.Name), fmt.Sprintf("Added yamls for %s", params.Name))
+	prLink, err := gitProvider.CreatePullRequestToUserRepo(userRepoRef, info.Spec.Branch, appHash, files, utils.GetCommitMessage(), fmt.Sprintf("wego add %s", info.Name), fmt.Sprintf("Added yamls for %s", info.Name))
 	if err != nil {
 		return fmt.Errorf("unable to create pull request: %w", err)
 	}
 	a.logger.Println("Pull Request created: %s\n", prLink.Get().WebURL)
 	return nil
 }
 
+func (a *App) getAppResourceInfo(app wego.Application, clusterName string) *AppResourceInfo {
```
jpellizzari, line 723: Can this just be a pure function with no receiver? It might signal the intent better.

pkg/services/app/add.go
```diff
@@ -692,14 +717,136 @@ func (a *App) createPullRequestToRepo(params AddParams, gitProvider gitproviders
...
+	result := "GitRepository"
+
+	if a.Spec.SourceType == "helm" {
+		result = "HelmRepository"
+	}
+
+	return result
+}
+
+func (a *AppResourceInfo) sinkKind() string {
```
jpellizzari: What does `sink` mean in this case?

jpellizzari (approved):
:+1: Only nits and questions. As I mentioned in Slack, similar changes around `SourceType` here: https://github.com/pulls?utf8=%E2%9C%93&q=is%3Aopen+is%3Apr+author%3Ajpellizzari

pkg/services/app/add.go
jrryjcksn: Good point. An earlier iteration was referencing something from the app and I neglected to make that change once it no longer did

pkg/services/app/add.go
jrryjcksn: Breaking each app automation into a "source" and a "sink" (input from repo, output to cluster). I'm happy to use some other terminology if you have any ideas.

jpellizzari: This appears to be returning an "automation" kind, which we are referring to as a `DeploymentType` elsewhere. Maybe `deployKind()` for now?

jrryjcksn: Sure, that seems reasonable to me

pkg/services/app/remove.go
```diff
@@ -0,0 +1,56 @@
...
+
+	wego "github.com/weaveworks/weave-gitops/api/v1alpha1"
+)
+
+type AutomationManifestPaths struct { // source for automation isn't currently stored
+	AppDirAutomationPath    string
+	TargetDirAutomationPath string
+}
+
+func dirExists(d string) bool {
```
josecordaz, line 18: small nit: this function already exists in utils, works for both files and directories

pkg/services/app/remove.go
```diff
@@ -0,0 +1,56 @@
...
+	info, err := os.Stat(d)
+	if os.IsNotExist(err) {
+		return false
+	}
+	return info.IsDir()
+}
+
+// findAppManifests locates all manifests associated with a specific application within a repository
+func findAppManifests(application wego.Application, repoDir string) ([][]byte, error) {
```
josecordaz, line 27: it seems this function is only used in the tests, show it be move it there ? 🤔

pkg/services/app/add.go
```diff
@@ -109,20 +115,22 @@ func (a *App) Add(params AddParams) error {
...
+
 	gitProvider, err := a.gitProviderFactory(params.GitProviderToken)
 	if err != nil {
 		return err
 	}
 
 	var secretRef string
 	if SourceType(params.SourceType) == SourceTypeGit {
-		secretRef, err = a.createAndUploadDeployKey(params, params.Url, clusterName, gitProvider)
+		secretRef, err = a.createAndUploadDeployKey(info, params.DryRun, info.Spec.URL, gitProvider)
```
josecordaz, line 127: might be a good idea to use info.Spec.URL within `createAndUploadDeployKey` as we are already passing in `info`

pkg/services/app/add.go
```diff
@@ -240,58 +258,57 @@ func (a *App) getGitRemoteUrl(params AddParams) (string, error) {
...
 	if err != nil {
 		return fmt.Errorf("could not create GitOps automation for .wego directory: %w", err)
 	}
 
 	// a local directory has not been passed, so we clone the repo passed in the --url
 	if params.Dir == "" {
-		a.logger.Actionf("Cloning %s", params.Url)
-		if err := a.cloneRepo(params.Url, params.Branch, params.DryRun); err != nil {
+		a.logger.Actionf("Cloning %s", info.Spec.URL)
+		if err := a.cloneRepo(info.Spec.URL, info.Spec.Branch, params.DryRun); err != nil {
```
josecordaz, line 287: would it be a good idea to remove the cloned repo when the user uses `--url` flag ? The cloning is happening in a temporary folder but I wonder if we should delete that temporary folder right away or if it would be ok to let the system delete it later? 🤔

pkg/services/app/add.go
```diff
@@ -240,58 +258,57 @@ func (a *App) getGitRemoteUrl(params AddParams) (string, error) {
...
 			return fmt.Errorf("failed to clone application repo: %w", err)
 		}
 	}
 
 	if !params.DryRun {
 		if !params.AutoMerge {
-			if err := a.createPullRequestToRepo(params, gitProvider, ".wego", params.Url, clusterName, appHash, appSpec, appGoat); err != nil {
+			if err := a.createPullRequestToRepo(info, gitProvider, info.Spec.URL, appHash, appSpec, appGoat); err != nil {
```
josecordaz, line 294: we could probably skip info.Spec.URL as we are passing in info

pkg/services/app/remove.go
jrryjcksn: Cool. Thanks for pointing that out!

pkg/services/app/remove.go
jrryjcksn: It's only used in tests right now (because I wrote the test for it first :-)). It will be used in the actual code later.

pkg/services/app/add.go
jrryjcksn: Wow. Funny we never noticed that before. :+1:

pkg/services/app/add.go
jrryjcksn: Actually, we can't here. This is called in two different places; once with the URL and once with the ConfigURL. We need to know which one to use.

pkg/services/app/add.go
jrryjcksn: Good catch, thanks

jrryjcksn: We're using that parameter to select either the URL or the ConfigURL and need it to decide which one

pkg/services/app/remove.go
jrryjcksn: Actually, after looking... I need to know it's a directory :man_shrugging:

pkg/services/app/remove_test.go
```diff
@@ -0,0 +1,134 @@
...
+	workloadPath1 := filepath.Join(dir, "kustomize", "one", "path", "to", "files")
+	workloadPath2 := filepath.Join(dir, "kustomize", "another", "path", "to", "more", "files")
+	if err := os.MkdirAll(workloadPath1, 0777); err != nil {
+		return "", err
+	}
+	if err := os.MkdirAll(workloadPath2, 0777); err != nil {
+		return "", err
+	}
+
+	if err := ioutil.WriteFile(filepath.Join(workloadPath1, "nginx.yaml"), []byte(workloadYaml1), 0644); err != nil {
```
luizbafilho, line 90: it's a minor thing, but we don't need to specify the whole file definition to verify if the Remove is working. you could just pass the string here `[]byte("file1")` and the same later and check it in the test.

I think that makes clear what is being actually tested and clean up the code since we don't actually need the actual definition for anything else.

luizbafilho (approved):

pkg/services/app/remove_test.go
jrryjcksn: That makes sense; you're right, the file contents are irrelevant.

Issue #494, opened 2021-07-19
Create a server-friendly logger implementation


jpellizzari:
Our current logging implementation uses `fmt.Print` to log to stdout. This approach will not work for operations that are initiated via the API.

We will need to associate log messages with a particular request, and create log levels that can be toggled via command-line arguments.

JamWils:
At team sync we decided to write up an ADR (Architecture Decision Record) for this work.

jpellizzari:
> At team sync we decided to write up an ADR (Architecture Decision Record) for this work.

Just a guide for how and when to use each of the log levels will suffice.

JamWils:
@palemtnrider, are you good with this?  It should at least clarify everything and if it doesn't feel like enough information we can always up it to a more concrete ADR.

jpellizzari:
Sorry I should have clarified:
>Just a guide for how and when to use each of the log levels will suffice.
 * In the form of a short ADR about logging

JamWils:
Thanks for the clarification!!!

PR #512, opened 2021-07-23
Adding a new api logger

luizbafilho:
**What changed?**
- Adding a new API logger

**Why?**
- The API server needs structured logging

**Note**
This is a band-aid solution, as we move to the new Server-Client architecture where all the actions will be performed by the api, we need a way to provide a stream of logs in real-time to the clients, web, and cli. 

A spike issue will be created to create a poc and a ADR for that.

**How did you test it?**

**Release notes**

**Documentation Changes**


pkg/logger/api.go
```diff
@@ -0,0 +1,56 @@
...
+func (l ApiLogger) Waitingf(format string, a ...interface{}) {
+	l.logger.Info(fmt.Sprintf(format, a...))
+}
+
+func (l ApiLogger) Successf(format string, a ...interface{}) {
+	l.logger.Info(fmt.Sprintf(format, a...))
+}
+
+func (l ApiLogger) Warningf(format string, a ...interface{}) {
+	l.logger.Info(fmt.Sprintf(format, a...))
```
jpellizzari, line 51: Is it possible to suppress everything but errors and/or warnings with this logging package? I would expect to see `log.Warn` here.

I read through the README of this package and it seems they have some strong opinions on NOT doing `log.Warn`, `log.Debug`, etc.

pkg/logger/api.go
```diff
@@ -0,0 +1,56 @@
...
+}
+
+func (l ApiLogger) Failuref(format string, a ...interface{}) {
+	l.logger.Info(fmt.Sprintf(format, a...))
```
J-Thompson12, line 55: why are all of these exactly the same? Is this something that will get fixed when we have an actual api solution?

jpellizzari (commented):
Can you hook this up to the API server and show the "before" and "after" for the log messages?

pkg/logger/logger.go
```diff
@@ -31,45 +26,3 @@ type Logger interface {
 	Warningf(format string, a ...interface{})
 	Failuref(format string, a ...interface{})
 }
```
J-Thompson12, line 28: This seems like it could be moved to cli.go so that we dont have this file anymore? Having this separate feels like it would be getting used for both logger types.

pkg/logger/api.go
luizbafilho: exacly

pkg/logger/logger.go
luizbafilho: This is the interface definition, the other two files are the actual implementation. So yes, the point is having it referenced by the two logger types

pkg/logger/api.go
luizbafilho: The approach is different as you've read, it uses numbers so there are no explicit log levels. And that's something I struggled with in this implementation, that's a bit hard to match. We can settle into something once we define the logging architecture.

luizbafilho:
Definitely I'll add it to the api

josecordaz (approved):
LGTM

jpellizzari:
Related to #494

PR #513, opened 2021-07-23
Add Auth ADR

jpellizzari:
This ADR seeks to outline the auth strategy for Weave GitOps Core


doc/adr/0005-wego-core-auth-strategy.md
```diff
@@ -0,0 +1,83 @@
...
+- [Bitbucket Implicit Grant](https://developer.atlassian.com/cloud/bitbucket/oauth-2/#2--implicit-grant--4-2-)
+  - No `CLIENT_SECRET` required here
+
+### Kubernetes Cluster
+
+All operations will need to go through an API server that is installed on the cluster via the `wego gitops install` command. This API server will run with a service account that provides the ability to do the following:
+
+- Create `Namespace` resources
+- Create `Application`, `Source`, and `Kustomize|Helm` resources
+- Read all Kubernetes objects, in all namespaces
```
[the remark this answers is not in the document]
jpellizzari: In order to get a list of all the things that have been "reconciled" by an Application. Would there be a namespace that would be prohibited?

doc/adr/0005-wego-core-auth-strategy.md
```diff
@@ -0,0 +1,83 @@
...
+
+These deploy keys are outside the scope of a git repo, so we will need a higher level of access in the form of a Git Provider access token to add deploy keys programmatically.
+
+### Git Provider
+
+Operations such as creating pull requests (or "merge requests" in the case of Gitlab), adding deploy keys, and retrieving user data will need to be completed via HTTP requests. Each major Git Provider has some form of OAuth flow:
+
+- [GitHub OAuth Flow](https://docs.github.com/en/developers/apps/building-oauth-apps/authorizing-oauth-apps#web-application-flow)
+  - Notice that `localhost` redirect_uris are permitted: https://docs.github.com/en/developers/apps/building-oauth-apps/authorizing-oauth-apps#localhost-redirect-urls
+  - This approach requires a `CLIENT_SECRET` be embedded in the binary, thus exposing the secret leaking by decompiling the binary
```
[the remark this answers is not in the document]
jpellizzari: They don't do this type of auth in their project (to my knowledge). The Device Flow noted below would NOT require a `CLIENT_SECRET`.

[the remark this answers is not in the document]
palemtnrider: We would only use this one for Authentication and in those cases we will leverage Dex and have the user provide their oauth client_id and client_secret to Dex.

doc/adr/0005-wego-core-auth-strategy.md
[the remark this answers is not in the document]
palemtnrider: I think our SA will need full access to `wego-system` but not all things in the cluster.  I _guess_  could have a read-only role to facilitate this.  However, I think this is out of scope for this ADR.  This ADR should focus on getting a token to associate deploy-keys with repos and create PRs/MRs on behalf of a user.

doc/adr/0005-wego-core-auth-strategy.md
```diff
@@ -0,0 +1,83 @@
...
+
+For this reason, we want to utilize the more short-lived OAuth tokens to avoid exposing a personal access token. We also don't want to persist these tokens longer than we have to.
+
+In addition, flows that require a `CLIENT_SECRET` will require us to embed it in the binary, which leaves us open to exposing it via decompilation.
+
+### UX Concerns
+
+Given that the Weave GitOps UI and API server may be running on a user's cluster, we may not always know the OAuth callback URL ahead of time. For this reason, we will need to use OAuth flows that do not require a static callback URLs. The flows listed above in the Git Providers section fit that constraint.
+
+We will need to ensure that the callback URL is configurable by the user via environment variables. This will allow them to run and expose the UI/API servers on whichever endpoint they choose.
```
palemtnrider, line 61: The flows to get a token requires a redirect_url and not a callback_url.  Additionally, we can't rely on customers allowing inbound traffic from the Git server.

doc/adr/0005-wego-core-auth-strategy.md
```diff
@@ -0,0 +1,83 @@
...
+
+## Design
+
+Since most modern Git Providers will support a form of browser-based OAuth, **we will utilize the browser for both the CLI and UI authentication with Git Providers**.
+
+In the case of the CLI, we can utilize a short-lived browser session that will recieve the OAuth callback and pass the token to the CLI:
```
palemtnrider, line 67: We don't need the local server running.  The CLI will act as the user agent to get the token.  Unfortunately, all 3 Git servers will require a different flow.

doc/adr/0005-wego-core-auth-strategy.md
```diff
@@ -0,0 +1,83 @@
...
+
+![UI Auth Diagram](ui_auth.svg)
+
+## Decision
+
+Implement the flow(s) defined in the "Design" section
+
+## Consequences
+
+A SaaS application would be able to hide their `CLIENT_SECRET` behind their firewalled infrastructure. We do not have that luxury, so we will have to ship the `CLIENT_SECRET` with the binary. The only flow that requires this is the GitHub OAuth Flow, which gives us the alternative Device Flow specially built for our exact use case. We may be able to utilize their Device Flow as a one-off (since other providers do not have that feature).
```
palemtnrider, line 81: For acquiring a token, we don't need the client_secret.  Implicit grant, device flow, and PKCE don't require it.

palemtnrider (changes requested):
Thanks for writing this up.  I believe we need a few changes before it's ready.

doc/adr/0005-wego-core-auth-strategy.md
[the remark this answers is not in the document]
jpellizzari: Fair enough.

doc/adr/0005-wego-core-auth-strategy.md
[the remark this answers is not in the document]
jpellizzari: Dex does not exist in our universe (WeGO core). We need to be able to authenticate with the Git Provider in the "Getting Started" user flow, which means we cannot require them to install Dex. That is what prompted this document.

doc/adr/0005-wego-core-auth-strategy.md
jpellizzari: >The flows to get a token requires a redirect_url and not a callback_url
Is this just a nit around using the correct terminology, or is there some other differentiation I am missing?

doc/adr/0005-wego-core-auth-strategy.md
jpellizzari: How would the Git Provider redirect back to a CLI with the OAuth code?

doc/adr/0005-wego-core-auth-strategy.md
jpellizzari: It says that right in the paragraph? 

> The only flow that requires this is the GitHub OAuth Flow, which gives us the alternative Device Flow specially built for our exact use case

doc/adr/0005-wego-core-auth-strategy.md
jpellizzari: > Additionally, we can't rely on customers allowing inbound traffic from the Git server

I'm confused by the term "Inbound Traffic" here. No TCP connections will be made into the cluster. The browser does the redirecting.

doc/adr/0005-wego-core-auth-strategy.md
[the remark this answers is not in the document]
bigkevmcd: There are also two orthogonal authentications...

 * Application and K8s authentication
 * GitHub/GitLab/Bitbucket authentication

For example, we want folks to authenticate to login to the Wego UI, and we want to validate that they have access to RBACed K8s objects, this is "Application and K8s authentication"

We also want to be able to use the APIs in remote Git Hosting Services e.g. GitHub, for this, we need a Personal Access Token (or GitHub Bot token).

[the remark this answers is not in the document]
jpellizzari: @bigkevmcd Those concerns are called out elsewhere in the doc. PTAL.

doc/adr/0005-wego-core-auth-strategy.md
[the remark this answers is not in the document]
jpellizzari: Scoped down to `wego-system`

doc/adr/0005-wego-core-auth-strategy.md
jpellizzari: From 1-1 chat, @palemtnrider and I are aligned on this. We don't rely on inbound TCP connectivity.

doc/adr/0005-wego-core-auth-strategy.md
```diff
@@ -0,0 +1,122 @@
...
+
+### Kubernetes Cluster
+
+All operations will need to go through an API server that is installed on the cluster via the `wego gitops install` command. This API server will run with a service account that provides the ability to do the following:
+
+- Create `Namespace` resources
+- Create `Application`, `Source`, and `Kustomize|Helm` resources
+- Read Kubernetes objects in the `wego-system` namespace
+
+### Security Concerns
```
jpellizzari, line 49: @jpellizzari Add "inject sensitive stuff in CI instead of checking it into version control"

doc/adr/0005-wego-core-auth-strategy.md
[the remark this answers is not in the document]
jpellizzari: 1-1 chat with @palemtnrider, we are ok with the assumption that Dex does NOT exist on the cluster. Marking as resolved.

doc/adr/0005-wego-core-auth-strategy.md
```diff
@@ -0,0 +1,128 @@
...
+
+Each of the Git Providers listed above support some sort of "personal access token" that allow for "impersonation" style authorization. These tokens never expire, and provide "sudo" access to the user's Github account.
+
+For this reason, we want to utilize the more short-lived OAuth tokens to avoid exposing a personal access token. We also don't want to persist these tokens longer than we have to.
+
+In addition, flows that require a `CLIENT_SECRET` will require us to embed it in the binary, which leaves us open to exposing it via decompilation.
+
```
jpellizzari, line 56:
```suggestion
CI will inject any sensitive information into the binary to avoid keeping data in version control in a public repo. We are still susceptible to leaking data via decompilation.
```

doc/adr/0005-wego-core-auth-strategy.md
```diff
@@ -0,0 +1,128 @@
...
+TLDR:
+
+1. The user does OAuth with their Git Provider
+2. We use the resulting access_token to push a Deploy Key to the repository (giving us `git` permissions)
+3. We create a pull request for the repository via the Git Provider HTTP API
+
+For browser security, we will convert the Git Provider OAuth token to a JSON Web Token (JWT) to protect against Cross-site Scripting (XSS) attacks. The encrypted JWT will allow a malicious script to authenticate with the Weave GitOps API only, whereas passing the unencrypted OAuth token to the browser would allow a malicious script to authenticate with the Github API.
+
+Additionally, we do not plan on adding third-party scripts to the Weave GitOps UI to minimize the surface area for XSS attacks. This does NOT, however, account for NPM modules or other dependencies that we add to our app at build time.
+
```
jpellizzari, line 86:
```suggestion
CI will inject any sensitive information into the binary to avoid keeping data in version control in a public repo. We are still susceptible to leaking data via decompilation.
```

doc/adr/0005-wego-core-auth-strategy.md
palemtnrider: Jordan and I discussed this, the API server will be the URL the browser is redirected to.  When the browser calls this server, it will store the token which will then be retrieved from our CLI/UI.

doc/adr/0005-wego-core-auth-strategy.md
```diff
@@ -0,0 +1,130 @@
...
+Alternatively, pass in a token on standard input by using --with-token. The minimum required scopes for the token are: "repo", "read:org".
+```
+
+## Decision
+
+Implement the flow(s) defined in the "Design" section
+
+## Consequences
+
```
palemtnrider, line 127: We no longer need to put utilize the CLIENT_SECRET for our git token operations.  You can remove the info about the client_secret in the binary.

palemtnrider (approved):
LGTM.  Just a couple of minor corrections.  Approving now assuming those will be addressed prior to merging.

doc/adr/0005-wego-core-auth-strategy.md
```diff
@@ -0,0 +1,130 @@
...
+Each of the Git Providers listed above support some sort of "personal access token" that allow for "impersonation" style authorization. These tokens never expire, and provide "sudo" access to the user's Github account.
+
+For this reason, we want to utilize the more short-lived OAuth tokens to avoid exposing a personal access token. We also don't want to persist these tokens longer than we have to.
+
+In addition, flows that require a `CLIENT_SECRET` will require us to embed it in the binary, which leaves us open to exposing it via decompilation.
+
+### UX Concerns
+
+Given that the Weave GitOps UI and API server may be running on a user's cluster, we may not always know the OAuth callback URL ahead of time. For this reason, we will need to use OAuth flows that do not require a static `redirect_uri`. The flows listed above in the Git Providers section fit that constraint.
+
```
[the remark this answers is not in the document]
jpellizzari: Yes, that is correct. If a user wants to make the UI/API public-facing, they will need to provide the `redirect_uri` as an environment variable.

I will add some info.

doc/adr/0005-wego-core-auth-strategy.md
```diff
@@ -0,0 +1,125 @@
...
+![UI Auth Diagram](ui_auth.svg)
+
+TLDR:
+
+1. The user does OAuth with their Git Provider
+2. We use the resulting access_token to push a Deploy Key to the repository (giving us `git` permissions)
+3. We create a pull request for the repository via the Git Provider HTTP API
+
+For browser security, we will convert the Git Provider OAuth token to a JSON Web Token (JWT) to protect against Cross-site Scripting (XSS) attacks. The encrypted JWT will allow a malicious script to authenticate with the Weave GitOps API only, whereas passing the unencrypted OAuth token to the browser would allow a malicious script to authenticate with the Github API.
```
bigkevmcd, line 80: Would `wego` generate a keypair for encrypting the JWT, if so, how and where is that stored?

doc/adr/0005-wego-core-auth-strategy.md
```diff
@@ -0,0 +1,125 @@
...
+
+Additionally, we do not plan on adding third-party scripts to the Weave GitOps UI to minimize the surface area for XSS attacks. This does NOT, however, account for NPM modules or other dependencies that we add to our app at build time.
+
+CI will inject any sensitive information into the binary to avoid keeping data in version control in a public repo. We are still susceptible to leaking data via decompilation.
+
+## Prior Art
+
+This authentication approach is inspired by other CLI tools that have very smooth user experiences. For example, the `gcloud` CLI for Google Cloud Platform utilizes the browser to authenciate the user:
```
bigkevmcd, line 88:
```suggestion
This authentication approach is inspired by other CLI tools that have very smooth user experiences. For example, the `gcloud` CLI for Google Cloud Platform utilizes the browser to authenticate the user:
```

doc/adr/0005-wego-core-auth-strategy.md
```diff
@@ -0,0 +1,125 @@
...
+
+```
+$ gcloud auth login
+
+Your browser has been opened to visit:
+
+https://accounts.google.com/o/oauth2/auth?response_type=code&client_id=1234.apps.googleusercontent.com&redirect_uri=http%3A%2F%2Flocalhost%3A8085%2F&scope=...
+
+Opening in existing browser session.
+
```
bigkevmcd, line 98: I should point out that the `gcloud` client code provides their `CLIENT_ID` and `SECRET` in the Python code.

_lib/googlecloudsdk/core/config.py_
```python
# TODO(b/35848109): Distribute a clientsecrets.json to avoid putting it in code.
CLOUDSDK_CLIENT_ID = '<redacted>.apps.googleusercontent.com'
CLOUDSDK_CLIENT_NOTSOSECRET = '<redacted>'
```

doc/adr/0005-wego-core-auth-strategy.md
```diff
@@ -0,0 +1,125 @@
...
+The operations for the git repository are the typically the ones done with the `git` CLI, ie: `git clone`, `git commit`, `git push`. We assume that git providers will allow for repository-specific keys to be added; GitHub and Gitlab calls these "deploy keys", other providers might call them "access keys". These are typically ssh key-pairs.
+
+These deploy keys are outside the scope of a git repo, so we will need a higher level of access in the form of a Git Provider access token to add deploy keys programmatically.
+
+### Git Provider
+
+Operations such as creating pull requests (or "merge requests" in the case of Gitlab), adding deploy keys, and retrieving user data will need to be completed via HTTP requests. Each major Git Provider has some form of OAuth flow:
+
+- [Github Device Flow](https://docs.github.com/en/developers/apps/building-github-apps/identifying-and-authorizing-users-for-github-apps#device-flow)
+  - This approach doesn't require a `CLIENT_SECRET` or callback page
```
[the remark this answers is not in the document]
jpellizzari: > But, it does require a CLIENT_ID which is issued by GitHub for a specific GitHub app.

I guess I let myself get confused by the word `SECRET` in `CLIENT_SECRET`. No matter what we do, the binary will ship everything one needs to identify as our GitHub app.

> Would we provide a single CLIENT_ID for all customers to use, or instruct them to create an app and provide one?

We would provide a single CLIENT_ID for all customers to use. I wouldn't want to send users to the GitHub App creation screen in order to use `wego`.

doc/adr/0005-wego-core-auth-strategy.md
jpellizzari: Good question. Can we generate that at runtime? Every time the app restarts, the user would have to re-authenticate.

Outside of generating at runtime, it would need to work like all of the other individually-identifiable information: injected into the binary by CI.

doc/adr/0005-wego-core-auth-strategy.md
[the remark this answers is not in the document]
jpellizzari: Noted in the UX concerns section.

doc/adr/0005-wego-core-auth-strategy.md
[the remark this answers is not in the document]
bigkevmcd: What if the customer isn't using GitHub?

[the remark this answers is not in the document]
jpellizzari: > What if the customer isn't using GitHub?

There are two other Git Providers in the next two bullet points.

If you are implying that some customers might want to use some Git Provider other than the ones we support, we can make the OAuth settings configurable.

We can also push people to a GitHub/Gitlab/Bitbucket page to create their own app with their own `CLIENT_ID`. IMO that is a worse UX, but if we feel the security trade-off is worth it, that might work.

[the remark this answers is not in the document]
bigkevmcd: I'm more thinking that we'd need to embed CLIENT_IDs for probably three different Git hosting services, never mind the cases for enterprises that run "private" installations.

doc/adr/0005-wego-core-auth-strategy.md
jpellizzari: Removed

doc/adr/0005-wego-core-auth-strategy.md
```diff
@@ -0,0 +1,125 @@
...
+
+In addition, flows that require a `CLIENT_SECRET` will require us to embed it in the binary, which leaves us open to exposing it via decompilation.
+
+### UX Concerns
+
+Given that the Weave GitOps UI and API server may be running on a user's cluster, we may not always know the OAuth callback URL ahead of time. For this reason, we will need to use OAuth flows that do not require a static `redirect_uri`. The flows listed above in the Git Providers section fit that constraint.
+
+We will need to ensure that the `redirect_uri` is configurable by the user via environment variables. This will allow users to run and expose the UI/API servers on whichever endpoint they choose.
+
+## Design
```
jpellizzari, line 60: @jpellizzari TODO list alternatives that were considered

doc/adr/0005-wego-core-auth-strategy.md
```diff
@@ -0,0 +1,125 @@
...
+
+### Kubernetes Cluster
+
+All operations will need to go through an API server that is installed on the cluster via the `wego gitops install` command. This API server will run with a service account that provides the ability to do the following:
+
+- Create `Namespace` resources
+- Create `Application`, `Source`, and `Kustomize|Helm` resources
+- Read Kubernetes objects in the `wego-system` namespace
+
+### Security Concerns
```
jpellizzari, line 46: @jpellizzari TODO what do we do if client_id is compromised

jpellizzari: @jpellizzari TODO how do we store JWT encryption secret

jpellizzari:
@bigkevmcd Updated with the action items from our review. PTAL.

See this commit for things that were changed: https://github.com/weaveworks/weave-gitops/pull/513/commits/733335e4cba5afab9fe52a93dca95c17c21fb61b

I may have jumped out the window with the "Advanced Use Case" section, but it felt right to call it out.

doc/adr/0005-wego-core-auth-strategy.md
jpellizzari: Added a section on this.

doc/adr/0005-wego-core-auth-strategy.md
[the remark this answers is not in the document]
jpellizzari: I did some more thinking/research on this: every OAuth request will have the `client_id` in the request URL, so the `client_id` is public knowledge. I added a section on what sort of attack vector(s) this opens up.

markramm:
Looks good to me.

jpellizzari:
@markramm did you merge `main` onto my branch intentionally?

markramm (approved):
Looks good to me

sympatheticmoose (approved):

PR #518, opened 2021-07-25
Fixed error creating test repositories.

josecordaz:
Closes: #518  

**What changed?**
Added timeout when session should exit as the default time was 1s, resulting in false positives because hub takes more than that to finish.

Added cases to fail when :
- local repo already exists
- repo in github.com already exists
- retry repo creation when getting exit code different from 0

**Why?**
Fix this error happening randomly:
image
It was failing with `File already exists` when running locally.
It wasn't failing if the repo already existed.

**How did you test it?**
I reran 3 times the CI pipeline without any issue.

**Release notes**

**Documentation Changes**


jpellizzari (approved):

test/acceptance/test/utils.go
```diff
@@ -339,12 +343,26 @@ func initAndCreateEmptyRepo(appRepoName string, IsPrivateRepo bool) string {
...
-	Eventually(session).Should(gexec.Exit())
+	Eventually(session, 10).Should(gexec.Exit())
+	Expect(session.ExitCode()).Should(Equal(0))
+	Expect(session.Err).ShouldNot(gbytes.Say(fmt.Sprintf("mkdir: /tmp/%s: File exists", appRepoName)))
+	Expect(session.Err).ShouldNot(gbytes.Say("Existing repository detected"))
+	if os.Getenv("CI") == "true" {
+		Expect(session.Out).Should(gbytes.Say(fmt.Sprintf(`Initialized empty Git repository in /tmp/%s/.git/
+Updating origin
+https://github.com/%s/%s`, appRepoName, GITHUB_ORG, appRepoName)))
+	} else {
```
J-Thompson12, line 354: Why do we have this if statement in the acceptance tests.  Shouldn't the result always be the same?

PR #520, opened 2021-07-26
Use wego pkg constants instead of redeclaring

### LINKED ISSUE #507: Use the DeploymentType and SourceType from application_types here instead of defining again
we should use the ones from application_types here instead of defining them again.

_Originally posted by @palemtnrider in https://github.com/weaveworks/weave-gitops/pull/497#discussion_r674987609_

jpellizzari:
Closes #507 

We were declaring `SourceType` and `DeploymentType` in both the `wego` package, as well as in the `services/app` package. This removes the `services/app/ declarations in favor of the `wego` ones.

### DIFF
--- a/pkg/services/app/add.go
+++ b/pkg/services/app/add.go
@@ -124,7 +124,7 @@ func (a *App) Add(params AddParams) error {
 	}
 
 	var secretRef string
-	if SourceType(params.SourceType) == SourceTypeGit {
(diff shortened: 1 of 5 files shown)


J-Thompson12 (approved):
LGTM

jrryjcksn (approved):
Good to clean this up!

PR #521, opened 2021-07-26
validate flux binary

### LINKED ISSUE #115: verify supply chain dependencies
**Short description**
As a security person, I want to ensure we validate all binaries/scripts downloaded and used in our product so that we can be assured we are running the correct code within our product.

This stems from an issue we luckily avoided with the codecov coverage uploader script.  tldr; they weren't verifying a bash script they downloaded and used.  More details at  https://about.codecov.io/security-update/ 

**Acceptance criteria**
- [ ] the build verifies the download of flux is the one we expect

J-Thompson12:
closes #115 Added a check in make to validate the flux binary.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -47,8 +47,8 @@ install: bin bin/$(BINARY_NAME)_ui
 
 # Clean up images and binaries
 clean:
-	rm -f bin/wego pkg/flux/bin/flux
-	rm -f pkg/flux/bin/fluxchecksums.txt
+	rm -f bin/wego
+	rm -rf pkg/flux/bin/
 	rm -rf cmd/ui/dist
 	rm -rf coverage
 	rm -rf node_modules
--- a/Makefile
+++ b/Makefile
@@ -48,7 +48,7 @@ install: bin bin/$(BINARY_NAME)_ui
 # Clean up images and binaries
 clean:
 	rm -f bin/wego pkg/flux/bin/flux
-	rm -f bin/wego pkg/flux/bin/fluxchecksums.txt
+	rm -f pkg/flux/bin/fluxchecksums.txt
 	rm -rf cmd/ui/dist
 	rm -rf coverage
 	rm -rf node_modules
--- a/Makefile
+++ b/Makefile
@@ -48,6 +48,7 @@ install: bin bin/$(BINARY_NAME)_ui
 # Clean up images and binaries
 clean:
 	rm -f bin/wego pkg/flux/bin/flux
+	rm -f bin/wego pkg/flux/bin/fluxchecksums.txt
 	rm -rf cmd/ui/dist
 	rm -rf coverage
 	rm -rf node_modules
--- a/tools/dependencies.toml
+++ b/tools/dependencies.toml
@@ -3,6 +3,11 @@ version="0.16.0"
 tarpath="https://github.com/fluxcd/flux2/releases/download/v${version}/flux_${version}_${goos}_${goarch}.tar.gz"
 bindir="../pkg/flux/bin"
 
+[fluxchecksums]
+version="0.16.0"
+txtpath="https://github.com/fluxcd/flux2/releases/download/v${version}/flux_${version}_checksums.txt"
+bindir="../pkg/flux/bin"
+
 [envtest]
 version="1.19.2"
 special_tarpath="https://storage.googleapis.com/kubebuilder-tools/kubebuilder-tools-${version}-${goos}-${goarch}.tar.gz;kubebuilder/bin"
--- a/tools/dependencies.toml
+++ b/tools/dependencies.toml
@@ -1,10 +1,6 @@
 [flux]
 version="0.16.0"
 tarpath="https://github.com/fluxcd/flux2/releases/download/v${version}/flux_${version}_${goos}_${goarch}.tar.gz"
-bindir="../pkg/flux/bin"
-
-[fluxchecksums]
-version="0.16.0"
 txtpath="https://github.com/fluxcd/flux2/releases/download/v${version}/flux_${version}_checksums.txt"
 bindir="../pkg/flux/bin"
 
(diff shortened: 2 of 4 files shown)


tools/functions.sh
```diff
@@ -108,8 +108,7 @@ do_curl_tarball() {
     dldir="$(mktempdir)"
     mkdir "${dldir}/${cmd}"
     do_curl "${dldir}/${cmd}.tar.gz" "${url}"
-    tar -C "${dldir}/${cmd}" -xvf "${dldir}/${cmd}.tar.gz"
```
J-Thompson12, line 111: I am not really sure why this is in there but it caused the hash to be different from what was in the checksum.

Makefile
```diff
@@ -48,6 +48,7 @@ install: bin bin/$(BINARY_NAME)_ui
 # Clean up images and binaries
 clean:
 	rm -f bin/wego pkg/flux/bin/flux
+	rm -f bin/wego pkg/flux/bin/fluxchecksums.txt
```
palemtnrider, line 51: This will try and delete bin/wego again. 
I think it reads better to put the flux (bin, checksum) removal on their own line.

tools/functions.sh
palemtnrider: This expands the tarball into the directory.  Your changes put the tarball in the directory w/o expanding it.

tools/dependencies.toml
```diff
@@ -3,6 +3,11 @@ version="0.16.0"
 tarpath="https://github.com/fluxcd/flux2/releases/download/v${version}/flux_${version}_${goos}_${goarch}.tar.gz"
 bindir="../pkg/flux/bin"
 
+[fluxchecksums]
```
palemtnrider, line 6: I think it would be cleaner to add the checksum txt file to the [flux] section instead of creating a new download here.  It will require less change (version number) as we update the flux versions.

tools/download-deps.sh
```diff
@@ -98,6 +103,19 @@ download_dependency() {
     "${fetch}" "${tool}" "${url_and_path}" "${custom_bindir:-$bin_dir}"
 }
 
+validate_flux_binary() {
+    cd ../pkg/flux/bin
+    local flux_hash
+    hash_result=$(openssl dgst -sha256 flux)
```
palemtnrider, line 109: You might be able to simplify this with `if ! grep $((openssl dgst -sha256 flux | cut -d ' ' -f 2) fluxchecksums.txt`...

tools/functions.sh
```diff
@@ -125,3 +124,17 @@ do_curl_tarball_with_path() {
...
+
+do_curl_txt() {
+    local cmd="${1}"
+    local url="${2}"
+    local default_path="${HOME}/.wego/bin"
+    local path="${3:-${default_path}}"/"${cmd}"
+
+    dldir="$(mktempdir)"
+    mkdir "${dldir}/${cmd}"
+    do_curl "${dldir}/${cmd}.txt" "${url}"
```
palemtnrider, line 136: We should be able to download the text file into the proper location instead of moving it after the download.

palemtnrider (changes requested):
Thanks for adding this logic to start validating our downloads.  I've suggested a few changes that I believe simplify the impl.

tools/functions.sh
```diff
@@ -125,3 +127,21 @@ do_curl_tarball_with_path() {
     tar -C "${dldir}/${cmd}" -xvf "${dldir}/${cmd}.tar.gz"
     mv "${dldir}/${cmd}/${url_and_path[1]}" "${path}/${cmd}"
 }
+
+do_curl_txt() {
+    local cmd="${1}"
```
palemtnrider, line 132: We don't use the cmd variable so it can be removed.

tools/functions.sh
```diff
@@ -125,3 +127,21 @@ do_curl_tarball_with_path() {
...
+    local url="${2}"
+    local default_path="${HOME}/.wego/bin"
```
palemtnrider, line 134: Not using the default_path variable so it can be removed.

tools/functions.sh
```diff
@@ -108,6 +108,8 @@ do_curl_tarball() {
     dldir="$(mktempdir)"
     mkdir "${dldir}/${cmd}"
     do_curl "${dldir}/${cmd}.tar.gz" "${url}"
+    ## need to validate flux here because unpacking the tar changes the hash
+    validate_flux "${path}.txt" "${dldir}/${cmd}.tar.gz"
```
palemtnrider, line 112: Odd to call a function named validate_flux in a generic do_curl_tarball function.  We should either call it to validate checksum or move to the calling function.  i.e., in download_dependency - which might be cleaner as we know there that the configuration provided a checksum txt file.

Makefile
```diff
@@ -48,6 +48,7 @@ install: bin bin/$(BINARY_NAME)_ui
 # Clean up images and binaries
 clean:
 	rm -f bin/wego pkg/flux/bin/flux
+	rm -f pkg/flux/bin/fluxchecksums.txt
```
palemtnrider, line 51: nit - have the first line remove bin/wego and this line remove everything under pkg/flux - unless we want to keep files in that dir other than flux and fluxchecksums.txt

palemtnrider (changes requested):
I like the most recent changes.  Just a few more suggestions...

tools/functions.sh
J-Thompson12: I thought about moving it there but then we would have to change where we do the tarbal unpacking. I did think about moving the unpacked tarball then validating it then unpacking it.

Makefile
J-Thompson12: there are no other files

tools/functions.sh
J-Thompson12: I am using it to set the path. Like the other functions are doing.

tools/functions.sh
J-Thompson12: I am using it to set the path. Like the other functions are doing.

palemtnrider (approved):

PR #523, opened 2021-07-26
Remove docker scanning CI jobs

jpellizzari:
Removes some jobs that are extraneous. We do not (currently) run in a docker container, so there is no need to have source code scanning for non-existent containers.


josecordaz (approved):
LGTM

Issue #256, opened 2021-06-07
Remove Application from one cluster via PR


markramm:
As an application operator I would like to be able to run wego application remove [app] to should  not leave workloads running but remove the gitops management, so I can clean up after tests. 

--remove-workload flag is a required flag

Ideally this will issue a PR that removes the customization in .wego to delete the files.   

In the short term this may require an imperative command as well.   

If there is a user initiated clean up step in the short term that must be documented.

markramm:
As an application operator I would like to be able to run` wego application remove [app]` to remove gitops control of that app.  By default this should leave workloads running but remove the gitops management.

markramm:
Clean up UX so people don't get confused between remove and delete.

bia:
How about `delete` and `remove-automation` or `deautomate`  ? `unautomate` ?

markramm:
https://github.com/fluxcd/flux2/discussions/763 has some more information.

markramm:
Remove the app always, and always require the --force flag.

jrryjcksn:
There is no indication (that I have found) in this story what the `--force` flag is supposed to do -- only that it is always required. If it means, "also remove the workloads from the cluster", then I feel it is poorly named. If it means something else, please describe its function.

Also, it is hard to interpret this sentence:

```
As an application operator I would like to be able to run wego application remove [app] to should  not leave workloads running but remove the gitops management, so I can clean up after tests. 
```

In any case, I am currently unable to determine what is being requested here which is why I moved it back to "Refine"

markramm:
Let's change it to --remove-workload

By default it would just remove the automation, and because that is hard, the --remove-workload flag is (for now) always required.

JamWils:
@jrryjcksn and @rokshana-b does this need a QA label?

jpellizzari:
This is open because we don't yet support `remove` via PR.

jpellizzari:
Implementation notes:
* Remove the entry from the `user/kustomization` for a given cluster
* Remove the `apps/<my app>` directory from the gitops repo

josecordaz:
This is the work plan I have in mind for this ticket. Let me know if I should consider something different.

![image.png](https://images.zenhubusercontent.com/606389311ead0cc4788a9a72/b53f969a-b4b1-449d-a01f-53570301af1a) 

[...]

PR #529, opened 2021-07-27
Collect all generated resources (both applied to cluster and written to git) for removal

jrryjcksn:
Part of #256 

**What changed?**

- Full unit testing of resource collection using actual generated manifests
- Added `Osys` client to allow mocking os operations (e.g. UserHomeDir, Exit)
- Removed shims, handlers, and command support that is no longer needed

**Why?**
- The unit testing ensures we will clean up everything correctly in `app remove`
- The refactoring tidies up the code base

**How did you test it?**
- Unit tests
- 

**Release notes**
No

**Documentation Changes**
No


cmd/wego/gitops/cmd.go
```diff
@@ -72,7 +73,8 @@ func init() {
 
 func installRunCmd(cmd *cobra.Command, args []string) error {
 	cliRunner := &runner.CLIRunner{}
-	fluxClient := flux.New(cliRunner)
+	osysClient := osys.New()
```
josecordaz, line 76: small nit. I would suggest to use this pattern also in `cmd/wego/flux/cmd.go` to be consistent.

josecordaz (approved):
LGTM

cmd/wego/gitops/cmd.go
jrryjcksn: Sure, okay. I only originally didn't because the osysclient was used in two places but that's no big deal. :+1:

PR #538, opened 2021-07-28
Remove source-type CLI arg

jpellizzari:
The `--source-type` arg was added naively here #497 . We already fill out this parameter elsewhere by detecting the `--chart` argument


jrryjcksn (approved):
LGTM

rokshana-b (approved):

PR #540, opened 2021-07-29
250 more remove

jrryjcksn:
**What changed?**
- Added initial `wego app remove` command (with auto-merge only)
- Refactored to share code across commands

**Why?**
It's in the plan

**How did you test it?**
New unit tests

**Release notes**
It will be when complete

**Documentation Changes**
Not yet

### DIFF
--- a/cmd/wego/app/remove/cmd.go
+++ b/cmd/wego/app/remove/cmd.go
@@ -1,11 +1,10 @@
-package add
+package remove
 
-// Provides support for adding an application to wego managment.
+// Provides support for removing an application from wego management.
 
 import (
 	"fmt"
 	"os"
-	"path/filepath"
 	"strings"
 
 	"github.com/lithammer/dedent"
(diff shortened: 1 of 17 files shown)


cmd/wego/app/remove/cmd.go
```diff
@@ -0,0 +1,97 @@
...
+	"github.com/weaveworks/weave-gitops/pkg/osys"
+	"github.com/weaveworks/weave-gitops/pkg/runner"
+	"github.com/weaveworks/weave-gitops/pkg/services/app"
+	"github.com/weaveworks/weave-gitops/pkg/utils"
+)
+
+var params app.RemoveParams
+
+var Cmd = &cobra.Command{
+	Use:   "remove [--auto-merge <true|false>] [--private-key <keyfile>] <app name>",
```
J-Thompson12, line 27: I thought auto merge had to always be set to true here? If so we shouldnt show the option.

pkg/kube/kube.go
```diff
@@ -104,6 +105,21 @@ func (k *KubeClient) Delete(manifests []byte, namespace string) ([]byte, error)
...
 }
 
+func (k *KubeClient) DeleteByName(name, kind, namespace string) ([]byte, error) {
+	args := []string{
+		"delete",
+		"--namespace", namespace,
+		kind, name,
+	}
+
+	out, err := k.runKubectlCmd(args)
```
J-Thompson12, line 115: I have seen this on the other kube commands. Is there a reason we dont just return k.runKubectlCmd?

pkg/kube/kubehttp.go
```diff
@@ -129,6 +129,10 @@ func (c *KubeHTTP) Delete(manifests []byte, namespace string) ([]byte, error) {
 	return nil, errors.New("not implemented")
 }
 
+func (c *KubeHTTP) DeleteByName(name, kind, namespace string) ([]byte, error) {
+	return nil, errors.New("not implemented")
```
jpellizzari, line 133: We will probably need this implemented at some point. What might that look like? Is it as simple as deleting something from the cluster?

pkg/kube/kube.go
jrryjcksn: Now that you mention it... no, not that I can see. I think I must have just copied the body and replaced the args. I'll fix that.

jpellizzari (approved):

J-Thompson12 (approved):
LGTM

pkg/kube/kubehttp.go
jrryjcksn: It would have to be based on the dynamic interface but, yes... I figured both `Delete` functions could be taken on as a different task. I'll add an issue

Issue #536, opened 2021-07-28
Spike: Demonstrate using wego-core as a library of another application


jpellizzari:
Related to https://github.com/weaveworks/weave-gitops-private/pull/25

We plan on using weave-gitops as a library in other projects. We need to demonstrate that is possible.

Deliverables for this spike:
* "Embed" the Weave Gitops UI in a third-party project
* Run the requisite backend services to make the UI possible ^^

Constraints:
* Must be a build-time dependency that (roughly) works like other NPM and Go modules

jpellizzari:
See the linked PR for the changes that need to be made internally.

Example repo that wraps Weave GitOps Core: https://github.com/jpellizzari/wego-wrapper-demo

PR #542, opened 2021-07-30
Allow for Weave GitOps to be called as a library

jpellizzari:
Related to #536 

Demo Repo: https://github.com/jpellizzari/wego-wrapper-demo


jpellizzari:
POC only, closing

Issue #527, opened 2021-07-27
Run go-fmt in CI to catch un-formatted code


jpellizzari:
There have been a couple of instances where un-formatted code has made it onto `main`. Our linter may not be checking for formatting problems (nor should it). The issue with committing un-formatted code is that the next person who DOES run a formatter will have extra changes in their pull request, thus muddying the intent of their PR.

We should run `go-fmt` on our code in CI a "check only" mode and have it exit if it finds formatting discrepancies. This will fail PR builds and prevent un-formatted code from being merged.

https://pkg.go.dev/cmd/gofmt#hdr-Examples

`gofmt -e` should output errors and exit if any are found.

Contributors on the project are encouraged to run `gofmt` on save. Most editors will have a version of this feature for go.

PR #552, opened 2021-08-02
add gofmt to test github action

J-Thompson12:
> Describe what has changed in this PR 

Closes at #527 Added gofmt check to github action test section

> Tell your future self why have you made these changes

To make sure all code is formatted

> How have you verified this change? Tested locally? Added unit test/integration/acceptance test(s)? 

using github actions

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -43,6 +43,8 @@ jobs:
       with:
         version: v1.29
         args: --timeout=600s
+    - name: Format
+      run: if [ "$(gofmt -l . | wc -l)" -gt 0 ]; then exit 1; fi
 
   build:
     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -43,6 +43,8 @@ jobs:
       with:
         version: v1.29
         args: --timeout=600s
+    - name: Format
+      run: if [ "$(gofmt -l . | wc -l)" -gt 0 ]; then exit 1; fi
```
jpellizzari, line 47: What does the output look like if this fails?

J-Thompson12: If the condition hits it just fails as exit code 1 but its the only thing run there. I realize you are right i should output the files that need formatting.

jpellizzari:
@J-Thompson12 One more thing I forgot to add: can you encapsulate this in a `make` target? It would be reproduce locally.

jpellizzari (approved):
Nice work!

PR #555, opened 2021-08-02
update weave-gitops objects to wego

J-Thompson12:
> Describe what has changed in this PR

Rename kubernetes obects from weave-gitops to wego. Also renamed the deploy key (it seemed like it also should be wego)

> Tell your future self why have you made these changes

To follow naming guide

### DIFF
--- a/pkg/gitproviders/provider.go
+++ b/pkg/gitproviders/provider.go
@@ -118,7 +118,7 @@ func (p defaultGitProvider) CreateRepository(name string, owner string, private 
 }
 
 func (p defaultGitProvider) DeployKeyExists(owner, repoName string) (bool, error) {
-	deployKeyName := "weave-gitops-deploy-key"
+	deployKeyName := "wego-deploy-key"
 
 	ownerType, err := p.GetAccountType(owner)
 	if err != nil {
@@ -167,7 +167,7 @@ func (p defaultGitProvider) DeployKeyExists(owner, repoName string) (bool, error
 }
 
 func (p defaultGitProvider) UploadDeployKey(owner, repoName string, deployKey []byte) error {
-	deployKeyName := "weave-gitops-deploy-key"
+	deployKeyName := "wego-deploy-key"
 	deployKeyInfo := gitprovider.DeployKeyInfo{
 		Name: deployKeyName,
 		Key:  deployKey,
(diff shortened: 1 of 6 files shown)


josecordaz (approved):
LGTM

pkg/gitproviders/provider.go
```diff
@@ -118,7 +118,7 @@ func (p defaultGitProvider) CreateRepository(name string, owner string, private
 }
 
 func (p defaultGitProvider) DeployKeyExists(owner, repoName string) (bool, error) {
-	deployKeyName := "weave-gitops-deploy-key"
+	deployKeyName := "wego-deploy-key"
```
jpellizzari, line 121: Can we pull this into a constant to be shared between any function that needs it? If we want to change this in the future, someone will need to remember all the places were it is used.

As it is now, I call this the "magic string" pattern where by the app breaks if someone forgets to change the string in all the places.

pkg/kube/kube.go
```diff
@@ -230,7 +230,7 @@ func (k *KubeClient) runKubectlCmdWithInput(args []string, input []byte) ([]byte
 	return out, nil
 }
 func (k *KubeClient) LabelExistsInCluster(ctx context.Context, label string) error {
-	cmd := []string{"get", "app", "-l", fmt.Sprintf("weave-gitops.weave.works/app-identifier=%s", label), "--all-namespaces"}
+	cmd := []string{"get", "app", "-l", fmt.Sprintf("wego.weave.works/app-identifier=%s", label), "--all-namespaces"}
```
jpellizzari, line 233: Should this be `app.WeGOAppIdentifierLabelKey`>

jpellizzari (commented):
I know it was already in the codebase, but let's try and get rid of the "magic strings" peppered in different places if possible.

pkg/kube/kube.go
J-Thompson12: It doesn't matter much because of the commit discussion this morning we will be removing this label

pkg/gitproviders/provider.go
J-Thompson12: I know @jrryjcksn was doing something around the deploy key. @jrryjcksn are you changing any of this? If not i can just move it to a const.

PR #556, opened 2021-08-02
Don't create a deploy key for a public repository

### LINKED ISSUE #551: Weave GitOps should not require a deploy key for a public repository
**Describe the bug**
Currently, wego creates and uploads deploy keys for public repositories. This prevents use of applications from upstream public repositories (like sock-shop).

**Environment**
 - wego: v0.2.1

**To Reproduce**

Run `wego app add` against a public repository. Note that a deploy key is created and uploaded.

jrryjcksn:
Resolves #551 

**What changed?**
If a user provides a public repository to wego, we no longer create a deploy key; instead, we hand `flux create source git` an `http` URL and don't create a secret

**Why?**
It's basically wrong to require creation of an unnecessary key and the current exclusive use of ssh URLs prevents adding an app from an upstream public repository (like sock-shop) for which we have no associated SSH keys.

**How did you test it?**
Manually adding the sock-shop app from the `github.com/microservices-demo/microservices-demo.git` repository

**Release notes**
Users would not have to change anything but some scenarios would now work that would have failed in the past

**Documentation Changes**
No

### DIFF
--- a/pkg/flux/flux.go
+++ b/pkg/flux/flux.go
@@ -91,12 +91,16 @@ func (f *FluxClient) Uninstall(namespace string, dryRun bool) error {
 func (f *FluxClient) CreateSourceGit(name string, url string, branch string, secretRef string, namespace string) ([]byte, error) {
 	args := []string{
 		"create", "source", "git", name,
-		"--url", url,
 		"--branch", branch,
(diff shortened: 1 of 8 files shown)


josecordaz (approved):
LGTM

PR #557, opened 2021-08-03
use repository default branch if not overridden

### LINKED ISSUE #341: wego app add should default to the current HEAD branch ref when called from within a Git repo
**Describe the bug**
`wego app app .` defauts to the `main` branch when the `--branch` option is not explicitly set.

```
$ wego app add .
Updating parameters from environment... using URL: 'ssh://git@github.com/sbernheim/podinfo.git' of origin from git config...

done

Checking cluster status... FluxInstalled

Uploading deploy key
---
apiVersion: source.toolkit.fluxcd.io/v1beta1
kind: GitRepository
metadata:
  name: podinfo
  namespace: wego-system
spec:
  interval: 30s
  ref:
    branch: main
  secretRef:
    name: podinfo
  url: ssh://git@github.com/sbernheim/podinfo.git

---
```

If the instant repo does not have a `main` branch, this will result in an error on the resulting GitRepository resource in the target cluster.

```
$ kg gitrepo -A
NAMESPACE     NAME      URL                                          READY   STATUS                                                                                                            AGE
wego-system   podinfo   ssh://git@github.com/sbernheim/podinfo.git   False   unable to clone 'ssh://git@github.com/sbernheim/podinfo.git', error: couldn't find remote ref "refs/heads/main"   33s
```

Pushing a `main` branch can easily resolve this error.

```
$ git checkout -b main 
Switched to a new branch 'main'
$ git push -u origin main
Total 0 (delta 0), reused 0 (delta 0)
remote: 
remote: Create a pull request for 'main' on GitHub by visiting:
remote:      https://github.com/sbernheim/podinfo/pull/new/main
remote: 
To github.com:sbernheim/podinfo.git
 * [new branch]      main -> main
Branch 'main' set up to track remote branch 'main' from 'origin'.

<... after some time ...>

$ kg gitrepo -A
NAMESPACE     NAME      URL                                          READY   STATUS                                                            AGE
wego-system   podinfo   ssh://git@github.com/sbernheim/podinfo.git   True    Fetched revision: main/5ac18a31750cd88b07504f74f770fde2bbfd3fcd   6m11s
```

There is nothing magic about either the `master` or `main` branches, and user application repos may not have either branch.  It would be safer to read the current HEAD ref from the local Git repository and use that as the default branch when calling `wego app add` from within a local repo.

```
$ git branch --show-current
current-branch-name
```

This would disintermediate between `main` and `master` default branches and offer more confidence that the expected branch actually exists in the GitHub remote repo.

**Environment**
 - wego: v0.0.5
 - kubernetes:
    - [ x] KinD - v0.8.1 go1.14.4 darwin/amd64
 ```
$ kubectl cluster-info
Kubernetes master is running at https://127.0.0.1:52803
KubeDNS is running at https://127.0.0.1:52803/api/v1/namespaces/kube-system/services/kube-dns:dns/proxy
```
 - Browser + version: N/A
 - 
**Affects versions**

**To Reproduce**
Steps provided in the description above.

**Expected behavior**
The default GitRepository manifest generated by `wego app add` should reference the current HEAD ref of the local repository when run from within a Git repo.

**Additional context**
Encountered while running through the Getting Started Docs for v0.0.5 during today's GitOps WOUG.

### PR DESCRIPTION
**Pull request by jrryjcksn**
Resolves #341 

**What changed?**
We now consult the repository to determine the default branch if the user does not explicitly specify a branch.

**Why?**
Always defaulting to `main` breaks if the repository has a different default branch such as `master` (which is still the default for many repositories)

**How did you test it?**
Unit tests and manual tests using an upstream repository with `master` as its default branch (the sock shop demo)

**Release notes**
Users will no longer need to override the branch if the repository does not default to `main`

**Documentation Changes**
The cli example for `wego app add` should change to not call out `main` as the default

### CODE REVIEW

File: pkg/services/app/add.go [hunk 1]
```diff
@@ -493,6 +529,27 @@ func (a *App) commitAndPush(filters ...func(string) bool) error {
 	return nil
 }
 
+func (a *App) getRepoInfo(repoUrl string, gitProvider gitproviders.GitProvider) (*gitprovider.RepositoryInfo, error) {
```
josecordaz, line 532: `getRepoInfo` and `getDefaultBranch` functions seem more go-git-providers related, I would suggest to move them to the `gitproviders` package. There is already a function named `GetRepoInfo` which receives `account-type`, `owner` and `repoName`, we could create another one with a slightly different name to reflect the case when receiving the repoURL, maybe GetRepoInfoFromURL or similar.

jrryjcksn: Okay, that makes sense. I wasn't really sure how much we wanted to put specific use cases in the go-git-providers interface (vs a single `GetRepoInfo` function that we then use in various places). I'll make the change.

jrryjcksn: Done

josecordaz (approved):
LGTM

Issue #545, opened 2021-08-02
`wego app add --auto-merge .` silently fails (on the second attempt)


squaremo:
**Describe the bug**

If wego app add --auto-merge is unable to push to the upstream, it will report the error on the first attempt, but appear to succeed on subsequent attempts.

**Environment**
 - wego: 0.2.2

**To Reproduce**

 1. install wego in a cluster; fork the example repo per the get started guide
 2. create a fresh SSH key with no password: `ssh-keygen -f bogus`
 3. in the cloned, forked git repo try adding the app: `wego app add . --auto-merge --private-key ./bogus`. It will fail, since the key doesn't grant access to the repo.
 4. try the same command again; it will appear to succeed, with the last message `✔ App manifests are up to date`. Check the upstream repo: no new commit.

**Expected behavior**

I'd expect it to fail on the second attempt in the same way as it failed on the first.

PR #558, opened 2021-08-03
Validate ssh key

josecordaz:
Fix: #545 

**What changed?**
SSH key validation via clone test with the key of cloning just 1 commit so it doesn't take long for large repositories.

**Why?**
The problem was that if we use an unauthorized ssh key when attempting to add an app, we end up having flux components that never reconcile because we apply them to the cluster right before pushing them. Which is where we need to have a valid ssh key so the push does not fail.

**How did you test it?**
Manually using an unauthorized ssh key and seeing a proper message like this `Error: failed to add the app : error validating git repo access ssh: handshake failed: ssh: unable to authenticate, attempted methods [none publickey], no supported methods remain`

**Release notes**

**Documentation Changes**


pkg/git/gogit.go
```diff
@@ -276,6 +277,34 @@ func (g *GoGit) Head() (string, error) {
...
+		RemoteName:    gogit.DefaultRemoteName,
+		ReferenceName: branchRef,
+		SingleBranch:  true,
+		NoCheckout:    false,
+		Progress:      nil,
+		Depth:         1,
+		Tags:          gogit.NoTags,
+	})
+
+	if err != nil && !(errors.Is(err, transport.ErrEmptyRemoteRepository) ||
```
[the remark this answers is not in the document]
JamWils: Found this in the comments above the function:

```
// PlainCloneContext a repository into the path with the given options, isBare
// defines if the new repository will be bare or normal. If the path is not empty
// ErrRepositoryAlreadyExists is returned.
```
How do you want to handle ErrRepositoryAlreadyExists?  Do you want to ignore that error too?

pkg/git/gogit.go
```diff
@@ -276,6 +277,34 @@ func (g *GoGit) Head() (string, error) {
...
+
+	path, err := ioutil.TempDir("", "temp-src")
+	if err != nil {
+		return err
+	}
+	defer os.RemoveAll(path)
+
+	branchRef := plumbing.NewBranchReferenceName(branch)
+	_, err = gogit.PlainCloneContext(ctx, path, true, &gogit.CloneOptions{
+		URL:           url,
```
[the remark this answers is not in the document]
josecordaz: @JamWils  you mean validate the format of the input value? like the structure that defines what an actual URL is? Right now the command fails like this `Error: failed to add the app : could not generate deploy key: could not get owner from url notvalidURL.git` with an invalid url value, which might seem ok because it describes(maybe no too clearly) what the problem is.

pkg/services/app/add.go
```diff
@@ -120,6 +120,13 @@ func (a *App) Add(params AddParams) error {
 		return fmt.Errorf("could not update parameters: %w", err)
 	}
 
+	if params.SourceType != string(wego.SourceTypeHelm) {
```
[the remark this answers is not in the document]
josecordaz: It seems there is no logic reason right now. However a probable explanation is that we follow this patter on fields that are used by cobra to read parameters because cobra can't read them using custom data types, or at least not in the way we are reading them with cobra right now. I'll fix that.

pkg/git/gogit.go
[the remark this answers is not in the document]
josecordaz: I just copy/pasted the ones used in the Clone method. I'll try to do some research about it

pkg/services/app/add.go
```diff
@@ -120,6 +120,13 @@ func (a *App) Add(params AddParams) error {
...
+		err = a.git.ValidateAccess(ctx, params.Url, params.Branch)
+		if err != nil {
+			return err
```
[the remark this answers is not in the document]
josecordaz: this is already wrapped up inside the `ValidateAccess` method

pkg/git/gogit.go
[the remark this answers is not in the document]
JamWils: I guess I meant more, do we allow any arbitrary URL to be passed into here or do we have limits.  E.g. do we have only specific hosts we want to let through, etc.  There is the url.URL parse function that could check that it is simply a valid URL in the first place.  Make sense?

pkg/services/app/add.go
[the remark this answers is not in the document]
JamWils: True, but if validate access is called from multiple places then it could be hard to track down where exactly the error occurred.  I'm not sure what the standard practice is here, but wrapping it here with context that it is being called from add like `return fmt.Errorf("app add ValidateAccess %w");`

pkg/git/gogit.go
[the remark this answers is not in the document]
josecordaz: yes, it does. We currently allow any arbitrary value for the URL. It doesn't exist any validation at this moment.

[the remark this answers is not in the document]
JamWils: We chatted f2f and @josecordaz is going to validate the URL in another location.

pkg/services/app/add.go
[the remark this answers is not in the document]
JamWils: Also, should the if-statement be `params.SourceType == wego.SourceTypeGit`?  If we add new source types besides helm and git in the future then this will ValidateAccess on those new source types.  This seems very specific to git.

pkg/git/gogit_test.go
```diff
@@ -84,6 +84,18 @@ var _ = Describe("Clone", func() {
...
 
+var _ = Describe("ValidateAccess", func() {
+	It("validate access to a given repository successfully", func() {
+		err := gitClient.ValidateAccess(context.Background(), "https://github.com/githubtraining/hellogitworld", "master")
+		Expect(err).ShouldNot(HaveOccurred())
+	})
+
+	It("fails to validate access to a possible private repository", func() {
+		err := gitClient.ValidateAccess(context.Background(), "https://github.com/notexisted/repo", "master")
+		Expect(err).Should(HaveOccurred())
```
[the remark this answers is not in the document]
josecordaz: I did a test to try to see if the Clone method would fail differently on either empty repo or non-existed branch using an invalid ssh key but it always fails with the same error `ssh: handshake failed: ssh: unable to authenticate, attempted methods [none publickey], no supported methods remain`. So there is no need to check for those specific errors.

pkg/git/gogit.go
```diff
@@ -34,15 +35,28 @@ import (
...
 )
 
+type GotGit struct{}
+
+func (g *GotGit) PlainCloneContext(ctx context.Context, path string, isBare bool, o *gogit.CloneOptions) (*gogit.Repository, error) {
+	return gogit.PlainCloneContext(ctx, path, isBare, o)
+}
+
+//counterfeiter:generate . GoGitI
+type GoGitI interface {
```
jpellizzari, line 45: Don't add the `I` here. We know its an interface :wink: .

pkg/git/gogit.go
```diff
@@ -34,15 +35,28 @@ import (
...
+
 type GoGit struct {
 	path       string
 	auth       transport.AuthMethod
 	repository *gogit.Repository
+	gitwrapper GoGitI
 }
 
-func New(auth transport.AuthMethod) *GoGit {
+func New(auth transport.AuthMethod, wrapper GoGitI) *GoGit {
```
jpellizzari, line 56: This seems strange. We are returning a `*GoGit` but also passing in a `*GoGit`. What is the goal here?

pkg/git/gogit.go
JamWils: @josecordaz I would move this interface to a separate package like "wrapper" and drop the `I`.

JamWils: @josecordaz I would also put the GotGit{} struct in the wrapper package and create a constructor that is public which returns the interface.

pkg/git/gogit_test.go
```diff
@@ -20,7 +25,7 @@ var (
 )
 
 var _ = BeforeEach(func() {
-	gitClient = git.New(nil)
+	gitClient = git.New(nil, &git.GotGit{})
```
JamWils, line 28: I see you are making the fake within the test.  I would just pass in the fake up here and then you can set up the mock on a per tests basis.

pkg/git/gogit_test.go
```diff
@@ -84,6 +89,31 @@ var _ = Describe("Clone", func() {
...
+
+		fakeGoGit := &gitfakes.FakeGoGitI{}
+
+		fakeGoGit.PlainCloneContextReturns(nil, transport.ErrEmptyRemoteRepository)
+
+		fakeGitClient := git.New(nil, fakeGoGit)
+
+		err := fakeGitClient.ValidateAccess(context.Background(), "https://github.com/githubtraining/hellogitworld", "master")
+
+		Expect(err).ShouldNot(HaveOccurred())
```
JamWils, line 108: We should really have more concrete assertions on the errors.  Which error occurred?  Check for that instead of the fact that it just occurred.

JamWils: Nevermind, I read this wrong.

pkg/git/gogit_test.go
```diff
@@ -84,6 +89,31 @@ var _ = Describe("Clone", func() {
...
+	})
+
+	It("fails to validate access to a possible private repository", func() {
+		err := gitClient.ValidateAccess(context.Background(), "https://github.com/notexisted/repo", "master")
```
JamWils, line 112: I get this test, but you are testing the internal of `PlainCloneContextReturns`.  At this point you could use the fake and return any error that isn't `transport.ErrEmptyRemoteRespository` and return

`fakeGoGit.PlainCloneContextReturns(nil, errors.New("ka-boom")`

That way if gogit changes internally the test is still valid.  Of course, if you want to return an error specifically about auth then that is a different case on my point above is moot.

pkg/git/gogit.go
josecordaz: @jpellizzari actually I didn't know how to name it, as I already have a struct named GoGit, any suggestion ?

luizbafilho (changes requested):
I think you can simplify this code by just using the `Clone` function we already have, if you read the code they do pretty much the same thing, this way we would avoid duplication.

pkg/git/gogit_test.go
josecordaz: yes, I'm trying to test an auth error.

josecordaz:
> I think you can simplify this code by just using the `Clone` function we already have, if you read the code they do pretty much the same thing, this way we would avoid duplication.

Right, that is a really good suggestion. Actually I also had it. But then I realized that we could move the clone closer to the beginning of `Add` but for some scenarios we don't clone all the time and if we do, it might be a heavy repo that could increase the time of the execution, that's why this "other clone" has `Depth: 1`. What we could do, and probably this is what you meant(sorry if you did and I didn't get it), is to adjust Clone to work for both cases. 🤔

josecordaz:
> > I think you can simplify this code by just using the `Clone` function we already have, if you read the code they do pretty much the same thing, this way we would avoid duplication.
> 
> Right, that is a really good suggestion. Actually I also had it. But then I realized that we could move the clone closer to the beginning of `Add` but for some scenarios we don't clone all the time and if we do, it might be a heavy repo that could increase the time of the execution, that's why this "other clone" has `Depth: 1`. What we could do, and probably this is what you meant(sorry if you did and I didn't get it), is to adjust Clone to work for both cases. 🤔

NVM I'm having seconds thoughts because we are setting `path` and `repository` fields in `Clone`, if I use that method I'm afraid I could mess the logic. I guess there will be no issue because those fields will be overwritten later in the code, but I'm not sure, it doesn't seem to be like a good practice, but I could be wrong. What do you guys think @JamWils @luizbafilho ?

luizbafilho:
> > > I think you can simplify this code by just using the `Clone` function we already have, if you read the code they do pretty much the same thing, this way we would avoid duplication.
> > 
> > 
> > Right, that is a really good suggestion. Actually I also had it. But then I realized that we could move the clone closer to the beginning of `Add` but for some scenarios we don't clone all the time and if we do, it might be a heavy repo that could increase the time of the execution, that's why this "other clone" has `Depth: 1`. What we could do, and probably this is what you meant(sorry if you did and I didn't get it), is to adjust Clone to work for both cases. 🤔
> 
> NVM I'm having seconds thoughts because we are setting `path` and `repository` fields in `Clone`, if I use that method I'm afraid I could mess the logic. I guess there will be no issue because those fields will be overwritten later in the code, but I'm not sure, it doesn't seem to be like a good practice, but I could be wrong. What do you guys think @JamWils @luizbafilho ?

yep I had the same concern, but as you said it's being overridden in other calls, once I tried to avoid that, but I faced a couple of issues due to the execution flow, and I couldn't make it work. As we are moving to an API architecture, we need to revisit that because there'll be concurrent calls and that variable setting will not work. But I think it's out of scope on this PR, for now you should be fine.

josecordaz:
@luizbafilho I think I was able to avoid the writing to `path` and `repository`. Let's see if the integration tests pass. If they don't I'll also try adding just `depth` to `Clone`.

luizbafilho (approved):

pkg/git/gogit.go
```diff
@@ -297,6 +314,21 @@ func (g *GoGit) GetRemoteUrl(dir string, remoteName string) (string, error) {
...
+func (g *GoGit) ValidateAccess(ctx context.Context, url string, branch string) error {
+
+	path, err := ioutil.TempDir("", "temp-src")
+	if err != nil {
+		return fmt.Errorf("error creating temporary folder %w", err)
+	}
+	defer os.RemoveAll(path)
+
+	_, err = g.clone(ctx, path, url, branch, 1)
+	if err != nil && !errors.Is(err, transport.ErrEmptyRemoteRepository) {
```
JamWils, line 326: What about `ErrRepositoryAlreadyExists`, do we want to ignore that error as well?

JamWils (changes requested):
Just want final verification on the error returned by PlainCloneContext

pkg/git/gogit.go
josecordaz: Good point. It seems we can only reproduce that error if we call Init or/and Clone twice using the same path https://github.com/go-git/go-git/blob/master/repository_test.go#L122. This call to clone always use a random folder, probably this error is not very likely to occur 🤔. But I could be missing something.

JamWils: It might be worth actually checking, it's a low friction add and the "probably this error is not very likely to occur" will get us eventually :)

josecordaz: sounds good

josecordaz: after further research, we found out that `PlainCloneContext` could return `ErrRepositoryAlreadyExists` before doing an actual ssh key validation so we don't want to ignore it.

JamWils: Nice work digging into this @josecordaz

JamWils:
LGTM, approved.

JamWils (approved):

Issue #197, opened 2021-05-21
Validate or populate .Add() params


josecordaz:
Inputs to the `pkg/services/app` public functions are currently not validated. We should either validate that all inputs are present, or populate input defaults for the user. This is currently not biting us because we rely on the `cobra` package to populate the default arguments for us. That package will not be able to help with HTTP API calls.

In the case of validation, we should return an error to the user indicating the remediation steps. I am of the opinion that we should have explicit API inputs to avoid unintended behaviors for the user.

PR #561, opened 2021-08-03
Set default values for path and deployment type if Add method is not used via CLI.

josecordaz:
Fix: #197 

**What changed?**
We now set default values for  path and deployment type in case Add is not used via CLI

**Why?**
In the case of adding an app via api where we are not going to use a command we needed a way to define those default values.

**How did you test it?**
Test suites in CI

**Release notes**

**Documentation Changes**


pkg/services/app/add.go
```diff
@@ -77,6 +77,32 @@ type AddParams struct {
...
+	DefaultBranch         = "main"
+	DefaultDeploymentType = "kustomize"
+	DefaultChart          = ""
+	DefaultPrivateKey     = ""
+	DefaultAppConfigURL   = ""
+	DefaultDryRun         = false
+	DefaultAutoMerge      = false
+)
+
+func (a AddParams) PopulateDefaultValues() {
```
jpellizzari, line 93: We should be using this `func` somewhere right?

josecordaz: Correct, this meant to be used when we implement Add in the api-service.

jpellizzari: Can we use it now in the `.Add()` service? The API will probably just consume that service.

jpellizzari: Can we use it now in the `.Add()` service? The API will probably just consume that service.

jpellizzari: Also, this doesn't need  a receiver. We want to pass in a pointer to the `AddParams` because we are modifying it.

jpellizzari: @josecordaz PTAL at these comments ^^

josecordaz: @jpellizzari if definitely can use it in the `.Add()` service but that will overwrite any parameter set by the user using different values from the default ones.  Maybe I need to populate the fields only if they are empty ?

jpellizzari: I guess I am confused because this is a function that never gets called. Also, since we are passing by value and this doesn't return anything, does it actually change the `a` variable?

This is also not a situation to use a method/receiver. This can be modeled as a pure function.

jpellizzari (approved):
:+1:

PR #564, opened 2021-08-03
Remove deps file from version control

jpellizzari:
The `.deps` file is used for `Makefile` shenanigans to get dependencies to download only once. It should not have been added to version control.


josecordaz (approved):

jpellizzari:
Closing this as it will cause more trouble than its worth; this is being taken care of in #566

PR #386, opened 2021-06-24
GitHub Actions: pin to SHA

morancj:
Add a GitHub Action to require actions are pinned to a SHA, not a tag.

Per https://docs.github.com/en/actions/learn-github-actions/security-hardening-for-github-actions#using-third-party-actions, we should pin all actions to a commit SHA. This change enforces that.

I considered adding https://github.com/mheap/pin-github-action as a development package (helps future devs pin to SHA's), but that seemed out of scope for this PR. Those changes are in reverted commit 631481e. If you want to use them locally, you can test with e.g:

<details><summary>Show Linux</summary>

```shell
➜ secret-tool store --label="token" provider GitHub profile pin-github-action

➜ for FILE in .github/workflows/*.yaml;\                    
 do GH_ADMIN_TOKEN=$( secret-tool lookup provider GitHub profile pin-github-action )\
 npm run-script pin-github-action "${FILE}";\
 done
```

</details>

<details><summary>Show macOS (untested)</summary>

```shell
➜ security add-generic-password -T "/usr/bin/security" -a $LOGNAME -s GitHub -G profile -G pin-github-action -w

➜ for FILE in .github/workflows/*.yaml;\                    
 do GH_ADMIN_TOKEN=$( security find-generic-password -s GitHub -G profile -G pin-github-action )\
 npm run-script pin-github-action "${FILE}";\
 done
```

</details>

The GitHub Token should only have `repo:public_repo` scope and should only be required to mitigate rate limits.

Tested against another repo

**Release notes**
Actions must now be pinned to a specific SHA, not a tag.

Documentation changes only required for maintainers, and if you want to switch to nvm and revert 18f1a55.

PR #565, opened 2021-08-03
automate flux2 version upgrade

josecordaz:
**What changed?**
PR automation to upgrade flux

We would need to execute `.github/scripts/current_and_old_flux_versions.s` in case we want to debug the parameters passed in to the Github action creating the PR.

**Why?**
It would be easier to update flux

**How did you test it?**
I used the same yaml config on isolated private repo

**Release notes**

**Documentation Changes**

Example from test repo of how the PR will look like:
image
image


.github/workflows/upgrade-flux.yaml
```diff
@@ -0,0 +1,83 @@
...
+          {
+            # diff change
+            git diff
+            # export PR_BODY for PR and commit
+            echo "::set-output name=pr_body::$PR_BODY"
+            echo "::set-output name=pr_title::Upgrade flux version from $currentFluxVersion to $latestFluxVersion"
+          }
+      - name: Create Pull Request
+        id: cpr
+        uses: peter-evans/create-pull-request@v3
```
jpellizzari, line 66: I think we need to lock these to specific hashes now? @morancj can you confirm?

.github/workflows/upgrade-flux.yaml
```diff
@@ -0,0 +1,83 @@
...
+jobs:
+  update-components:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v2
+      - name: Setup Go
+        uses: actions/setup-go@v2
+        with:
+          go-version: 1.16.x
```
jpellizzari, line 19: Nit: we can lock this to prevent inconsistencies from happening in CI.

.github/workflows/upgrade-flux.yaml
```diff
@@ -0,0 +1,83 @@
+name: Update Components
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: "0 * * * *"
+  push:
+    branches: [main]
```
jpellizzari, line 8: Does this automatically run when someone pushes to `main`? I think we only want to run this on the `cron` job, once a day (if I am reading that correctly).

jpellizzari (commented):
The job itself looks great! Just a question about the triggers

.github/workflows/upgrade-flux.yaml
josecordaz: you are right

jpellizzari (commented):
Sorry @josecordaz I should have caught this before, would you mind putting this logic into a script file that we can call from our laptops? The problem with putting this in a Github action is that we cannot run it locally.

In general, we want to be able to reproduce everything that CI does locally.

jpellizzari (approved):
LGTM, modulo one change: move the script to `tools/` and rename to `get-flux-update-pr.sh`

Also, I'm not super great at bash, so maybe we can get another set of :eyes: on this before merging.

.github/workflows/upgrade-flux.yaml
morancj: Yes, please. [zgosalvez/github-actions-ensure-sha-pinned-actions](https://github.com/zgosalvez/github-actions-ensure-sha-pinned-actions) can help. For some commentary, see https://github.com/weaveworks/weave-gitops/pull/386.

tools/get-flux-update-pr.sh
```diff
@@ -0,0 +1,43 @@
...
+
+echo "Latest flux version $latestFluxVersion"
+echo "Current flux version $currentFluxVersion"
+
+if [[ "$latestFluxVersion" == "$currentFluxVersion" ]]; then
+  echo "No changes needed"
+  exit 0
+fi
+
+newContent=$(sed -e "/version\=\"[0-9\.]*\"/{s//version=\"$latestFluxVersion\"/;:a" -e '$!N;$!ba' -e '}' tools/dependencies.toml)
```
jrryjcksn, line 13: I think this has the same issue in that it would fail if we reordered the file. Maybe something like:
`newContent=$(cat /tmp/dependencies.toml | tr '\n' '\r' | sed -e "/\[flux\]\rversion\=\"[0-9\.]*\"/{s//\[flux\]\rversion=\"$latestFluxVersion\"/;:a" -e '$!N;$!ba' -e '}' | tr '\r' '\n')`

jrryjcksn (approved):
A couple of comments but LGTM

tools/get-flux-update-pr.sh
```diff
@@ -1,6 +1,6 @@
 latestFluxVersion="$(curl -s --request GET --url "https://api.github.com/repos/fluxcd/flux2/releases?per_page=1" | jq . | jq '.[0] | .tag_name' | jq -r | sed -e 's/v//')"
 
-currentFluxVersion="$(sed -n 's/version="\([0-9\.]*\)"/\1/p' tools/dependencies.toml | head -1)"
+currentFluxVersion="$(tools/bin/stoml tools/dependencies.toml flux.version)"
```
morancj, line 3: Nice! Does `stoml` work well?

jrryjcksn: We've been using it for a long time, starting a while back in WKP. It's so simple that it's hard to know what could go wrong :-)

tools/get-flux-update-pr.sh
josecordaz: just changed `/tmp` with `/tools` and worked perfect, thanks 🙂

tools/get-flux-update-pr.sh
josecordaz: yes it does :)

PR #566, opened 2021-08-04
Support SSH agent for git authentication

squaremo:
**What changed?**

This adds support for [`ssh-agent`](https://en.wikipedia.org/wiki/Ssh-agent) to the `wego app add` command. If ssh-agent isn't running, or doesn't have a key that works, it will fall back to guessing a key file to use. If you explicitly tell it a key file, it'll just use that, and fail if it doesn't work.

**Why?**

Many people using SSH with git will be running `ssh-agent`, and their expectation will be that SSH authentication just works, especially if they've just been using git. Being asked for your private key passphrase in that circumstance is jarring.

I've also made the passphrase prompt look more like what git does, so it will seem familiar; and, it will now only ask for a password when it needs it, rather than up front.

**How did you test it?**

I tried it -- I ran `wego app add .` before and after the change, and with or without ssh-agent available (and with or without a key added). There don't appear to be tests already, at least not in the obvious place, so I didn't feel like I was making things markedly worse by not including automated tests for this new thing.

**Release notes**

 - `wego app add` will now attempt to use keys from [ssh-agent](https://en.wikipedia.org/wiki/Ssh-agent) for authentication, if it is available

**Documentation Changes**

It might be mentioned in the getting started guide; but, that guide doesn't mention being asked for a private key password, so perhaps it would be considered a distracting detail.


jrryjcksn:
This looks like great functionality but the add command has been significantly refactored since you started this (to share the auth code with `wego app remove`). All of the private key code has been moved to `pkg/osys/osys.go`...

jrryjcksn:
I'm happy to take on the update (particularly since I did the refactoring :-))

jpellizzari:
Slightly related regarding the `.deps` file: https://github.com/weaveworks/weave-gitops/pull/564

jrryjcksn (approved):

luizbafilho (approved):

Makefile
```diff
@@ -52,7 +52,7 @@ clean:
 	rm -rf cmd/ui/dist
 	rm -rf coverage
 	rm -rf node_modules
-	rm .deps
+	rm -f .deps
```
jpellizzari, line 55: :partying_face:

package-lock.json
```diff
@@ -1,14745 +1,8 @@
 {
   "name": "weave-gitops",
   "version": "0.0.1",
-  "lockfileVersion": 2,
+  "lockfileVersion": 1,
   "requires": true,
```
jpellizzari, line 5: I would not expect anything to change here. You can probably just checkout this file on `main` to resolve.

cmd/wego/app/remove/cmd.go
```diff
@@ -78,6 +67,11 @@ func runCmd(cmd *cobra.Command, args []string) error {
 
 	params.GitProviderToken = token
 
+	authMethod, authErr := osysClient.SelectAuthMethod(params.PrivateKey)
+	if authErr != nil {
+		return authErr
```
jpellizzari, line 72: Shadowing `err` is probably ok here.

jpellizzari (approved):
LGTM. Thanks for the extra effort to get this in.

squaremo:
How come there are repeated commits? https://github.com/weaveworks/weave-gitops/pull/566/commits/193cc92e6c9970b103ce04bc75aeccd19189cbdb looks like it's purely a whitespace change, but has the commit message from another change. I'm surprised there's whitespace changes at all -- there aren't any in the original commits. (I reckon just `git commit --amend` to undo accidental whitespace changes).

squaremo:
And what is going on with the commits marked as authored by a bot account, but clearly prepared by a human?

PR #567, opened 2021-08-04
Moving middleware tests to its package

luizbafilho:
**What changed?**
- Moving middleware tests to it's package

**Why?**

**How did you test it?**

**Release notes**

**Documentation Changes**


jpellizzari (changes requested):
Please remove this ping service. It is going to cause confusion. If we have to use a fully implemented gRPC service to test middleware, we may as well just test it with the `applications` service.

luizbafilho:
The reasoning for the ping service is to minimize mocking, as you start to actually implement the applications service and add all the calls it might make it hard to work with, but I'm well removing it, at least for now I don't think it adds a lot of value.

PR #569, opened 2021-08-04
Add Github Auth Device Flow

jpellizzari:
Adds an authentication flow to the `wego app add` command.

Key notes:
* Removes the `ssh` key lookup entirely from the `add` command. Instead, we use a personal access token or temporary access token obtained via the CLI auth flow. Other `git` operations can be done with the deploy key that we push up.
* Lays the ground work for other Git Provider auth flows to be added
* This represents the first step to enabling API/UI auth

**To test, ensure you unset your `GITHUB_TOKEN` env variable**

Here is what the user will see when they run the command without a `GITHUB_TOKEN` set:

```
$ wego app add .

Visit this URL to authenticate with Github:

https://github.com/login/device

Type the following code into the page at the URL above: 82AB-434E

Waiting for authentication flow completion...

```

Then the user enters the code:
Screenshot from 2021-08-04 15-32-06

Then on to the Auth page:

Screenshot from 2021-08-04 15-31-38

Once they visit the page and enter the code, their access token is returned:

```
Authentication successful!

Adding application:
```

### DIFF
--- a/cmd/wego/app/add/cmd.go
+++ b/cmd/wego/app/add/cmd.go
@@ -85,15 +85,33 @@ func runCmd(cmd *cobra.Command, args []string) error {
 	osysClient := osys.New()
 
 	token, err := osysClient.GetGitProviderToken()
-	if err != nil {
-		return err
+
+	if err == osys.ErrNoGitProviderTokenSet {
+		// No provider token set, we need to do the auth flow.
+		url := params.Url
+		if url == "" {
+			// Find the url using an unauthenticated git client. We just need to read the URL.
+			// params.Dir must be defined here because we already checked for it above.
+			url, err = git.New(nil).GetRemoteUrl(params.Dir, "origin")
+			if err != nil {
+				return fmt.Errorf("could not get remote url for directory %s: %w", params.Dir, err)
+			}
+		}
+		// DoAppRepoCLIAuth will take over the CLI and block until the flow is complete.
+		token, err = app.DoAppRepoCLIAuth(url, osysClient.Stdout())
+		if err != nil {
+			return fmt.Errorf("could not complete auth flow: %w", err)
+		}
+	} else if err != nil {
+		// We didn't detect a NoGitProviderSet error, something else went wrong.
+		return fmt.Errorf("could not get access token: %w", err)
 	}
 
 	params.GitProviderToken = token
 
 	authMethod, err := osysClient.SelectAuthMethod(params.PrivateKey)
 	if err != nil {
-		return err
+		return fmt.Errorf("error selecting auth method: %w", err)
 	}
 
 	cliRunner := &runner.CLIRunner{}
--- a/pkg/auth/github.go
+++ b/pkg/auth/github.go
@@ -1,0 +1,174 @@
+package auth
+
+import (
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"net/http"
+	"net/url"
+	"time"
+)
+
+// codeResponse represents response body from the Github API
+type codeResponse struct {
+	DeviceCode     string `json:"device_code"`
+	UserCode       string `json:"user_code"`
+	VerficationURI string `json:"verification_uri"`
+	Interval       int    `json:"interval"`
+}
+
+// Uniquely identifies us as a GitHub app.
+// This does not need to be obfuscated because it is publicly available
+// to anyone who does an OAuth request via wego.
+// See the auth ADR for more details:
+// https://github.com/weaveworks/weave-gitops/blob/main/doc/adr/0005-wego-core-auth-strategy.md#design
+const WeGOGithubClientID = "edcb13588d46f254052c"
+
+// Encapsulate shared logic between doCodeRequest and doAuthRequest
+func doRequest(req *http.Request, client *http.Client) ([]byte, error) {
+	req.Header.Set("Accept", "application/json")
+
+	res, err := client.Do(req)
+	if err != nil {
+		return nil, err
+	}
+
+	if res.StatusCode != http.StatusOK {
+		// err is falsey even on 4XX or 5XX
+		return nil, err
+	}
+
+	rb, err := ioutil.ReadAll(res.Body)
+	if err != nil {
+		return nil, err
+	}
+
+	return rb, nil
+}
+
+const codeRequestURL = "https://github.com/login/device/code?%s"
+
+// doCodeRequest does the initial request of the Device Flow
+func doCodeRequest(client *http.Client, scope string) (*codeResponse, error) {
+	query := url.Values.Encode(map[string][]string{
+		"client_id": {WeGOGithubClientID},
+		"scope":     {scope},
+	})
+
+	req, err := http.NewRequest("POST", fmt.Sprintf(codeRequestURL, query), nil)
+	if err != nil {
+		return nil, err
+	}
+
+	b, err := doRequest(req, client)
+	if err != nil {
+		return nil, fmt.Errorf("error doing code request: %w", err)
+	}
+
+	d := &codeResponse{}
+
+	if err := json.Unmarshal(b, d); err != nil {
+		return nil, fmt.Errorf("could not unmarshal code response: %w", err)
+	}
+
+	return d, nil
+}
+
+var ErrAuthPending = errors.New("auth pending")
+
+const accessTokenUrl = "https://github.com/login/oauth/access_token?%s"
+const githubRequiredGrantType = "urn:ietf:params:oauth:grant-type:device_code"
+
+// It appears we need `repo` scope, which is VERY permissive.
+// We need to be able to push a deploy key and merge commits. No other scopes matched.
+// Available scopes: https://docs.github.com/en/developers/apps/building-oauth-apps/scopes-for-oauth-apps
+const githubOAuthScope = "repo"
+
+type githubAuthResponse struct {
+	AccessToken string `json:"access_token"`
+	Error       string `json:"error"`
+}
+
+// doAuthRequest is used to poll for the status of the device flow.
+func doAuthRequest(client *http.Client, deviceCode string) (string, error) {
+	query := url.Values.Encode(map[string][]string{
+		"client_id":   {WeGOGithubClientID},
+		"device_code": {deviceCode},
+		"grant_type":  {githubRequiredGrantType},
+	})
+	url := fmt.Sprintf(accessTokenUrl, query)
+
+	req, err := http.NewRequest("POST", url, nil)
+	if err != nil {
+		return "", fmt.Errorf("could not create auth request: %w", err)
+	}
+
+	b, err := doRequest(req, client)
+	if err != nil {
+		return "", fmt.Errorf("error doing auth request: %w", err)
+	}
+
+	p := githubAuthResponse{}
+
+	if err := json.Unmarshal(b, &p); err != nil {
+		return "", fmt.Errorf("err marshaling request body: %w", err)
+	}
+
+	if p.Error == "authorization_pending" {
+		// This is expected until the user completes the auth flow.
+		return "", ErrAuthPending
+	}
+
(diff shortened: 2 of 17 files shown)


josecordaz (approved):
💯

jrryjcksn (commented):
The flow looks great, but I'm still wondering about how the normal git operations are supposed to work here. I would have thought we would need to create the git client w/ our generated deploy key :thinking:  I'm happy to be enlightened.

pkg/auth/github.go
```diff
@@ -0,0 +1,163 @@
...
+
+// NewGithubDeviceFlowHandler returns a function which will initiate the Github Device Flow for the CLI.
+func NewGithubDeviceFlowHandler(client *http.Client) BlockingCLIAuthHandler {
+	return func(ctx context.Context, w io.Writer) (string, error) {
+		codeRes, err := doCodeRequest(client, githubOAuthScope)
+		if err != nil {
+			return "", fmt.Errorf("could not do code request: %w", err)
+		}
+
+		fmt.Println()
```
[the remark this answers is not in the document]
jpellizzari: Nice catch. Should be `fmt.FPrintln(w)`.

pkg/auth/github.go
```diff
@@ -0,0 +1,163 @@
...
+const githubOAuthScope = "repo"
+
+type githubAuthResponse struct {
+	AccessToken string `json:"access_token"`
+	Error       string `json:"error"`
+}
+
+// doAuthRequest is used to poll for the status of the device flow.
+func doAuthRequest(client *http.Client, deviceCode string) (string, error) {
+	url := fmt.Sprintf(accessTokenUrlTemplate, WeGOGithubClientID, deviceCode, githubRequredGrantType)
```
JamWils, line 91: FYI: url.URL is an alternative way to do this and a lot cleaner for more complex URLs.

JamWils (commented):
I did a cursory review.  I haven't looked at the end-to-end flow yet.

cmd/wego/app/add/cmd.go
```diff
@@ -81,24 +80,40 @@ func runCmd(cmd *cobra.Command, args []string) error {
...
+			if url == "" {
+				// Find the url.
+				// params.Dir must be defined here because we already checked for it above.
+				url, err = gitClient.GetRemoteUrl(params.Dir, "origin")
+				if err != nil {
+					return fmt.Errorf("could not get remote url for directory %s: %w", params.Dir, err)
+				}
+			}
+			// DoAppRepoCLIAuth will take over the CLI and block until the flow is complete.
+			token, err = app.DoAppRepoCLIAuth(url, os.Stdout)
```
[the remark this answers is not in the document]
jpellizzari: I assumed shadowing was ok in this case, as we return immediately if its not nil. My golang knowledge may have a gap here. Any literature you can point me to?

cmd/wego/app/add/cmd.go
```diff
@@ -81,24 +80,40 @@ func runCmd(cmd *cobra.Command, args []string) error {
...
 	if err != nil {
-		return err
+		if err == osys.ErrNoGitProviderTokenSet {
+			// No provider token set, we need to do the auth flow.
+			url := params.Url
+
+			if url == "" {
+				// Find the url.
+				// params.Dir must be defined here because we already checked for it above.
+				url, err = gitClient.GetRemoteUrl(params.Dir, "origin")
```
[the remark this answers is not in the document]
jpellizzari: This was lifted up from inside a function that already defaulted to `origin`. I thought it was better to explicitly call it out here. Not sure if it gets referenced anywhere else at the moment.

pkg/auth/github.go
jpellizzari: I looked around for some sort of way to build a query from `map[string]string`. Is there a builtin for that?

pkg/gitproviders/provider.go
```diff
@@ -471,3 +472,23 @@ func (p defaultGitProvider) waitUntilRepoCreated(ownerType ProviderAccountType,
...
+// The raw URL is assumed to be something like ssh://git@github.com/myorg/myrepo.git.
+// The common `git clone` variant of `git@github.com:myorg/myrepo.git` is not supported.
+func DetectGitProviderFromUrl(raw string) (GitProviderName, error) {
+	u, err := url.Parse(raw)
+	if err != nil {
+		return "", fmt.Errorf("could not parse git repo url %q", raw)
+	}
+
+	switch u.Hostname() {
+	case "github.com":
```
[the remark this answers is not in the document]
jpellizzari: I had considered that, but these aren't referenced anywhere else, so they are effectively constant here. I can go either way though.

pkg/auth/github.go
jpellizzari: Looks like this might work: https://pkg.go.dev/net/url#Values.Encode

jpellizzari: Refactored to use the `Values.Encode()` function

jpellizzari:
This has changed a bit. As of this PR, we don't try to use the pushed deploy key for `git` operations. Followup issue: https://github.com/weaveworks/weave-gitops/issues/579

cmd/wego/app/add/cmd.go
```diff
@@ -84,15 +84,33 @@ func runCmd(cmd *cobra.Command, args []string) error {
...
+		if url == "" {
+			// Find the url using an unauthenticated git client. We just need to read the URL.
+			// params.Dir must be defined here because we already checked for it above.
+			url, err = git.New(nil).GetRemoteUrl(params.Dir, "origin")
+			if err != nil {
+				return fmt.Errorf("could not get remote url for directory %s: %w", params.Dir, err)
+			}
+		}
+		// DoAppRepoCLIAuth will take over the CLI and block until the flow is complete.
+		token, err = app.DoAppRepoCLIAuth(url, os.Stdout)
```
JamWils, line 100: The shadow variable is still here, can you change this to `authErr`.  I can tell you from experience that these side effects are a pain to debug.

jpellizzari: Please see my other comment on the PR regarding when shadowing is ok.

pkg/gitproviders/provider.go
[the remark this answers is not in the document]
sympatheticmoose: super cursory look, but this might have negative impact trying to work with GHE which I believe would use a structure of github.<ORG>.com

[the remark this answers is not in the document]
jpellizzari: It should return the correct hostname, based on my understanding of how the `url.URL` library works. In this case, we don't currently support Github Enterprise, so we would error here as "not supported" (correctly IMO).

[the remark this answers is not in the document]
sympatheticmoose: We should be looking to support GHE and on-prem versions of the major git providers, given this is what a good proportion of enterprise developers will be using. We can of course address this in a separate issue / later effort to keep moving with this.

We currently just claim support for GitHub, we can fall back on that, my point is this implementation may explicitly will tie us to erroring on GHE, perhaps unnecessarily, and is likely something we'll end up needing to change.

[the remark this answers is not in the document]
jpellizzari: > We should be looking to support GHE and on-prem versions of the major git providers

No argument on that, but we don't currently.

> this implementation may explicitly will tie us to erroring on GHE

That is the idea. As of this PR it should error. If no other changes go into the code base, it should always error.

cmd/wego/app/add/cmd.go
jrryjcksn, line 100: This is really a nit (and it might be unlikely that we would want to mock this particular case) but the `osys` client has a `osysClient.Stdout()` function that could be mocked so I've tried to use that everywhere except the main function of a cli command just in case we want to swap it out :man_shrugging:

pkg/auth/github.go
```diff
@@ -0,0 +1,172 @@
...
+		return nil, fmt.Errorf("could not unmarshal code response: %w", err)
+	}
+
+	return d, nil
+}
+
+var ErrAuthPending = errors.New("auth pending")
+
+const accessTokenUrl = "https://github.com/login/oauth/access_token?%s"
+const githubRequredGrantType = "urn:ietf:params:oauth:grant-type:device_code"
```
jrryjcksn, line 81: NIT: Requred -> Required?

pkg/auth/github.go
```diff
@@ -0,0 +1,172 @@
...
+}
+
+// doCodeRequest does the initial request of the Device Flow
+func doCodeRequest(client *http.Client, scope string) (*codeResponse, error) {
+	query := url.Values.Encode(map[string][]string{
+		"client_id": {WeGOGithubClientID},
+		"scope":     {scope},
+	})
+
+	req, err := http.NewRequest("POST", fmt.Sprintf("https://github.com/login/device/code?%s", query), nil)
```
jrryjcksn, line 59: Given that we're using a constant for the `accessTokenUrl`, should we be using one for the device code url?

pkg/auth/github.go
```diff
@@ -0,0 +1,172 @@
...
+	}
+
+	p := githubAuthResponse{}
+
+	if err := json.Unmarshal(b, &p); err != nil {
+		return "", fmt.Errorf("err marshaling request body: %w", err)
+	}
+
+	if p.Error == "authorization_pending" {
+		// This is expect until the user completes the auth flow.
```
jrryjcksn, line 119: NIT: expected?

pkg/auth/github.go
```diff
@@ -0,0 +1,172 @@
...
+				return "", err
+			}
+
+			ticker.Stop()
+			fmt.Fprintf(w, "Authentication successful!\n\n")
+			return authToken, nil
+
+		}
+
+		return "", nil
```
jrryjcksn, line 170: Mostly a nit (since it looks like this is unreachable) but if we somehow got here, we probably shouldn't return nil for the error (or at least a comment to the effect that it's unreachable so that a cursory examination of the code won't make it appear that this is a valid exit :thinking:

pkg/services/app/add.go
```diff
@@ -292,7 +292,7 @@ func (a *App) getGitRemoteUrl(params AddParams) (string, error) {
 		return "", fmt.Errorf("remote config in %s does not have an url", params.Dir)
 	}
 
-	return sanitizeRepoUrl(urls[0]), nil
+	return utils.SanitizeRepoUrl(urls[0]), nil
```
jrryjcksn, line 295: Seems like this whole function should either be replaced by your new `gogit.GetRemoteUrl` or be implemented in terms of it...

jrryjcksn (commented):
Some minor nits and one leftover redundant function

pkg/services/app/add.go
jpellizzari: I am feeling like we need to add some type safety to these URLs and handling git repository data in general. There are too many strings getting passed around. Probably outside the scope of this PR. I will create a follow up.

cmd/wego/app/add/cmd.go
jpellizzari: Took a look and that call returns a `*File`, which I guess implements the `io.Writer`, so it compiles but seems weird. I'll go with your suggestion so it doesn't bite someone trying to test.

pkg/services/app/add.go
jrryjcksn: I didn't mean the `SanitizeRepoUrl` function, I meant the function containing it `func (a *App) getGitRemoteUrl(params AddParams) (string, error)` (it basically duplicates `GetRemoteUrl`). I couldn't put a comment on the actual code above because it didn't change :-) I should have been more clear what I was referring to but that was the meaning of "whole function".

jpellizzari: I think we are in agreement. We rely on a URL being run through this function for lots of things to work. Same with cluster name. We can't just use any string, we need a `SanitizedURL` type or something.

jrryjcksn: :+1:

Issue #571, opened 2021-08-05
`wego app add` fails when the repo name contains underscore characters


josecordaz:
**Describe the bug**
`wego app add` fails when the repo name contains underscore characters like in `wego app add --url https://github.com/josecordaz/test_repo_with_manifests_path --path ./mycustompathtomanifests --auto-merge --app-config-url https://github.com/josecordaz/test-config-repo`

**Environment**
 - wego: [e.g. v0.1.0]
 - kubernetes:  [e.g. 1.20.4]
    - [ ] KinD - _version_]
    - [x] k3d - v4.4.3
    - [ ] cloud [e.g., EKS, AKS]  _version_
    - [ ] other - _name_ _version_
 - Browser + version: [e.g. chrome 74, safari 12, firefox 87]
 - 
**Affects versions**
v0.2.2

**To Reproduce**
Having a cluster git weave-gitops runtime installed.
1. Add an app with underscore characters being part of the repo name like in here `wego app add --url https://github.com/josecordaz/test_repo_with_manifests_path --path ./mycustompathtomanifests --auto-merge --app-config-url https://github.com/josecordaz/test-config-repo`

**Expected behavior**
It should not throw any error messages

**Config and Logs**
If applicable, add logs to help explain your problem. _please compress the output before attaching_ 
- [ ] wego config
- [ ] `wego` cli call output (and parameters)
- [ ] wego-controller logs
- [ ] Events from wego-* namespaces
- [ ] `kubectl cluster-info dump` 
- [ ] Prometheus alerts

**Screenshots**
![image.png](https://images.zenhubusercontent.com/606389311ead0cc4788a9a72/a197c47b-45e9-41a4-a912-1bb593f0b4c9)If applicable, add screenshots to help explain your problem.

**Additional context**

PR #572, opened 2021-08-05
Sanitize repository name to match flux name validation on git secrets

josecordaz:
Fixes: #571 

**What changed?**
Replace underscore characters with dash characters.

**Why?**
Flux doesn't accept git secret names with underscore values

**How did you test it?**
Manually using a repository with underscore characters in its name

**Release notes**

**Documentation Changes**


jpellizzari (approved):
LGTM. Does the name of this PR still apply? Keep in mind these PR titles get turned into release notes.

josecordaz:
@jpellizzari what do you think about the title change?

jpellizzari:
@josecordaz LGTM

pkg/services/app/add.go
```diff
@@ -851,7 +851,9 @@ func (a *AppResourceInfo) appResourceName() string {
 }
 
 func (a *AppResourceInfo) appSecretName(repoURL string) string {
-	return fmt.Sprintf("wego-%s-%s", a.targetName, utils.UrlToRepoName(repoURL))
+	repoName := utils.UrlToRepoName(repoURL)
+	repoName = strings.ReplaceAll(repoName, "_", "-")
```
JamWils, line 855: Can you write a unit test for this function?

JamWils: It might be worth make a small utility function called SanitizeSecretName in the `utils` package.  There are other places we might want to use this functionality in the future.

PR #573, opened 2021-08-05
fix flux validate check

J-Thompson12:
fixed a sneaky error when running make all.  The if check on a checksum file was not valid.

removed checksum to see if validate doesnt run and put it back so it would run

### DIFF
--- a/tools/functions.sh
+++ b/tools/functions.sh
@@ -109,7 +109,9 @@ do_curl_tarball() {
     dldir="$(mktempdir)"
     mkdir "${dldir}/${cmd}"
     do_curl "${dldir}/${cmd}.tar.gz" "${url}"
-    if ! ${checksums_path} == ""; then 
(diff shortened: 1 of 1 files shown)


tools/functions.sh
```diff
@@ -109,7 +109,9 @@ do_curl_tarball() {
     dldir="$(mktempdir)"
     mkdir "${dldir}/${cmd}"
     do_curl "${dldir}/${cmd}.tar.gz" "${url}"
-    if ! ${checksums_path} == ""; then 
+    ## if checksum_path is set validate tarball
+    if [ ${checksum_path:+1} ]
```
jrryjcksn, line 113: This works but seems a bit subtle. I had to go back and relearn what `:+` does :-)  I think the most common idiom here would be:

`if [ -n "${checksum_path}" ]`

jrryjcksn (approved):

PR #578, opened 2021-08-05
add --helm-release-target-namespace for helm releases and validate all namespaces

### LINKED ISSUE #472: WeGO does not provide support for user-specified application/workload namespace
**Describe the bug**
User cannot specify custom namespace to install applications. In a scenario where the user specifies a namespace for wego-automation AND provides _another_ namespace for workloads, the command would look something like this:

`wego app add --app-config-url=<url> --namespace=<wego-automation> --namespace<workloadNamespace>`

This is confusing from a user's perspective. A better option would be to allow support for installation of apps under user-specified namespace.

**To Reproduce**
1. Download wego
2. `wego gitops install --namespace=<wego-automation>`
3. `wego add app --url=<helmRepoUrl> --chart=<chart> --app-config-url=<url> --namespace<workloadNamespace> --namespace=<wego-automation>`

jrryjcksn:
Resolves #472 

**What changed?**
We now support passing `--helm-release-target-namespace <ns>` to specify the namespace in which to deploy a helm chart. We also apply standard Kubernetes validation to all namespace arguments.

**Why?**
It's unreasonable to ask users to edit their generated `HelmRelease` resources to add a target namespace and that isn't even possible in the `config=NONE` case.

**How did you test it?**
Unit tests and installed `sock-shop` from a helm chart; prior to the fix, all the pods were in the `wego-system` namespace.

**Release notes**
No changes to user environments are required.

**Documentation Changes**
The new argument should be documented in the CLI reference.

### DIFF
--- a/api/v1alpha1/application_types.go
+++ b/api/v1alpha1/application_types.go
@@ -38,6 +38,8 @@ type ApplicationSpec struct {
 	DeploymentType DeploymentType `json:"deployment_type,omitempty"`
 	// SourceType is the type of repository containing the app manifests
 	SourceType SourceType `json:"source_type,omitempty"`
+	// HelmTargetNamespace is the namespace in which to deploy an added Helm Chart
+	HelmTargetNamespace string `json:"helm_target_namespace,omitempty"`
 }
 
 // +kubebuilder:validation:Enum=helm;kustomize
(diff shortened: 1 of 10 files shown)


josecordaz (approved):
LGTM

PR #582, opened 2021-08-06
Automate test for wego app pause/unpause

rokshana-b:
**What changed?**
- Added test for `wego app pause` and its idempotency
- Added test for `wego app unpause` and its idempotency

### DIFF
--- a/test/acceptance/test/add_tests.go
+++ b/test/acceptance/test/add_tests.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"os"
 	"regexp"
+	"strconv"
 
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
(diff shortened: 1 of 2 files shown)


test/acceptance/test/utils.go
```diff
@@ -508,3 +514,41 @@ func mergePR(repoAbsolutePath, prLink string) {
...
+func waitForReplicaCreation(namespace string, replicasSetValue int, timeout time.Duration) error {
+
+	replica := strconv.Itoa(replicasSetValue)
+	pollInterval := 5
+	if timeout < 5*time.Second {
+		timeout = 5 * time.Second
+	}
+	timeoutInSeconds := int(timeout.Seconds())
+
+	for i := pollInterval; i < timeoutInSeconds; i += pollInterval {
```
josecordaz, line 535: there is a function in utils doing this already called `WaitUntil`  it also does retries, the advantage is that it will return the error if it keeps failing in the period of time defined by the timeout. So then we can assert it and it will print out the error if it is different from what are expecting.

test/acceptance/test/add_tests.go
```diff
@@ -870,6 +875,58 @@ var _ = Describe("Weave GitOps Add Tests", func() {
...
+		By("When I pause an app: "+appName1, func() {
+			pauseOutput, _ = runCommandAndReturnStringOutput(WEGO_BIN_PATH + " app pause " + appName1)
+		})
+
+		By("Then I should see pause message", func() {
+			Expect(pauseOutput).To(ContainSubstring("gitops automation paused for " + appName1))
+		})
+
+		By("And changes to the app files should not be synchronized", func() {
+			appManifestFile1, _ = runCommandAndReturnStringOutput("cd " + repoAbsolutePath1 + " && ls")
```
josecordaz, line 888: small nit: it might be a good idea to assert the error here. If this fails for some reason we probably won't know why it failed.

iahmad9 (approved):
Once Jose's comments are fixed, it looks good to me.

PR #588, opened 2021-08-09
Replace github token with one that works

J-Thompson12:
Replace github token with one that actually works. The old token didnt have push permissions. Created a new secret that has push and pull access 

**Why?**

**How did you test it?**
Tested by running the workflow on this branch. It worked.

**Release notes**

**Documentation Changes**

### DIFF
--- a/.github/workflows/repostats.yaml
+++ b/.github/workflows/repostats.yaml
@@ -20,4 +20,4 @@ jobs:
         uses: jgehrcke/github-repo-stats@dab4a915b37a7521cd54033a3147daaeb868ec5c
         with:
           repository: ${{ matrix.statsRepo }}
-          ghtoken: ${{ secrets.GHRS_GITHUB_TOKEN }}
+          ghtoken: ${{ secrets.GHRS_TOKEN }}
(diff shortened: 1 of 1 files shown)


.github/workflows/repostats.yaml
```diff
@@ -20,4 +20,4 @@ jobs:
         uses: jgehrcke/github-repo-stats@dab4a915b37a7521cd54033a3147daaeb868ec5c
         with:
           repository: ${{ matrix.statsRepo }}
-          ghtoken: ${{ secrets.github_token }}
+          ghtoken: ${{ secrets.GHRS_GITHUB_TOKEN }}
```
jpellizzari, line 23: What does `GHRS` mean?

jpellizzari (approved):

.github/workflows/repostats.yaml
J-Thompson12: Git hub repo stats

J-Thompson12: which i see now makes the github part of github token a little weird

jpellizzari: Nit: In this case, no need to save keystrokes. We can just list out exactly what it is: `REPO_STATS_TOKEN` is pretty explicit. It will save the next person who comes along the headache of figuring out what it means.

PR #590, opened 2021-08-10
ADR api-service

josecordaz:
**What changed?**
ADR api-service


doc/adr/0007-api-service.md
```diff
@@ -0,0 +1,106 @@
...
+
+The path to access `wego-api` will be `127.0.0.1:63701` 
+
+This process of exposing `wego-api` will be executed for each command that is executed from CLI. With the exception of the `wego ui run` command that will keep `wego-api` exposure execution in the background for the duration of that command session, so that the UI can access the resource.
+
+### Automate the build of docker images from the `wego-api`.
+
+For production environments, it will be needed a docker repository, a token with permissions to push the image, and public access to those images.
+
+When a `wego` release is made, a Github action will be triggered that will make the corresponding release in the wego-api's docker repository using the same tag used for the` wego` release.
```
jpellizzari, line 85: Lets specify that we want to use the Github container registry for this rather than docker.

jpellizzari (approved):
Nice work!

doc/adr/0007-api-service.md
```diff
@@ -0,0 +1,106 @@
...
+
+resources: ["helmrepositories.source.toolkit.fluxcd.io"]
+verbs: ["get","create"]
+```
+
+These resources will be applied when installing the GitOps engine.
+
+### Enable access to the wego-api from the client-side.
+
+To enable access to the `wego-api` from the client machine, we will use `kubectl` by running this command: `kubectl -n NAMESPACE port-forward deployment/wego-api :8283`. An example of the output would be:
```
palemtnrider, line 71: The user should have the option of the local port to expose.

palemtnrider: Is the user expected to open the port prior to running the command(s)? Or is the port forwarding done on their behalf?

doc/adr/0007-api-service.md
```diff
@@ -0,0 +1,106 @@
...
+
+To enable access to the `wego-api` from the client machine, we will use `kubectl` by running this command: `kubectl -n NAMESPACE port-forward deployment/wego-api :8283`. An example of the output would be: 
+```
+Forwarding from 127.0.0.1:63701 -> 8283
+Forwarding from [::1]:63701 -> 8283
+```
+
+The path to access `wego-api` will be `127.0.0.1:63701` 
+
+This process of exposing `wego-api` will be executed for each command that is executed from CLI. With the exception of the `wego ui run` command that will keep `wego-api` exposure execution in the background for the duration of that command session, so that the UI can access the resource.
```
palemtnrider, line 79: Please elaborate on the `wego ui run` command.  e.g., does it open the port to the service, then open a browser to a certain address that loads the SPA from the API service?

doc/adr/0007-api-service.md
palemtnrider: It's worth mentioning that another ADR will cover API authentication/authorization

doc/adr/0007-api-service.md
```diff
@@ -0,0 +1,106 @@
...
+
+The part that needs to be run in this image is the HTTP server which is initialized in the `wego ui run` command.
+
+To continue communicating with the Kubernetes API, it will be required to change the authentication method to one that allows communication from within the cluster called in-cluster client configuration.
+
+Kubectl and flux will be added to this docker image as well.
+
+### Generate RBAC-based permissions to the wego-api and apply them when installing the GitOps runtime.
+
+Since the wego-api needs to access resources within the cluster, RBAC permissions will be applied so that there are no security gaps. In consideration that the functionality of the wego-api will be specified only in actions within the namespace where the GitOps runtime is executed, the Role and RoleBinding resources will be used. Since the scope of this type of resource is by namespace.
```
palemtnrider, line 38: Will `wego gitops install` install this into the cluster?  If so, please add note about that.  If not, can you elaborate on how it is installed?

doc/adr/0007-api-service.md
```diff
@@ -0,0 +1,106 @@
...
+
+### Create wego-api docker image and update authentication method.
+
+For the wego-api to run within the cluster, a docker image will need to exist that can run it.
+
+The part that needs to be run in this image is the HTTP server which is initialized in the `wego ui run` command.
+
+To continue communicating with the Kubernetes API, it will be required to change the authentication method to one that allows communication from within the cluster called in-cluster client configuration.
+
+Kubectl and flux will be added to this docker image as well.
```
palemtnrider, line 34: I understand flux.  Why is kubectl bundled in the API server?

palemtnrider: The CLI will still require flux for `wego gitops install`.

doc/adr/0007-api-service.md
```diff
@@ -0,0 +1,106 @@
...
+The recommendation is to use `garden.io` because they use yaml files to set the configuration which is something the team is already used to. Rather than `tilt` that relies on python which would enforce learning that language. 
+
+### Reuse the local build of the wego-api inside Github actions for testing.
+
+Since creating a pull request in `wego` triggers several tests. It will also be required to use the construction system of the previous point to create an image that is completely directed for the tests.
+
+To ensure that the construction of images of the wego-api in Github actions is as fast as possible, a mechanism will be implemented that allows sharing the context of the docker builder. One option that seems simple to implement is https://github.com/docker/build-push-action.
+
+## Consequences
+
```
palemtnrider, line 102: Please elaborate on current CLI actions that will be updated.  i.e. version should not report the version of the API server.  Does status change to interact with the API?  etc.

palemtnrider (commented):
Thank you for writing this up.  I have a few questions/comments/suggestions before approving.

doc/adr/0007-api-service.md
josecordaz: We could add the option that the user can choose the port but we might not be getting any additional benefit. The way this command works without specifying local port(`LOCAL_PORT:EXTERNAL_PORT` like `:8383`), is that it will forward a connection to random and not-used local port to the port on the wego-api pod. More details here. https://kubernetes.io/docs/tasks/access-application-cluster/port-forward-access-application-cluster/#let-kubectl-choose-local-port. Actually I don't think we even need a service, we might just need a deployment 🤔

josecordaz: Not sure about the API authentication/authorization. `kubectl port-forward` does that automatically. Do you mean like an extra layer where only wego CLI or UI can access to it ?

doc/adr/0007-api-service.md
josecordaz: There is a note about that at the end of this section `These resources will be applied when installing the GitOps engine.` https://github.com/weaveworks/weave-gitops/pull/590/files#diff-d9dd63abc835dd1b94cd996e0c5cc69d16c4126a4faccad2f6b1ee3934ecf4faR67.

doc/adr/0007-api-service.md
josecordaz: we actually don't need `kubectl` right now

doc/adr/0007-api-service.md
josecordaz: Initially the only command using `wego-api` will be `wego ui run` I'll fix the ADR to mentions this.

doc/adr/0007-api-service.md
```diff
@@ -0,0 +1,107 @@
...
+## Decision
+
+The proposal consists of extracting the API related to applications, which we will call `API-service`, execute it within the cluster through using deployment and expose it through a service. The API-service will be the only entity that will be able to interact directly with the Kubernetes API. In other words, the API-service will expose the necessary options to interact with wego applications, abstracting the logic that interacts with the Kubernetes API.
+
+To illustrate, you could say that the `wego` command line could tell the API-service "Add this application" and it, in turn, would tell Kubernetes "Apply these manifest files. "
+
+For security reasons, permissions will be implemented through RBAC so that the API-service only has access to the resources it needs.
+
+With this proposal, the use of CLI options would still be required, however, it would be an intermediate step to eventually do
```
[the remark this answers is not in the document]
josecordaz: You are right

doc/adr/0007-api-service.md
```diff
@@ -0,0 +1,107 @@
...
+- Apply
+- Delete
+- DeleteByName
+- LabelExistsInCluster
+
+### Create API-service docker image and update authentication method.
+
+For the API-service to run within the cluster, a docker image will need to exist that can run it.
+
+The part that needs to be run in this image is the HTTP server which is initialized in the `wego ui run` command.
```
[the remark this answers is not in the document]
josecordaz: The installation of `wego-api` (previously called `API-service`) will happen on wego installation (`wego gitops install`)  We need to dynamically tell the wego-api deployment which version of the image to use. https://github.com/weaveworks/weave-gitops/pull/590/files#diff-d9dd63abc835dd1b94cd996e0c5cc69d16c4126a4faccad2f6b1ee3934ecf4faR91

doc/adr/0007-api-service.md
```diff
@@ -0,0 +1,107 @@
...
+
+### Automate the build of docker images from the `API-service`.
+
+For production environments, it will be needed a docker repository, a token with permissions to push the image, and public access to those images.
+
+When a `wego` release is made, a Github action will be triggered that will make the corresponding release in the API-service's docker repository using the same tag used for the` wego` release.
+
+At runtime, we will use the version LD flag to tell `wego` which tag to use in the `API-service` deployment.
```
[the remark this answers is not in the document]
josecordaz: We us LD flags to build wego binary

doc/adr/0007-api-service.md
```diff
@@ -0,0 +1,107 @@
...
+
+For production environments, it will be needed a docker repository, a token with permissions to push the image, and public access to those images.
+
+When a `wego` release is made, a Github action will be triggered that will make the corresponding release in the API-service's docker repository using the same tag used for the` wego` release.
+
+At runtime, we will use the version LD flag to tell `wego` which tag to use in the `API-service` deployment. 
+
+### Use source to build locally the API-service docker image for development mode.
+
+Part of the mechanism from the previous point will be used so that it can also be applied in development mode. And thus avoid wasting time publishing / downloading to a docker registry. Currently, there are many tools on the market to achieve this task, to mention a few there are `garden.io` and` tilt`.
```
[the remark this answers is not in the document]
josecordaz: I forgot to mentioned about the two-stage builds in the previous point. But actually there is no need to do it.

doc/adr/0007-api-service.md
palemtnrider: > Please elaborate on current CLI actions that will be updated. i.e. version should not report the version of the API server. Does status change to interact with the API? etc.

I meant to say version **should** report the version of the API server.  I realize only the UI will utilize it for now but we will want to inform the user about it and version is the first place.

palemtnrider (commented):
Looking really good.  I just had a few additional comments.

palemtnrider (approved):
LGTM after the latest round of changes.

JamWils:
LGTM

PR #597, opened 2021-08-11
add list commits for cli

### LINKED ISSUE #562: Implement `get-commits` logic
Based on the output from this spike #514 

It should:
* Show the commits for an application when the user does `wego app get-commits`
* Return a list of commits from the `ListCommits` API response

Not doing:
* Commit pagination (yet)

J-Thompson12:
**What changed?**
Closes #562
Added get-commits cli command to return the latest 10 commits for an app. `wego app <app-name> get commits`
To get the command that way the only way i saw how was to add `get commits` as  arguments and put it all under the main app cmd.go. I also built it in a way that it should be easy to add more options to get which product mentioned may be something we pursue. If there is a better way to do this I am super open to it.

**How did you test it?**
Added unit test for getting user and org repo commits

**Release notes**

**Documentation Changes**
https://github.com/weaveworks/weave-gitops-docs/pull/65


pkg/services/app/commit.go
```diff
@@ -0,0 +1,75 @@
...
+
+	if accountType == gitproviders.AccountTypeUser {
+		userRepoRef := gitproviders.NewUserRepositoryRef(github.DefaultDomain, owner, params.Name)
+		commits, err = gitProvider.GetCommitsFromUserRepo(userRepoRef, app.Spec.Branch)
+		if err != nil {
+			return nil, fmt.Errorf("unable to get Commits for user repo: %s", err)
+		}
+	} else {
+		repoName := utils.UrlToRepoName(app.Spec.URL)
+		orgRepoRef := gitproviders.NewOrgRepositoryRef(github.DefaultDomain, owner, repoName)
```
josecordaz, line 54: it seems like we could use the same variable to pass in the repository name to `NewUserRepositoryRef` and `NewOrgRepositoryRef` in lines 47 and 54

pkg/services/app/commit.go
```diff
@@ -0,0 +1,75 @@
...
+		}
+	} else {
+		orgRepoRef := gitproviders.NewOrgRepositoryRef(github.DefaultDomain, owner, repoName)
+		commits, err = gitProvider.GetCommitsFromOrgRepo(orgRepoRef, app.Spec.Branch)
+		if err != nil {
+			return nil, fmt.Errorf("unable to get Commits for org repo: %s", err)
+		}
+	}
+
+	printCommitTable(commits)
```
jpellizzari, line 61: This printing logic needs to be moved out to the `cmd`. We do not want to print these to the server logs.

pkg/services/app/commit.go
```diff
@@ -0,0 +1,75 @@
...
+
+func printCommitTable(commits []gitprovider.Commit) {
+	header := []string{"Commit Hash", "Author", "Message", "Created At"}
+	rows := [][]string{}
+	for _, commit := range commits {
+		c := commit.Get()
+		rows = append(rows, []string{c.Sha, c.Author, c.Message, c.CreatedAt.String()})
+	}
+
+	utils.PrintTable(os.Stdout, header, rows)
```
jpellizzari, line 74: We don't want to print to `stdout` by default. We should be able to pass in a `logger.Logger` as a dependency here. The logic of where to print the output can be moved out into the `cmd` module.

pkg/services/app/commit.go
```diff
@@ -0,0 +1,75 @@
...
+	}
+
+	var commits []gitprovider.Commit
+	repoName := utils.UrlToRepoName(app.Spec.URL)
+
+	if accountType == gitproviders.AccountTypeUser {
+		userRepoRef := gitproviders.NewUserRepositoryRef(github.DefaultDomain, owner, repoName)
+		commits, err = gitProvider.GetCommitsFromUserRepo(userRepoRef, app.Spec.Branch)
+		if err != nil {
+			return nil, fmt.Errorf("unable to get Commits for user repo: %s", err)
```
jpellizzari, line 51: Probably needs to be a `%w` directive.

pkg/services/app/commit.go
```diff
@@ -0,0 +1,75 @@
...
+		}
+	} else {
+		orgRepoRef := gitproviders.NewOrgRepositoryRef(github.DefaultDomain, owner, repoName)
+		commits, err = gitProvider.GetCommitsFromOrgRepo(orgRepoRef, app.Spec.Branch)
+		if err != nil {
+			return nil, fmt.Errorf("unable to get Commits for org repo: %s", err)
```
jpellizzari, line 57:
```suggestion
			return nil, fmt.Errorf("unable to get Commits for org repo: %w", err)
```

pkg/services/app/commit.go
```diff
@@ -0,0 +1,75 @@
...
+	ctx := context.Background()
+
+	app, err := a.kube.GetApplication(ctx, types.NamespacedName{Name: params.Name, Namespace: params.Namespace})
+	if err != nil {
+		return nil, err
+	}
+
+	owner, err := utils.GetOwnerFromUrl(app.Spec.URL)
+	if err != nil {
+		return nil, fmt.Errorf("failed to retrieve owner: %w", err)
```
jpellizzari, line 31: Fantastic work on these error message throughout :+1:

pkg/utils/common.go
```diff
@@ -9,6 +9,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/olekukonko/tablewriter"
```
jpellizzari, line 12: :+1:

jpellizzari (commented):
Some comments around utilizing this on the server eventually. We can do the HTTP implementation of this in a later PR; just make sure we have a follow up issue created.

pkg/gitproviders/provider.go
```diff
@@ -436,6 +438,47 @@ func (p defaultGitProvider) CreatePullRequestToOrgRepo(orgRepRef gitprovider.Org
...
+	}
+
+	// currently locking the commit list at 10. May discuss pagination options later.
+	commits, err := ur.Commits().ListPage(ctx, targetBranch, 10, 0)
+	if err != nil {
+		return nil, fmt.Errorf("error getting commits for repo[%s] err [%s]", userRepRef.String(), err)
+	}
+
+	if len(commits) == 0 {
+		return nil, fmt.Errorf("targetBranch[%s] has no commits", targetBranch)
```
jpellizzari, line 456:
```suggestion
		return nil, fmt.Errorf("targetBranch [%s] has no commits", targetBranch)
```

pkg/gitproviders/provider.go
```diff
@@ -436,6 +438,47 @@ func (p defaultGitProvider) CreatePullRequestToOrgRepo(orgRepRef gitprovider.Org
...
+		return nil, fmt.Errorf("error getting info for repo [%s] err [%s]", userRepRef.String(), err)
+	}
+
+	// currently locking the commit list at 10. May discuss pagination options later.
+	commits, err := ur.Commits().ListPage(ctx, targetBranch, 10, 0)
+	if err != nil {
+		return nil, fmt.Errorf("error getting commits for repo[%s] err [%s]", userRepRef.String(), err)
+	}
+
+	if len(commits) == 0 {
```
jpellizzari, line 455: Is there ever a case where a branch has no commits? Is this truly an error?

J-Thompson12: Yeah that was left from a test when i was having issues. Nice catch

pkg/services/app/commit.go
```diff
@@ -0,0 +1,75 @@
...
+	"k8s.io/apimachinery/pkg/types"
+)
+
+type CommitParams struct {
+	Name             string
+	Namespace        string
+	GitProviderToken string
+}
+
+func (a *App) GetCommits(params CommitParams) ([]gitprovider.Commit, error) {
```
palemtnrider, line 21: Adding a godoc comment here will help others know what this function does - since it is exported.

pkg/services/app/commit.go
```diff
@@ -0,0 +1,75 @@
...
+	ctx := context.Background()
+
+	app, err := a.kube.GetApplication(ctx, types.NamespacedName{Name: params.Name, Namespace: params.Namespace})
+	if err != nil {
+		return nil, err
```
palemtnrider, line 26: Please add a message on this error.

pkg/gitproviders/provider.go
```diff
@@ -436,6 +438,47 @@ func (p defaultGitProvider) CreatePullRequestToOrgRepo(orgRepRef gitprovider.Org
...
+		return nil, fmt.Errorf("targetBranch[%s] has no commits", targetBranch)
+	}
+
+	return commits, nil
+}
+
+func (p defaultGitProvider) GetCommitsFromOrgRepo(orgRepRef gitprovider.OrgRepositoryRef, targetBranch string) ([]gitprovider.Commit, error) {
```
palemtnrider, line 462: Could use a godoc comment.

pkg/gitproviders/provider.go
```diff
@@ -436,6 +438,47 @@ func (p defaultGitProvider) CreatePullRequestToOrgRepo(orgRepRef gitprovider.Org
...
+	ctx := context.Background()
+
+	ur, err := p.provider.OrgRepositories().Get(ctx, orgRepRef)
+	if err != nil {
+		return nil, fmt.Errorf("error getting info for repo [%s] err [%s]", orgRepRef.String(), err)
+	}
+
+	// currently locking the commit list at 10. May discuss pagination options later.
+	commits, err := ur.Commits().ListPage(ctx, targetBranch, 10, 0)
```
palemtnrider, line 471: nit - the function will be more generic if the caller can specify the number of commits to return. Of course, that will mean guarding against a really large number.

pkg/gitproviders/provider.go
```diff
@@ -436,6 +438,47 @@ func (p defaultGitProvider) CreatePullRequestToOrgRepo(orgRepRef gitprovider.Org
...
+func (p defaultGitProvider) GetCommitsFromUserRepo(userRepRef gitprovider.UserRepositoryRef, targetBranch string) ([]gitprovider.Commit, error) {
+	ctx := context.Background()
+
+	ur, err := p.provider.UserRepositories().Get(ctx, userRepRef)
+	if err != nil {
+		return nil, fmt.Errorf("error getting info for repo [%s] err [%s]", userRepRef.String(), err)
+	}
+
+	// currently locking the commit list at 10. May discuss pagination options later.
+	commits, err := ur.Commits().ListPage(ctx, targetBranch, 10, 0)
```
palemtnrider, line 450: nit - the function will be more generic if the caller can specify the number of commits to return. Of course, that will mean guarding against a really large number.

pkg/gitproviders/provider.go
```diff
@@ -436,6 +438,47 @@ func (p defaultGitProvider) CreatePullRequestToOrgRepo(orgRepRef gitprovider.Org
 	return pr, nil
 }
 
+func (p defaultGitProvider) GetCommitsFromUserRepo(userRepRef gitprovider.UserRepositoryRef, targetBranch string) ([]gitprovider.Commit, error) {
```
palemtnrider, line 441: Please add godoc since this function is exported.

palemtnrider (commented):
Looking good.  A few minor comments.

One question, since we are pretty explicit about on 10 commits, I'd like to see a test on that.  Not sure if that can be done in the unit test or will require and acceptance test...

pkg/gitproviders/provider.go
J-Thompson12: The decision was made to lock it at 10. Figured it would be safer to not allow it until we decided what to do next.

pkg/logger/api.go
```diff
@@ -54,3 +54,8 @@ func (l ApiLogger) Warningf(format string, a ...interface{}) {
 func (l ApiLogger) Failuref(format string, a ...interface{}) {
 	l.logger.Info(fmt.Sprintf(format, a...))
 }
+
+func (l ApiLogger) Write(p []byte) (n int, err error) {
+	// Not implemented yet
```
jpellizzari, line 59: Not OK, we need to implement this. I realize I set a bad example on the `KubeHTTP` library, but there were specific reasons why I left that unimplemented.

The implementation for this can be very simple:

```golang
l.logger.Info(string(p))
```

pkg/logger/logger.go
```diff
@@ -33,4 +33,5 @@ type Logger interface {
 	Successf(format string, a ...interface{})
 	Warningf(format string, a ...interface{})
 	Failuref(format string, a ...interface{})
+	Write(p []byte) (n int, err error)
```
jpellizzari, line 36: We probably need a comment here on why this was added. It can be as simple as something like "added to satisfy `io.Writer` interface.

To add more context: this `interface` probably needs to be re-worked. Forcing the user of this interface to make a decision about which method to use just to get a specific ascii symbol at the front of the log seems like overkill.

pkg/utils/common.go
```diff
@@ -109,3 +111,20 @@ func SanitizeRepoUrl(url string) string {
 
 	return url
 }
+func PrintTable(logger logger.Logger, header []string, rows [][]string) {
```
jpellizzari, line 114: Does this really need a `logger.Logger`, or will any `io.Writer` work?

jpellizzari (approved):
Giving a :+1: with the caveat that we need to implement the `APILogger.Write` function.

I blame myself for setting a bad example, leaving an un-implemented method in the `KubeHTTP` library. In general, that is not an acceptable practice. I will create an issue to address `KubeHTTP`.

cmd/wego/app/cmd.go
```diff
@@ -42,3 +58,60 @@ func init() {
 	ApplicationCmd.AddCommand(pause.Cmd)
 	ApplicationCmd.AddCommand(unpause.Cmd)
 }
+
+func runCmd(cmd *cobra.Command, args []string) error {
+	params := app.CommitParams{}
+	params.Name = args[0]
+	params.Namespace, _ = cmd.Parent().Flags().GetString("namespace")
```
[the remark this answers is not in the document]
J-Thompson12: Agreed but i couldnt see a way to do command arg comand

pkg/utils/common.go
J-Thompson12: Any io.writer will work

jpellizzari: The more open the function signature, the more useful this code will be. I suggest we only require the exact interface that is required.

PR #599, opened 2021-08-12
calculate appHash instead of storing as label

### LINKED ISSUE #541: User changing the branch for an application
If a user changes the branch for an application then the wego-label that we are using to keep from creating duplicate applications is now out of date. The label is created from a hash of path, branch, and url of an application. We need to determine a way to change/validate the label when the branch is changed.

J-Thompson12:
closes #541 

**What changed?**
We no longer store the app hash as a label in case the user changes the branch name. The hash is now calculated on the fly and all applications are checked to see if they match the app hash.

**Why?**
If the user changes the branch name after application is created it causes the label to be out of date. Since we dont have a way to know this happened we now calculate the hash everytime. 

**How did you test it?**
tested manually by adding multiple of the same application and unit tests exist

**Release notes**

**Documentation Changes**
No


jpellizzari (approved):
LGTM. I have an implementation for `AppExistsInCluster` for the `KubeHTTP` struct on my branch.

PR #600, opened 2021-08-12
Add UI embedding ADR

jpellizzari:
Adds an ADR describing how we will embed the Weave GitOps UI and HTTP services in other applications

### DIFF
--- a/doc/adr/0007-embedding-weave-gitops-core.md
+++ b/doc/adr/0007-embedding-weave-gitops-core.md
@@ -4,7 +4,7 @@ Date: 2021-08-12
 
 ## Status
 
-Proposal
+Accepted
 
 ## Problem
 
--- a/doc/adr/0007-embedding-weave-gitops-core.md
+++ b/doc/adr/0007-embedding-weave-gitops-core.md
@@ -1,0 +1,195 @@
+# 7. Embedding Weave GitOps Core
+
+Date: 2021-08-12
+
+## Status
+
+Proposal
+
+## Problem
+
+We want to be able to use the Weave GitOps Core UI elements and support services in other projects. These projects might be other editions of Weave GitOps, or other third-party environments.
+
+Weave GitOps will need to be designed in such as way as to support this type of installation, and provide the necessary facilities to make embedding the UI as simple as possible.
+
+## Design
+
+### WeGo Core UI as a library
+
+The Weave GitOps Core repository will contain an `index.js` file that will define the public API for including WeGO UI components in other projects. That file might look something like this:
+
+```javascript
+import _Timestamp from "./components/Timestamp.tsx";
+import _AppContextProvider from "./contexts/AppContext";
+import { Applications as appsClient } from "./lib/api/applications/applications.pb";
+import _Theme from "./lib/theme";
+import _ApplicationDetail from "./pages/ApplicationDetail";
+import _Applications from "./pages/Applications";
+
+export const Timestamp = _Timestamp;
+export const theme = _Theme;
+export const AppContextProvider = _AppContextProvider;
+export const Applications = _Applications;
+export const ApplicationDetail = _ApplicationDetail;
+export const applicationsClient = appsClient;
+```
+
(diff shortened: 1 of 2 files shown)


AlinaGoaga (commented):
Hey Jordan! Thanks a lot for putting this together! Two questions for you (looping in @foot too): 
- I am not sure how the contexts and nav will work because a provider could have several children underneath using its data (for example, in our case, the template provider data is used in cluster components and template components) so linking one route to one provider would limit access to this data? https://github.com/weaveworks/wks/blob/master/ui-cra/src/components/index.tsx (this is our setup) How do we see this working with multiple contexts? I understand the idea is to keep things as decoupled as possible but with contexts needing each other I am not sure how we can do that. I am also thinking about what David I think was saying to have features in Enterprise where you can deploy an app on multiple clusters which would require these to interact and have access to each other's data.
- I think in terms of nav and the other shell we've got a few elements that are already done so if you're happy with that we will look to move them over? We'll need to decide on the styles as I think we've gone separate ways a bit. Looking fwd to your thoughts! :D 
Thank you!

jpellizzari:
> I am not sure how the contexts and nav will work because a provider could have several children underneath using its data (for example, in our case, the template provider data is used in cluster components and template components) so linking one route to one provider would limit access to this data?

There is no need to scope a `React.Context` provider to a single route. `Context` is "type safe" and unique, so you can mix and match contexts no problem. So you can have a component like this:
```typescript

function MyApp() {
  <ContextProviderA>
    <ContextProviderB>
      <MyComponent />
      <OtherComponent />
    <ContextProviderB />
   <ContextProviderA />
}
```
^^ If `<MyComponent />`  only needs things from `ContextA`, `ContextB` will be ignored. 

> I think in terms of nav and the other shell we've got a few elements that are already so if you're happy with that we will look to move them over?

No need to move them over yet. As soon as we move components, local development gets much harder. Also, our current plan is to have WeGO Enterprise wrap WeGO Core, so dependencies will flow one direction. So you would be moving the component, just to `import` it again.

AlinaGoaga:
Hey Jordan! Yes for sure, that's what we have atm in our application, the wrapping of all components in the contexts. I guess my question was more around that link resolver that you're passing to the context. It gave me the impression that we were trying to link one route to one context. That's the bit that's a bit unclear. How will the multiple contexts look with the link resolver? 
As for the second question I mean that our version of the outer layer has more elements in it so I was suggesting using those as part of core (breadcrumbs, logos, notification component, layout). The dependencies will indeed only flow one way core => enterprise. 
Screenshot 2021-08-13 at 15 13 45
 These will be part of core. Thanks!

jpellizzari:
> How will the multiple contexts look with the link resolver?

The `linkResolver` just translates the incoming URL to another URL. So in the case where the user clicks on a link that has this href:

`/mypath/12345` It can get translated to: `/other_path/12345` in the other application. This keeps the routing logic decoupled for `<Link ... />` components.

It only gets used by the `<Link />` component.

> our version of the outer layer has more elements in it so I was suggesting using those as part of core (breadcrumbs, logos, notification component, layout).

We aren't ready to utilize these components yet in WG. No need to move them just yet.

foot:
> That way, we don't have to rely on routing being an exact match in both projects.

We'll try and keep this aligned between the projects for deep linking from.. docs, things. But its nice to have that flexibility just in case.

But this all makes sense and LGTM! ✨ 

For the initial integration its fine to keep the components as they are, in the future it would be nice to have a common table/buttons etc to keep look and feel consistent between pages (applications vs clusters etc) in WGE too.

jpellizzari:
> For the initial integration its fine to keep the components as they are, in the future it would be nice to have a common table/buttons etc to keep look and feel consistent between pages (applications vs clusters etc) in WGE too.

Agreed. Moving things around will be an ongoing effort.

JamWils (approved):
Met as a team to discuss the ADR.  We are moving forward with this set up.  :lgtm:

PR #624, opened 2021-08-20
Fix deploy

### LINKED ISSUE #608: Complete api part for get commits
https://github.com/weaveworks/weave-gitops/pull/597 Just implemented the cli portion. The api portion needs to be done now. 

The grpc structure
```
message commit {
    string   commit_hash     = 1;  // The hash of the commit
    string   author                 = 2;  // The author of the commit
    string   date                     = 3;  // The date the commit was made. the date type is subject to change.
    string commit_message = 4;  // The commit message
}

message ListCommitsRequest {
    string namespace = 1;  // The namespace the application is in
    string name = 2;  // The application name
}

message ListCommitResponse {
   repeated Commit commits = 1; // A list of commits
}
```

### LINKED ISSUE #623: Gaps in `get commits` output
When I run `get commits`, I get strange blank lines in the output.

```
COMMIT HASH                             	AUTHOR           	MESSAGE                                                                     	CREATED AT                    
fb9e54e2d245619a5b4c519ce4d1ce1666089c54	Weave Gitops     	Add App manifests                                                           	2021-08-04 23:09:51 +0000 UTC	
3be00abc592a785c3f37f1df7aacfda71a8da063	Weave Gitops     	Add App manifests                                                           	2021-08-04 21:30:23 +0000 UTC	
b41e4f70faab45aa739f15db1122ee14a45eeb28	Jordan Pellizzari	remove                                                                      	2021-08-03 23:15:07 +0000 UTC	
f846bfd668436c9288a062fb6aad520a6a0e3083	Jordan Pellizzari	Merge pull request #4 from jpellizzari/wego-b2fd72bf29f5759243beb496871aa724	2021-07-28 13:24:01 +0000 UTC	
                                        	                 	                                                                            	                             	
                                        	                 	wego add podinfo                                                            	                             	
14fda37cb743b71cd7d7e4af41d68b5fb7c75827	Jordan Pellizzari	wego app add https://github.com/jpellizzari/podinfo ./                      	2021-07-28 13:23:30 +0000 UTC	
3a9d4792559a779b771d59ffe4ebb72815a13825	Weave Gitops     	Add App manifests                                                           	2021-07-20 19:24:24 +0000 UTC	
abdfd5d530b3804babacd752112c714b4ab897e5	Weave Gitops     	Add App manifests                                                           	2021-07-20 18:59:47 +0000 UTC	
d6b365a7ef07db511d8bfb81210ebe3465b726ed	Jordan Pellizzari	Merge pull request #1 from jpellizzari/wego-4c3282a70e4a1d6244aa99b2e856ef83	2021-07-20 18:52:22 +0000 UTC	
                                        	                 	                                                                            	                             	
                                        	                 	wego add podinfo                                                            	                             	
680b9e30f8ed5c526a91139b0b55d06aafef4586	Jordan Pellizzari	wego app add  ./                                                            	2021-07-20 18:49:46 +0000 UTC	
320c9f7955a09d74a000954d25bca80b75cbe10e	Weave Gitops     	Add App manifests                                                           	2021-07-01 16:39:45 +0000 UTC	
```

We should also improve the output format. Here are some ideas:

* Commit Hash should be a short hash
* Columns should be ordered so that message is last, since it takes up the most amount of space
* Timestamps should be relative or at least more human readable
* "Message" should be truncated so it only shows the first 50 characters (if it doesn't truncate already) with an ellipsis at the end to indicate the truncation

J-Thompson12:
**What changed?**
Added commits proto and server.  Added pagination and reformatted commits output
closes #608 closes #623 

**Why?**
UI can now get commits list with pagination

**How did you test it?**
unit tests and tested by hand

**Release notes**

**Documentation Changes**


jpellizzari (changes requested):
Commits is not a separate service. Please move this into the `applications.proto` file. I think that will resolve a lot of the merge conflicts as well.

api/applications/applications.proto
```diff
@@ -9,6 +9,8 @@ import "google/api/annotations.proto";
 
 import "protoc-gen-openapiv2/options/annotations.proto";
 
+import "api/applications/commits.proto";
```
jpellizzari, line 12: Sorry if this wasn't clear. Move all the stuff from `commits.proto` into this file please. As it is here, this is creating a second "service" with its own client. We don't want users to have to use two different client libraries to interact with the Applications service.

api/applications/commits.proto
```diff
@@ -0,0 +1,43 @@
...
+    string   date        = 2;  // The date the commit was made.
+    string   author      = 3;  // The author of the commit
+    string   message     = 4;  // The commit message
+}
+
+message ListCommitsRequest {
+    string name               = 1;  // The application name
+    string namespace          = 2;  // The namespace the application is in
+    int32 page_size           = 3;
+    // Optional. A pagination token returned from a previous call
```
jpellizzari, line 33: Nice job on these doc comments.

pkg/server/server.go
```diff
@@ -16,37 +17,44 @@ import (
...
+	app *app.App
 }
 
-// An ApplicationsConfig allows for the customization of an ApplicationsServer.
+// An ServerConfig allows for the customization of an ApplicationsServer.
 // Use the DefaultConfig() to use the default dependencies.
-type ApplicationsConfig struct {
-	Logger     logr.Logger
-	KubeClient kube.Kube
+type ServerConfig struct {
```
jpellizzari, line 43: I think the WeGO enterprise team is relying on this. This might break their integration. What was the reason for this change?

pkg/server/server.go
```diff
@@ -167,6 +177,73 @@ func (s *server) GetApplication(ctx context.Context, msg *pb.GetApplicationReque
...
+	//Remove when middleware is done
+	type contextVals struct {
+		Token *oauth2.Token
+	}
+
+	vals := contextVals{Token: &oauth2.Token{AccessToken: "temptoken"}}
+	ctx = context.WithValue(ctx, tokenKey, vals)
+
+	//Temporary solution to get this to build until middleware is done
+	extractToken := func(ctx context.Context) (string, error) {
```
jpellizzari, line 191: Does this utilize anything from the outer scope of the handler? This is a closure pattern, but I don't see where it is closing over the outer scope variables. Can we move this out of the handler into a local function? You can move it right above this `ListCommits` handler if you want so its easily readable.

Same with the `contextVals` struct.

pkg/server/server_test.go
```diff
@@ -217,6 +226,71 @@ var _ = Describe("Applications handler", func() {
...
+					Branch: "main",
+					Path:   "./k8s",
+					URL:    "github.com/test",
+				},
+			}, nil
+		}
+
+		appSrv := app.New(logger.NewCLILogger(os.Stderr), nil, nil, kubeClient, nil)
+		gitProviders := &gitprovidersfakes.FakeGitProvider{}
+		var commits []gitprovider.Commit
```
jpellizzari, line 249:
```suggestion
		commits := []gitprovider.Commit{commitInfo: gitprovider.CommitInfo{Author: "testAuthor"}}
```

No need to use `append` below if you are going to declare a literal. I (typically) only use `append` in a loop.

pkg/services/app/app.go
```diff
@@ -37,6 +38,8 @@ type AppService interface {
 	Add(params AddParams) error
 	// Get returns a given applicaiton
 	Get(name types.NamespacedName) (*wego.Application, error)
+	// GetCommits returns a list of commits for an application
```
jpellizzari, line 41: :tada:

jpellizzari (changes requested):
A couple major things:
* See my comment regarding multiple proto files. Happy to chat more about this one.
* Watch for breaking package API changes. I guess we haven't formalized which packages cannot change, so maybe we need to do that.

Other minor things:
* It looks like most of the `server` struct fields You can move the `server_test` package into `server` if the only reason for making the `ApplicationsConfig` fields public is for testing. I probably should have set it up that way to begin with.

api/applications/applications.proto
J-Thompson12: This actually only has one service that handles both. I added commits under the application service.

pkg/server/server.go
J-Thompson12: I thought it was done because it only handled the application proto. I wanted to make it cover both. If thats not the case i will certainly change it back.

api/applications/applications.proto
jpellizzari: On second look, you are correct. But we show commits in the context of an application. `/v1/commits` is a bit weird since we don't have a top-level `commit` object, like we do for `application`.

Can you make a case for why they should be separate files?

J-Thompson12: Other than its just cleaner to read and maintain there isnt really another reason why. Ohhhh did you want the url to be v1/application/<appname>/commits because that makes way more sense than the way i did it.

jpellizzari: > Other than its just cleaner to read and maintain

I would argue one file is easier to read and maintain that two, especially since a lot of boilerplate is repeated between them.

> Ohhhh did you want the url to be v1/application/appname/commits because that makes way more sense than the way i did it.

Yes please. URLs are a little weird for us since we aren't using REST ourselves, and we supply a swagger file for auto-generated clients, but I think the `/application/:name/commits` is pretty sane/predictable.

PR #626, opened 2021-08-20
Suppress output during unit tests and fix ginkgo deprecations

### LINKED ISSUE #366: Suppress output while running tests
Currently, it is difficult to understand and troubleshoot unit-test output due to all the log messages that appear when tests are run. Adding a new `fmt.Println` to log a variable, for example, is useless due to the sea of things being logged to console. 

The only thing that should be output during test runs are the failing tests or the status of the tests themselves, plus any extra debug logging.

jrryjcksn:
Resolves #366 

**What changed?**
- Use a fake logger to avoid normal output from add, etc.
- Removed spurious Printf calls
- Fixed async gingko tests to deal with deprecations
- Use a fake osys so messages to stderr get swallowed
- Removed superfluous `writeHeader` call

**Why?**
To clean up test output

**How did you test it?**
Ran the unit tests

**Release notes**
No changes

**Documentation Changes**
No


pkg/auth/github_test.go
```diff
@@ -66,8 +66,6 @@ var _ = Describe("Github Device Flow", func() {
 				})
 				Expect(err).NotTo(HaveOccurred())
 			}
-
-			w.WriteHeader(http.StatusBadRequest)
```
jpellizzari, line 70: :+1:

jpellizzari (approved):
LGTM. Is it just the `go test` output that appears after this?

PR #632, opened 2021-08-23
Add helm charts for weave-gitops

Skarlso:
Helm charts for minimum installation for weave-gitops.


Callisto13:
Context: https://docs.google.com/document/d/1dr6cNUViTrr25LWae-m1dZO0KHYUF_GP3HT3bQeIckg/edit#

jpellizzari:
@palemtnrider Would you mind taking a look at this? I don't have enough context to review

palemtnrider:
Sure thing.

palemtnrider (changes requested):
I think this is a good start.  However, we will need to generate the templates dynamically based on the output from `wego gitops install`.   I believe dynamically generating it will also remove the hard-coded flux version.  If not, we will need the flux version pulled from the dependencies.toml file.

The application version needs to be pulled from the environment/Makefile

I corrected a minor error in the template - and I noticed the test using --dry-run doesn't fail but when trying to install the chart in a cluster it does fail.  We will need to look at the tests.

I tried installing the helm chart (with the fix mentioned above) and my `wego app add . --app-config-url NONE` failed.  The logs from the GitRepository object prints
```
error: no kind "GitRepository" is registered for version "source.toolkit.fluxcd.io/v1beta1" in scheme "k8s.io/kubectl/pkg/scheme/scheme.go:28"
```
I'm not certain what's causing that.

Skarlso:
@palemtnrider thanks for taking a look!

This chart and values ARE the result of running wego gitops install. :)

I'll take a look at the install failure. Sadly, I seem to be getting some other error with kube proxy locally right now. :/ So I'll try to reproduce this.

Skarlso:
Okay, so my env is working now... 

I installed the helm charts and I see no error so far:

Screenshot 2021-08-24 at 10 20 01

I tried running that command:

```
➜  weave-gitops git:(helm_chart) ✗ ./bin/wego app add . --app-config-url NONE
Adding application:

Name: weave-gitops
URL: ssh://git@github.com/weaveworks/weave-gitops.git
Path: ./
Branch: main
Type: kustomize

◎ Checking cluster status
✔ Wego installed
✚ Generating Source manifest
✚ Generating GitOps automation manifests
✚ Generating Application spec manifest
► Applying manifests to the cluster
```

So.... ?

Callisto13:
> we will need to generate the templates dynamically based on the output from wego gitops install. 

Automated generation of charts after wego changes will be done in a later PR (@Skarlso it is part of this task to have a `make` command, or whatever they have going on, to easily generate new charts). This is just the bare minimum to get us started as the existence of this chart blocks everything else.

As for the failure: could be a flake? could be @Skarlso has things existing on his cluster already?

Skarlso:
@Callisto13 nope, cluster is completely new and literally just created with `kind create cluster` and updated kind and refreshed registry. ¯\_(ツ)_/¯ 
In any case, slight change of plan, this chart will now be generated and thus will not have any values in it to have automation around this scenario.

Skarlso:
@palemtnrider Also, the --debug and --dry-run are there to just verify that the helm chart generation works. If there would be any errors in the yaml or the values or anything like that, the helm install would fail. It's not testing if the actual installation works, or not. It's just testing the that helm charts are okay. Testing if the installation works should be handled separately.

palemtnrider:
> @palemtnrider Also, the --debug and --dry-run are there to just verify that the helm chart generation works. If there would be any errors in the yaml or the values or anything like that, the helm install would fail. It's not testing if the actual installation works, or not. It's just testing the that helm charts are okay. Testing if the installation works should be handled separately.
Thanks for the additional context.  I suspected that and was surprised when the --dry-run passed but it failed to install do to the name missing on the chart. I guess that was a valid chart but not actually valid kubernetes manifests.  Is there any harm in running with --debug and no --dry-run?  I think the chart we generate should installing into k8s successfully.

Makefile
```diff
@@ -143,3 +143,8 @@ crd:
 # Check go format
 check-format: 
 	if [ ! -z "$(FORMAT_LIST)" ] ; then echo invalid format at: ${FORMAT_LIST} && exit 1; fi
+
+generate-helm: wego
```
palemtnrider, line 147: Nice!

Please update the clean target to remove the generated file.

Makefile
```diff
@@ -143,3 +143,8 @@ crd:
...
+	kind create cluster --name wego-helm-generator
```
palemtnrider, line 148: It would be nice to save the current kubectl context and set it back at the end of this make target. Otherwise, the user has their kubectl pointing to a cluster that no longer exists.

.github/workflows/helm-chart-test.yaml
```diff
@@ -0,0 +1,20 @@
...
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - name: Run check
+        run: |
+          helm lint helm/weave-gitops
+          kind create cluster
+          helm install --dry-run --debug weave-gitops helm/weave-gitops
+          kind delete cluster
```
palemtnrider, line 20: FYI not having a newline at the end of the file causes diff (and git apply) to warn the user.  Also, if someone were to use >> to append to this file they wouldn't get what they expected.  Do you have strong objections to adding newlines back in?

palemtnrider (approved):
A couple of minor suggestions and then LGTM.

.github/workflows/helm-chart-test.yaml
Skarlso: Nope, I'll add it.

Makefile
Skarlso: Technically, this file is the template to install wego gitops. Why would we want to clean that? It should just exist there and regenerated in case of changes to CRDs or other Manifest files, right?

Callisto13: The Chart will end up being stored and then pulled from S3, so there may not be a reason to have it actually live here full-time.

Skarlso: 👍

PR #659, opened 2021-08-24
Make wait tests reliable by abstracting out the time update

### LINKED ISSUE #576: "Verify WaitUntil" test cases relies on actual time
I occasionally get a test failure that looks like this:

```
• Failure [0.004 seconds]
Test common utils
/home/jordan/Projects/weave-gitops/pkg/utils/common_test.go:40
  Verify WaitUntil prints out proper messages after succeeding at second attempt in less than 1 millisecond [It]
  /home/jordan/Projects/weave-gitops/pkg/utils/common_test.go:56

  Expected
      <time.Duration>: 3750636
  to be <=
      <time.Duration>: 3000000

  /home/jordan/Projects/weave-gitops/pkg/utils/common_test.go:69
------------------------------
•••••

Summarizing 1 Failure:

[Fail] Test common utils [It] Verify WaitUntil prints out proper messages after succeeding at second attempt in less than 1 millisecond 
/home/jordan/Projects/weave-gitops/pkg/utils/common_test.go:69
```

It is generally problematic to rely on actual time passing to validate that code is working in automated tests. We need to refactor this test to be testable synchronously.

jrryjcksn:
Resolves #576 

**What changed?**
`utils.WaitUntil` is now a wrapper around a `timedRepeat` function that accepts a time updater. The tests use reliable addition-based updaters while `utils.WaitUntil` passes in:

```
func(currentTime time.Time) time.Time {
	time.Sleep(poll)
	return time.Now()
},
```

**Why?**
So that unit tests won't fail due to an unusual timing delay

**How did you test it?**
The unit tests were updated to test the actual repeat and check functionality with reliable updaters.

**Release notes**
No

**Documentation Changes**
No

### DIFF
--- a/pkg/utils/common.go
+++ b/pkg/utils/common.go
@@ -35,14 +35,32 @@ func Exists(filePath string) bool {
 
 // WaitUntil runs checkDone until a timeout is reached
 func WaitUntil(out io.Writer, poll, timeout time.Duration, checkDone func() error) error {
-	for start := time.Now(); time.Since(start) < timeout; time.Sleep(poll) {
+	_, err := timedRepeat(
+		out,
+		time.Now(),
+		poll,
+		timeout,
+		func(currentTime time.Time) time.Time {
+			time.Sleep(poll)
+			return time.Now()
+		},
+		checkDone)
+	return err
+}
+
+// timedRepeat runs checkDone until a timeout is reached by updating the current time via a specified operation
+func timedRepeat(out io.Writer, start time.Time, poll, timeout time.Duration, updater func(currentTime time.Time) time.Time, checkDone func() error) (time.Time, error) {
+	currentTime := start
+	endTime := currentTime.Add(timeout)
+
+	for ; currentTime.Before(endTime); currentTime = updater(currentTime) {
 		err := checkDone()
 		if err == nil {
-			return nil
+			return currentTime, nil
 		}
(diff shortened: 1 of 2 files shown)


pkg/utils/common.go
```diff
@@ -35,14 +35,32 @@ func Exists(filePath string) bool {
...
+		checkDone)
+	return err
+}
+
+// timedRepeat runs checkDone until a timeout is reached by updating the current time via a specified operation
+func timedRepeat(out io.Writer, start time.Time, poll, timeout time.Duration, updater func(currentTime time.Time) time.Time, checkDone func() error) (time.Time, error) {
+	currentTime := start
+	endTime := currentTime.Add(timeout)
+
+	for ; currentTime.Before(endTime); currentTime = updater(currentTime) {
```
jpellizzari, line 56: Confused by this `for` syntax and what it is trying to tell me. I didn't know you could do this with the `;`, and it looks like an infinite loop to me. 

I guess it relies on the user `sleep`'ing, which is what we are going for here.

jpellizzari (approved):
A bit confused by the approach, but if it solves the issue and has test coverage, then :+1:

pkg/utils/common.go
jrryjcksn: It's basically like this go example: https://tour.golang.org/flowcontrol/2

No initializer, test = "current time < end time", update current time

Issue #592, opened 2021-08-11
Use deploy keys for Add() using app-config-repo


jpellizzari:
Follow up from #587

Adding the ability to use deploy keys in the above PR proved to be too large a change to make at a single time. Many of the unit tests around the `.Add()` command rely on specific function calls to be run, rather than checking for the outcomes of those calls.

The work to use deploy keys for external repos will require a significant refactor of those tests, possibly replacing them with new ones to preserve code coverage.

Acceptance Criteria:
* After this, we shouldn't need to look at the users environment for `git` operations

Note: Acceptance tests should still work, but they may be doing an unnecessary ssh key lookup. We should handle that in a follow up issue. We also need to remove any dead code that this creates.

jpellizzari:
Not supporting two different Git Providers

jpellizzari:
Related to #581

PR #687, opened 2021-08-30
Use deploy keys everywhere

jrryjcksn:
Closes: #592 

**What changed?**
We now use only deploy keys for auth; no more user private keys

**Why?**
We have no need to reference a user's personal keys

**How did you test it?**
Unit tests, manual tests, acceptance tests

**Release notes**
No

**Documentation Changes**
Yes, we will need to remove the references to private keys


pkg/services/app/app.go
```diff
@@ -49,55 +48,30 @@ type AppService interface {
...
-	GitProviderFactory func(token string) (gitproviders.GitProvider, error)
-	// TODO: @jpellizzari adding this as a temporary stop-gap to maintain the current behavior for external config repos.
-	// As of https://github.com/weaveworks/weave-gitops/pull/587,
-	// we are not addressing this case yet. Many of the unit tests check for exact function call
-	// arguments, which will get skipped when the AuthService is used (and therefore tests will fail).
-	// Follow up issue where this will be addressed: https://github.com/weaveworks/weave-gitops/issues/592
-	temporaryGitClientFactory func(osysClient osys.Osys, privKeypath string) (git.Git, error)
+	Osys        osys.Osys
+	AppGit      git.Git
+	ConfigGit   git.Git
```
jpellizzari, line 53: :tada:

pkg/services/auth/auth.go
```diff
@@ -81,9 +117,42 @@ func NewAuthService(fluxClient flux.Flux, k8sClient client.Client, providerName
 	}, nil
 }
 
-// SetupDeployKey creates a git.Git client instrumented with existing or generated deploy keys.
+// GetGitProvider returns the GitProvider associated with the AuthService instance
+func (a *authSvc) GetGitProvider() gitproviders.GitProvider {
+	return a.gitProvider
```
jpellizzari, line 122: I'm not seeing where `gitProvider` gets set. Do we need to check for `nil` here?

pkg/services/auth/auth.go
```diff
@@ -81,9 +117,42 @@ func NewAuthService(fluxClient flux.Flux, k8sClient client.Client, providerName
...
+		Namespace: namespace,
+	}
+
+	pubKey, keyErr := a.setupDeployKey(ctx, secretName, targetName, normalizedUrl)
+	if keyErr != nil {
+		return nil, fmt.Errorf("error setting up deploy keys: %w", keyErr)
+	}
+
+	if pubKey == nil {
+		// Don't return git.New(pubkey, wrapper.NewGoGit()), nil here. It will fail
```
jpellizzari, line 144: Wild. All `nils` are NOT created equal?

pkg/cliutils/utils.go
```diff
@@ -0,0 +1,124 @@
...
+
+	if isExternalConfig {
+		configRepoClient, configRepoErr := authsvc.CreateGitClient(ctx, targetName, namespace, utils.SanitizeRepoUrl(configUrl))
+		if configRepoErr != nil {
+			return nil, nil, nil, configRepoErr
+		}
+
+		configClient = configRepoClient
+	} else {
+		configClient = appClient
```
jpellizzari, line 102: :+1:

jpellizzari (commented):
Looks excellent. Love the abstraction and cleanliness.

@jrryjcksn For the `ApplicationsServer` struct, I had originally thought an `App` service singleton would hang off the server. I think instead, it will need to be instantiated on every request (after this PR). We won't have a Git Provider token when the API server starts up.

If you agree, we should probably remove it as a struct field from the `ApplicationsServer` object. WDYT?

pkg/services/auth/auth.go
jrryjcksn: It's now an argument to `NewAuthService`. I guess that could be called with nil but is that something we need to guard against?

pkg/services/auth/auth.go
jrryjcksn: Right? That one puzzled me for a while

pkg/services/auth/auth.go
jpellizzari: Ok I must have missed it. I think the constructor is enough protection.

pkg/services/app/app.go
```diff
@@ -49,55 +48,30 @@ type AppService interface {
...
+	Flux        flux.Flux
+	Kube        kube.Kube
+	Logger      logger.Logger
+	GitProvider gitproviders.GitProvider
```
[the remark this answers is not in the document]
jrryjcksn: That's an interesting point. I had definitely been viewing the `App` as a disposable entity. I didn't realize the plan was to keep it around. How much of an issue is that from your point of view?

jrryjcksn:
> Looks excellent. Love the abstraction and cleanliness.
> 
> @jrryjcksn For the `ApplicationsServer` struct, I had originally thought an `App` service singleton would hang off the server. I think instead, it will need to be instantiated on every request (after this PR). We won't have a Git Provider token when the API server starts up.
> 
> If you agree, we should probably remove it as a struct field from the `ApplicationsServer` object. WDYT?

As mentioned in a comment above, I wasn't really aware of this plan. If you're okay with instantiating for each request, I certainly agree that we can remove it from the `ApplicationsServer`; otherwise, I'll revisit the provider change

pkg/services/app/app.go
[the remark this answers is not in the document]
jpellizzari: I don't think its a huge issue. It's a pretty common pattern to make a new class on every request. Testing becomes harder, however.

This was the reason for the GitProviderFactory pattern we were using before. If "User A" makes a request with their token, we need to create a new GitProvider client with their token, since we don't want "User B" to use the same token.

jpellizzari:
> As mentioned in a comment above, I wasn't really aware of this plan. If you're okay with instantiating for each request, I certainly agree that we can remove it from the ApplicationsServer; otherwise, I'll revisit the provider change

I am okay with it. We may need to get creating with unit tests, since the App service will be closed-over by the handler, but there are ways to handle that.

jrryjcksn:
> > As mentioned in a comment above, I wasn't really aware of this plan. If you're okay with instantiating for each request, I certainly agree that we can remove it from the ApplicationsServer; otherwise, I'll revisit the provider change
> 
> I am okay with it. We may need to get creative with unit tests, since the App service will be closed-over by the handler, but there are ways to handle that.

Actually, having gone back and looked again, we were already creating the git client(s) ahead of time and passing them into the app so I think we were already down this road. :man_shrugging:

pkg/server/server.go
```diff
@@ -95,7 +88,12 @@ func NewApplicationsHandler(ctx context.Context, cfg *ApplicationConfig, opts ..
 }
 
 func (s *applicationServer) ListApplications(ctx context.Context, msg *pb.ListApplicationsRequest) (*pb.ListApplicationsResponse, error) {
-	apps, err := s.app.Kube.GetApplications(ctx, msg.GetNamespace())
+	kubeService, kubeError := s.appFactory.GetKubeService()
+	if kubeError != nil {
+		return nil, fmt.Errorf("failed to create kube service: %w", kubeError)
+	}
+
+	apps, err := kubeService.GetApplications(ctx, msg.GetNamespace())
```
jpellizzari, line 96:
```suggestion
	apps, err := kubeService.GetApplications(ctx, msg.Namespace)
```

pkg/apputils/utils.go
```diff
@@ -0,0 +1,191 @@
...
+
+type DefaultAppFactory struct {
+}
+
+func (f *DefaultAppFactory) GetAppService(ctx context.Context, name, namespace string) (*app.App, error) {
+	return GetAppService(ctx, name, namespace)
+}
+
+func (f *DefaultAppFactory) GetKubeService() (kube.Kube, error) {
+	kubeClient, _, kubeErr := kube.NewKubeHTTPClient()
```
jpellizzari, line 36: Not sure what the implications of multiple k8s clients will be. Each will make a separate TCP connection and teardown when its done (I assume). It might already do that with the singleton client.

In any case, something to keep an eye on.

jpellizzari (approved):
LGTM. The factory pattern should solve my early questions about tenancy. Nice work!

pkg/apputils/utils.go
jrryjcksn: Sounds good. We also do this in the paths with an app service so it will require some more work if it turns out to be a significant bottleneck.

pkg/server/server.go
```diff
@@ -95,7 +88,12 @@ func NewApplicationsHandler(ctx context.Context, cfg *ApplicationConfig, opts ..
 }
 
 func (s *applicationServer) ListApplications(ctx context.Context, msg *pb.ListApplicationsRequest) (*pb.ListApplicationsResponse, error) {
-	apps, err := s.app.Kube.GetApplications(ctx, msg.GetNamespace())
+	kubeService, kubeError := s.appFactory.GetKubeService()
```
JamWils, line 91: Minor: I do think that we don't need this on a per call basis.  It might be safer though.  The code is already here.

pkg/apputils/utils.go
```diff
@@ -0,0 +1,191 @@
...
+
+	appClient, configClient, gitProvider, clientErr := getGitClients(ctx, url, configUrl, namespace, isHelmRepository)
+	if clientErr != nil {
+		return nil, fmt.Errorf("error getting git clients: %w", clientErr)
+	}
+
+	return app.New(logger, appClient, configClient, gitProvider, fluxClient, kubeClient, osysClient), nil
+}
+
+func getGitClientsForApp(ctx context.Context, appName, namespace string) (git.Git, git.Git, gitproviders.GitProvider, error) {
```
JamWils, line 106: Idea: the passing around of `git.Git, git.Git, gitproviders.GitProvider` is so common I wonder if a dedicated struct makes sense?

JamWils (commented):
Overall my biggest change is the `*App` to `AppService` on the constructor.  I almost made it a blocker, but given the sheer scope of changes around it I would think we should include it in this PR as well.  

I would still like to review the unit test changes compared to the code changes, but I have not gotten to that point yet.  Great work on this refactoring.  There was a ton of stuff cleaned up.

pkg/apputils/utils.go
```diff
@@ -0,0 +1,191 @@
...
+	if baseClientErr != nil {
+		return nil, fmt.Errorf("error initializing clients: %w", baseClientErr)
+	}
+
+	appClient, configClient, gitProvider, clientErr := getGitClientsForApp(ctx, appName, namespace)
+	if clientErr != nil {
+		return nil, fmt.Errorf("error getting git clients: %w", clientErr)
+	}
+
+	return app.New(logger, appClient, configClient, gitProvider, fluxClient, kubeClient, osysClient), nil
```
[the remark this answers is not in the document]
jrryjcksn: Ha! I actually went back to do that but the application server was referencing the kube client hanging off the app. It wasn't until later that I added `GetKubeService` for the path in which there was no app service available. I can go back and return the interface and have the application server methods pull the kube service separately.

pkg/apputils/utils.go
```diff
@@ -0,0 +1,191 @@
...
+
+	var providerUrl string
+
+	switch {
+	case !isHelmRepository:
+		providerUrl = url
+	case isExternalConfig:
+		providerUrl = configUrl
+	default:
+		return nil, nil, nil, nil
```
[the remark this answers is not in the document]
jrryjcksn: I'll think about this, but the path you're referring to is the "helm chart from helm repository" path with no external configuration which has no git activity at all and we will always have acceptance tests for that case as long as we maintain the current "app-config-url" model. This isn't for Pause/Unpause/Status. They all get normal git clients with this change (unless they are referencing helm charts from helm repos, in which case there is no git involved). It's not clear we won't want status, for example, to access git at some time in the future.

cmd/wego/app/cmd.go
```diff
@@ -77,21 +71,12 @@ func runCmd(cmd *cobra.Command, args []string) error {
...
-	logger := logger.NewCLILogger(os.Stdout)
-	if err := app.IsClusterReady(logger, kube); err != nil {
-		return err
+	appService, appError := apputils.GetAppService(ctx, params.Name, params.Namespace)
+	if appError != nil {
+		return fmt.Errorf("failed to create app service: %w", appError)
 	}
 
-	appContent, err := kubeClient.GetApplication(ctx, types.NamespacedName{Name: params.Name, Namespace: params.Namespace})
+	appContent, err := appService.Kube.GetApplication(ctx, types.NamespacedName{Name: params.Name, Namespace: params.Namespace})
```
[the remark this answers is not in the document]
jrryjcksn: Because the app service doesn't _have_ a `GetApplication` method.

pkg/server/server_test.go
```diff
@@ -92,7 +91,14 @@ var _ = Describe("ApplicationsServer", func() {
 			BeforeEach(func() {
 				log = makeFakeLogr()
 				kubeClient = &kubefakes.FakeKube{}
-				appsSrv = NewApplicationsServer(&ApplicationConfig{App: app.New(nil, nil, nil, kubeClient, nil)})
+				appFactory := &apputilsfakes.FakeAppFactory{}
+				appFactory.GetAppServiceStub = func(ctx context.Context, name, namespace string) (*app.App, error) {
+					return app.New(nil, nil, nil, nil, nil, kubeClient, nil), nil
```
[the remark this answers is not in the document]
jrryjcksn: Because the application server wouldn't know to create the mocked `AppService`. We configure it via the factory so that when it goes to create an `AppService` during a call, it creates the mock.

pkg/apputils/utils.go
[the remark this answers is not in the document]
jrryjcksn: The reason for this function is that the `ListApplications` call has no application context from which to create an `AppService`. The other methods retrieve an `AppService` by looking up all the relevant information in the `Application` resource from the cluster. So, we needed a way to get a Kube client without access to an `AppService`

pkg/apputils/utils.go
jrryjcksn: Maybe? It's only two functions in one file :man_shrugging:

pkg/gitproviders/provider.go
```diff
@@ -265,6 +266,23 @@ func (p defaultGitProvider) GetDefaultBranch(url string) (string, error) {
 	return "main", nil
 }
 
+func (p defaultGitProvider) GetRepoVisibility(url string) (*gitprovider.RepositoryVisibility, error) {
```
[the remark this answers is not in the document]
jrryjcksn: K

pkg/apputils/utils.go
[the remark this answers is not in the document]
JamWils: Actually, this might become an issue when we implement multi-cluster.

pkg/apputils/utils.go
JamWils: And then they are passed into `app.App`.

pkg/apputils/utils.go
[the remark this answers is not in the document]
JamWils: I definitely missed that.  I know this is WAY out of scope for this PR, but it almost seems like there should be two separate implementations of `AppService` as well `app.GitService` and `app.HelmService`.  The two shouldn't be mixed like they are.  What happens when an app is composed of something else entirely?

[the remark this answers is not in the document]
JamWils: > we will always have acceptance tests for that case as long as we maintain the current "app-config-url" model.

As a team I think we need to move out of this mindset.  There should also be unit tests proving this fact.

pkg/apputils/utils.go
JamWils: Chatted f2f, not pull this thread on this PR given the large number of changes.

pkg/server/server.go
JamWils: Talked f2f and we want to limit the scope of the refactoring.  This is good.

cmd/wego/app/cmd.go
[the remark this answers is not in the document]
JamWils: This was fixed.

pkg/services/app/add_test.go
```diff
@@ -368,24 +270,9 @@ stringData:
 			}
 		})
 
-		Describe("avoids deploy key for helm", func() {
```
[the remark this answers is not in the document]
JamWils: Chatted on slack, this is corrrect.

Issue #690, opened 2021-08-30
npm package is not exporting anything for commonjs


foot:
Trying to consume the library we get runtime errors in the browser about `ApplicationDetail` etc being `undefined`.

_Note: I'm not sure if this a completely valid way of validating the export mechanism, but:_

After running `npm run build:lib`:

```
$ node
Welcome to Node.js v14.15.1.
Type ".help" for more information.
> require('./dist/index.js')
{}
```

Adding the `main` field to package.json as suggested in https://v2.parceljs.org/getting-started/library/ (and I think we were doing this before?):

```diff
diff --git a/package.json b/package.json
index b93d5bb..7aebc9c 100644
--- a/package.json
+++ b/package.json
@@ -2,6 +2,7 @@
   "name": "@weaveworks/weave-gitops",
   "version": "0.2.4",
   "description": "Weave GitOps core",
+  "main": "dist/index.js",
   "scripts": {
     "preinstall": "npm install --package-lock-only --ignore-scripts && npx npm-force-resolutions",
     "build": "parcel build --no-source-maps ui/index.html --dist-dir cmd/wego/ui/run/dist",
```

Works better:

```
$ node
Welcome to Node.js v14.15.1.
Type ".help" for more information.
> require('./dist/index.js')
{
  theme: [Getter],
  applicationsClient: [Getter],
  AppContextProvider: [Getter],
  ApplicationDetail: [Getter],
  Applications: [Getter]
}
```

And we can consume the lib using webpack now too and everything is happy in the browser.

foot:
However, `dist/main.js` is not a correct path w/ our current build mechanism of `cp package.json dist` and might need some tweaking.

PR #693, opened 2021-08-31
Adds multiple build targets and configurations to package.json

foot:
Closes: #690 

**What changed?**

Splits parcel build into multiple build targets, the library one now exposes commonjs and doesn't bundle react and node_modules.

**Why?**

As far as I understand it `parcel` infers how you want to build something based on what package.json looks like.
- If `main` is declared in `package.json` then it flips into "library mode" https://v2.parceljs.org/features/targets/#library-targets and builds commonjs artifacts
- However if `main` is declared then you get funny build errors when trying to build the wego UI..

I'm not sure if you can have a single package.json that builds for ui _and_ library. (?) There are build flags that you can pass to `parcel build` to customize things, but some required options are missing like `outputFormat: commonjs`.

Using `targets` seems to be a way to escape a single `package.json` inferred build config and specify things serparately.

**Other things we could try**

- **esm** (?), maybe we could build something that works for both.

**How did you test it?**

```
./make
./bin/wego run ui
open localhost:9001
```

seems to work

```
npm start
open localhost:4567
```

seems to work

**Release notes**

**Documentation Changes**

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,10 +2,25 @@
   "name": "@weaveworks/weave-gitops",
   "version": "0.2.4",
   "description": "Weave GitOps core",
+  "targets": {
+    "default": {
+      "distDir": "cmd/wego/ui/run/dist",
+      "source": "ui/index.html",
+      "sourceMap": false
+    },
+    "lib": {
+      "includeNodeModules": false,
+      "isLibrary": true,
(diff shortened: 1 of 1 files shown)


jpellizzari (approved):
Nice!

PR #694, opened 2021-08-31
Remove token arg

### LINKED ISSUE #377: Add automation around keeping in sync with Flux releases
As part of our discussion on vulnerability tracking, we decided it would be valuable to stay in sync with Flux binary and controller releases.

Automation can take several forms. Some ideas:
* A new GH issue gets made in the weave-gitops repo when a flux version is released, prompting someone to upgrade wego
* A PR is opened with the new flux version "embedded" for review

Some prior research from [this issue](https://github.com/fluxcd/webui/issues/42):

>@jpellizzari yes we need to keep the UI in-sync with flux2 release, in the same way we do for terraform provider. Some >examples:
>- https://github.com/fluxcd/flux2/blob/main/.github/workflows/update.yaml
>- https://github.com/fluxcd/terraform-provider-flux/blob/main/.github/workflows/update.yaml

>_Originally posted by @stefanprodan in https://github.com/fluxcd/webui/issues/41#issuecomment-840634274_

jpellizzari:
Closes #377 

Uses the `GITHUB_TOKEN` default to create a PR on a cron job.

### DIFF
--- a/.github/workflows/upgrade-flux.yaml
+++ b/.github/workflows/upgrade-flux.yaml
@@ -29,7 +29,7 @@ jobs:
         id: cpr
         uses: peter-evans/create-pull-request@9825ae65b1cb54b543b938503728b432a0176d29 # v3
         with:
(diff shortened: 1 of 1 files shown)


josecordaz (approved):
LGTM

PR #696, opened 2021-08-31
Add useApplications to the UI exports

AlinaGoaga:
**What changed?**
The UI is also exporting the `useApplications` hook and `LoadingPage` now. 
Updated Loader style to match what we do in WE in terms of wrapping content. 

**Why?**
The export is needed in WE to have access to the number of applications.
The LoadingPage will be used in WE instead of what we have atm to ensure consistency.


jpellizzari (approved):

AlinaGoaga:
Sorry @jpellizzari realised today that there's a loading page that comes with the apps which looks a bit different to what we have. I've made some adjustments and we'll use this one (added to the exports as well). Thanks!

jpellizzari (approved):

Issue #666, opened 2021-08-25
Remove remaining hardcoded references to github providername


jpellizzari:
We want it to be easy to add new providers. So all of the hardcoded references to gihtub need to be made generic.

PR #701, opened 2021-08-31
add gitlab for --app-config-url=none path

J-Thompson12:
Related To: #666

**What changed?**
Removes hardcoded github code and allows the path --a
until the deploy key stuff is setup for add the outcome is `Error: error setting up deploy keys: error getting account type: could not get account type GET https://gitlab.com/api/v4/groups/J-Thompson12: 404 {message: 404 Group Not Found}`

**Why?**

**How did you test it?**
manually and acceptance tests

**Release notes**

**Documentation Changes**

### DIFF
--- a/cmd/wego/app/add/cmd.go
+++ b/cmd/wego/app/add/cmd.go
@@ -113,24 +113,28 @@ func runCmd(cmd *cobra.Command, args []string) error {
 		}
 	}
 
-	var providerName gitproviders.GitProviderName
+	var normalizedUrl gitproviders.NormalizedRepoURL
 	// We re-use the same --url flag for both git and helm sources.
 	// There isn't really a concept of "provider" in helm charts, and there is nothing to push.
 	// Assume charts are always public and no auth needs to be done.
 	if !isHelmRepository {
-		providerName, err = gitproviders.DetectGitProviderFromUrl(repoUrlString)
+		normalizedUrl, err = gitproviders.NewNormalizedRepoURL(repoUrlString)
 		if err != nil {
-			return fmt.Errorf("error detecting git provider: %w", err)
+			return fmt.Errorf("error creating normalized url: %w", err)
 		}
+		params.GitProviderName = normalizedUrl.Provider()
+	} else {
+		// helm doesnt use the gitprovider but it still needs to be set to something for now
+		params.GitProviderName = gitproviders.GitProviderGitHub
 	}
 
-	token, tokenErr := osysClient.GetGitProviderToken()
+	token, tokenErr := osysClient.GetGitProviderToken(params.GitProviderName)
 
 	if !isHelmRepository && tokenErr == osys.ErrNoGitProviderTokenSet {
 		// No provider token set, we need to do the auth flow.
-		authHandler, err := auth.NewAuthCLIHandler(providerName)
+		authHandler, err := auth.NewAuthCLIHandler(params.GitProviderName)
 		if err != nil {
-			return fmt.Errorf("could not get auth handler for provider %s: %w", providerName, err)
+			return fmt.Errorf("could not get auth handler for provider %s: %w", params.GitProviderName, err)
 		}
 
 		token, err = authHandler(ctx, osysClient.Stdout())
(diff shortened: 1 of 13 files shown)


cmd/wego/app/add/cmd.go
```diff
@@ -113,24 +113,28 @@ func runCmd(cmd *cobra.Command, args []string) error {
...
-		providerName, err = gitproviders.DetectGitProviderFromUrl(repoUrlString)
+		normalizedUrl, err = gitproviders.NewNormalizedRepoURL(repoUrlString)
 		if err != nil {
-			return fmt.Errorf("error detecting git provider: %w", err)
+			return fmt.Errorf("error creating normalized url: %w", err)
 		}
+		params.GitProviderName = normalizedUrl.Provider()
+	} else {
+		// helm doesnt use the gitprovider but it still needs to be set to something for now
+		params.GitProviderName = gitproviders.GitProviderGitHub
```
jpellizzari, line 128: @J-Thompson12 How would you feel about a `gitproviders.GitProviderNone` or something like that. I would also be okay with setting this to some zero value.

Do you have a sense for where this is required? Is it only in the `auth.NewAuthCLIHandler` or is it other places?

cmd/wego/app/add/cmd.go
```diff
@@ -113,24 +113,28 @@ func runCmd(cmd *cobra.Command, args []string) error {
 		}
 	}
 
-	var providerName gitproviders.GitProviderName
+	var normalizedUrl gitproviders.NormalizedRepoURL
 	// We re-use the same --url flag for both git and helm sources.
 	// There isn't really a concept of "provider" in helm charts, and there is nothing to push.
 	// Assume charts are always public and no auth needs to be done.
 	if !isHelmRepository {
-		providerName, err = gitproviders.DetectGitProviderFromUrl(repoUrlString)
```
jpellizzari, line 121: Do we still need the `DetectGitProviderFromUrl` after this? Maybe it can be made private now?

pkg/services/app/remove.go
```diff
@@ -99,7 +100,7 @@ func (a *App) getConfigUrlAndBranch(info *AppResourceInfo, token string) (string
 	if cloneURL == string(ConfigTypeUserRepo) {
 		cloneURL = info.Spec.URL
 	} else {
-		gitProvider, err := a.GitProviderFactory(token)
+		gitProvider, err := a.GitProviderFactory(token, gitproviders.GitProviderGitHub)
```
jpellizzari, line 103: Only supporting Github in `remove` for now?

jpellizzari (commented):
Looks good. The original issue (#666) was to prove that `wego app add` would work with a Gitlab repo. Do we think that will be working as of this PR?

cmd/wego/app/add/cmd.go
J-Thompson12: its being used here buildGitProvider() to create the git provider. I can certainly find a way to skip all of that but the current behavior is to look for GITHUB_TOKEN even if its helm. i am pretty sure all this behavior will change soon with @jrryjcksn changes.

pkg/services/app/remove.go
J-Thompson12: Yeah with my understanding of the ticket it was just for getting add to work in the app-config-url=none path.

cmd/wego/app/add/cmd.go
J-Thompson12: we get the providername in NewNormalizedRepoURL(). So instead of doing it twice i just use the normalizedUrl to return it.

jpellizzari: I guess I am asking if that function needs to be exported?

cmd/wego/app/add/cmd.go
jpellizzari: Ok nevermind. We can leave Github as the default then.

jpellizzari (approved):

jpellizzari:
@J-Thompson12 LGTM, but lets wait for another reviewer to take a look. Can you create yourself a Gitlab org and test manually for now?

It's ok to add the Gitlab story in small pieces, and this PR makes the code base better, so don't feel like you have to solve the whole Gitlab story in one go. Let's keep the issue open until we know that Gitlab works.

Issue #644, opened 2021-08-23
Create authentication plumbing HTTP methods


jpellizzari:
Per the auth ADR, we will need a couple of handlers to complete our auth feature(s). These will express themselves as RPC methods on the `Applications` service (this may be moved to a separate service in the future).

Auth ADR: https://github.com/weaveworks/weave-gitops/blob/main/doc/adr/0005-wego-core-auth-strategy.md

Some examples from a previous implementation. The actual implementation may differ:

#### Authenticate
```
    /**
    * Authenticate exchanges a code recieved from an OAuth2 callback for a user token
    */
    rpc Authenticate(AuthenticateRequest) returns (AuthenticateResponse) {
        option (google.api.http) = {
            post : "/v1/authenticate/{provider_name}"
            body: "*"
        };
    }

message AuthenticateRequest {
    string provider_name = 1;  // The name of the OAuth provider
    string access_token          = 2;
}

message AuthenticateResponse {
    string token = 1; // The token that was exchanged for the callback code with the OAuth2 provider
}

```
Should accept and access token and return a converted WeGO JWT with the Git Provider access token embedded. the browser will store the WeGO JWT string and use it in future API requests.

Previous closed PR for reference: https://github.com/weaveworks/weave-gitops/pull/481/files#diff-ea5977af0f12b795d8d25e75b822ad72b0f57db936dd535c6aa8e2922bb85804R51

PR #704, opened 2021-08-31
Added validate method to Application server

josecordaz:
Closes: #644 

**What changed?**
Added end point Validate to generate JTW tokens

**Why?**
This was added as a first stage to start securing the access to the api

**How did you test it?**
Added unit tests

**Release notes**

**Documentation Changes**


api/applications/applications.proto
```diff
@@ -20,6 +20,16 @@ option (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_swagger) = {
 };
 
 service Applications {
+    /**
+    * Authenticate exchanges a code received from an OAuth2 callback for a user token
```
jpellizzari, line 24: Let re-work this comment. It doesn't really do this in our case. This will end up as auto-generated API documentation.

api/applications/applications.proto
```diff
@@ -64,6 +74,15 @@ message Application {
     repeated Condition deployment_conditions = 5;  // A list of conditions for the Kustomization or HelmRelease for this application
 }
 
+message AuthenticateRequest {
+    string provider_name = 1;  // The name of the OAuth provider
+    string access_token          = 2;
```
jpellizzari, line 79: Need comments here, and lets try to get these aligned.

api/applications/applications.proto
```diff
@@ -64,6 +74,15 @@ message Application {
...
+}
+
+message AuthenticateResponse {
+    string token = 1; // The token that was exchanged for the callback code with the OAuth2 provider
```
jpellizzari, line 83: Comment needs a re-work

pkg/server/server.go
```diff
@@ -308,3 +315,23 @@ func mapConditions(conditions []metav1.Condition) []*pb.Condition {
 
 	return out
 }
+
+//Until the middleware is done this function will not be able to get the token and will fail
+func (s *applicationServer) Authenticate(ctx context.Context, r *pb.AuthenticateRequest) (*pb.AuthenticateResponse, error) {
+
+	if !strings.HasPrefix(github.DefaultDomain, r.ProviderName) &&
```
jpellizzari, line 322: Let's use `NormalizedRepoURL` here

pkg/server/server.go
```diff
@@ -308,3 +315,23 @@ func mapConditions(conditions []metav1.Condition) []*pb.Condition {
 
 	return out
 }
+
+//Until the middleware is done this function will not be able to get the token and will fail
+func (s *applicationServer) Authenticate(ctx context.Context, r *pb.AuthenticateRequest) (*pb.AuthenticateResponse, error) {
```
jpellizzari, line 320: Nit: the convention we have been using is to label the request payload `msg`. `r` is generally reserved for an `http.Request` object.

pkg/services/auth/jwt.go
```diff
@@ -0,0 +1,67 @@
...
+
+// Claims is a custom JWT claims that contains some token information
+type Claims struct {
+	jwt.StandardClaims
+	Provider      string `json:"provider"`
+	ProviderToken string `json:"provider_token"`
+}
+
+// Generate generates and signs a new token
+func Generate(secretKey string, expirationTime time.Duration, providerName, providerToken string) (string, error) {
```
jpellizzari, line 24: We have a typed provider name now. I believe its `gitproviders.GitProviderNameGithub`

pkg/services/auth/jwt_test.go
```diff
@@ -0,0 +1,33 @@
...
+
+		claims, err := Verify(SecretKey, jwtToken)
+		Expect(err).NotTo(HaveOccurred())
+
+		Expect(claims.Provider).To(Equal(provider))
+		Expect(claims.ProviderToken).To(Equal(token))
+
+		time.Sleep(time.Second)
+		claims, err = Verify(SecretKey, jwtToken)
+		Expect(err.Error()).To(MatchRegexp("invalid token: token is expired by.*"))
```
jpellizzari, line 28: Let's avoid checking a specific error message. Can we check for a specific error `type` instead?

pkg/server/server_test.go
```diff
@@ -79,6 +82,45 @@ var _ = Describe("ApplicationsServer", func() {
...
+	It("Authorize fails on wrong provider", func() {
+		ctx := context.Background()
+		provider := "wrong provider"
+		token := "token"
+
+		_, err := appsClient.Authenticate(ctx, &pb.AuthenticateRequest{
+			ProviderName: provider,
+			AccessToken:  token,
+		})
+		Expect(err.Error()).NotTo(Equal(fmt.Sprintf("unknown provider name %s, expecting github or gitlab", provider)))
```
jpellizzari, line 110: Seems like an odd assertion here. We are asserting that an error is NOT like this string? So it could be anything. Custom error types will make this safer.

pkg/server/server_test.go
```diff
@@ -79,6 +82,45 @@ var _ = Describe("ApplicationsServer", func() {
...
+		Expect(err).NotTo(HaveOccurred())
+
+		Expect(res.Token).To(Equal(expectedToken))
+	})
+	It("Authorize fails on wrong provider", func() {
+		ctx := context.Background()
+		provider := "wrong provider"
+		token := "token"
+
+		_, err := appsClient.Authenticate(ctx, &pb.AuthenticateRequest{
```
jpellizzari, line 106: We should be testing for a specific HTTP error code here

jpellizzari (commented):
Looking good. I listed some changes that I think will really improve maintainability.

pkg/server/server.go
josecordaz: not sure I understand this 🤔, what I'm trying to do is just to make sure the provider name passed in is either `github` or `gitlab` using the constants from `go-git-providers` 🙂

pkg/services/auth/jwt_test.go
josecordaz: that sounds like a good idea, but not really sure how to do it when the error is dynamic 🤔 . That is why I'm using RegExp here. It could fail with `invalid token: token is expired by 4s` or `invalid token: token is expired by 3s`.

pkg/server/server_test.go
josecordaz: good catch 🙂

pkg/server/server_test.go
josecordaz: not sure I understand, this is one of the possible errors `rpc error: code = Unknown desc = unknown provider name wrongprovider, expecting github or gitlab`

jpellizzari: Sorry I thought we had a reference to HTTP errors already. 

You can use the runtime package to return a specific error code. An error case might look like this:
```
	if err != nil {
		return nil, &runtime.HTTPStatusError{HTTPStatus: http.StatusUnauthorized, Err: err}
	}
```

We need to ensure the correct error code is being returned. The UI and API users will configure their behavior based on this code.

pkg/services/auth/jwt_test.go
jpellizzari: There is no need to include the token expiration delta in the message. That might actually be a security risk.

pkg/server/server.go
jpellizzari: `NormalizedRepoURL` has some helper methods to return a type-safe git provider name: https://github.com/weaveworks/weave-gitops/blob/main/pkg/gitproviders/provider.go#L624

jpellizzari: Nevermind I misunderstood. Let's try using the typesafe provider names rather than looking at string prefixes

pkg/services/auth/jwt_test.go
josecordaz: I'll replace the error msg coming from the jwt library with `unauthorized token`, not sure if that would be accurate 🤔

jpellizzari: Testing for another library's error string is a sure fire way to have flaky tests. If the other lib changes the error message by 1 character, unit tests will suddenly break once we upgrade that dependency.

josecordaz: @jpellizzari I have receive feedback from James to assert error messages in previous PRs instead of just doing `.HaveOccurred()`. But I could be misunderstanding something. I'm not sure what I have to do here. @JamWils would you mind to help me clear this up ?

pkg/services/auth/jwt.go
```diff
@@ -0,0 +1,73 @@
...
+
+// Claims is a custom JWT claims that contains some token information
+type Claims struct {
+	jwt.StandardClaims
+	Provider      gitproviders.GitProviderName `json:"provider"`
+	ProviderToken string                       `json:"provider_token"`
+}
+
+// Generate generates and signs a new token
+func Generate(secretKey string, expirationTime time.Duration, providerName gitproviders.GitProviderName, providerToken string) (string, error) {
```
luizbafilho, line 27: I think `Generate` it's a bit too generic, I think `GenerateJWT` is probably easier to read and  disambiguate from the provider token

pkg/services/auth/jwt.go
```diff
@@ -0,0 +1,73 @@
...
+		Provider:      providerName,
+		ProviderToken: providerToken,
+	}
+
+	token := jwt.NewWithClaims(jwt.SigningMethodHS512, claims)
+	return token.SignedString([]byte(secretKey))
+}
+
+// Verify verifies the access token string and return a user claim if the token is valid
+func Verify(secretKey string, accessToken string) (*Claims, error) {
```
luizbafilho, line 41: Same as `Generate`

pkg/services/auth/jwt.go
```diff
@@ -0,0 +1,73 @@
...
+			_, ok := token.Method.(*jwt.SigningMethodHMAC)
+			if !ok {
+				return nil, fmt.Errorf("unexpected token signing method")
+			}
+
+			return []byte(secretKey), nil
+		},
+	)
+
+	if err != nil {
```
luizbafilho, line 55: Instead of checking error strings, we could use their error types and have a more robust error checking: https://pkg.go.dev/github.com/golang-jwt/jwt#example-Parse-ErrorChecking

luizbafilho (changes requested):
Looks good, only a few comments and the big one is moving from `github.com/dgrijalva/jwt-go` to `https://github.com/golang-jwt/jwt` the former is no longer maintained and its code was migrated to the latter.

pkg/services/auth/jwt.go
```diff
@@ -0,0 +1,78 @@
...
+		if ve, ok := err.(*jwt.ValidationError); ok {
```
luizbafilho, line 59: Looking at this again, if we have an error we'll have an Unauthorized token anyway, what do you think? should we always return that `ErrUnauthorizedToken` and not check what type of error is that since we only care that the valid or not?

So we would become just:

```
if err != nil {
   return nil, ErrUnauthorizedToken
}
```

pkg/services/auth/jwt_test.go
JamWils: Yes, I told @josecordaz to also verify error messages as well.  The example we were looking at when we chatted was you couldn't tell which tests were testing which error path.  By verifying the error messages you actually were reducing flakiness in the test.  I also prefer specificity with assertions.  It looks like we need to have a conversation as a team what the expectations are, pun intended.  Jordan I do agree with you about testing another library's error string.  To eliminate flakiness I would use a pattern such as:

```
func TestSomething(t *T) {
    ....
    err := MyDailyFunc("", "")
    
    expectedErr := json.Unmarshal(badData, &obj)
   t.Assert(err, Equals, fmt.Errorf("MyDailyFunc json unmarshal fail: %w, expectedErr))
}
```

In this instance, I actually propose we verify the string, not for my normal specificity reasons (error messages shouldn't change that often anyway), but for security reasons.  We want to lock-in the contract for the test and if someone changes the error message and puts sensitive information in it that we don't want returned then the test will fail.

We can revisit assertions outside of this PR.  I made my case but leave the final call up to you and Jose.

pkg/services/auth/jwt.go
JamWils: I had a chat with @josecordaz f2f and we talked about putting `Generate` and `Verify` in an interface so he can mock it out and call it from the `ApplicationServer` more easily.

jpellizzari: Why mock it out?

luizbafilho:
on the move to https://github.com/golang-jwt/jwt let's use the latest release v4, here is the migration guide https://github.com/golang-jwt/jwt/blob/main/MIGRATION_GUIDE.md. Quite simple since it's backward compatibly.

pkg/services/auth/jwt.go
josecordaz: Good point.  I'm thinking in the user consuming the api, should the user get a specific error to acknowledge that the api  call was done wrong?. I know we don't want to return `ValidationErrorExpired` because that exposes how long ago the token expired (security reasons) .  I guess this comes down to which errors we consider are security issues. Do all errors that `ParseWithClaims` returns would be considered security issues? @jpellizzari @JamWils any thoughts you might have ?

jpellizzari: > Do all errors that ParseWithClaims returns would be considered security issues?

Probably? This error happens when someone submits a malformed token, which **is** an `Unauthorized` token. I would be okay with returning an `ErrUnauthorizedToken` here.

pkg/services/auth/jwt_test.go
jpellizzari: I like the solution @josecordaz has now with keeping the assertion in sync with the error string. The important thing is to verify that the the error you are receiving is the one you expect, so we agree there.

I'm trying to to have tests break when trivial error messages changes happen. If as assertion has a string literal, you will have to update the assertion every time the error message changes. Having both the assertion AND the error message refer to the same string is :+1:

jpellizzari (approved):
LGTM, but lets get one more green checkmark before merging.

pkg/services/auth/jwt.go
JamWils: That way he doesn't have to mess with the internals of this function.  In my mind, Generate unit tests should focus on the internals.  By mocking it out from `server_test.go` he is able to just focus on testing `application.Server.Authenticate(...)`.  From there he is able to test both paths easily:

1. `Generate` returns a valid token
2. `Generate` returns an error

Without the mock, how does he test the sad path?  He can try to force `Generate` into an error state, but that introduces a cascading effect of failures within the test suites that can get annoying very quickly.  If `Generate` evolves because we sign the JWT with a certificate or encrypt information, then the tests in `server_test.go` will most likely fail and need to be updated as well because they are dependent on the internal implementation of `Genereate`.  

It's a trade off, disconnecting the two means as an engineeer updates `Generate` the engineer can focus on making sure that `jwt_test.go` has full coverage and the tests will clearly show which units have broken.  `server_test.go` is still good because ultimately it only cares with a valid token or error is returned.  End-to-end tests can test the full flow.  It's a trade off but it simplifies the code and makes it easier to test in my opinion.

pkg/services/auth/jwt.go
```diff
@@ -0,0 +1,87 @@
...
+type JWTClient interface {
+	GenerateJWT(secretKey string, expirationTime time.Duration, providerName gitproviders.GitProviderName, providerToken string) (string, error)
+	VerifyJWT(secretKey string, accessToken string) (*Claims, error)
+}
+
+func NewJwtClient() JWTClient {
+	return &internalJwtClient{}
+}
+
+type internalJwtClient struct {
```
bigkevmcd, line 41: I think this should be `internalJWTClient` ?

https://github.com/golang/go/wiki/CodeReviewComments#initialisms

pkg/services/auth/jwt.go
```diff
@@ -0,0 +1,87 @@
...
+
+// Claims is a custom JWT claims that contains some token information
+type Claims struct {
+	jwt.StandardClaims
+	Provider      gitproviders.GitProviderName `json:"provider"`
+	ProviderToken string                       `json:"provider_token"`
+}
+
+//counterfeiter:generate . JWTClient
+type JWTClient interface {
```
bigkevmcd, line 32: This should be documented (it's exported).

pkg/services/auth/jwt.go
```diff
@@ -0,0 +1,87 @@
...
+	GenerateJWT(secretKey string, expirationTime time.Duration, providerName gitproviders.GitProviderName, providerToken string) (string, error)
+	VerifyJWT(secretKey string, accessToken string) (*Claims, error)
+}
+
+func NewJwtClient() JWTClient {
```
bigkevmcd, line 37: Missing a doc comment.

pkg/services/auth/jwt.go
```diff
@@ -0,0 +1,87 @@
...
+	return &internalJwtClient{}
```
bigkevmcd, line 38: I think instead of using a package global, this should accept the random string.

And change the instantiation to provide the secure string to use.

```go
cfg := ApplicationConfig{App: app.New(nil, nil, nil, k, nil), JwtClient: auth.NewJwtClient(<insert random string here>)}
```

pkg/server/server.go
```diff
@@ -309,3 +324,25 @@ func mapConditions(conditions []metav1.Condition) []*pb.Condition {
 
 	return out
 }
+
+var BadErrProvider = errors.New("wrong provider name")
+
+// Authenticate generates and returns a jwt token using git provider name and git provider token
+func (s *applicationServer) Authenticate(_ context.Context, msg *pb.AuthenticateRequest) (*pb.AuthenticateResponse, error) {
+
+	if string(gitproviders.GitProviderGitHub) != msg.ProviderName &&
```
bigkevmcd, line 333: I'm not sure there's a lot of value in the [`GitProviderName`](https://github.com/weaveworks/weave-gitops/blob/cf1654f34912bae653be91df69d1e3d855624a90/pkg/gitproviders/factory.go#L15) type here as you can see it complicates comparisons.

I'd probably switch back to a `string` which is what the Go stdlib does https://pkg.go.dev/net/http#pkg-constants

pkg/server/server.go
```diff
@@ -309,3 +324,25 @@ func mapConditions(conditions []metav1.Condition) []*pb.Condition {
...
+		string(gitproviders.GitProviderGitLab) != msg.ProviderName {
+		return nil, status.Errorf(codes.InvalidArgument, "%s expected github or gitlab, got %s", BadErrProvider.Error(), msg.ProviderName)
```
bigkevmcd, line 335: You don't need to call `.Error()` on the error here, because you're using `%s` https://pkg.go.dev/fmt

> 3. If the %v verb is used with the # flag (%#v) and the operand implements the GoStringer interface, that will be invoked.

So, because `BadErrProvider` implements `.Error()` it will be converted to a string automatically.

pkg/server/server_test.go
```diff
@@ -79,6 +87,51 @@ var _ = Describe("ApplicationsServer", func() {
...
+	It("Authorize fails on empty provider token", func() {
+		ctx := context.Background()
+		provider := "github"
+
+		_, err := appsClient.Authenticate(ctx, &pb.AuthenticateRequest{
+			ProviderName: provider,
+			AccessToken:  "",
+		})
+
+		Expect(err.Error()).To(ContainSubstring(ErrEmptyAccessToken.Error()))
```
bigkevmcd, line 131: Bonus points: I think I'd write a custom matcher for these https://onsi.github.io/gomega/#adding-your-own-matchers that would look like `Expect(err).To(MatchGRPCError(codes.InvalidArgument, ErrEmptyAccessToken))`

pkg/server/server_test.go
```diff
@@ -171,6 +224,46 @@ var _ = Describe("ApplicationsServer", func() {
...
+				Expect(err).NotTo(HaveOccurred())
+
+				ts := httptest.NewServer(httpHandler)
+				defer ts.Close()
+
+				// A valid URL for our server
+				path := "/v1/authenticate/github"
+				url := ts.URL + path
+
+				res, err := http.Post(url, "application/json", strings.NewReader("{\"accessToken\":\"sometoken\"}"))
```
bigkevmcd, line 250: You could simplify this to...

```go
strings.NewReader(`{"accessToken":"sometoken"}`)
```

Which would remove the string quoting.

pkg/server/server_test.go
```diff
@@ -171,6 +224,46 @@ var _ = Describe("ApplicationsServer", func() {
...
+				Expect(err).NotTo(HaveOccurred())
+				Expect(res.StatusCode).To(Equal(http.StatusInternalServerError))
+
+				response := make(map[string]interface{})
```
bigkevmcd, line 254: You could probably use [`MatchJSON`](https://onsi.github.io/gomega/#matchjsonjson-interface) here

bigkevmcd (commented):
Some comments.

luizbafilho (approved):

PR #709, opened 2021-09-02
Revert "Helm chart (#664)"

Callisto13:
This PR reverts the work done to add a wego core Helm Chart to this repo.

As pitch-black's work is throwing up some complexities, we realise it is probably better for now to keep everything in our repo while we figure out the kinks, rather than hit y'all with a bunch of tiny random PRs. We can decide later where to keep everything once things are tied up in a pretty bow.


jpellizzari (approved):
:crying_cat_face:

Issue #649, opened 2021-08-24
Add docker build and CI job for `wego-app`


josecordaz:
**Short description**
As a developer I want to have a `dockerfile` so that I can run `cmd/wego-server/main.go` logic inside a container running on the cluster. We also want to push this image to CI on the `release` job.

**Acceptance criteria**
- [x] Dockerfile should use a multi-stage build to build both the UI and binary.
- [x] The runtime container should be very minimal and contain only the api runtime binary
- [ ] The image should push to the container registry during the `release` job, and only during that job
- [ ] `$ docker run ghcr.io/weaveworks/weave-gitops wego ui run` should download the image and start an API server
- [ ] Should be built via a `Makefile` target

PR #710, opened 2021-09-02
Building docker image on release pipeline

luizbafilho:
Closes: #649 

**What changed?**
- Builds a docker image and publishes it in Github Container Registry

**Why?**

**How did you test it?**

**Release notes**

**Documentation Changes**


Dockerfile
```diff
@@ -0,0 +1,22 @@
+# UI build
+FROM node:14-buster AS ui
```
jpellizzari, line 2: What is `buster`?

Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+
+# Go build
+FROM golang:1.16 AS go-build
+COPY . /app
+COPY --from=ui /home/app/cmd/wego/ui/run/dist/ /app/cmd/wego/ui/run/dist/
+WORKDIR /app
+RUN make dependencies && make bin
+
+# Distroless
+FROM gcr.io/distroless/base
```
jpellizzari, line 20: Not familiar with `distroless`. Can you help me understand why you chose this image?

Makefile
```diff
@@ -36,11 +36,11 @@ unit-tests: dependencies cmd/wego/ui/run/dist/index.html
 debug:
 	go build -ldflags $(LDFLAGS) -o bin/$(BINARY_NAME) -gcflags='all=-N -l' cmd/wego/*.go
 
-bin: ui
+bin:
```
jpellizzari, line 39: After a `make clean`, I think `make bin` will now fail to compile without the UI job since it wants UI files for embedding. Does that still happen on this branch?

package.json
```diff
@@ -20,9 +20,9 @@
   "scripts": {
     "preinstall": "npm install --package-lock-only --ignore-scripts && npx npm-force-resolutions",
     "build": "parcel build --target default",
-    "build:lib": "parcel build --target lib",
+    "build:lib": "npx parcel build --target lib",
     "typedefs": "npx -p typescript tsc ui/**/*.ts ui/**/*.tsx --skipLibCheck --esModuleInterop --jsx react-jsx --declaration --allowJs --emitDeclarationOnly --outDir dist",
-    "start": "parcel serve --port 4567 ui/index.html",
+    "start": "npx parcel serve --port 4567 ui/index.html",
```
jpellizzari, line 25: Why was the change to `npx` necessary?

jpellizzari (commented):
Looks pretty close, just some questions before giving a :+1:

Makefile
luizbafilho: It will, but that was the only way I could find without having creating something like `docker-bin` or so, since the go stage doesn't have node in it

jpellizzari: I think its okay as long as this is documented in the README and we let everyone know

Dockerfile
luizbafilho: simpler to maintain: https://blog.baeke.info/2021/03/28/distroless-or-scratch-for-go-apps/

Dockerfile
luizbafilho: the `node:14` tag has a pretty old version of `glibc` causing the node build to fail.

package.json
luizbafilho: not really, removing it.

jpellizzari (approved):
Looks good :+1: . @josecordaz PTAL

jpellizzari (commented):
Sorry I approved a bit too quickly. Can we get a `make` target to build the container image?

Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+COPY --from=go-build /app/bin/wego /wego
+ENTRYPOINT ["/wego"]
```
josecordaz, line 22: I'm seeing that the docker image has ENTRYPOINT ["/wego"] not sure how we are going to start the api 🤔 Is there a way to tell the deployment to replace that entrypoint ? 🙂

luizbafilho: in a k8s deployment you set `args` to `["ui", "run"]`, or running locally 

`docker run -it --rm ghcr.io/weaveworks/wego-app:latest ui run`

Issue #606, opened 2021-08-12
Show a reconciliation graph in the UI for a given application


jpellizzari:
Balsamiq Wireframes: https://balsamiq.cloud/s5qd59g/p340wpr/rA93F

Show a list of the objects that are created as a result of a `Kustomization` or `HelmRelease`. The Flux UI has a version of this implemented:

UI Code: https://github.com/fluxcd/webui/blob/main/ui/components/ReconciliationGraph.tsx

Backend Code: https://github.com/fluxcd/webui/blob/main/pkg/clustersserver/clustersserver.go#L301

Recursive lookup: https://github.com/fluxcd/webui/blob/main/ui/lib/hooks/kustomizations.ts#L115

PR #711, opened 2021-09-02
Add API methods for rendering parent-child object relationships for an Application

jpellizzari:
Closes: #606  

Adds the necessary backend methods to render a list of k8s objects that were created from an `Application`, and a first draft of the Graph visualization.

Screenshot from 2021-09-03 11-56-07

@foot @AlinaGoaga There is a breaking change to the `useApplications` hook where the user of that hook will have to run `listApplications` instead of it being available be default. Here is the new API:

```js
  const [applications, setApplications] = React.useState<Application[]>([]);
  const { listApplications, loading } = useApplications();

  React.useEffect(() => {
    listApplications().then((res) => setApplications(res as Application[]));
  }, []);
```


pkg/server/server.go
```diff
@@ -251,6 +264,109 @@ func (s *applicationServer) ListCommits(ctx context.Context, msg *pb.ListCommits
...
+Items:
+	for _, obj := range list.Items {
+
+		refs := obj.GetOwnerReferences()
+
+		for _, ref := range refs {
+			if ref.UID != types.UID(msg.ParentUid) {
+				// This is not the child we are looking for.
+				// Skip the rest of the operations in Items loops.
+				// The is effectively an early return.
```
josecordaz, line 345: typo?

josecordaz (approved):
LGTM

foot:
This is quite vertical in our test cluster. We're using the App to reconcile all our CAPI clusters though which might not be the right model 
image

foot (approved):
All looks good 👏 ! ✨

ui/pages/ApplicationDetail.tsx
```diff
@@ -62,20 +62,27 @@ function ApplicationDetail({ className, name }: Props) {
...
+      {/* <DataTable
         sortFields={["name"]}
         fields={[
           { label: "Name", value: "name" },
           { label: "Kind", value: (v) => v.groupVersionKind.kind },
           { label: "Status", value: "status" },
         ]}
         rows={reconciledObjects}
-      />
+      /> */}
```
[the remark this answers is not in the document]
jpellizzari: It probably needs to switch between tabular and graph. This will make a triumphant return

pkg/server/server.go
```diff
@@ -282,6 +299,109 @@ func (s *applicationServer) ListCommits(ctx context.Context, msg *pb.ListCommits
...
+			Name:      obj.GetName(),
+			Namespace: obj.GetNamespace(),
+			Status:    res.Status.String(),
+			Uid:       string(obj.GetUID()),
+		})
+	}
+	return &pb.GetReconciledObjectsRes{Objects: objects}, nil
+}
+
+func (s *applicationServer) GetChildObjects(ctx context.Context, msg *pb.GetChildObjectsReq) (*pb.GetChildObjectsRes, error) {
```
jrryjcksn, line 356: This works as it is, but it seems like it would be a nicer abstraction if we maintained the types of child objects separately so that a user of this function could just pass in the parent object without having to specify the child type(s); that would also support an eventual recursive descendant search

jrryjcksn (approved):
LGTM w/ a couple of minor comments.

api/applications/applications.proto
```diff
@@ -53,15 +75,18 @@ message Condition {
...
 message Application {
-    string   name                            = 1;  // The name of the application
-    string   path                            = 2;  // The file path where the k8s yaml files for this application are stored.
-    string   url                             = 3;  // The git repository URL for this application
-    repeated Condition source_conditions     = 4;  // A list of conditions for the Source related to this Application
-    repeated Condition deployment_conditions = 5;  // A list of conditions for the Kustomization or HelmRelease for this application
+    string                    name                            = 1;  // The name of the application
+    string                    path                            = 2;  // The file path where the k8s yaml files for this application are stored.
+    string                    url                             = 3;  // The git repository URL for this application
+    repeated                  Condition source_conditions     = 4;  // A list of conditions for the Source related to this Application
```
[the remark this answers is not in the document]
jpellizzari: :thinking: We are...

pkg/server/server.go
jpellizzari: @jrryjcksn Kubernetes won't support that. An object doesn't know the types of children it has unfortunately.

jpellizzari: Oh I see what you are saying. We store this in the UI right now. It can be moved to the backend, but we don't know ALL of the parent-child relationships, so it made sense to push this onto the user.

jpellizzari: In the case of a CRD, maybe its impossible for us to know what the relationships are, which makes me lean towards the current API design of asking the user to specify the objects they care about.

Issue #677, opened 2021-08-26
Unsuccessful installation can leave components on cluster which cannot be uninstalled using wego gitops unistall


sympatheticmoose:
**Describe the bug**
Whilst testing Weave GitOps on Red Hat CodeReady Containers (OpenShift) I hit an issue as a result of scc settings. Most stuff got installed except the source controller fails to start, so the verification failed in the final steps of `wego gitops install`. As verification fails, Weave GitOps seems to have no idea of the components that were deployed to the cluster. Therefore `wego gitops uninstall` fails as it claims theres nothing there

**Environment**
 - wego: [e.g. v0.2.2]
 - kubernetes:  [v1.21.1+38b3ecc]
    - OpenShift 4.8.4

**To Reproduce**
Steps to reproduce the behavior: (outlining my steps, there may be a simpler way to force this error)
1. Download [CRC](https://developers.redhat.com/products/codeready-containers/overview) & copy pull secret for later 
2. Download oc ([mac](https://downloads-openshift-console.apps-crc.testing/amd64/mac/oc.zip) / [linux](https://downloads-openshift-console.apps-crc.testing/amd64/linux/oc.tar) )
3. `crc setup`, `crc start`, grab a cup of tea,`eval $(crc oc-env)`
4. Login as kubeadmin using the information provided in the terminal `oc login -u kubeadmin -p <password> https://api.crc.testing:6443`
5. Create project `oc new-project wego-system` (probably optional, required if you want to get this to install properly)
6. Install weave gitops - `wego gitops install` and wait for it to fail
6. Check deployment to verify there is stuff there - `oc get all`
7. Try and uninstall - `wego gitops uninstall`
8. See error

**Expected behavior**
If we can **safely** attempt cleanup, we should provide this as an option either following the validation failure (not automatic I would assume as folks may want to grab info to understand the failure) and/or wego gitops uninstall should be able to remove this partly installed deployment. My concern is if there are any risks to someone who has installed the GitOps Toolkit through some other means, maybe this is a non-concern if we fail to install in that case.

**Screenshots**
<img width="670" alt="Screenshot 2021-08-24 at 14 03 08" src="https://user-images.githubusercontent.com/19860021/130988864-f3ffbb0e-96c7-44bd-946e-9f08d977622b.png">

**Additional context**
The reason this fails is the additional steps that need to be configured for OpenShift as described in the OperatorHub description for [flux](https://operatorhub.io/operator/flux)

[...]

PR #716, opened 2021-09-03
Update "wego gitops uninstall" to run to completion even if original install failed

jrryjcksn:
Closes: #677 

**What changed?**
`wego gitops uninstall` used to check for a complete wego installation before continuing. This prevented removing a partial installation.

**Why?**
So a failed installation can be undone

**How did you test it?**
Unit and integration tests

**Release notes**
No

**Documentation Changes**
No

### DIFF
--- a/pkg/kube/kubehttp.go
+++ b/pkg/kube/kubehttp.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"strings"
 
 	v1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -133,12 +134,18 @@ func (k *KubeHTTP) GetClusterStatus(ctx context.Context) ClusterStatus {
 		Namespace: "kube-system",
 	}
 
+	notFoundErr := `deployments.apps "coredns" not found`
+
 	if err := k.Client.Get(ctx, coreDnsName, &dep); err != nil {
-		// Couldn't find the coredns deployment.
-		// We don't know what state the cluster is in.
+		// Some clusters don't have 'coredns'; if we get a "not found" error, we know we
+		// can talk to the cluster
+		if strings.Contains(err.Error(), notFoundErr) {
+			return Unmodified
+		}
+
 		return Unknown
 	} else {
-		// Request for the coredns namespace was successfull.
+		// Request for the coredns namespace was successful.
 		return Unmodified
 	}
 }
--- a/pkg/kube/kubehttp.go
+++ b/pkg/kube/kubehttp.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"strings"
 
 	v1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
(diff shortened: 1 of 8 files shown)


pkg/kube/kubehttp.go
```diff
@@ -133,12 +134,18 @@ func (k *KubeHTTP) GetClusterStatus(ctx context.Context) ClusterStatus {
...
 	}
 
+	notFoundErr := `deployments.apps "coredns" not found`
+
 	if err := k.Client.Get(ctx, coreDnsName, &dep); err != nil {
-		// Couldn't find the coredns deployment.
-		// We don't know what state the cluster is in.
+		// Some clusters don't have 'coredns'; if we get a "not found" error, we know we
+		// can talk to the cluster
+		if strings.Contains(err.Error(), notFoundErr) {
```
jpellizzari, line 142: Might check out `apierrors.IsNotFound()`

pkg/services/gitops/uninstall_test.go
```diff
@@ -22,38 +73,91 @@ var _ = Describe("Uninstall", func() {
...
+	It("Generates an error if flux uninstall fails with partial or no flux installed", func() {
+		kubeClient.GetClusterStatusStub = func(ctx context.Context) kube.ClusterStatus {
+			return kube.Unmodified
+		}
+
+		checkFluxUninstallFailure()
+	})
+
+	It("Generates an error if CRD uninstall fails with wego installed", func() {
+		checkAppCRDUninstallFailure()
```
jpellizzari, line 144: I'm having a tough time reasoning about these tests. I would guess this is trying to stay DRY, but a test case with no assertions and no arguments into a helper function makes me scratch my head a little bit.

pkg/services/gitops/uninstall_test.go
```diff
@@ -14,6 +16,55 @@ import (
...
+	Expect(kubeClient.GetClusterStatusCallCount()).To(Equal(1))
+	Expect(fluxClient.UninstallCallCount()).To(Equal(1))
+	namespace, dryRun := fluxClient.UninstallArgsForCall(0)
+	Expect(namespace).To(Equal("wego-system"))
+	Expect(dryRun).To(Equal(false))
+}
+
+func checkAppCRDUninstallFailure() {
+	crdErrMsg := "CRD uninstall failed"
+	globalErrMsg := "errors occurred during uninstall; the original state of the cluster may not be completely restored"
```
jpellizzari, line 45: Is this a string that needs to stay in sync with the other error message for this tests to pass? A custom error type might help here.

jpellizzari (approved):
LGTM, only nits.

pkg/kube/kubehttp.go
jrryjcksn: Thanks! Was not aware of that.

pkg/services/gitops/uninstall_test.go
jrryjcksn: Understood... the only difference between the test cases, though, is the cluster status and this case has it set by default. I should probably set it explicitly here to make that clear.

pkg/services/gitops/uninstall_test.go
jrryjcksn: Yep. Good call.

pkg/kube/kubehttp.go
jrryjcksn: Doh! In the same file.

PR #720, opened 2021-09-03
remove hardcoded github mentions

J-Thompson12:
Closes: #666

**What changed?**
Removes remaining hardcoded references to github as the providername.

**Why?**
This makes it easier to add new providers

**How did you test it?**
make sure acceptance tests still work

**Release notes**

**Documentation Changes**

### DIFF
--- a/pkg/gitproviders/provider.go
+++ b/pkg/gitproviders/provider.go
@@ -42,12 +42,11 @@ type GitProvider interface {
 	GetCommitsFromUserRepo(userRepRef gitprovider.UserRepositoryRef, targetBranch string, pageSize int, pageToken int) ([]gitprovider.Commit, error)
 	GetCommitsFromOrgRepo(orgRepRef gitprovider.OrgRepositoryRef, targetBranch string, pageSize int, pageToken int) ([]gitprovider.Commit, error)
 	GetAccountType(owner string) (ProviderAccountType, error)
-	GetProviderName() GitProviderName
+	GetProviderDomain() string
 }
 
 // making sure it implements the interface
 var _ GitProvider = defaultGitProvider{}
-var providerName GitProviderName
 
 type defaultGitProvider struct {
 	provider gitprovider.Client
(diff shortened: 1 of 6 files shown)


pkg/gitproviders/provider.go
```diff
@@ -498,6 +500,10 @@ func (p defaultGitProvider) GetCommitsFromOrgRepo(orgRepRef gitprovider.OrgRepos
 	return commits, nil
 }
 
+func (p defaultGitProvider) GetProviderName() GitProviderName {
+	return providerName
```
jpellizzari, line 504: :warning: Global State alarm bells :warning:

What happens when two concurrent HTTP requests try to set this variable at the same time?

jpellizzari:
Does this PR mean that Gitlab will be working, as defined in #666

J-Thompson12:
@jpellizzari gitlab will not work until we have the deploy key stuff in place. Now that we dont look for a token there is no way get it working.

jpellizzari:
@J-Thompson12 The reason I ask is because this PR says it "closes" the issue. Please change it to "Related to"

pkg/services/app/add.go
```diff
@@ -686,7 +685,7 @@ func (a *App) createPullRequestToRepo(info *AppResourceInfo, repo string, appHas
 		return nil
 	}
 
-	userRepoRef := gitproviders.NewUserRepositoryRef(github.DefaultDomain, owner, repoName)
+	userRepoRef := gitproviders.NewUserRepositoryRef(string(a.GitProvider.GetProviderDomain()), owner, repoName)
```
jpellizzari, line 688: Converting to a string here when it looks like this is already a string?

jpellizzari (approved):

pkg/services/app/add.go
J-Thompson12: yeah that is something left over from the first draft. Sorry I missed that.

Issue #684, opened 2021-08-30
Inconsistent test: Verify that wego can deploy a helm app from a git repo with app-config-url set to NONE 


jpellizzari:
The acceptance test labelled as `Verify that wego can deploy a helm app from a git repo with app-config-url set to NONE ` is not deterministic.

https://github.com/weaveworks/weave-gitops/pull/660/checks?check_run_id=3463127082

This PR run only touched UI stuff ^^

Running the tests a second time usually succeeds.

PR #722, opened 2021-09-03
Using go-git-providers to create test repos instead of hub

josecordaz:
Closes: #684 

**What changed?**
- Using go-git-providers to create test repositories.
- It now waits for the right time after getting rate limit errors.

**Why?**
- `hub` returns confusing error messages resulting in not knowing if the repository was created properly.
- go-git-providers return the amount of time it needs to wait before getting more quota after a rate limit issue.

**How did you test it?**
Re-running integration tests manually many times.

**Release notes**

**Documentation Changes**


test/acceptance/test/utils.go
```diff
@@ -624,3 +605,85 @@ func takeScreenshot() string {
...
+		gitprovider.WithOAuth2Token(os.Getenv("GITHUB_TOKEN")),
+		gitprovider.WithDestructiveAPICalls(true),
+	)
+	if err != nil {
+		return err
+	}
+
+	ctx := context.Background()
+	fmt.Printf("creating repo %s ...\n", repoName)
+	if err := utils.WaitUntil(os.Stdout, time.Second, time.Second*30, func() error {
```
JamWils, line 661: ⚠️ Can you move `time.Second*30` into a const.  Is this the default timeout that is typically used?  If so, then it sounds like a good name for the const.

test/acceptance/test/utils.go
```diff
@@ -624,3 +605,85 @@ func takeScreenshot() string {
...
+
+func createRepository(repoName string, private bool) error {
+
+	visibility := gitprovider.RepositoryVisibilityPrivate
+	if !private {
+		visibility = gitprovider.RepositoryVisibilityPublic
+	}
+
+	description := "Weave Gitops test repo"
+	defaultBranch := "main"
```
JamWils, line 632: ⚠️ This feels like it should be a const too

test/acceptance/test/utils.go
```diff
@@ -624,3 +605,85 @@ func takeScreenshot() string {
...
+		return duration, nil
+	}
+
+	return 0, fmt.Errorf("could not found a rate reset on string: %s", errOutput)
+}
+
+func createRepository(repoName string, private bool) error {
+
+	visibility := gitprovider.RepositoryVisibilityPrivate
+	if !private {
```
JamWils, line 627: 💡 positives are always easier to read.  Why not reverse this logic and have if the if statement just be `if private {}`?

JamWils (approved):
Just some minor comments on cleaning up the code.  Other than that LGTM

test/acceptance/test/utils.go
```diff
@@ -624,3 +605,85 @@ func takeScreenshot() string {
...
+			return fmt.Errorf("retry after waiting for rate quota")
+		}
+		return err
+	}); err != nil {
+		return fmt.Errorf("error creating repo %s", err)
+	}
+	fmt.Printf("repo %s created ...\n", repoName)
+
+	fmt.Printf("validating access to the repo %s ...\n", repoName)
+	err = utils.WaitUntil(os.Stdout, time.Second, time.Second*30, func() error {
```
JamWils, line 679: If you turn time.Second*30 into a const per the above suggestion then you can update it here too.

test/acceptance/test/utils.go
rokshana-b: @josecordaz There's already a `const` you can reuse from utils file: `const APP_REMOVAL_TIMEOUT time.Duration = 30 * time.Second` You can rename this `const` to `THIRTY_SECOND_TIMEOUT`

PR #732, opened 2021-09-07
Add dockerized library integration tests

### LINKED ISSUE #691: Create automated test for library integration points
There are some library integration points that other teams are relying on to integrate WeGO Core into their projects. We currently do not have an automated way to test for breakages of that contract.

It should:
1. Clone some outside repo that integrates WeGO Core like a real project would. Something like this: https://github.com/jpellizzari/wego-wrapper-demo
2. Import the WeGO `ApplicationServer` and mux it with another route/service
3. Ensure that the external app compiles
4. Some sort of runtime test

jpellizzari:
Closes #691

Adds a test that will report when the "public" API for `wego` go libraries break consuming implementations.

To test, rename something like the `server.DefaultConfig()` function and see the `make lib-test` target fail with an exit code of 2.

Uses this repo as a reference implementation: https://github.com/weaveworks-gitops-test/wego-library-test

Output looks like this when something breaks the API contract:
Screenshot from 2021-09-07 17-37-23

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -63,6 +63,7 @@ clean: ## Clean up images and binaries
 	rm -rf coverage
 	rm -rf node_modules
 	rm -f .deps
+	rm -rf dist
 
 fmt: ## Run go fmt against code
 	go fmt ./...
@@ -127,7 +128,16 @@ cmd/wego/ui/run/dist/index.html: cmd/wego/ui/run/dist
 cmd/wego/ui/run/dist/main.js:
 	npm run build
 
-dist/index.js:
+# Runs a test to raise errors if the integration between WeGO Core and EE is 
+# in danger of breaking due to package API changes.
+# See the test/library dockerfile and test.sh script for more info.
+lib-test: dependencies
+	docker build -t wego-library-test -f test/library/libtest.dockerfile $(CURRENT_DIR)/test/library
+	docker run -e GITHUB_TOKEN=$$GITHUB_TOKEN -i --rm \
+		-v $(CURRENT_DIR):/go/src/github.com/weaveworks/weave-gitops \
+		 wego-library-test
+
+dist/index.js: ui/index.ts
 	npm run build:lib && cp package.json dist
 
 dist/index.d.ts:
--- a/test/library/test.sh
+++ b/test/library/test.sh
@@ -1,0 +1,10 @@
+#!/bin/bash
+set -e
+set -x
+
+git clone https://github.com/weaveworks-gitops-test/wego-library-test.git test/library/wego-library-test
+cd test/library/wego-library-test
+npm ci
+npm run build
+go mod tidy
+go build main.go
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -259,3 +259,17 @@ jobs:
         name: acceptance-test-artifacts
         path: ${{ env.ARTIFACTS_BASE_DIR }}
         retention-days: 1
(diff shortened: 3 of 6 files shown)


josecordaz (approved):
LGTM

foot (approved):
\o/, this is great, love the implementation. Simple.

test/library/test.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+set -e
+set -x
+
+git clone https://github.com/weaveworks-gitops-test/wego-library-test.git test/library/wego-library-test
+cd test/library/wego-library-test
+npm ci
```
palemtnrider, line 7: When running this locally, it fails with 
> npm ci
npm ERR! code E401
npm ERR! 401 Unauthorized - GET https://npm.pkg.github.com/download/@weaveworks/weave-gitops/0.0.4/fc8d34755955e02e2004edab9e4dd88f22aaf46b933498f048f0abb256f87644

Is it intentional that it pulls package version 0.0.4?  This surprised me (but I don't completely grasp how the test verifies the change in APIs).  Please add some comments in the Makefile describing what this is checking.  It might be worthwhile to add a bit of info to the readme as well.

palemtnrider (commented):
Should I be able to run this locally?  It fails on my mac.

Makefile
```diff
@@ -128,6 +128,12 @@ cmd/wego/ui/run/dist/index.html: cmd/wego/ui/run/dist
 cmd/wego/ui/run/dist/main.js:
 	npm run build
 
+lib-test: dependencies
+	docker build -t wego-library-test -f test/library/libtest.dockerfile $(CURRENT_DIR)/test/library
+	docker run -e GITHUB_TOKEN=$$GITHUB_TOKEN -i \
```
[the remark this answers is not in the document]
jpellizzari: :+1: fixed

test/library/test.sh
jpellizzari: > Is it intentional that it pulls package version 0.0.4? 
I hadn't realized that, but this test also doesn't seek to prove the UI stuff works (we have some  test coverage on that already). That begs the question as to why we are installing the JS module in the test repo.

This comment gave me an idea of how to include the JS stuff in this integration test. I'll make a time-boxed effort to get that working.

jpellizzari: Gave up on adding `index.ts` coverage here, it wasn't working, and we have unit test coverage there.

Makefile
```diff
@@ -127,7 +128,16 @@ cmd/wego/ui/run/dist/index.html: cmd/wego/ui/run/dist
 cmd/wego/ui/run/dist/main.js:
 	npm run build
 
-dist/index.js:
+# Runs a test to raise errors if the integration between WeGO Core and EE is 
+# in danger of breaking due to package API changes.
+# See the test/library dockerfile and test.sh script for more info.
```
palemtnrider, line 133:
```suggestion
# See the test/library dockerfile and test.sh script for more info.
# The GITHUB_TOKEN requires read:packages permissions as the UI publishes a UI packages on GH.
```

test/library/test.sh
palemtnrider: The problem with the GH token is it requires read:packages permission

palemtnrider (approved):
One additional note on the permissions required for the GITHUB_TOKEN in order to run the tests locally.  Other than that, LGTM

jpellizzari:
Merging before acceptance tests pass, since no app code was changed.

PR #734, opened 2021-09-07
Update source scan to include go module dependencies

josecordaz:
Closes: 
N/A

**What changed?**
Adding a step to the scan GH Action to install go 

**Why?**
FOSSA needs it to generate the go dependencies for the project.

**How did you test it?**
Manually executed the action and checked the FOSSA dashboard

**Release notes**
N/A

**Documentation Changes**
N/A

### DIFF
--- a/main.go
+++ b/main.go
@@ -2,9 +2,12 @@ package main
 
 import (
 	"fmt"
+
+	log "github.com/sirupsen/logrus"
 )
 
 func main() {
+	log.Debug("just a test")
 }
 
 func test() string {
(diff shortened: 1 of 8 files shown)


main.go
```diff
@@ -2,9 +2,12 @@ package main
 
 import (
 	"fmt"
+
+	log "github.com/sirupsen/logrus"
```
jpellizzari, line 6: :thinking: Do we need a third-party library for this?

jpellizzari (approved):

main.go
palemtnrider: We don't need these changes at all.  I'll remove the file from the PR.  Good catch.

Issue #648, opened 2021-08-24
Pass commit url as part of get commits


J-Thompson12:
get commits needs to pass a link to the commit. The best possible solution is if this can be returned from go-git-providers with everything else. If this isnt possible then the link will need to be generated in some way.

If go-git-providers does not provide this, we will need to synthesize this by detecting the provider and the URL for supported git providers.

PR #737, opened 2021-09-08
Add commit URL to ListCommits API  and CLI response 

J-Thompson12:
Closes: #648

**What changed?**
Added commit url to commit info.  go-git-providers updated `github.com/go-logr/logr` with a breaking change but `sigs.k8s.io/controller-runtime` has not updated to the same version. This is why I have replaced the versions for go-logr.

**Why?**

**How did you test it?**
manually

**Release notes**

**Documentation Changes**


pkg/utils/common.go
```diff
@@ -173,6 +173,12 @@ func ConvertCommitHashToShort(hash string) string {
 	return hash[:7]
 }
 
+func BuildCommitURL(url string, commitHash string) string {
+	url = strings.Replace(url, "ssh://git@", "https://", 1)
+	url = strings.Replace(url, ".git", "", 1)
+	return fmt.Sprintf("%s/commit/%s", url, commitHash)
```
jpellizzari, line 179: This works for both Github and Gitlab?

pkg/utils/common_test.go
```diff
@@ -119,6 +119,15 @@ error occurred some error, retrying in 1s
 		Expect(stdout).To(Equal("my output"))
 
 	})
+
+	It("Verify BuildCommitURL works on supported providers", func() {
+		githubURL := "ssh://git@github.com/J-Thompson12/nginx.git"
```
jpellizzari, line 124: Can we add a case for Gitlab?

pkg/utils/common.go
```diff
@@ -173,6 +173,12 @@ func ConvertCommitHashToShort(hash string) string {
 	return hash[:7]
 }
 
+func BuildCommitURL(url string, commitHash string) string {
```
jpellizzari, line 176: Does it make sense to use a `NormalizedRepoURL` here?

pkg/utils/common.go
J-Thompson12: No it doesnt.  There is work needed to normalize gitlab url when we support it. So that work could effect whats done here

pkg/utils/common.go
J-Thompson12: I thought about that and I think when we have gitlab it will make sense. I was kinda hoping my talks with the go-git-provider team to fix the logr issues would be done before we need to deal with gitlab.

pkg/utils/common.go
J-Thompson12: worked it out with yiannis and now it supports gitlab

J-Thompson12:
I do have one product question. go-git-providers generates the url using the big commit hash. This makes the output a little larger on the cli. I could change it to the short hash if thats desired. It doesnt really effect the frontend though.

jpellizzari:
> It doesnt really effect the frontend though.

It does, pretty majorly:
Screenshot from 2021-09-09 16-06-58

Short hash is the way to go, especially if we have a link.

jpellizzari (commented):
Please title this PR to read like a commit message. Remember these PR titles  get converted into release notes.

pkg/server/server_test.go
```diff
@@ -607,6 +607,7 @@ func testCommit() gitprovider.CommitInfo {
 		Author:    "testauthor",
 		Message:   "some awesome commit",
 		CreatedAt: time.Now(),
+		URL:       "http://github.com/testrepo/commit/2349898723987989234",
```
jpellizzari, line 610: Are we checking for this in a test anywhere?

jpellizzari (approved):
LGTM, but some tests would be good.

"Add commit URL to ListCommits API response" would be a good PR title (and release note).

pkg/server/server_test.go
J-Thompson12: I am now

PR #743, opened 2021-09-08
Fix non-deterministic behavior between `npm i` and `npm ci`

jpellizzari:
There was an issue where `npm ci` seemed to not run `force-resolutions` as expected. This change makes it explicitly called in the `make all` target.

`force-resolutions` is used to avoid security vulnerabilities in our JS third-party packages.


josecordaz (approved):
LGTM

PR #749, opened 2021-09-09
Show git diff after running make all on release workflow instead

josecordaz:
Closes: 

**What changed?**
Added `git diff` to release workflow

**Why?**
To help on debugging. Initially GoReleaser was failing because we had a bad dirty state due to NPM modules being updated during the build.

**How did you test it?**

**Release notes**

**Documentation Changes**

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -26,7 +26,9 @@ jobs:
           echo "BRANCH=$(git rev-parse --abbrev-ref HEAD)" >> $GITHUB_ENV
           echo "FLUX_VERSION=$($(pwd)/tools/bin/stoml $(pwd)/tools/dependencies.toml flux.version)" >> $GITHUB_ENV
       - name: "Make All"
-        run: make all
+        run: |
+          make all
+          git --no-pager diff
       - name: Run GoReleaser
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yaml
```diff
@@ -26,7 +26,9 @@ jobs:
           echo "BRANCH=$(git rev-parse --abbrev-ref HEAD)" >> $GITHUB_ENV
           echo "FLUX_VERSION=$($(pwd)/tools/bin/stoml $(pwd)/tools/dependencies.toml flux.version)" >> $GITHUB_ENV
       - name: "Make All"
-        run: make all
+        run: |
+          make all
```
jpellizzari, line 30: How about splitting this into a separate step? That way its easier to tell which command fails.

jpellizzari (approved):
LGTM, but we should include the historical context of why this was changed. In this specific case @palemtnrider found the PR but it didn't contain any details.

Something like:
`GoReleaser was failing because we had a bad dirty state due to NPM modules being updated during the build`

Remember, we are writing messages to our future selves with these PR comments.

jpellizzari (approved):

Issue #616, opened 2021-08-17
WeGO fails to reconcile when folder containing app manifests is deleted


rokshana-b:
**Describe the bug**
WeGO fails to delete apps running inside containers after the folder containing the app manifest files is deleted from the git repository. Once an app is added with `--path` field specified, WeGO relies on the specified folder to exist in order to continue **"gitopsing"**. If the all the files in the folder are deleted, github automatically deletes the folder. This confuses WeGO since the directory specified during `wego app add` no longer exists. Please look at the **Steps To Reproduce** section for more clarity.

**Environment**
 - wego: v0.2.2

**To Reproduce**
1. Create a git repo and save all the app manifest files in a sub-directory (do not save files at the root)
1. Have a cluster up and running
2. Install wego components by running `wego gitops install`
3. Run `wego app add . --path=<path-to-sub-directory> --auto-merge=true`
4. See app get added and containers being created (Run `kubectl get pods -A`)
5. Delete all manifest files from the sub-directory and watch the directory get removed from the git repo
6. Run `kubectl get pods -A` and see the containers still running

**Screenshots**

![image.png](https://images.zenhubusercontent.com/606389351ead0c83798a9a73/0e516d00-ead5-44f3-83de-4e7db411c0c2)

**Additional context**
Is this expected behavior? The `wego app status` is giving the right message but the containers still persist. If the directory no longer exists, then the containers that got created should get destroyed.

jrryjcksn:
This appears to be the following flux issue: https://github.com/fluxcd/kustomize-controller/issues/187
which they do not appear to be willing to fix. I believe we have basically three options:

1. Document that users should always maintain a manifest in each app directory
2. Write a dummy manifest into the directory ourselves at add time (maybe a config map), but that could still fail if a user removed it
3. Have a controller that watches app directories and removes matching resources if a directory goes empty

or, I guess, some combination of the three

jpellizzari:
Can we use a `.keep` file?

jpellizzari:
Consensus: Option 1 is good enough for now (document that users shouldn't delete the last thing)

PR #761, opened 2021-09-14
Deleting last manifest causes sync failure

jrryjcksn:
Closes: #616 

**What changed?**
We now add a `.keep` file to the app repository for the `--app-config-url=NONE` and `--app-config-url=<url>` cases so that removing the last app manifest does not break flux synchronization. We don't do this for `--app-config-url=""` (the config in app repo case) because the `.wego` directory performs the same function.

**Why?**
To prevent confusing failures for users that believe they are undoing an app by removing all manifests.

**How did you test it?**
Unit tests and manual tests

**Release notes**
No

**Documentation Changes**
Yes, we will need to talk about the extra pull request being made in the app repository (or extra commit in the auto-merge case)

### DIFF
--- a/pkg/services/app/add.go
+++ b/pkg/services/app/add.go
@@ -176,7 +176,7 @@ func (a *App) Add(params AddParams) error {
 
 	switch strings.ToUpper(info.Spec.ConfigURL) {
 	case string(ConfigTypeNone):
-		return a.addAppWithNoConfigRepo(info, params.DryRun, secretRef, appHash)
+		return a.addAppWithNoConfigRepo(info, params, secretRef, appHash)
 	case string(ConfigTypeUserRepo):
 		return a.addAppWithConfigInAppRepo(info, params, secretRef, appHash)
 	default:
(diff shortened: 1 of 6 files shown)


pkg/services/app/add.go
```diff
@@ -577,6 +623,21 @@ func (a *App) cloneRepo(client git.Git, url string, branch string, dryRun bool)
 	}, nil
 }
 
+const keepFileContents = `This file exists so that removing all manifests from your application will not cause
```
jpellizzari, line 626: :sparkles:

pkg/services/app/add.go
```diff
@@ -633,6 +694,59 @@ func generateResourceName(url string) string {
...
+		return nil
+	}
+
+	userRepoRef := gitproviders.NewUserRepositoryRef(a.GitProvider.GetProviderDomain(), owner, repoName)
+
+	prLink, err := a.GitProvider.CreatePullRequestToUserRepo(userRepoRef, info.Spec.Branch, pullRequestBranchName, files, utils.GetCommitMessage(), fmt.Sprintf("wego add %s", info.Name), fmt.Sprintf("Added sentinel file for %s", info.Name))
+	if err != nil {
+		return fmt.Errorf("unable to create sentinel file pull request: %w", err)
+	}
+	a.Logger.Println("Sentinel Pull Request created: %s\n", prLink.Get().WebURL)
```
jpellizzari, line 746: What does `Sentinel` mean?

pkg/services/app/add.go
```diff
@@ -633,6 +694,59 @@ func generateResourceName(url string) string {
 	return hashNameIfTooLong(strings.ReplaceAll(utils.UrlToRepoName(url), "_", "-"))
 }
 
+func (a *App) createKeepFilePullRequest(info *AppResourceInfo, cloneNeeded bool) error {
```
jpellizzari, line 697: This was a bit alarming. Can't we just add the `.keep` file to the list of files we already commit?

pkg/services/app/add_test.go
```diff
@@ -950,6 +953,356 @@ var _ = Describe("Add", func() {
...
+				gitProviders.CreatePullRequestToOrgRepoStub = func(orgRepRef gitprovider.OrgRepositoryRef, targetBranch string, newBranch string, files []gitprovider.CommitFile, commitMessage string, prTitle string, prDescription string) (gitprovider.PullRequest, error) {
+					return dummyPullRequest{}, nil
+				}
+
+				gitProviders.CreatePullRequestToUserRepoStub = func(userRepRef gitprovider.UserRepositoryRef, targetBranch string, newBranch string, files []gitprovider.CommitFile, commitMessage string, prTitle string, prDescription string) (gitprovider.PullRequest, error) {
+					return dummyPullRequest{}, nil
+				}
+			})
+
+			DescribeTable("error during keep file write is propagated", func(conf testConfig, errorExpected bool) {
```
jpellizzari, line 995: :+1: on the boatload of test cases

jpellizzari (commented):
I guess I wasn't expecting the extra PR that was needed. I would have expected this `.keep` to get added whenever all the `yamls` got added.

pkg/services/app/add.go
jrryjcksn:
```
    custodian, guard, guardian, keeper, lookout, minder, picket, sentry, warden, warder, watch, watcher, watchman
```

jrryjcksn:
> I guess I wasn't expecting the extra PR that was needed. I would have expected this `.keep` to get added whenever all the `yamls` got added.

That was why I thought Product might be interested. It's the app repo that has the problem, not the config repo. It's so a user removing all her manifests won't run into the issue. In the case where we store the config in the app repo, we're not generating an additional PR since the `.wego` directory acts as a sentinel.

pkg/services/app/add.go
jrryjcksn: Not really. The .keep file is in the app repo while the other files are in the config repo. We don't need to do this at all in the default case where the automation is stored in the _app_ repo; only in the `--app-config-url=NONE` and `--app-config-url=<url>` case.

jpellizzari:
Discussed on a call; we decided that adding extra PRs JUST to ensure users don't hit this problem was too burdensome. We will close this, document the "gotcha", and revisit after some directory structure changes (which may invalidate this anyway).

jrryjcksn:
We decided to handle this via documentation for now. It may be reopened later if the problem appears to be more common than anticipated.

Issue #757, opened 2021-09-14
Cannot use weave-gitops@v0.2.5 as a go module


foot:
**Describe the bug**

`flux` is included as a `go:embed` directive but its not included in the source archive that is `go get`'d by the go module that depends on weave-gitops: https://proxy.golang.org/github.com/weaveworks/weave-gitops/@v/v0.2.5.zip

I would guess this was introduced when flux was moved into the apiutils and so importing https://github.com/weaveworks/weave-gitops/blob/main/pkg/server/server.go now also includes `"github.com/weaveworks/weave-gitops/pkg/flux"` via apiutils in https://github.com/weaveworks/weave-gitops/pull/687 (maybe, not sure)

**Environment**

```
$ go version
go version go1.17 darwin/amd64
```

**Affects versions**

`v0.2.5`

**To Reproduce**

```bash
$ git clone https://github.com/weaveworks-gitops-test/wego-library-test
$ vim go.mod # remove the replace directive for weave-gitops
$ go build main.go
../../go/pkg/mod/github.com/weaveworks/weave-gitops@v0.2.5/pkg/flux/setup.go:11:12: pattern bin/flux: no matching files found
```

**Expected behavior**

Be able to include the `wego_server "github.com/weaveworks/weave-gitops/pkg/server"` from another project and build successfully

jpellizzari:
@foot and I discussed, and we think the best remediation will be to let consumers of the `weave-gitops` packages specify their own path to the flux binary

PR #762, opened 2021-09-14
Allow flux binary to be overridden via an environment variable at setup time

jrryjcksn:
Closes: #757 

**What changed?**
We now allow the flux binary used during `weave-gitops` setup to be specified via an environment variable so that the `weave-gitops` code can be used as a library. The flux binary is currently stored as an embedded binary which does not propagate when pulled via `go.mod`. After pulling the code as a module, setting `WEAVE_GITOPS_FLUX_BIN_PATH` to point at a flux binary will cause the referenced binary to be installed for use by weave-gitops.

**Why?**
To allow the code to be used as a library

**How did you test it?**
Unit tests

**Release notes**
No

**Documentation Changes**
No

### DIFF
--- a/pkg/flux/flux.go
+++ b/pkg/flux/flux.go
@@ -32,7 +32,7 @@ type Flux interface {
 	GetLatestStatusAllNamespaces() ([]string, error)
 }
 
-const fluxBinaryPathEnvVar = "FLUX_BIN_PATH"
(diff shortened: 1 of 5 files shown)


jpellizzari (approved):
LGTM, digging the test coverage. Added @foot as a reviewer.

PR #772, opened 2021-09-16
Enabling linter: wsl

luizbafilho:
**What changed?**
- Enables the Whitespace Linter
- Nothing was changed besides adding and removing lines

**Why?**
`golangci-lint` provides a variety of linter that can help us to keep code consistency in place, and minimize friction during reviews by helping to fix nits comments that happen quite often.

Here is the full list of linters: https://golangci-lint.run/usage/linters/

Here are the rules this linter will enforce: https://github.com/bombsimon/wsl/blob/master/doc/rules.md#anonymous-switch-statements-should-never-be-cuddled

**How did you test it?**

**Release notes**

**Documentation Changes**


jpellizzari:
@luizbafilho Is this rule enforceable within `gofmt`? I rely on my editor to format on save, and if these rules cannot be enforced by `gofmt` (or some other formatter we all run), then it will be a pretty major blocker for me.

luizbafilho:
You'd have to integrate `golangci-lint` in your editor: https://golangci-lint.run/usage/integrations/

jpellizzari (commented):
If this requires an external linting runtime outside of `gofmt`, please add a note in the README. This also needs a `Makefile` target to run the job in a standardized way.

Overall, not super excited about having to run two formatters (`gofmt` and `golanci-lint`). If this is only in an effort to avoid "nits" in PR feedback, another option would be to all agree as reviewers that giving feedback on whitespace (when we are all using `gofmt`) is super low value.

jpellizzari:
> I think you are confusing gofmt with linter, those are not the same thing, we already have golanglint-ci enable in the project, and it's already catching some stuff, you are compiling with it, I'm just adding an additional one initially. It's not just about whitespace.

If there are linting errors around whitespace, can I get `golangci-lint` to fix them for me?

luizbafilho:
> > I think you are confusing gofmt with linter, those are not the same thing, we already have golanglint-ci enable in the project, and it's already catching some stuff, you are compiling with it, I'm just adding an additional one initially. It's not just about whitespace.
> 
> If there are linting errors around whitespace, can I get `golangci-lint` to fix them for me?

nope, that linter specifically does not provide autofix, others do.

jpellizzari (approved):

PR #774, opened 2021-09-17
Pass version to deployment template as argument

jpellizzari:
Having an `ErrImagePull` state on a pod seems to slow down tests by double. This fix ensures a `latest` tag is pushed during a release (although users should be discouraged from using that tag).

For acceptance tests, they will pull down the `latest` tag for now as a stop gap. I have already pushed a `latest` image from my laptop so that we don't have to wait for a release for the tag to be present.

Also fixes an error in the FE docker image build where `npm install` was being used incorrectly.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -76,5 +76,7 @@ jobs:
         with:
           context: .
           push: true
-          tags: ${{ env.REGISTRY }}/weaveworks/${{ env.IMAGE_NAME }}:${{ github.event.release.tag_name }}
+          tags:
+            ${{ env.REGISTRY }}/weaveworks/${{ env.IMAGE_NAME }}:${{ github.event.release.tag_name }}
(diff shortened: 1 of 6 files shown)


luizbafilho (approved):

josecordaz (approved):
LGTM

josecordaz:
@Jordan Pellizzari @rokshana @iahmad There is a conflict in some conditions using kubectl wait where they wait for "ALL" pods to be in a ready state to continue the execution of the tests, like this one https://github.com/weaveworks/weave-gitops/blob/main/test/acceptance/test/utils.go#L560 . That is why fixing the image pull error fixes the issue. I tried to temporarily avoid that problem by excluding that specific pod (wego-app https://github.com/weaveworks/weave-gitops/blob/main/test/acceptance/test/utils.go#L568) but it seems there are more places that need that condition. What is interesting is that we didn't seem to catch that timeout error, not sure if we are missing errors handlers somewhere.

Issue #780, opened 2021-09-20
Gitops no longer allows add for applications from repositories with default branches other than main


jrryjcksn:
**Describe the bug**
The addition of an app default branch in `add/cmd.go` prevents gitops from determining the default branch for a repository. There is a code path in `add.go` which checks for an empty branch and consults the repository for the default branch. This is being prevented by the default value.

**Environment**
 - gitops: v0.2.5

**To Reproduce**
1. Attempt to add an application from a repository with a default branch other than main (e.g. master)

PR #781, opened 2021-09-20
fixes for: 780, redundant status messages, and mention of wego in cluster status

jrryjcksn:
Closes: #780 

**What changed?**
- Status calls no longer mention "wego"
- We once again use the correct default branch for a repository if no branch is specified
- Added an acceptance test for default branch behavior

**Why?**
- Fix for leftover "wego" reference
- To make sure default branches work correctly

**How did you test it?**
Manual tests, new acceptance test

**Release notes**
No

**Documentation Changes**
The gitops doc changes are in flight


J-Thompson12 (approved):

pkg/services/app/add.go
```diff
@@ -135,7 +136,7 @@ func (a *App) Add(params AddParams) error {
 
 	a.printAddSummary(params)
 
-	if err := IsClusterReady(a.Logger, a.Kube); err != nil {
+	if err := IsClusterReady(&loggerfakes.FakeLogger{}, a.Kube); err != nil {
```
jpellizzari, line 139: Is the fake logger here intentional?

test/acceptance/test/utils.go
```diff
@@ -216,14 +216,18 @@ func ResetOrCreateClusterWithName(namespace string, deleteWegoRuntime bool, clus
 	return clusterName, nil
 }
 
-func initAndCreateEmptyRepo(appRepoName string, isPrivateRepo bool) string {
+func initAndCreateEmptyRepoWithMainDefaultBranch(appRepoName string, isPrivateRepo bool) string {
+	return initAndCreateEmptyRepo(appRepoName, DEFAULT_BRANCH_NAME, isPrivateRepo)
```
jpellizzari, line 220: Seems odd to wrap this when it's only saving one argument that would already be a `const`. I guess it is saving us from passing `DEFAULT_BRANCH_NAME` in everywhere?

jpellizzari (approved):
LGTM, only nits

pkg/services/app/add.go
jrryjcksn: Yes, because we now check the status up front when we set up the git clients so we were getting all the status info printed out twice. I didn't want to entirely remove the check in `add` for when it gets used by the UI, so I passed in a fake logger.

test/acceptance/test/utils.go
jrryjcksn: Well, I needed to be able to call it in _one_ test with a different branch and thought it would be unfortunate to have to pass the same argument in _every call but one_ (and then one would have to go look to determine what branch was being passed).

Issue #634, opened 2021-08-23
Create "app add" API endpoint


jpellizzari:
We will want to add an application via the UI, so therefore we need an API route to add an app. Users may also want to consume the endpoint to add applications programatically.

Acceptance Critera:
* Receive a POST with the necessary JSON fields filled out
* Validates that all necessary fields are present
* Adds the application to the cluster
* Extract a token from the request context for use with the AuthService
* Add BOTH unit tests and integration tests (test with actual the actual HTTP client all the way through to the `testenv` cluster)

PR #788, opened 2021-09-21
Simplify apputils app service creation

jpellizzari:
I did some of this work as part of #634, but it made sense to pull it into a separate PR. This tries to flatten the calls in `apputils` a little bit.

The `fake.Client` stuff was cherry-picked from another PR and will get wiped out from the diff on rebase. It can be ignored.

### DIFF
--- a/api/v1alpha1/application_types.go
+++ b/api/v1alpha1/application_types.go
@@ -87,6 +87,10 @@ type Application struct {
 	Status ApplicationStatus `json:"status,omitempty"`
 }
 
+func (a *Application) IsHelmRepository() bool {
+	return a.Spec.DeploymentType == DeploymentTypeHelm
+}
+
 //+kubebuilder:object:root=true
 
 // ApplicationList contains a list of Application
--- a/api/v1alpha1/application_types.go
+++ b/api/v1alpha1/application_types.go
@@ -88,7 +88,7 @@ type Application struct {
 }
 
 func (a *Application) IsHelmRepository() bool {
-	return a.Spec.DeploymentType == DeploymentTypeHelm
+	return a.Spec.SourceType == SourceTypeHelm
 }
 
 //+kubebuilder:object:root=true
--- a/cmd/gitops/app/remove/cmd.go
+++ b/cmd/gitops/app/remove/cmd.go
@@ -54,7 +54,7 @@ func runCmd(cmd *cobra.Command, args []string) error {
 		return fmt.Errorf("failed to fetch app %q: %w", params.Name, err)
 	}
 
-	appService, appError := apputils.GetAppService(ctx, appObj.Spec.URL, appObj.Spec.ConfigURL, appObj.Namespace, appObj.IsHelmRepository(), params.DryRun)
+	appService, appError := apputils.GetAppService(ctx, appObj)
 	if appError != nil {
 		return fmt.Errorf("failed to create app service: %w", appError)
 	}
--- a/cmd/gitops/app/remove/cmd.go
+++ b/cmd/gitops/app/remove/cmd.go
@@ -12,7 +12,6 @@ import (
 	"github.com/spf13/cobra"
 	"github.com/weaveworks/weave-gitops/cmd/gitops/version"
 	"github.com/weaveworks/weave-gitops/pkg/apputils"
-	"github.com/weaveworks/weave-gitops/pkg/kube"
 	"github.com/weaveworks/weave-gitops/pkg/services/app"
 	"github.com/weaveworks/weave-gitops/pkg/utils"
 	"k8s.io/apimachinery/pkg/types"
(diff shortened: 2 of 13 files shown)


cmd/gitops/app/remove/cmd.go
```diff
@@ -48,7 +50,17 @@ func runCmd(cmd *cobra.Command, args []string) error {
 	params.Name = args[0]
 	params.Namespace, _ = cmd.Parent().Flags().GetString("namespace")
 
-	appService, appError := apputils.GetAppService(ctx, params.Name, params.Namespace)
+	kube, _, err := kube.NewKubeHTTPClient()
```
jrryjcksn, line 53: I don't understand why we would want to introduce so much redundant code. It trigger my DRY detector. If any of those steps change, all sorts of commands need to be updated. Additionally, people writing commands have to know a lot more about the underlying support.

jrryjcksn (commented):
The test app stuff seems better to me but I disagree with the changes to `GetAppService`. I think introducing unnecessary redundancy has a detrimental effect on code in general. :man_shrugging:

api/v1alpha1/application_types.go
```diff
@@ -87,6 +87,10 @@ type Application struct {
 	Status ApplicationStatus `json:"status,omitempty"`
 }
 
+func (a *Application) IsHelmRepository() bool {
+	return a.Spec.DeploymentType == DeploymentTypeHelm
```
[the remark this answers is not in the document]
jpellizzari: Fixed

cmd/gitops/app/remove/cmd.go
jpellizzari: @jrryjcksn Good call, updated to move that stuff into a helper. The main impetus for this change was to remove the two `GetAppService` functions that were **almost** identical. PTAL.

jrryjcksn (commented):
I'm okay with this but I think it sort of inverts things. `add` presents a unique special case because it has no existing app to query out. The original code had a special case for `add` for that reason. Now, we make every `non-add` command pay the price for `add` being special. I assume you're intending to make the API cleaner (which I appreciate) but I have a fairly strong aversion to boilerplate code. What if we did something like:

```
type AppServiceDataFetcher interface {
    GetAppServiceData() (AppServiceData, error)
}

type AppServiceData struct {
    URL, ConfigURL, Namespace string
    IsHelmRepository bool
}

type AppRef struct {
    Name, Namespace string
}

func (ar AppRef) GetAppServiceData() (AppServiceData, error) {
	kube, _, kubeErr := kube.NewKubeHTTPClient()
	if kubeErr != nil {
		return AppServiceData{}, fmt.Errorf("error creating k8s http client: %w", kubeErr)
	}

	app, appErr := kube.GetApplication(ctx, types.NamespacedName{Namespace: ar.Namespace, Name: ar.Name})
	if appErr != nil {
		return AppServiceData{}, fmt.Errorf("could not retrieve application: %w", appErr)
	}

    return AppServiceData{
        URL: app.URL,
        ConfigURL: app.ConfigURL,
        Namespace: ar.Namespace,
        IsHelmRepository: app.Spec.SourceType == wego.SourceTypeHelm,
    }, nil
}

func (asd AppServiceData) GetAppServiceData() (AppServiceData, error) {
    return asd, nil
}

// Then GetAppService could look like:
func GetAppService(ctx context.Context, appDataFetcher apputils.AppServiceDataFetcher) (app.AppService, error) {
    appData, appDataErr := appDataFetcher.GetAppServiceData()
    if appDataErr != nil {
        return nil, fmt.Errorf("error retrieving app data: %w", appDataErr)
    }
    
	osysClient, fluxClient, kubeClient, logger, baseClientErr := GetBaseClients()
	if baseClientErr != nil {
		return nil, fmt.Errorf("error initializing clients: %w", baseClientErr)
	}

	appClient, configClient, gitProvider, clientErr := getGitClients(ctx, appData)
	if clientErr != nil {
		return nil, fmt.Errorf("error getting git clients: %w", clientErr)
	}

	return app.New(ctx, logger, appClient, configClient, gitProvider, fluxClient, kubeClient, osysClient), nil
}

    // and add would look like:
    appService, appError := apputils.GetAppService(
        ctx,
        apputils.AppServiceData{URL: params.Url, ConfigURL: params.AppConfigUrl, IsHelmRepository: isHelmRepository, Namespace: namespace})
	if appError != nil {
		return fmt.Errorf("failed to create app service: %w", appError)
	}

    // and remove, etc. would look like:
    appService, appError := apputils.GetAppService(ctx, apputils.AppRef(appName, namespace))
	if appError != nil {
		return fmt.Errorf("failed to create app service: %w", appError)
	}


```

jpellizzari:
Anywhere that we would call `FetchAppByName` would need to have a `GetAppServiceData`, so I'm not sure what that saves in terms of boilerplate.

I'll try with that approach and see how it looks.

jrryjcksn:
It wouldn't; it would pass an `AppRef` with name and namespace and the GetAppService method would fetch the app

jrryjcksn:
Do you have an issue with my `AppServiceDataFetcher` approach? I think it gets around all the issues we had (boilerplate in the commands and multiple versions of `GetAppService`, etc. :thinking:

jpellizzari:
> Do you have an issue with my `AppServiceDataFetcher` approach? I think it gets around all the issues we had (boilerplate in the commands and multiple versions of `GetAppService`, etc. thinking

I'm not quite ready for a review yet, but trying to make it work. It seems circuitous and tougher to understand. Still working though.

jpellizzari:
This is turning into more trouble than its worth. I don't want to spend a bunch of time playing whack-a-mole with brittle acceptance tests that look for specific error messages. Closing.

Issue #750, opened 2021-09-10
Let the user know they can skip auth if they set a GITHUB_TOKEN


jpellizzari:
While watching a user use `wego` live, they kept having to run through the Github device flow. It might not be obvious to someone that they can skip that by setting a token in their environment.

Acceptance Criteria:
* If a user does not have a `GIT{HUB|LAB}_TOKEN` env var set, warn them in the CLI output that they can set one to skip the Auth flow

PR #789, opened 2021-09-21
Inform the user that setting the appropriate token allows skipping auth flow

jrryjcksn:
Closes: #750 

**What changed?**
If the auth service fails to find a token, a log message is emitted to inform the user that she may bypass the auth flow by setting it

**Why?**
So that a user will not repeatedly go through the auth flow for no reason

**How did you test it?**

**Release notes**

**Documentation Changes**

### DIFF
--- a/pkg/services/auth/auth.go
+++ b/pkg/services/auth/auth.go
@@ -51,10 +51,19 @@ func GetGitProvider(ctx context.Context, url string) (gitproviders.GitProvider, 
 		return nil, fmt.Errorf("error detecting git provider: %w", providerNameErr)
 	}
 
-	token, tokenErr := osysClient.GetGitProviderToken()
+	tokenVarName, varNameErr := getTokenVarName(providerName)
+	if varNameErr != nil {
+		return nil, fmt.Errorf("could not determine git provider token name: %w", varNameErr)
+	}
+
+	token, tokenErr := osysClient.GetGitProviderToken(tokenVarName)
 
 	if tokenErr == osys.ErrNoGitProviderTokenSet {
 		// No provider token set, we need to do the auth flow.
+
+		logger := logger.NewCLILogger(osysClient.Stdout())
+		logger.Printf("Setting the %q environment variable to a valid token will allow ongoing use of the CLI without requiring a browser-based auth flow...\n", tokenVarName)
+
 		authHandler, authErr := NewAuthCLIHandler(providerName)
 		if authErr != nil {
 			return nil, fmt.Errorf("could not get auth handler for provider %s: %w", providerName, authErr)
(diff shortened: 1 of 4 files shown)


pkg/services/auth/auth.go
```diff
@@ -44,21 +44,38 @@ func NewAuthCLIHandler(name gitproviders.GitProviderName) (BlockingCLIAuthHandle
...
-	token, tokenErr := osysClient.GetGitProviderToken()
+	authHandler, authErr := NewAuthCLIHandler(providerName)
+	if authErr != nil {
+		return nil, fmt.Errorf("could not get auth handler for provider %s: %w", providerName, authErr)
+	}
+
+	osysClient := osys.New()
+	logger := logger.NewCLILogger(osysClient.Stdout())
+
+	return getGitProviderWithClients(ctx, providerName, osysClient, authHandler, logger)
```
jpellizzari, line 60: I only see the used once. Does it need to be a helper, or can we just do this work here in this function?

jpellizzari: I looked at it longer and realized this is a test-ability change.

pkg/services/auth/auth.go
```diff
@@ -44,21 +44,38 @@ func NewAuthCLIHandler(name gitproviders.GitProviderName) (BlockingCLIAuthHandle
...
+
+	token, tokenErr := osysClient.GetGitProviderToken(tokenVarName)
 
 	if tokenErr == osys.ErrNoGitProviderTokenSet {
 		// No provider token set, we need to do the auth flow.
-		authHandler, authErr := NewAuthCLIHandler(providerName)
-		if authErr != nil {
-			return nil, fmt.Errorf("could not get auth handler for provider %s: %w", providerName, authErr)
-		}
+		logger.Printf("Setting the %q environment variable to a valid token will allow ongoing use of the CLI without requiring a browser-based auth flow...\n", tokenVarName)
```
jpellizzari, line 78: What about `logger.Warnf` to render the little :warning: icon to get someone's attention?

pkg/services/auth/auth_test.go
```diff
@@ -113,5 +116,71 @@ var _ = Describe("auth", func() {
...
+			Context("informs the user that she can use a token for auth", func() {
+				BeforeEach(func() {
+					osysClient = &osysfakes.FakeOsys{
+						GetGitProviderTokenStub: func(tokenVarName string) (string, error) {
+							return "", osys.ErrNoGitProviderTokenSet
+						},
+					}
+				})
+
+				It("generates the correct token info message for GitHub", func() {
```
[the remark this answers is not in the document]
jrryjcksn: I'm on the fence. Let me see what that would look like. A few more cases and it would be absolutely be the right model; one or two less and it would also be clear :-)

pkg/services/auth/auth.go
jrryjcksn: I didn't realize that would generate the icon. Definitely worth switching over.

pkg/services/auth/auth_test.go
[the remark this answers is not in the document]
jrryjcksn: It was definitely worth it.

PR #792, opened 2021-09-21
Leave image as wego-app until we completely purge "wego"

jrryjcksn:
**What changed?**
This undoes the beginning of moving our UI image to `gitops-app` from `wego-app`. There are a variety of other things to update in that move.

**Why?**
So that a released version works correctly

**How did you test it?**
There are current no tests that are affected by this as the image is only pushed during release. The wego-app deployment always fails with an `ErrImagePull` (this should be fixed with the addition of `garden.io`)

**Release notes**
No

**Documentation Changes**
No

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -6,7 +6,7 @@ on:
 
 env:
   REGISTRY: ghcr.io
-  IMAGE_NAME: gitops-app
+  IMAGE_NAME: wego-app
 
 jobs:
   goreleaser:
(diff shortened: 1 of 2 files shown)


jpellizzari:
@jrryjcksn So I understand it, we are not naming the container image `gitops-app` **yet**, correct?

josecordaz (approved):
LGTM

jrryjcksn:
That's right. I'd like to make those changes once we can test locally

PR #793, opened 2021-09-21
Only check for flux upgrades once a week

jpellizzari:
We will realistically only upgrade flux once per release, and we release every two weeks. A once-a-week check should be fine. This cron expression is `At 00:00 on Sunday.`


J-Thompson12:
I feel like you are missing some changes.

jpellizzari:
> I feel like you are missing some changes.

What the heck happened there? Nice catch. Updated.

josecordaz (approved):
LGTM

Issue #776, opened 2021-09-17
Installation into non-default namespace fails


jrryjcksn:
**Describe the bug**
If you attempt to install wego into a namespace other than `wego-system`, it will fail with this error:
```
Error: could not apply manifest: failed applying {"apiVersion":"v1","kind":"ServiceAccount","metadata":{"name":"wego-app-service-account","namespace":"wego-system"},"secrets":[{"name":"wego-app-service-account-token"}]}: the namespace of the object (wego-system) does not match the namespace on the request (<namespace>)
```
**Environment**
 - wego: v0.2.5

PR #795, opened 2021-09-22
Removed namespace from wego-app manifests

josecordaz:
Added assertion to integration test using non-default ns to make sure it fails when error output is not empty

Closes: #776 

**What changed?**
Removed wego-system ns from wego-app manifests

**Why?**
It was causing an error at installation time when using a namespace different from the default

**How did you test it?**
Manually and  by fixing the integration test for this particular case

**Release notes**

**Documentation Changes**

### DIFF
--- a/manifests/crds/wego.weave.works_apps.yaml
+++ b/manifests/crds/wego.weave.works_apps.yaml
@@ -78,9 +78,3 @@ spec:
     storage: true
     subresources:
       status: {}
-status:
-  acceptedNames:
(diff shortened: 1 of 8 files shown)


luizbafilho (approved):

rokshana-b (approved):

talaverant:
Hi @josecordaz ! If I understand correctly, we can install wego on a non-default namespace? How can we do that? Thanks, Jose

PR #800, opened 2021-09-23
Fix goreleaser binary and docker push job

jpellizzari:
`goreleaser.yaml` still had "wego" as the binary name. Docker push `tags:` is a CSV list


J-Thompson12 (approved):
Of course. I forgot jerry undid these changes

PR #801, opened 2021-09-23
Use fake k8s client instead of full etcd cluster

jpellizzari:
On recommendation from @hiddeco, Uses a lighter `fake` client. Hopefully reduces the incidents of `cannot listen: port already in use'

### DIFF
--- a/pkg/services/auth/auth_suite_test.go
+++ b/pkg/services/auth/auth_suite_test.go
@@ -5,8 +5,9 @@ import (
 
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
-	"github.com/weaveworks/weave-gitops/pkg/testutils"
+	"github.com/weaveworks/weave-gitops/pkg/kube"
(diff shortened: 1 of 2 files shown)


J-Thompson12 (approved):
Hope this works.

jpellizzari:
> Hope this works.

@J-Thompson12 Why do you say that? Were you hitting that error frequently?

luizbafilho:
It worked great, I always got the error. Should that be removed from https://github.com/weaveworks/weave-gitops/blob/main/pkg/kube/kube_suite_test.go#L31 either?

luizbafilho (approved):

jpellizzari:
> It worked great, I always got the error. Should that be removed from https://github.com/weaveworks/weave-gitops/blob/main/pkg/kube/kube_suite_test.go#L31 either?

No, we still need the full client for the `.Apply()` logic. This fake client won't do all the fancy logic we need for that operation.

Issue #601, opened 2021-08-12
Skip authenticate on `--dry-run`


jpellizzari:
We do the authentication flow even when the `--dry-run` flag is present, and we probably shouldn't do that. The user just wants to see how their manifests will look, so presenting them with an authentication flow is a bit strange and jarring.

palemtnrider:
In the dry-run output we should print a message about skipping the authentication check with the git server (GitHub, GitLab, etc)

jpellizzari:
Harder to write a test for this one with the `apputils` change. This might require a slightly larger refactor to make that package more testable.

`apputils` is also low-hanging fruit for unit test coverage increase.

PR #806, opened 2021-09-27
Implementing DryRun gitprovider

luizbafilho:
Closes: #601 

**What changed?**
- Avoid auth flow and deployment keys upload when dryrun
- Renaming `err` vars to Go convention

**Why?**
- minimize user friction

**How did you test it?**

**Release notes**

**Documentation Changes**

### DIFF
--- a/pkg/gitproviders/dryrun.go
+++ b/pkg/gitproviders/dryrun.go
@@ -6,7 +6,7 @@ type dryrunProvider struct {
 	provider GitProvider
 }
 
-func NewDryRun(_ Config) (GitProvider, error) {
+func NewDryRun() (GitProvider, error) {
 	provider, err := New(Config{
 		Provider: GitProviderGitHub,
 		Token:    "dummy",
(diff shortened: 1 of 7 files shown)


jpellizzari:
@luizbafilho Does this close #601 ?

luizbafilho:
yes, updated the description

pkg/gitproviders/dryrun.go
```diff
@@ -0,0 +1,83 @@
+package gitproviders
+
+import "github.com/fluxcd/go-git-providers/gitprovider"
+
+type dryrunProvider struct {
+	provider GitProvider
+}
+
+func NewDryRun(_ Config) (GitProvider, error) {
```
jpellizzari, line 9: Why do we need a `Config` object as an argument here?

jpellizzari (commented):
Is it possible to add test coverage for this?

pkg/gitproviders/dryrun.go
luizbafilho: we don't, removing it.

luizbafilho:
> Is it possible to add test coverage for this?

as you mentioned in the issue, is quite hard to require a big refactor, which I think is better done in a separate issue.

But I could add some to the dryrun implementation indeed. Not really testing anything but it will help bump the coverage.

jpellizzari:
> But I could add some to the dryrun implementation indeed. Not really testing anything but it will help bump the coverage.

Not looking to bump the coverage for its own sake, only to prove that this is working programtically. I'll leave it up to you as to whether it makes sense to add some sort of automated test.

jpellizzari (approved):

PR #807, opened 2021-09-27
Add template command

yiannistri:
Closes: weaveworks/weave-gitops-enterprise#92

**What changed?**
Added a new command `gitops get template` that shows a list of CAPI templates.

**Why?**
This is part of the initiative to bring Weave GitOps Enterprise commands into Core.

**How did you test it?**
Tested locally + unit tests.

**Release notes**
A new command `gitops get template` will appear.

**Documentation Changes**
Follow-up PRs are needed to update the docs accordingly.


cmd/gitops/get/cmd.go
```diff
@@ -0,0 +1,21 @@
...
+
+import (
+	"github.com/go-resty/resty/v2"
+	"github.com/spf13/cobra"
+	"github.com/weaveworks/weave-gitops/cmd/gitops/get/templates"
+)
+
+func GetCommand(endpoint string, client *resty.Client) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "get",
```
jpellizzari, line 11: @J-Thompson12 Can you confirm this will work with our current implementation of `get commits`?

pkg/adapters/http.go
```diff
@@ -0,0 +1,81 @@
...
+
+// An HTTP client of the cluster service.
+type HttpClient struct {
+	baseURI *url.URL
+	client  *resty.Client
+}
+
+// NewHttpClient creates a new HTTP client of the cluster service.
+// The endpoint is expected to be an absolute HTTP URI.
+func NewHttpClient(endpoint string, client *resty.Client, out io.Writer) (*HttpClient, error) {
```
jpellizzari, line 25: This `adapters` package is only relevant for `templates`, correct? If so, should we put this code in `templates`?

jpellizzari (commented):
LGTM, minus one question about what to do with `adapters`. Since the `HttpClient` object is only relevant to `templates`, should we put it in that package?

cmd/gitops/get/cmd.go
jrryjcksn: Our current `get commits` is under `app`: `gitops app <appname> get commits`

cmd/gitops/main.go
```diff
@@ -95,14 +98,15 @@ func main() {
 	fluxClient.SetupBin()
 	rootCmd.PersistentFlags().BoolVarP(&options.verbose, "verbose", "v", false, "Enable verbose output")
 	rootCmd.PersistentFlags().String("namespace", "wego-system", "gitops runtime namespace")
+	rootCmd.PersistentFlags().StringVarP(&options.endpoint, "endpoint", "e", os.Getenv("WEAVE_GITOPS_ENTERPRISE_API_URL"), "The Weave GitOps Enterprise HTTP API endpoint")
```
jrryjcksn, line 101: Minor: I think people might find it a bit confusing to have `--endpoint` as an argument for every command. In fairness, the same could be said of `--namespace`, but that applies _almost_ everywhere.

pkg/adapters/http.go
```diff
@@ -0,0 +1,81 @@
...
+	res, err := c.client.R().
+		SetHeader("Accept", "application/json").
+		SetResult(&templateList).
+		Get(endpoint)
+
+	if err != nil {
+		return nil, fmt.Errorf("unable to GET templates from %q: %w", res.Request.URL, err)
+	}
+
+	if res.StatusCode() != http.StatusOK {
```
jrryjcksn, line 68: Not really an issue with the code... just for my education: When would we git a non-ok status but no error in this context?

jrryjcksn (approved):
LGTM (w/ one minor comment)

pkg/adapters/http.go
yiannistri: Currently yes but it would soon be used to retrieve `clusters` too. The intent of that package is to keep any adapter specific code (HTTP/gRPC etc.) in one place separate from the command/presentation logic. The two packages `adapters` and `templates` are then used together when the command is invoked. Later on, another package could be added i.e. `clusters` that deals with CAPI cluster commands and that could also be used together with the `adapters` package to retrieve and display cluster information.

I don't feel too strongly about this though, let me know if you think it complicates things.

cmd/gitops/get/templates/cmd.go
```diff
@@ -0,0 +1,40 @@
...
+	"github.com/go-resty/resty/v2"
+	"github.com/spf13/cobra"
+	"github.com/weaveworks/weave-gitops/pkg/adapters"
+	"github.com/weaveworks/weave-gitops/pkg/templates"
+	"k8s.io/cli-runtime/pkg/printers"
+)
+
+func TemplateCommand(endpoint string, client *resty.Client) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "template",
```
foot, line 15: Easy way to add an alias for `templates` too?

cmd/gitops/main.go
yiannistri: I agree that it can be confusing but they don't need to worry about it if they are not enterprise users. Maybe we can make that part more clear in the parameter description? The alternative would be to include that in every enterprise command.

jpellizzari (approved):

pkg/adapters/http.go
yiannistri: This check is here more for completeness really. It's just a way to ensure that the response we got was a 200 as per the API contract. I guess a different status code could be returned without an error but that would likely be a bug on the server side. The previous error check would account for transport errors I guess.

PR #817, opened 2021-09-28
Get Commits returns empty commit list instead of an error

J-Thompson12:
**What changed?**
Add tests for get commits

**Why?**

**How did you test it?**
Manually and added commit tests

**Release notes**

**Documentation Changes**

### DIFF
--- a/pkg/gitproviders/provider.go
+++ b/pkg/gitproviders/provider.go
@@ -493,7 +493,7 @@ func (p defaultGitProvider) GetCommitsFromUserRepo(userRepRef gitprovider.UserRe
 	// currently locking the commit list at 10. May discuss pagination options later.
 	commits, err := ur.Commits().ListPage(ctx, targetBranch, pageSize, pageToken)
 	if err != nil {
-		if strings.Contains(err.Error(), "409 Git Repository is empty") {
+		if isEmptyRepoError(err) {
 			return []gitprovider.Commit{}, nil
 		}
 
@@ -515,7 +515,7 @@ func (p defaultGitProvider) GetCommitsFromOrgRepo(orgRepRef gitprovider.OrgRepos
 	// currently locking the commit list at 10. May discuss pagination options later.
 	commits, err := ur.Commits().ListPage(ctx, targetBranch, pageSize, pageToken)
 	if err != nil {
-		if strings.Contains(err.Error(), "409 Git Repository is empty") {
+		if isEmptyRepoError(err) {
 			return []gitprovider.Commit{}, nil
 		}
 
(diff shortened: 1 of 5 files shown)


jpellizzari (changes requested):
Please format this PR title like release notes

pkg/services/app/commit_test.go
```diff
@@ -0,0 +1,98 @@
...
+		gitProviders.GetCommitsFromUserRepoStub = func(gitprovider.UserRepositoryRef, string, int, int) ([]gitprovider.Commit, error) {
+			return commits, nil
+		}
+
+		commit, err := appSrv.GetCommits(commitParams, application)
+		Expect(err).ShouldNot(HaveOccurred())
+		Expect(len(commit)).To(Equal(1))
+	})
+
+	It("gets commits for a user app", func() {
```
jpellizzari, line 38: Same name as the test above it?

pkg/services/app/commit_test.go
```diff
@@ -0,0 +1,98 @@
...
+			Namespace: "wego-system",
+		}
+
+		application := &wego.Application{
+			Spec: wego.ApplicationSpec{SourceType: wego.SourceTypeHelm},
+		}
+
+		_, err := appSrv.GetCommits(commitParams, application)
+		Expect(err).Should(HaveOccurred())
+		Expect(err.Error()).Should(Equal("unable to get commits for a helm chart"))
```
jpellizzari, line 74: Can we use an error type here instead of looking for an exact string match? If someone wraps the error or changes one character, the test fails.

pkg/gitproviders/provider.go
```diff
@@ -511,6 +515,10 @@ func (p defaultGitProvider) GetCommitsFromOrgRepo(orgRepRef gitprovider.OrgRepos
 	// currently locking the commit list at 10. May discuss pagination options later.
 	commits, err := ur.Commits().ListPage(ctx, targetBranch, pageSize, pageToken)
 	if err != nil {
+		if strings.Contains(err.Error(), "409 Git Repository is empty") {
```
jpellizzari, line 518: Why is this block repeated?

J-Thompson12: One is for org repo and one is for use repo. I guess I can combine them.

jpellizzari: One way to avoid repeating would be to do something like this:

```suggestion
		if isEmptyRepoError(err) {
```

jpellizzari (approved):

pkg/services/app/commit_test.go
J-Thompson12: Yep copy paste got me.

pkg/services/app/commit_test.go
J-Thompson12: I will follow what I see in other places just checking if an err occured.

Issue #656, opened 2021-08-24
Implement garden.io for local development of wego-app


josecordaz:
**Short description**
As a developer I want to implement garden.io so that I can have fast deployments of wego-api locally.

Garden will allow us to develop API features quickly while the `wego-app` is running on the cluster.

Success looks like: I can run a command and a `wego-app` instance runs on the cluster; when I update some `.go` files, the server reloads with the changes.

Acceptance Criteria:
* One-time installation instructions are documented in the README
* Some sort of configuration file is present in the `weave-gitops` repo that tells Garden what to do
* WeGO devs are able to run a script or Makefile target to get started
* Updating `.go` code automatically reloads the container/pod/deployment running in the cluster
* Rebuilds should not take longer than a few seconds to complete (if its taking too long to rebuild, we need to re-evaluate)
* The `wego-app` needs to be accessible via `localhost` so that UI development can happen

jpellizzari:
Putting this on hold for now. Closing.

PR #819, opened 2021-09-28
Added garden-io configuration for local development

josecordaz:
Closes: #656 

**What changed?**
Added garden-io configuration for local development

**Why?**
To facilitate local development, like automatic deployment to the cluster once the source code changes.

**How did you test it?**
By manually changing source code and making sure the deployment changed inside the cluster.

**Release notes**

**Documentation Changes**


README.md
```diff
@@ -76,6 +76,54 @@ For VSCode, use these editor configuration flags:
         "--fast",
     ],
 ```
+## Development on kubernetes
```
jpellizzari, line 79: :+1: For the README notes

README.md
```diff
@@ -76,6 +76,54 @@ For VSCode, use these editor configuration flags:
...
+
+If you need to develop while gitops is running inside a kubernetes cluster you can use garden.io (Install Garden [Install garden-io](https://docs.garden.io/getting-started/1-installation))
+
+Garden version `0.12.26` was used to generate the configuration files for this repository.
+
+Only requirement is to run `gitops install` beforehand as need some kubernetes resources to be present in the cluster first.
+
+Once Garden is installed you just need to run:
+```
+garden deploy --dev wego-app-service
```
jpellizzari, line 89: Can we make this a `Makefile` target to ensure everyone users it the right way?

README.md
```diff
@@ -76,6 +76,54 @@ For VSCode, use these editor configuration flags:
...
+```
+
+### Ingress configuration to expose gitops locally
+
+Different configuration is needed depending on the type of cluster used:
+
+For kind clusters we need to configure nginx at creation time:
+```bash
+cat <<EOF | kind create cluster --config=-                                                                                                             ─╯
```
jpellizzari, line 98: Let's standardize this into a `Makefile` target

cmd/gitops/ui/run/cmd.go
```diff
@@ -93,7 +93,7 @@ func runCmd(cmd *cobra.Command, args []string) error {
 		}
 	}()
 
-	if isatty.IsTerminal(os.Stdout.Fd()) {
+	if isatty.IsTerminal(os.Stdout.Fd()) && os.Getenv("SKIP_OPEN_BROWSER") == "" {
```
jpellizzari, line 96: :question: How would you feel about making this a CLI argument instead? That way, a user can specify the flag on the CLI and it gets printed in the help.

manifests/garden/deployment.yaml
```diff
@@ -0,0 +1,23 @@
+apiVersion: apps/v1
+kind: Deployment
```
jpellizzari, line 2: Why do we need a new deployment for this? We already have a deployment defined for `wego-app`.

manifests/garden/ingress.yaml
```diff
@@ -0,0 +1,18 @@
+apiVersion: networking.k8s.io/v1
+kind: Ingress
```
jpellizzari, line 2: Can we just use a `NodePort` service instead?

Dockerfile-dev
```diff
@@ -0,0 +1,20 @@
+FROM golang:1.16-buster
```
jpellizzari, line 1: :bulb: We can use the convention `dev.dockerfile` here for the name of this file.

Dockerfile-dev
```diff
@@ -0,0 +1,20 @@
...
+
+RUN apt-get update -y && apt-get install -y build-essential
+
+WORKDIR $GOPATH/src/github.com/weaveworks/weave-gitops
+
+RUN curl -sSfL https://raw.githubusercontent.com/cosmtrek/air/master/install.sh | sh -s
```
jpellizzari, line 7: What is `air`?

jpellizzari (commented):
Initial thoughts before pulling it down to try it out

README.md
josecordaz: @jpellizzari what would be a good name for the job? I'm thinking in `make dev` ?

cmd/gitops/ui/run/cmd.go
josecordaz: I wasn't completely sure about this approach initially, but I do like this idea

README.md
jpellizzari: How about we change the current `make api-dev` target to `make local-dev` and add a new target that is `make cluster-dev`?

manifests/garden/deployment.yaml
josecordaz: The way Garden works is that it doesn't allow to have the same file in both garden Modules. The specific way in which we embed wego-app manifests forces me to include this file in the container module. But also this file needs to be included in the Kubernetes module so it can apply it. I'm open to suggestions on how to address this in a different approach.

manifests/garden/ingress.yaml
josecordaz: Any particular reason to consider `NodePort` ?

manifests/garden/deployment.yaml
jpellizzari: Let me take a look and get back to you. There may be a way.

Dockerfile-dev
josecordaz: `air` is a tool that automatically rebuilds a binary when the source code changes. It uses a file named `.air.toml` to specify the configuration, like we can set which folders to watch/exclude.

jpellizzari: This works pretty well, and doesn't require a new tool and config file:

https://github.com/weaveworks/weave-gitops/blob/993c4723920d968e465958a314fa667ebb952be4/Makefile#L46

jpellizzari: You may even be able to run `make api-dev` in that container :exploding_head: 

In general though, this approach means there is no visual feedback that a rebuild is in progress.

manifests/garden/ingress.yaml
jpellizzari: It seems odd to use an `ingress` for development when we don't even use one in production

jpellizzari (commented):
In general, we cannot have two sets of yamls, one for garden and one for prod. As it is now, this won't work for us.

I will branch from this branch and see if I can get it to work with our existing yamls.

Dockerfile-dev
josecordaz: Thanks for the suggestion.

manifests/garden/ingress.yaml
josecordaz: The problem I was trying to fix with this is exposing the wego-api to be accessible locally. I'm curious if there is any particular advantage on using `NodePort` over `Ingress` ?

manifests/garden/deployment.yaml
josecordaz: This is fixed now

manifests/garden/ingress.yaml
josecordaz: I had another thought. I think using a `NodePort` service would involve a new service. But I'm trying to use the same service we already have and avoid having an extra one just for this.

Dockerfile-dev
josecordaz: Got a bit confused about which tool has no visual feedback. I played around with both tools and seems like air describes more details even when the file cannot be compiled, which find very helpful as we are working in development mode. 

This is the output I got using reflex:
Screen Shot 2021-09-30 at 12 52 25 AM

This is the output using air:
image

This is the change I tested with:
Screen Shot 2021-09-30 at 12 52 18 AM

README.md
```diff
@@ -76,6 +76,36 @@ For VSCode, use these editor configuration flags:
         "--fast",
     ],
 ```
+## Development on kubernetes
+
+If you need to develop while gitops is running inside a kubernetes cluster you can use garden.io (Install Garden [Install garden-io](https://docs.garden.io/getting-started/1-installation))
+
+Garden version `0.12.26` was used to generate the configuration files for this repository.
+
+Only requirement is to run `gitops install` beforehand as need some kubernetes resources to be present in the cluster first.
```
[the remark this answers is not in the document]
josecordaz: I just added the apps crd to the garden configuration to avoid the need of running `gitops install`

Dockerfile-dev
josecordaz: I'm happy to change this to use `reflex`.

manifests/garden/ingress.yaml
josecordaz: @jpellizzari  Something I would like to clarify here is if we want to expose the UI as well or just the server like `local-dev` target does?. I understood we should expose the UI as well. But I could be wrong. I was trying to avoid running two different commands to access the UI. Or if we just need to access the UI to work in backend modifications.

PR #821, opened 2021-09-29
Temporarily comment out gitlab acceptance tests

J-Thompson12:
**What changed?**
Temporarily comment out gitlab acceptance tests. The reason for this is that delete repo still uses hub and go-git-providers does not support repo deletion. We do not want to spam the test repos so these need to be commented out until a fix is made.

**Why?**

**How did you test it?**

**Release notes**

**Documentation Changes**

### DIFF
--- a/test/acceptance/test/add_tests.go
+++ b/test/acceptance/test/add_tests.go
@@ -205,72 +205,72 @@ var _ = Describe("Weave GitOps App Add Tests", func() {
 		})
 	})
 
-	It("Verify that gitops can deploy and remove a gitlab app after it is setup with an empty repo initially", func() {
-		var repoAbsolutePath string
-		private := true
-		tip := generateTestInputs()
-		appName := tip.appRepoName
-		var appRemoveOutput *gexec.Session
+	// It("Verify that gitops can deploy and remove a gitlab app after it is setup with an empty repo initially", func() {
+	// 	var repoAbsolutePath string
+	// 	private := true
+	// 	tip := generateTestInputs()
+	// 	appName := tip.appRepoName
+	// 	var appRemoveOutput *gexec.Session
 
-		addCommand := "app add . --auto-merge=true"
+	// 	addCommand := "app add . --auto-merge=true"
 
-		defer deleteRepo(tip.appRepoName, GITLAB_ORG)
-		defer deleteWorkload(tip.workloadName, tip.workloadNamespace)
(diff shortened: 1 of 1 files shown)


jpellizzari (commented):
How about using a `Skip()`? Otherwise the entire git history of these blocks will get hidden behind a single trivial commit that didn't actually change any logic.

Skip docs:
https://pkg.go.dev/github.com/onsi/ginkgo#Skip

test/acceptance/test/add_tests.go
```diff
@@ -205,72 +205,72 @@ var _ = Describe("Weave GitOps App Add Tests", func() {
 		})
 	})
 
-	It("Verify that gitops can deploy and remove a gitlab app after it is setup with an empty repo initially", func() {
-		var repoAbsolutePath string
-		private := true
-		tip := generateTestInputs()
-		appName := tip.appRepoName
-		var appRemoveOutput *gexec.Session
+	// It("Verify that gitops can deploy and remove a gitlab app after it is setup with an empty repo initially", func() {
```
jrryjcksn, line 208: :heavy_check_mark:  You can just put a `P` or `X` in front of an `It`, `Context`, etc. to make it get skipped. See "Pending Specs" here: 

https://onsi.github.io/ginkgo/

test/acceptance/test/utils.go
```diff
@@ -141,18 +141,18 @@ func getUniqueWorkload(placeHolderSuffix string, uniqueSuffix string) string {
...
-                           echo "%s" >> %s &&
-                           chmod 0600 %s &&
-                           ls -la %s &&
-						   ssh-keyscan gitlab.com >> ~/.ssh/known_hosts`, os.Getenv("GITLAB_KEY"), sshKeyPath, sshKeyPath, sshKeyPath))
-		session, err := gexec.Start(command, GinkgoWriter, GinkgoWriter)
-		Expect(err).ShouldNot(HaveOccurred())
-		Eventually(session).Should(gexec.Exit())
-	}
-}
+// func setupGitlabSSHKey(sshKeyPath string) {
```
jrryjcksn, line 144: If we use the pending specs, we don't need to comment this out

J-Thompson12:
Using skip does the job and is pretty clean

jrryjcksn:
Sure, either works. :man_shrugging: I didn't see Jordan's suggestion until after I posted mine

jrryjcksn (approved):
LGTM

jpellizzari (approved):

Issue #627, opened 2021-08-20
Create "wego update" command


jrryjcksn:
**Short description**
As a Weave GitOps user, I want to upgrade my existing Weave GitOps installation to the latest version.

**Acceptance criteria**

**Given** a new wego binary with version X.Y.Z
**And** a Weave GitOps installation running version >= X-1.0.0
**When** I invoke the `upgrade` command for the new binary
**Then** A PR is created in the platform repository containing the new GORT manifests
**And** the cluster is updated to the new version when the PR is merged

This assumes that none of the CRs generated during `wego app add` have been modified. If/when that occurs, we will need to update those in all relevant config repositories. See: Issue #70. 

**Design materials**
https://app.zenhub.com/workspaces/weaveworks-foss-603cb6b1d4dff9000e242dc4/issues/weaveworks/weave-gitops/70

sympatheticmoose:
A point on terminology - it would be useful for us if this could be `update` rather than `upgrade`. If we want to provide a CLI experience for upgrading from Weave GitOps Core to Weave GitOps Enterprise, `upgrade` is likely to make the most sense. So upgrade is to the higher tier, update is to the higher version.

PR #824, opened 2021-09-30
Add upgrade command

sarataha:
Closes: weaveworks/weave-gitops-enterprise#134

**What changed?**
Added a new command `gitops upgrade` used to upgrade to Weave GitOps Enterprise 

**Why?**
Part of the migrating Weave GitOps Enterprise commands to Weave GitOps Core.

**How did you test it?**
Manually tested it and used unit tests.

**Release notes**

**Documentation Changes**


jpellizzari:
@sarataha Code looks good at first glance, but I am a bit concerned about the `gitops upgrade` command and how it differentiates between WeGO core and WeGO enterprise. Let me discuss with the product team and @foot and I will get back to you for a real review.

sympatheticmoose:
In terms of naming, as referenced in this [issue](https://github.com/weaveworks/weave-gitops/issues/627#issuecomment-904734516) I'm happy to go with `upgrade` to describe a move in tiers from core to enterprise, and `update` as a version-to-version move. 

Update _generally_ seems accepted as a less drastic change of action compared to an upgrade, which would match our usage. 

We may discover through usage that this still confuses users and we need to come up with an alternative.

pkg/upgrade/upgrade.go
```diff
@@ -0,0 +1,289 @@
...
+
+	return strings.Trim(strings.TrimSuffix(strings.TrimSpace(repoEndpoint.Path), ".git"), "/"), nil
+}
+
+func cloneToTempDir(parentDir, gitURL, branch string, privKey []byte, w io.Writer) (*GitRepo, error) {
+	fmt.Fprintf(w, "Creating a temp directory...")
+
+	gitDir, err := ioutil.TempDir(parentDir, "git-")
+	if err != nil {
+		return nil, errors.Wrap(err, "TempDir")
```
jpellizzari, line 128: Does it make sense to use `fmt.Errorf` everywhere?

jpellizzari (commented):
Code looks good, but I am worried there isn't much test coverage here to reduce the risk of breaking this feature in the future. @foot Any opinions re test coverage?

pkg/upgrade/upgrade.go
```diff
@@ -0,0 +1,289 @@
...
+
+	fmt.Fprintf(w, "Checking if entitlement exists...\n")
+
+	entitlement, err := getEntitlement(clientset)
+	if err != nil {
+		return err
+	}
+
+	upgradeValues.Name = "weave-gitops-enterprise"
+	upgradeValues.Namespace = "wego-system"
```
J-Thompson12, line 63: This may be a misunderstanding on my part with how we are handling things in enterprise but in install we can pass in a namespace and do the install on that namespace.
```
func installRunCmd(cmd *cobra.Command, args []string) error {
	namespace, _ := cmd.Parent().Flags().GetString("namespace")
```
It looks to me like if someone did install in their own namespace it would screw up the upgrade.

J-Thompson12:
@sympatheticmoose I have a question as well about the cmd format. This looks to me like the command is to update one repo to enterprise. If this is the case then the command should follow the other commands `gitops app remove <app-name>` `gitops app status <app-name>` and be `gitops app upgrade <app-name>` Its very possible I am just misunderstanding how the upgrade to enterprise works because I dont know how the upgrade works.

pkg/upgrade/upgrade.go
sarataha: Yesss! Thank you for pointing this out 🙂

pkg/upgrade/upgrade.go
sarataha: I agree! We can make the namespace configurable and give it a default value 👍

sarataha: That totally makes sense to me! Thank you 🙏

sympatheticmoose:
> @sympatheticmoose I have a question as well about the cmd format. This looks to me like the command is to update one repo to enterprise. If this is the case then the command should follow the other commands `gitops app remove <app-name>` `gitops app status <app-name>` and be `gitops app upgrade <app-name>` Its very possible I am just misunderstanding how the upgrade to enterprise works because I dont know how the upgrade works.

I can't say I've looked at the code in any detail :) perhaps @sarataha could give a high level overview of the workflow? But this _should_ be installing the enterprise profile and replacing the UI etc. The `app` cmds are for user workloads.

pkg/upgrade/upgrade.go
```diff
@@ -0,0 +1,287 @@
...
+		if err != nil {
+			return err
+		}
+
+		fmt.Fprintf(w, "Deriving org/repo for PR as %v\n", githubRepoPath)
+		upgradeValues.RepoOrgAndName = githubRepoPath
+	}
+
+	if upgradeValues.GitRepository == "" {
+		gitRepositoryNameNamespace := "wego-system/" + strings.TrimSuffix(filepath.Base(repoURL), ".git")
```
luizbafilho, line 84: you'll need to use `upgradeValues.Namespace` instead of "wego-system" since user can pick another one.

pkg/upgrade/upgrade.go
```diff
@@ -0,0 +1,287 @@
...
+		return "", err
+	}
+
+	return strings.TrimSpace(stdout.String()), nil
+}
+
+func getEntitlement(clientset kubernetes.Interface) (*v1.Secret, error) {
+	var entitlement *v1.Secret
+
+	entitlement, err := clientset.CoreV1().Secrets("wego-system").Get(context.Background(), "weave-gitops-enterprise-credentials", metav1.GetOptions{})
```
luizbafilho, line 188: same as before using the namespace from UpgradeValues

luizbafilho (changes requested):

sarataha:
> > @sympatheticmoose I have a question as well about the cmd format. This looks to me like the command is to update one repo to enterprise. If this is the case then the command should follow the other commands `gitops app remove <app-name>` `gitops app status <app-name>` and be `gitops app upgrade <app-name>` Its very possible I am just misunderstanding how the upgrade to enterprise works because I dont know how the upgrade works.
> 
> I can't say I've looked at the code in any detail :) perhaps @sarataha could give a high level overview of the workflow? But this _should_ be installing the enterprise profile and replacing the UI etc. The `app` cmds are for user workloads.

That's right! :) We basically do as follows:

* Check for valid entitlements
* Add WGE profile to generate its manifests
* Create a PR against the repo to replace WG with WGE based on the new changes

Once the PR is merged it should install WGE.

pkg/upgrade/upgrade.go
sarataha: Yes right! 🙂 Thank you Luiz 🙏

foot:
Re: naming, as we're moving to _verb-noun_ I'd guess `gitops upgrade app $appName` is not far off and reserving the top level `upgrade` might not work?
- `gitops upgrade --system` ?
- `gitops upgrade --enterprise` ?

foot:
> Code looks good, but I am worried there isn't much test coverage here to reduce the risk of breaking this feature in the future. @foot Any opinions re test coverage?

Yes! We imported chunks of code from `pctl` on the initial pass but not the tests 🤦 . We can defo add some tests here. Some functionality here is very integrationy, and we haven't been embracing mocking very much instead relying on acceptance tests defined elsewhere.

Lets chat!

sympatheticmoose:
> `gitops upgrade app $appName` is not far off

@foot I was going to say no plans for it, then took another look at the proposed spreadsheet! I've reworked that to reference update as per my comments above. Also linked to this would be the changes to make add app idempotent which I believe in theory should allow an update/upgrade.

foot:
The test coverage is much better but not perfect, I had trouble faking the very top level due to having to instantiate clients after reading credentials / defaults from the env.

foot:
For sure, I've had quite positive experiences w/ ginkgo and wouldn't mind moving in that direction. If you guys are happy w/ it and think it will stick we'll have a chat about migrating, at least here and then in wge too.

jpellizzari:
> For sure, I've had quite positive experiences w/ ginkgo and wouldn't mind moving in that direction. If you guys are happy w/ it and think it will stick we'll have a chat about migrating, at least here and then in wge too.

I was kinda thinking about going the other way :thinking:. The work to convert it is probably not worth it. In any case, totally up to you guys.

PR #841, opened 2021-10-05
Remove concurrent test run setting

rokshana-b:
**What changed?**
I added a setting in the test.yaml file for concurrent test runs. It seems to be invalid and is causing test run problems. I'm removing that setting in this PR.

**Why?**
To stop github action from failing.


jrryjcksn (approved):
LGTM

Issue #768, opened 2021-09-15
Add/Remove App using the new `.weave-gitops` directory structure


jpellizzari:
From the ADR: https://github.com/weaveworks/weave-gitops-private/blob/main/docs/adrs/0007-directory-layout.md

Acceptance Criteria
* Make our repo write operations match the new file structure
* Ensure the behaviors listed in the [consequences section](https://github.com/weaveworks/weave-gitops-private/blob/main/docs/adrs/0007-directory-layout.md#consequences) are implemented
* Create a single entry-point for read/write operations from the `.weave-gitops` directory (ie create a `WeGODirectoryClient` object or some such)
* Ensure removing an app with the new directory structure works (does NOT include "remove via PR)
* Clean up existing references to the `.wego` directory and use the aforementioned client in their place ^^

Edit: After working through the multi-cluster stuff, here is a proposed `interface` that I have in my head:

```golang
type GitOpsDirectoryWriter interface {
	AddApplication(ctx context.Context, app models.Application, repo models.GitRepository) error
	AddCluster(ctx context.Context, cluster models.Cluster, repo models.GitRepository) error
}
```
A further idea where a some sort of `RepoWriter` service is passed as an argument:

```golang
type RepoWriter interface {
	Write(ctx context.Context, path string, file []byte) error
}

type GitOpsDirectoryWriter interface {
	AddApplication(ctx context.Context, app models.Application, w gitrepo.RepoWriter) error
	AddCluster(ctx context.Context, cluster models.Cluster, w gitrepo.RepoWriter) error
}
```

jpellizzari:
@palemtnrider Does this sufficiently capture the work that needs to be done for the ADR^^

palemtnrider:
Yes, that captures the majority of the work.  Couple questions - 

> * Create a single entry-point for read/write operations from the `.weave-gitops` directory (ie create a `WeGODirectoryClient` object or some such)
By this do you mean you want there to be a package/lib that handles the interaction with the directory structure?

jpellizzari:
> By this do you mean you want there to be a package/lib that handles the interaction with the directory structure?

@palemtnrider Yep. WDYT?

palemtnrider:
Additionally, we want to have AC for removing apps as this work isn't as straightforward in the new directory structure.

jpellizzari:
> Additionally, we want to have AC for removing apps as this work isn't as straightforward in the new directory structure.

Added

[...]

PR #845, opened 2021-10-06
Add beta commands with support for new directory layout

palemtnrider:
# Superseded by #882

Closes: #768 

**What changed?**
This adds a new command `beta` with sub-commands for adding applications and initializing clusters.  When initializing clusters - only external repositories  (GitOps repo) are supported.  Also, when initializing the GitOps runtime is persisted into the GitOps repo.  This allows users to recover a cluster and associated workloads by run install after the cluster is re-created.  

**Why?**
These changes are needed to solve the shortcomings and limitations of the existing GitOps repository layout.  

**How did you test it?**
Unit tests and manual execution

**Release notes**
The release notes will need to cover these changes.

**Documentation Changes**
Yes, we need to add documentation for the new beta command and its associated sub-commands.  Along with the new directory layout.

### DIFF
--- a/pkg/services/app/add.go
+++ b/pkg/services/app/add.go
@@ -325,7 +325,7 @@ func (a *App) addAppWithConfigInAppRepo(info *AppResourceInfo, params AddParams,
 	if params.Dir == "" {
 		a.Logger.Actionf("Cloning %s", info.Spec.URL)
 
-		remover, _, err := CloneRepo(a.ConfigGit, info.Spec.URL, info.Spec.Branch, params.DryRun)
+		remover, _, err := a.cloneRepo(a.ConfigGit, info.Spec.URL, info.Spec.Branch, params.DryRun)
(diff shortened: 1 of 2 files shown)


palemtnrider:
> After our parking lot discussion today, I talked over our strategy for `beta` commands with @JamWils. We think it is best to draw a line and say that "beta commands should NOT modify anything that is used by stable". This would mean it is okay for `beta` commands to copy-paste and modify any code that needs to be changed for a `beta` command to function.
> 
> @palemtnrider @luizbafilho Does that make sense and are you comfortable with keeping that constraint ^^ for this PR and going forward?
> 
> We may need to modify this PR to meet that constraint (I haven't read through code in detail).

I disagree with this approach and think further discussions are needed.

palemtnrider:
This PR from the forked repository has been **superseded by** #882

PR #820, opened 2021-09-29
GitProviders refactor

luizbafilho:
Closes: 

**What changed?**

**Why?**

**How did you test it?**

**Release notes**

**Documentation Changes**

PR #852, opened 2021-10-07
Fix chromedriver issues on CI

rokshana-b:
**What changed?**
Stabilize ChromeDriver version

**Why?**
ChromeDriver is causing failure on the CI


rokshana-b:
Issues in this PR were addressed and fixed in a different PR: https://github.com/weaveworks/weave-gitops/pull/820

Issue #786, opened 2021-09-21
Provide links to help and support


sympatheticmoose:
**Short description**
As a Weave GitOps user I want to easily find our documentation so that I can better understand features and troubleshoot if things go wrong.

I believe there is a lot more context help we could provide but good to start simple.

**Acceptance criteria**
- [ ] A link to https://docs.gitops.weave.works/ is available from all screens in the UI (i.e header/footer) under [Help] or [Docs]. 
- [ ] A reference is added to the CLI help text, i.e. on `wego help` An additional sentence in the description "To learn more, you can find our documentation at https://docs.gitops.weave.works/
-  [ ] STRETCH - Help could include a "Need commercial support?" with a link to sales@weave.works; in enterprise this could be replaced with links to both support@weave.works and the support portal.
- [ ] STRETCH - detects current version and generates appropriate URL for documentation version.

PR #874, opened 2021-10-12
Add footer with help links and add docs link to CLI help text

jpellizzari:
Closes: #786

Adds a footer component:
Screenshot from 2021-10-12 14-58-29

The footer is aware of the current version and links to the docs accordingly:
Screenshot from 2021-10-12 14-58-46

There is also a mechanism by which another project can turn off the `<Footer />` and/or substitute their own with a customized element via the`<Footer supportComponent={<MyElement />} />` prop.


jpellizzari:
@sympatheticmoose Changed to open links in a new tab

Issue #881, opened 2021-10-13
Beta commands for add app and install


palemtnrider:
**Short description**
As an app developer, I want to persist the gitops runtime so that the version of GitOps runtime is stored in a git repository and applications can be saved to the repository with the same repository to be deployed to my cluster.

**Acceptance criteria**
- [ ] existing commands
- [ ] all existing testing runs successfully (unit and acceptance)
- [ ] unit tests for the beta commands
- [ ] command help is available
- [ ] stories for follow on work are defined

**Design materials**

**TBD** define the changes to app add and install.

PR #882, opened 2021-10-13
Add beta commands with support for new directory layout

palemtnrider:
Closes: #881 

**What changed?**
This adds a new command `beta` with sub-commands for adding applications and initializing clusters.  When initializing clusters - only external repositories  (GitOps repo) are supported.  Also, when initializing the GitOps runtime is persisted into the GitOps repo.  This allows users to recover a cluster and associated workloads by run install after the cluster is re-created.  

**Why?**
These changes are needed to solve the shortcomings and limitations of the existing GitOps repository layout.  

**How did you test it?**
Unit tests and manual execution

**Release notes**
The release notes will need to cover these changes.

**Documentation Changes**
Yes, we need to add documentation for the new beta command and its associated sub-commands.  Along with the new directory layout.

### DIFF
--- a/pkg/git/git.go
+++ b/pkg/git/git.go
@@ -50,6 +50,12 @@ const WegoAppDir = "apps"
 // WegoClusterDir is where cluster information and manifests will live in the GitOps repo
 const WegoClusterDir = "clusters"
 
+// WegoClusterOSWorkloadDir is where OS workload manifests will live in the GitOps repo
+const WegoClusterOSWorkloadDir = "system"
+
+// WegoClusterUserWorloadDir is where user workload manifests will live in the GitOps repo
+const WegoClusterUserWorloadDir = "user"
+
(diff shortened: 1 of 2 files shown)


palemtnrider:
@jpellizzari   I'm really sorry, but I missed this comment - 
> Since we are moving to the new directory structure as of this PR (rather than via the beta command), I would like the implementation to match more closely with how we have spec'd it out in this ticket: #768 

Is it OK to put this in a follow-up story.

jpellizzari:
> @jpellizzari I'm really sorry, but I missed this comment -
> 
> > Since we are moving to the new directory structure as of this PR (rather than via the beta command), I would like the implementation to match more closely with how we have spec'd it out in this ticket: #768
> 
> Is it OK to put this in a follow-up story.

Works for me.

PR #892, opened 2021-10-15
Support for npm installing the ui-lib via github repo

foot:
**What changed?**

Adds a `prepare` step to the package.json [scripts section](https://docs.npmjs.com/cli/v7/using-npm/scripts#npm-install). npm/yarn will run `prepare` during `npm install weave-gitops` and so we can neatly bundle and generate typedefs in the consuming library's build process.

It also runs on `publish` which simplifies that process somewhat and we don't need to manually move files around.

**Why?**

Allows us to consume the weave-gitops ui library from git directly and from an arbitrary commit so that we can test changes more easily before they are merged to `main` or released.

**How did you test it?**

`npm publish --prepare` will now do the same thing that `make ui-lib` used to do.

**Caveats**

`prepare` is _also_ run on a vanilla `npm install` / `npm ci` (e.g. not only on `npm install weave-gitops` ([justification](https://stackoverflow.com/a/63293078)). This will slow down a few of the of the gh-action jobs. (10s on my macbook but gh-actions can sometimes be slow).

### DIFF
--- a/.github/workflows/frontend.yaml
+++ b/.github/workflows/frontend.yaml
@@ -47,4 +47,3 @@ jobs:
       - run: make ui-lint
       - run: make ui-test
       - run: make ui-audit
-      - run: make ui-lib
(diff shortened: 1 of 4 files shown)


foot:
This works after doing a `yarn cache clean` in the consumer but is a bit clunky w/ a few warnings around mixing yarn/npm.

foot:
Closing for now may explore more one day

Issue #612, opened 2021-08-16
Find and replace `wego-system` magic strings with a constant


jpellizzari:
Followup from #587

We have `wego-system` declared in various places throughout the code base. We should replace string-literals with the `wego.DefaultNamespace` constant where appropriate.

PR #896, opened 2021-10-15
Replace uses of "wego-system" with the wego.DefaultNamespace constant

jrryjcksn:
Closes: #612 

**What changed?**
We now use the `wego.DefaultNamespace` constant everywhere we mention the default namespace

**Why?**
Code cleanliness and maintainability

**How did you test it?**
Grep

**Release notes**
No

**Documentation Changes**
No

### DIFF
--- a/pkg/kube/kubehttp.go
+++ b/pkg/kube/kubehttp.go
@@ -45,7 +45,6 @@ func CreateScheme() *apiruntime.Scheme {
 	return scheme
 }
 
-const WeGONamespace = wego.DefaultNamespace
 const WeGOCRDName = "apps.wego.weave.works"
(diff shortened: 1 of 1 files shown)


jrryjcksn:
> There are a handful of other references to the hard-coded wego-system.
> 
>     * README.md - however this still says the command is `wego`
> 

Looking at that, we should probably get rid of the CLI reference in README.md altogether. It's way out-of-date and we now auto-generate the CLI reference in the docs dir. I'll bring that up at standup.

>     * application_types - has a DefaultNamespace const
> 

This is the definition we're using everywhere else; can't really get rid of this one :-) 

>     * role-binding.yaml

I didn't think generating this file programmatically was in scope for the story.

Issue #842, opened 2021-10-06
Spike: Determine how to break up acceptance tests to reduce wait time


jrryjcksn:
**Question**
What is the best way to divide up acceptance tests so that more of them can be run in parallel. Are there resource limits that constrain how we can break them apart?

**Short description**
Determine the degree to which we can break up tests and the best breakdown.

**Timebox**
8 hours
---

**Outcome**
I was able to significantly reduce acceptance test runtime by assigning a new label to approx. half the tests and creating an additional workflow check that selectively runs them. I also excluded smoke tests from the acceptance tests since they always run for every PR already. Ginkgo allows both `--focus=<regexp>` and `--skip=<regexp>` flags on the command line. For one of the checks, I used:
`ginkgo --skip='KustomizeTest|SmokeTest'` and for the other:
`ginkgo --focus='KustomizeTest'`

![image.png](https://images.zenhubusercontent.com/5db83c9473e8f70001d9f27a/66a770f1-87b7-4266-9bde-b0358bd071f7)

PR with experiment: https://github.com/weaveworks/weave-gitops/pull/930

PR #930, opened 2021-10-20
Split up acceptance tests into multiple checks

jrryjcksn:
Closes: 
- #842 

**What changed?**
- Tagged all "Kustomize" add tests (that aren't smoke tests)
- Created second acceptance test check to run only the Kustomize tagged tests
- Updated existing acceptance test check to skip Kustomize and Smoke tagged tests

**Why?**
Breaking the tests up allows them to be run in parallel as different checks so we only have to wait for the longest one (at this point ~24 minutes)

**How did you test it?**
The PR you are reading

**Release notes**
No

**Documentation Changes**
No

### DIFF
--- a/test/acceptance/test/add_tests.go
+++ b/test/acceptance/test/add_tests.go
@@ -21,6 +21,7 @@ import (
 var clusterName string
 
 var _ = Describe("Weave GitOps App Add Tests", func() {
+
 	deleteWegoRuntime := false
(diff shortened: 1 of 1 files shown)


rokshana-b:
>get all the configuration settings shared between the two jobs, then 👍 👍
>It's only a matter of time before someone runs into the "forgot to change an env setting in two places"

Agreed

Issue #911, opened 2021-10-18
Refactor `get` and `list` commands to use new v1 architecture


jpellizzari:
Our v1 architecture ADR is merged: https://github.com/weaveworks/weave-gitops/blob/main/doc/adr/0009-v1-architecture.md

We should begin the move to this architecture with the simplest use case possible, which would be the `get` and `list` commands. The work we do on the simple use cases should be built in such a way as to expedite development of more complicated use cases in the future.

Acceptance criteria:
* Acceptance tests should not require major changes
* Follows the `models`, `services`, `adapters` pattern(s) laid out in the ADR
* Contains unit test tests with a high level of coverage

PR #975, opened 2021-10-27
Add new App Fetcher

J-Thompson12:
Closes: #911 

**What changed?**
Refactors get and list to the new standards. List is wired up and being used but I was unable to remove the old code until add is refactored because the old code is being used there. Get is not being used yet because it needs others to be refactored that use that function.

**Why?**

**How did you test it?**
Added unit tests and tested manually

**Release notes**

**Documentation Changes**


pkg/server/server_test.go
```diff
@@ -774,15 +778,22 @@ var _ = Describe("ApplicationsServer", func() {
 
 			})
 			It("logs server errors", func() {
-				ts := httptest.NewServer(httpHandler)
-				defer ts.Close()
+				k8s := fake.NewClientBuilder().Build()
+				cfg := ApplicationsConfig{
+					AppFactory: appFactory,
```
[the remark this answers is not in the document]
J-Thompson12: @jpellizzari  So this comment is actually on old code. I am not passing in appFactory or k8s. I am only setting the cfg with log and fakeFetcher.

pkg/models/application.go
```diff
@@ -0,0 +1,27 @@
...
+)
+
+type SourceType string
+
+const (
+	SourceTypeGit  SourceType = "git"
+	SourceTypeHelm SourceType = "helm"
+)
+
+type Application struct {
```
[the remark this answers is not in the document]
jpellizzari: @jrryjcksn I saw a duplicate comment in the timeline here so I deleted it, and it appears to have removed BOTH comments. Here is what I was trying to say (from memory):

> If there are no objections, I can do that part in my stuff; I have to obtain those paths somewhere in order to do the manifest generation anyway.

I do have objections to bringing path writing stuff into `models.Application`. I would expect `WeGODirectoryWriter` to accept a `models.Application` as an argument and write it to the correct directory structure.

Sorry about the confusion there, I wasn't trying to delete my comment.

pkg/services/applicationv2/fetcher.go
```diff
@@ -0,0 +1,107 @@
...
+}
+
+func translateApp(app wego.Application) (models.Application, error) {
+	var (
+		appRepoUrl    gitproviders.RepoURL
+		configRepoUrl gitproviders.RepoURL
+		err           error
+	)
+
+	if wego.DeploymentType(app.Spec.SourceType) == wego.DeploymentType(wego.SourceTypeGit) {
```
jrryjcksn, line 82: Should these be `app.Spec.DeploymentType` and `wego.DeploymentTypeGit`?

pkg/models/application.go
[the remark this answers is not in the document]
jrryjcksn: I didn't want any path writing stuff in the model. I just wanted the model to be able to return what the paths are... The writing would still be separate. The paths are just derived from the model data. Also, it could be fine to move that into the manifest generation (which also needs to know what the paths are).

pkg/models/application.go
```diff
@@ -0,0 +1,43 @@
...
+	Name                string
+	Namespace           string
+	SourceURL           gitproviders.RepoURL
```
jrryjcksn, line 31: This won't work in all cases because we have helm repository sources which have no git provider info. The RepoURL creation will fail.

pkg/services/applicationv2/fetcher.go
```diff
@@ -0,0 +1,107 @@
...
+
+	a, err := translateApp(*app)
+	if err != nil {
+		return models.Application{}, err
+	}
+
+	return a, nil
+}
+
+func (f fetcher) List(ctx context.Context, namespace string) ([]models.Application, error) {
```
JamWils, line 54: Minor: namespace isn't used in this function remove it as a parameter.

jpellizzari:
> `kube.Client` is already an interface. I would just create a fake for it in `vendorfakes`. Is there value in them? I would argue yes, especially for `Get` since you are handling the error in different ways depending on what you get back.

FYI `controller-runtime` provides a fake `client.Client` interface: https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.10.2/pkg/client/fake

pkg/services/applicationv2/fetcher.go
J-Thompson12: Well I copied this over from `getAppResourceInfo` so if its wrong I should change it both places.

pkg/models/application.go
jpellizzari: Good catch. From standup, my suggestion would be a new struct field on `models.Application` called `HelmChartURL` instead of trying to fit both `Git` and `Helm` URLs on the same field.

pkg/services/applicationv2/fetcher.go
jpellizzari: I think it needs to be used :thinking:

J-Thompson12: Yeah I have added it in and added a test to cover it.

pkg/models/application.go
J-Thompson12: Added a HelmSourceURL and changed SourceURL to GitSourceURL.

J-Thompson12:
@jpellizzari the function shared in FYI https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.10.2/pkg/client/fake is actually deprecated and recommends using NewClientBuilder like I am currently doing.

pkg/server/server_test.go
```diff
@@ -913,16 +921,25 @@ var _ = Describe("ApplicationsServer", func() {
...
-				ts := httptest.NewServer(httpHandler)
-				defer ts.Close()
-
 				errMsg := "there was a big problem"
-
+				fakeFetcher := &applicationv2fakes.FakeFetcher{}
 				// Pretend something went horribly wrong
-				kubeClient.GetApplicationsStub = func(c context.Context, s string) ([]wego.Application, error) {
-					return nil, errors.New(errMsg)
+				fakeFetcher.ListStub = func(c context.Context, s string) ([]models.Application, error) {
```
jpellizzari, line 927: :+1: Much more explicit unhappy-path simulation!

Nit: We recently found a slightly cleaner way of doing `fake` returns

```suggestion
				fakeFetcher.ListReturns([]models.Application{}, errors.New(errMsg))
```

pkg/services/applicationv2/fetcher_test.go
```diff
@@ -0,0 +1,101 @@
...
+			Expect(k8s.Create(ctx, app2)).To(Succeed())
+
+			gs := NewFetcher(k8s)
+
+			result, err := gs.List(ctx, app.Namespace)
+			Expect(err).NotTo(HaveOccurred())
+
+			Expect(len(result)).To(Equal(2))
+		})
+		It("lists does not list an application in a different namespace", func() {
```
jpellizzari, line 67: :+1: * :100:

pkg/models/application.go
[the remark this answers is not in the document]
jpellizzari: @jrryjcksn 

>  I just wanted the model to be able to return what the paths are

To me, that is inherently "path stuff" and only relevant in a context where we are trying to write an `Application` to a repo. I have a feeling you and I mostly agree on this after our chat today.

Can I follow up with you outside of this PR so Justin can get this thing merged? If yes, I will mark this as resolved.

jpellizzari (approved):
LGTM. First PR of the new architecture! Nice work.

pkg/models/application.go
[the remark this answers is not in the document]
jpellizzari: Chatted with Jerry and we are :+1:

jrryjcksn (approved):
:+1:

PR #1007, opened 2021-11-02
Moved middleware to pkg/server instead to avoid conflict with referen…

josecordaz:
…ce inside integration folder

Closes: 

**What changed?**

**Why?**

**How did you test it?**

**Release notes**

**Documentation Changes**


jpellizzari (approved):
I disagree with the premise of the original ticket (cc @JamWils), but the AC was to NOT have middleware accessible outside of the `server` packaage. That said, I don't think this should be in `internal`.

The middleware was split into it's own package so other HTTP server implementations could share the middleware behavior. We will need to move it back out again if we ever add a new HTTP service.

In any case, this change is fine for now as it removes another top-level `pkg` directory.

PR #1009, opened 2021-11-02
Add flag validation for WGE commands

yiannistri:
Closes: weaveworks/weave-gitops-enterprise#207

**What changed?**
- Refactor root command to allow for easier unit testing
- Set `SilenceUsage`/`SilenceErrors` for all WGE commands to `true`
- Added validation for endpoint and provider

**Why?**
To improve errors returned to the user

**How did you test it?**
Local testing + new unit tests

**Release notes**
N/A

**Documentation Changes**
N/A


pkg/wegoerrors/errors.go
```diff
@@ -0,0 +1,5 @@
+package wegoerrors
+
+import "errors"
+
+var ErrWGEHTTPApiEndpointNotSet = errors.New("the Weave GitOps Enterprise HTTP API endpoint flag (--endpoint) has not been set")
```
bigkevmcd, line 5: I suspect you could shorten this to just `ErrWGEEndpointNotSet` and maybe even `ErrNoWGEEndpoint` which would be more in keeping with the Go stdlib https://pkg.go.dev/net/http#pkg-variables and https://pkg.go.dev/io#pkg-variables

bigkevmcd: Oh, and I think this should be in the `cmd` package ?

Because it explicitly refers to a command-line option `--endpoint)` ?

foot:
Does this change any behaviour in the commands like `gitops add app` etc? 🤔

yiannistri:
I don't think it does as only the WGE commands were updated 🤔

pkg/wegoerrors/errors.go
yiannistri: Renamed and moved to `cmd` package.

cmd/gitops/cmderrors/errors.go
```diff
@@ -0,0 +1,5 @@
+package cmderrors
```
jpellizzari, line 1: :+1:

jpellizzari (approved):

foot (approved):
💯

Issue #589, opened 2021-08-10
`testspace-analytics` job shows failed on main


jpellizzari:

jpellizzari:
@iahmad9 Can you take a look at this?

PR #1019, opened 2021-11-03
setup kubebuilder before running unit tests.

iahmad9:
Closes: #589 

**What changed?**
Updated the deploy job to set the dependencies needed before running the unit tests, mainly kubebuilder.

**Why?**
Publishing the test results were failing.

**How did you test it?**
By running the deploy job on branch

**Release notes**

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -1,7 +1,7 @@
 on:
   push:
-    branches:
-      - main
+  #  branches:
+  #    - main
   workflow_dispatch:
 
 name: Deploy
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy.yml
```diff
@@ -1,7 +1,7 @@
 on:
   push:
-    branches:
-      - main
+  #  branches:
```
jpellizzari: This should only run on `main` correct?

jpellizzari (approved):
Thanks for looking at this!

PR #1023, opened 2021-11-03
Clarify the description of "namespace" in the UI and CLI

### LINKED ISSUE #898: Clearer guidance on namespace field in the UI for adding applications and --namespace flag in the CLI
The `--namespace` option within our CLI (and exposed through adding an application in the web UI) can be confusing for users who believe this will impact where their workloads (which they may associate as applications) will be deployed. 

Additionally, our description in the global help text implies that it is **only** where the GitOps Runtime is installed, when it can include where the GitOps Automation (source/kustomization etc) for an Application is stored through an `app add`. 

Would welcome suggestions or feedback on proposed changes - 
- CLI help text:`the namespace scope for this operation (default "wego-system").` <- matches Flux help text, 
- Web UI caption: `The namespace where GitOps automation objects will be stored.`

cc @bia @jpellizzari

joshri:
Closes #898 

Changes the description of the namespace field in the UI (ApplicationAdd), as well as the --namespace flag in gitops/root + README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,7 +49,7 @@ Available Commands:
 
 Flags:
   -h, --help               Help for gitops
-      --namespace string   The namespace scope for this operation (default wego-system).
+      --namespace string   The namespace scope for this operation (default "wego-system").
   -v, --verbose            Enable verbose output
 
 Use "gitops [command] --help" for more information about a command.
--- a/README.md
+++ b/README.md
@@ -49,7 +49,7 @@ Available Commands:
 
 Flags:
   -h, --help               Help for gitops
-      --namespace string   The namespace scope for this operation (default "wego-system").
+      --namespace string   The Kubernetes namespace in which Weave GitOps objects will be stored (default "wego-system").
   -v, --verbose            Enable verbose output
 
 Use "gitops [command] --help" for more information about a command.
--- a/README.md
+++ b/README.md
@@ -49,7 +49,7 @@ Available Commands:
 
 Flags:
   -h, --help               Help for gitops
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -49,7 +49,7 @@ Available Commands:
 
 Flags:
   -h, --help               Help for gitops
-      --namespace string   Weave GitOps runtime namespace (default "wego-system")
+      --namespace string   The namespace scope for this operation (default "wego-system").
```
jpellizzari, line 52: This is a change I did in a branch. Trying to be super specific about what this flag does.

```suggestion
      --namespace string   The Kubernetes namespace in which Weave GitOps objects will be stored (default "wego-system").
```

jpellizzari (commented):
Check out my suggestion and if you like it, let's use it everywhere. This highlights a need for some sort of application text catalog.

README.md
sympatheticmoose: @jpellizzari specific can be tricky... doing a `get/status` or `remove` isn't a "will be", it's "the objects are here already"

jpellizzari: @sympatheticmoose Not sure what you mean. We are open to suggestions on the text itself.

sympatheticmoose: I gave my suggestion already in the issue ;) I'm saying "will be stored" implies that this is always an action where I am about to store objects which don't yet exist into the given namespace. Which isn't the case for read or remove operations.

jpellizzari: OK I missed that. @joshri ^^

joshri: Got it 🌞

J-Thompson12 (approved):
LGTM

Issue #810, opened 2021-09-28
Deploying podinfo on second cluster throws failure: "kustomize build failed: accumulating resources: accumulation err='merging resources from './.wego/targets/.../podinfo-deploy-gitops-deploy.yaml'"


talaverant:
**Describe the bug**
When trying to deploy podinfo app on second cluster (EKS), I'm facing this failure:
kustomize build failed: accumulating resources: accumulation err='merging resources from './.wego/targets/minikube/podinfo-deploy/podinfo-deploy-gitops-deploy.yaml': may not add resource with an already registered id: source.toolkit.fluxcd.io_v1beta1_GitRepository|wego-system|podinfo-deploy': got file 'podinfo-deploy-gitops-deploy.yaml', but '/tmp/podinfo-deploy956763450/.wego/targets/minikube/podinfo-deploy/podinfo-deploy-gitops-deploy.yaml' must be a directory to be a root	        	False    

**Environment**
 - gitops: [e.g. v0.3.0]
 - kubernetes:  [e.g. 1.20]
    - [ ] cloud [EKS]  
 - Browser + version: [e.g. chrome 91.0.4472.101]
 -
**Affects versions**
v0.2.0 and v0.3.0

**To Reproduce**
(kubectl pointing at cluster a) gitops install
(kubectl pointing at cluster a) gitops app add . (in a local clone of a forked copy of podinfo-deploy)
(kubectl pointing at cluster B) gitops install
(kubectl pointing at cluster B) gitops app add . (same clone as before)

**Expected behavior**
Capacity to deploy an app into multiple clusters without failures

**Config and Logs**
Please let me know which logs are needed and how to get them.

**Screenshots**
<img width="1380" alt="Screenshot 2021-09-27 at 14 29 21" src="https://user-images.githubusercontent.com/53871388/135074105-0fdfef1e-7f47-4b50-9e50-ad80a1435e02.png">
<img width="1380" alt="Screenshot 2021-09-27 at 15 16 06" src="https://user-images.githubusercontent.com/53871388/135074124-ad4fe0dc-2131-4846-9610-7438223c5a43.png">
<img width="1373" alt="Screenshot 2021-09-28 at 08 52 31" src="https://user-images.githubusercontent.com/53871388/135074128-4f13c69e-1b36-46d5-bad3-e7a7b118e173.png">
[...]

PR #1024, opened 2021-11-04
768 new directory structure

jrryjcksn:
Closes: 
- #768 
- #810 

**What changed?**
- New directory structure meaning:
  - No "automation automation" created any more (handled by "user kustomization" created by install)
  - No manifests written directly to the cluster (only the "GORT" set up by install)
  - Install requires a repo in which to create the directory structure and store the GORT
  - No more `config=NONE`
  - All added apps must reference a config repo containing the directory structure. The "config in app repo" case is now just the same as the external repo case except that the config URL is the same as the source URL. You can still leave the repo URL unspecified; it will just end up being set to the source URL.
  - Every config repo contains a `.weave-gitops` directory (no more `.wego` nor config repo without a special dir)
- Add/remove broken up into new services:
  - GitOpsDirWriter service that has `AddApplication` and `RemoveApplication` commands that use:
  - Automation service that generates all manifests
  - RepoWriter service that creates PRs and writes / commits / pushes to git providers
- The new `models.Application` created by @J-Thompson12 is used everywhere

**Why?**
It's part of the product plan

**How did you test it?**
Unit, integration, acceptance tests

**Release notes**
Yes, it will change the way the app is used and what the user's repo will look like

**Documentation Changes**
Yes. I will pursue the doc change separately.


jrryjcksn:
A note. The `RepoWriter` operations don't yet have their own tests. They were always tested indirectly via add/remove before (and still are) but I want to circle back and add explicit tests for that service.

pkg/services/gitrepo/service.go
```diff
@@ -0,0 +1,128 @@
...
+	"io/ioutil"
+	"os"
+
+	"github.com/weaveworks/weave-gitops/pkg/git"
+	"github.com/weaveworks/weave-gitops/pkg/gitproviders"
+	"github.com/weaveworks/weave-gitops/pkg/logger"
+	"github.com/weaveworks/weave-gitops/pkg/services/automation"
+)
+
+type RepoWriter interface {
```
jpellizzari, line 15: Digging this :+1:. Much more general and easy for us to consume in the future.

pkg/services/gitrepo/service.go
```diff
@@ -0,0 +1,128 @@
...
+
+func (rw *RepoWriterSvc) CloneRepo(ctx context.Context, branch string) (func(), string, error) {
+	return CloneRepo(ctx, rw.GitClient, rw.URL, branch)
+}
+
+func (rw *RepoWriterSvc) CommitAndPush(ctx context.Context, commitMsg string, filters ...func(string) bool) error {
+	return CommitAndPush(ctx, rw.GitClient, commitMsg, rw.Logger, filters...)
+}
+
+func CommitAndPush(ctx context.Context, client git.Git, commitMsg string, logger logger.Logger, filters ...func(string) bool) error {
```
jpellizzari, line 75: Question: Why split this into it's own function? Is it meant to be used outside of a `RepoWriter`?

test/integration/server/add_test.go
```diff
@@ -86,36 +86,40 @@ var _ = Describe("AddApplication", func() {
...
 				},
+				Ignore: helpers.GetIgnoreSpec(),
 			}
 
 			expectedApp := wego.ApplicationSpec{
 				URL:            req.Url,
 				Branch:         req.Branch,
 				Path:           req.Path,
-				ConfigURL:      "",
+				ConfigURL:      req.Url,
```
jpellizzari, line 122: Question: so we automatically populate this for the user if they don't specify a `ConfigURL`, correct?

test/integration/server/helpers/helpers.go
```diff
@@ -94,40 +97,50 @@ func NewGithubClient(ctx context.Context, token string) *ghAPI.Client {
...
 
-func clusterPath(root, appName, clusterName, filename string) string {
-	return wegoPath(root, filepath.Join("targets", clusterName, appName, filename))
+func automationPath(root, appName string) string {
+	return appPath(root, appName, fmt.Sprintf("%s-gitops-deploy.yaml", appName))
 }
 
-func automationPath(root, appName, clusterName string) string {
-	return clusterPath(root, appName, clusterName, fmt.Sprintf("%s-gitops-deploy.yaml", appName))
+func sourcePath(root, appName string) string {
```
jpellizzari, line 115: It probably makes sense to migrate these to the new ones that I saw in the `/automation` package. Not necessary for this PR, I will create a follow up issue.

test/integration/server/helpers/helpers.go
```diff
@@ -160,6 +173,32 @@ func GenerateExpectedFS(req *pb.AddApplicationRequest, root, clusterName string,
...
+				Name:      req.Name,
+				Namespace: req.Namespace,
+			},
+			Resources: []string{
+				"app.yaml",
+				fmt.Sprintf("%s-gitops-deploy.yaml", req.Name),
+				fmt.Sprintf("%s-gitops-source.yaml", req.Name),
+			},
+		},
+		userKustPath(root, clusterName): &types.Kustomization{
```
jpellizzari, line 191: :+1:

pkg/services/gitopswriter/service.go
```diff
@@ -0,0 +1,207 @@
+package gitopswriter
```
jpellizzari, line 1: Nit: we can probably just name this file `pkg/services/gitopswriter/gitopswriter.go`, for the sole purpose of making it easier to find when fuzzy-searching in an editor.

Not a big deal either way though.

jpellizzari (commented):
Overall approach looks great! My only blockers are around the unit testing strategy (they are noted with "Blocker")

pkg/services/automation/automation_test.go
```diff
@@ -0,0 +1,354 @@
...
+					// A straight string/byte comparison doesn't account for un-ordered keys in yaml.
+					Expect(yaml.Unmarshal(out.Content, &result))
+
+					diff := cmp.Diff(result, myApp)
+					Expect(diff).To(Equal(""))
+
+					// Not entirely sure how to get gomega to pretty-print the output of `diff`,
+					// so we assert it should be empty above, and conditionally print the diff to make a nice assertion message.
+					// `diff` is a formatted string
+					if diff != "" {
```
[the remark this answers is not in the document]
jrryjcksn: Indeed. This test was carried over :-)

pkg/services/automation/service.go
```diff
@@ -0,0 +1,494 @@
...
+
+	ResourceKindApplication    ResourceKind = "Application"
+	ResourceKindSecret         ResourceKind = "Secret"
+	ResourceKindGitRepository  ResourceKind = "GitRepository"
+	ResourceKindHelmRepository ResourceKind = "HelmRepository"
+	ResourceKindKustomization  ResourceKind = "Kustomization"
+	ResourceKindHelmRelease    ResourceKind = "HelmRelease"
+)
+
+type AutomationService interface {
```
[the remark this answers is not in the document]
jrryjcksn: Okay

test/integration/server/helpers/helpers.go
jrryjcksn: Yeah, I wasn't sure about that. I've heard people complain in the past that using the same definitions as the ones used to generate can mask a bug. I'm mostly not worried about that but I decided not to make that change yet anyway.

pkg/services/gitopswriter/service.go
jrryjcksn: Another one where I wasn't sure what would be better... On the one hand, your point is well taken. On the other hand, this way people browsing through the directory know where the main service logic will be found :man_shrugging:

pkg/services/app/app.go
```diff
@@ -77,25 +77,25 @@ func (a *App) getDeploymentType(ctx context.Context, name string, namespace stri
 	return wego.DeploymentType(app.Spec.DeploymentType), nil
 }
 
-func (a *App) getSuspendedStatus(ctx context.Context, name, namespace string, deploymentType wego.DeploymentType) (bool, error) {
-	var automation client.Object
+func (a *AppSvc) getSuspendedStatus(ctx context.Context, name, namespace string, deploymentType wego.DeploymentType) (bool, error) {
+	var automationClient client.Object
```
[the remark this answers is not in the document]
jrryjcksn: I had to change the name because it conflicted with the `automation` package and landed on`automationClient` since it was a `client.Object`. I'm happy to change it.

pkg/services/app/remove.go
```diff
@@ -39,148 +36,18 @@ func (a *App) Remove(params RemoveParams) error {
...
-			manifests = append(manifests, manifest)
-		}
-
-		return nil
-	})
-
-	if err != nil {
-		return nil, err
-	}
+func (a *AppSvc) removeApp(ctx context.Context, app models.Application, clusterName string, autoMerge bool) error {
```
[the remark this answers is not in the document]
jrryjcksn: I mostly did that for two reasons:
- Consistency with `Add` (do the setup and then hand off)
- It leaves the actual "remove" operation as something that could be easily mocked / manipulated so someone could just look in the one place and see what needed to be done to invoke the `RemoveApplication` method.

I don't feel particularly strongly about it, though.

pkg/services/automation/automation_test.go
```diff
@@ -0,0 +1,354 @@
...
+	Context("add app with config in app repo", func() {
+		BeforeEach(func() {
+			app.GitSourceURL = createRepoURL("ssh://git@github.com/foo/bar.git")
+			app.ConfigURL = emptyRepoURL
+		})
+
+		Describe("generates source manifest", func() {
+			It("creates GitRepository when source type is git", func() {
+				app.SourceType = models.SourceTypeGit
+				_, err := automationSvc.GenerateAutomation(ctx, app, "test-cluster")
```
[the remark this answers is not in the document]
jrryjcksn: That's fair, but just FYI... I didn't really write those tests. Those are all the same tests we had before except that they take a `model.Application` and invoke `automation.GenerateAutomation` rather than `AddParams` and `app.Add`. Example:

```
		Describe("generates source manifest", func() {
			It("creates GitRepository when source type is git", func() {
				addParams.SourceType = wego.SourceTypeGit
				err := appSrv.Add(addParams)

				Expect(fluxClient.CreateSourceGitCallCount()).To(Equal(1))

				name, url, branch, secretRef, namespace := fluxClient.CreateSourceGitArgsForCall(0)
				Expect(name).To(Equal("bar"))
				Expect(url.String()).To(Equal("ssh://git@gitlab.com/group/subgroup/bar.git"))
				Expect(branch).To(Equal("main"))
				Expect(secretRef).To(Equal("wego-test-cluster-bar"))
				Expect(namespace).To(Equal(wego.DefaultNamespace))
			})
		})
```

I figured just migrating them would be okay given the huge set of changes that are already in here. WDYT about redoing the tests in a separate PR w/ the `RepoWriter` tests? If that causes too much heartburn, I can do it now.

pkg/services/gitrepo/service.go
jrryjcksn: It was already being used in `gitops/install.go` to do the initial repo setup

test/integration/server/add_test.go
jrryjcksn: Yes. That happens in `app.Add` when we validate / update the parameters.

pkg/services/gitopswriter/service.go
```diff
@@ -0,0 +1,207 @@
...
+
+	if err = dw.RepoWriter.Write(ctx, kManifest.Path, kManifest.Content); err != nil {
+		return fmt.Errorf("failed to write updated kustomize file: %w", err)
+	}
+
+	return dw.RepoWriter.CommitAndPush(ctx, RemoveCommitMessage)
+}
+
+func addKustomizeResources(app models.Application, repoDir, clusterName string, resources ...string) (automation.AutomationManifest, error) {
+	userKustomizationRepoPath := filepath.Join(git.WegoRoot, git.WegoClusterDir, clusterName, "user", "kustomization.yaml")
```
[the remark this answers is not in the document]
jrryjcksn: :+1:

pkg/services/gitopswriter/writer_add_test.go
```diff
@@ -0,0 +1,265 @@
...
+			It("writes the files to the disk", func() {
+				fluxClient.CreateSourceGitReturns(dummyGitSource, nil)
+				fluxClient.CreateKustomizationReturns([]byte("kustomization"), nil)
+
+				err := gitOpsDirWriter.AddApplication(ctx, app, "test-cluster", true)
+				Expect(err).ShouldNot(HaveOccurred())
+
+				Expect(gitClient.WriteCallCount()).To(Equal(5))
+
+				path, content := gitClient.WriteArgsForCall(0)
```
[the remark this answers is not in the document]
jrryjcksn: I see that two of the files aren't being checked (only the original three are) but if they were, wouldn't it be okay to check that the git client writes: <data> to <path> rather than having it write through to an actual git repo?

pkg/services/automation/automation_test.go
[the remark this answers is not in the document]
jpellizzari: @jrryjcksn Ok them I am cool with migrating them for now and "fixing" them later :+1:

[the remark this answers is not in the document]
jrryjcksn: I've been looking at this some more... Right now we're checking that we pass the right args to the `flux` commands for GitRepositories, HelmRepositories, Kustomizations, and HelmReleases. If we check the return value from the generation, we'll be fragile in the face of changes in the manifests that flux generates. Is that what we want to check here? I can make sure that the things we generate ourselves (app.yaml, app kustomize, and user kustomize) are correct but maybe we shouldn't be explicitly checking the output of flux here for the things _it_ generates :thinking:

pkg/services/gitopswriter/gitopswriter.go
```diff
@@ -0,0 +1,211 @@
...
+	defer remover()
+
+	kManifest, err := addKustomizeResources(app, repoDir, clusterName, appKustomizeReference(app))
+	if err != nil {
+		return err
+	}
+
+	manifests = append(manifests, kManifest)
+
+	if !autoMerge {
```
luizbafilho, line 63: minor: invert the logic to make easier to read

```go
if autoMerge {
```

jrryjcksn: :+1:

pkg/services/app/add.go
```diff
@@ -314,769 +207,57 @@ func (a *App) updateParametersIfNecessary(ctx context.Context, params AddParams)
...
+			return models.Application{}, err
 		}
 	}
 
-	if models.IsExternalConfigUrl(app.Spec.ConfigURL) {
-		configRepoUrl, err = gitproviders.NewRepoURL(app.Spec.ConfigURL)
+	var configURL gitproviders.RepoURL
+
+	switch strings.ToUpper(params.AppConfigUrl) {
+	case string(models.ConfigTypeNone):
```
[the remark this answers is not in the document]
jrryjcksn: Sure, that makes sense.

pkg/services/gitopswriter/writer_add_test.go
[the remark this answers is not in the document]
jpellizzari: Assuming these are existing tests, we don't need to fix it here. Blocker removed for me.

To answer the question: since the `It` block says "writes the files to disk", we should be able to check their contents. Unless this is going over the network?

test/integration/server/add_test.go
```diff
@@ -14,15 +14,15 @@ import (
 	. "github.com/onsi/gomega"
 
 	"github.com/fluxcd/go-git-providers/gitprovider"
-	kustomizev2 "github.com/fluxcd/kustomize-controller/api/v1beta2"
+	kustomizev1 "github.com/fluxcd/kustomize-controller/api/v1beta2"
```
jpellizzari, line 17: :+1:

test/integration/server/helpers/helpers.go
jpellizzari: Interesting take. I think I like that, since updating this test would indicate a breaking change for the user.

pkg/services/gitopswriter/writer_add_test.go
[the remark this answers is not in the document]
jpellizzari: On second look, are these existing tests that were moved @jrryjcksn ?

pkg/services/gitopswriter/gitopswriter.go
```diff
@@ -0,0 +1,213 @@
...
+)
+
+var _ GitOpsDirectoryWriter = &GitOpsDirectoryWriterSvc{}
+
+type GitOpsDirectoryWriter interface {
+	AddApplication(ctx context.Context, app models.Application, clusterName string, autoMerge bool) error
+	RemoveApplication(ctx context.Context, app models.Application, clusterName string, autoMerge bool) error
+}
+
+type GitOpsDirectoryWriterSvc struct {
```
jpellizzari, line 32: Nit: does this need to be exported? The constructor seems to encapsulate everything we need to add to the struct.

jpellizzari (approved):
LGTM. It's tough to tell what represents a new test and what was moved, so we can settle on a testing strategy in a follow up.

I would expect the `RepoWriter` to check the contents of a local repo against an expected file system, similar to how the `integration/server` tests work.

pkg/services/gitopswriter/gitopswriter.go
jrryjcksn: Good point.

jrryjcksn: BTW, GitHub doesn't want to let me reply to your question above about the moved tests. This is what I tried to say:

Yes, they were moved but the invocation was changed from calling app.Add to calling the gitopswriter.AddApplication directly. Only the tests that check parameter validation (and the original "new directory structure" tests) are left in add_test.go in app. The tests that were purely concerned with manifest generation moved to the automation service and the ones that tested "adding" but not add parameters got moved to the gitopswriter.

palemtnrider (commented):
Here are a few comments so far.  I haven't completed my review yet but wanted to share what I have so far.

cmd/gitops/install/cmd.go
```diff
@@ -44,32 +52,55 @@ If a previous version is installed, then an in-place upgrade will be performed.`
...
+	flux := flux.New(osysClient, &runner.CLIRunner{})
 
 	k, _, err := kube.NewKubeHTTPClient()
 	if err != nil {
 		return fmt.Errorf("error creating k8s http client: %w", err)
 	}
 
+	factory := services.NewFactory(flux, log)
+	providerClient := internal.NewGitProviderClient(osysClient.Stdout(), osysClient.LookupEnv, auth.NewAuthCLIHandler, log)
+
```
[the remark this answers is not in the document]
jrryjcksn: Sure.

pkg/models/application.go
```diff
@@ -39,6 +36,5 @@ type Application struct {
 }
 
 func IsExternalConfigUrl(url string) bool {
-	return strings.ToUpper(url) != string(ConfigTypeNone) &&
-		strings.ToUpper(url) != string(ConfigTypeUserRepo)
+	return strings.ToUpper(url) != string(ConfigTypeUserRepo)
```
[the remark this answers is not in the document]
jrryjcksn: Can do. I just didn't redo all this after the "NONE" case was removed.

pkg/osys/osys.go
```diff
@@ -40,6 +49,27 @@ func (o *OsysClient) Unsetenv(envVar string) error {
...
 }
 
+func (o *OsysClient) ReadDir(dirName string) ([]os.DirEntry, error) {
+	return os.ReadDir(dirName)
+}
+
+// The following three functions are used by both "add app" and "delete app".
+// They are here rather than in "utils" so they can use the (potentially mocked)
+// local versions of UserHomeDir, LookupEnv, and Stdin and so that they can also
+// be mocked (e.g. we might want to mock the private key password handing).
```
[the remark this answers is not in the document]
jrryjcksn: Yes, this was left over from quite a while ago

pkg/services/app/add.go
```diff
@@ -304,7 +193,7 @@ func (a *App) updateParametersIfNecessary(ctx context.Context, gitProvider gitpr
 		}
 	}
 
-	if nameTooLong(params.Name) {
+	if automation.ApplicationNameTooLong(params.Name) {
 		return params, fmt.Errorf("application name too long: %s; must be <= 63 characters", params.Name)
```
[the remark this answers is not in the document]
jrryjcksn: K

pkg/services/app/app_suite_test.go
```diff
@@ -46,7 +49,10 @@ var _ = BeforeEach(func() {
 		},
 	}
 
-	appSrv = New(context.Background(), &loggerfakes.FakeLogger{}, fluxClient, kubeClient)
+	appSrv = New(context.Background(), &loggerfakes.FakeLogger{}, fluxClient, kubeClient, osysClient)
```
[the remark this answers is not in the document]
jrryjcksn: :heavy_check_mark: Yep, that came in during a rebase

pkg/services/app/remove.go
```diff
@@ -40,147 +38,18 @@ func (a *App) Remove(configGit git.Git, gitProvider gitproviders.GitProvider, pa
...
-			return gitproviders.RepoURL{}, "", err
-		}
-	}
-
-	return repoUrl, branch, nil
-}
-
-func clusterDeleteError(name string, err error) error {
-	return fmt.Errorf("failed to delete resource: %s with error: %w", name, err)
+	return a.removeApp(ctx, configGit, gitProvider, app, clusterName, true)
```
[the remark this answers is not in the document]
jrryjcksn: We're going to add PR remove very soon. I guess I can say that in a comment :man_shrugging:

pkg/services/automation/automation_test.go
```diff
@@ -0,0 +1,453 @@
...
+				appManifest, nonApps := extractApp(app, results)
+				Expect(len(nonApps)).To(Equal(len(results) - 1))
+
+				wegoApp := AppToWegoApp(app)
+				wegoApp.ObjectMeta.Labels = map[string]string{
+					WeGOAppIdentifierLabelKey: GetAppHash(app),
+				}
+
+				bytes, err := yaml.Marshal(&wegoApp)
+				Expect(err).To(BeNil())
```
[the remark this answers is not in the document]
jrryjcksn: Honestly, no... (at least not based on any sort of principle). There are too many equivalent ways of saying things in Ginkgo / Gomega. We don't have a standard. Maybe we should select some standards as a team?

pkg/services/automation/generator.go
```diff
@@ -0,0 +1,494 @@
...
+
+func AppDeployName(a models.Application) string {
+	return a.Name
+}
+
+func AppResourceName(a models.Application) string {
+	return a.Name
+}
+
+type GeneratedSecretName string
```
palemtnrider, line 395: nit: I find the code easier to read when types are defined at the top of the file or at a minimum before their first use.  `GeneratedSecretName` is used in `getAppSecretRef()` at the top of the file

pkg/services/automation/generator.go
```diff
@@ -0,0 +1,494 @@
...
+
+	if *visibility != gitprovider.RepositoryVisibilityPublic {
+		secretRef = CreateRepoSecretName(clusterName, url)
+	}
+
+	return secretRef, nil
+}
+
+func (a *AutomationGen) generateAppAutomation(ctx context.Context, app models.Application, clusterName string) ([]AutomationManifest, error) {
+	a.Logger.Generatef("Generating GitOps automation manifests")
```
palemtnrider, line 108: This log applies to the `generateApplicationGoat` call

pkg/services/automation/generator.go
```diff
@@ -0,0 +1,494 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("could not create app.yaml for '%s': %w", app.Name, err)
+	}
+
+	appGoatManifest, err := a.generateApplicationGoat(app, clusterName)
+	if err != nil {
+		return nil, fmt.Errorf("could not create GitOps automation for '%s': %w", app.Name, err)
+	}
+
+	a.Logger.Generatef("Generating application spec manifest")
```
palemtnrider, line 120: This log call should be moved to before the call to `generateAppYaml`

pkg/services/automation/generator.go
```diff
@@ -0,0 +1,494 @@
...
+}
+
+func AutomationUserKustomizePath(clusterName string) string {
+	return filepath.Join(automationRoot, "clusters", clusterName, "user", "kustomization.yaml")
+}
+
+func AppSourceName(a models.Application) string {
+	return a.Name
+}
+
```
palemtnrider, line 386:
```suggestion


```
I didn't see this function being used.

pkg/services/automation/generator.go
```diff
@@ -0,0 +1,494 @@
...
+func AppDeployName(a models.Application) string {
+	return a.Name
+}
+
+func AppResourceName(a models.Application) string {
+	return a.Name
+}
```
palemtnrider, line 393:
```suggestion

```
I didn't see this function being used.

test/integration/server/add_test.go
```diff
@@ -87,36 +87,40 @@ var _ = Describe("AddApplication", func() {
 			actual, err := helpers.GetFilesForPullRequest(ctx, gh, org, sourceRepoName, fs)
 			Expect(err).NotTo(HaveOccurred())
 
-			expectedKustomization := kustomizev2.KustomizationSpec{
+			expectedKustomization := kustomizev1.KustomizationSpec{
 				// Flux adds a prepending `./` to path arguments that doen't already have it.
```
palemtnrider, line 91: nit: typo doen't -> doesn't

test/integration/server/add_test.go
```diff
@@ -262,34 +270,38 @@ var _ = Describe("AddApplication", func() {
...
 				Path:           req.Path,
-				ConfigURL:      "",
+				ConfigURL:      ref.GetCloneURL(gitprovider.TransportTypeSSH) + ".git",
 				DeploymentType: wego.DeploymentTypeKustomize,
 				SourceType:     wego.SourceTypeGit,
 			}
 
-			expectedKustomization := kustomizev2.KustomizationSpec{
+			expectedKustomization := kustomizev1.KustomizationSpec{
 				// Flux adds a prepending `./` to path arguments that doen't already have it.
```
palemtnrider, line 279: nit: Typo: doen't -> doesn't

test/integration/server/helpers/helpers.go
```diff
@@ -94,40 +97,50 @@ func NewGithubClient(ctx context.Context, token string) *ghAPI.Client {
 	return ghAPI.NewClient(tc)
 }
 
-const InAppRoot = ".wego"
-const ExternalConfigRoot = ""
+const InAppRoot = ".weave-gitops"
```
palemtnrider, line 100: Thoughts on exporting `automationRoot` from the automation package instead of defining it again here?

palemtnrider (commented):
Here are another set of comments.  I have a few more files to review...

test/integration/server/helpers/helpers.go
jrryjcksn: The idea here was to use a separate set of definitions so as not to just hide a bug by using the same calls that might contain the bug

pkg/services/automation/generator.go
jrryjcksn: :heavy_check_mark:

pkg/services/automation/generator.go
jrryjcksn: Neglected to move comments after refactoring

pkg/services/automation/generator.go
jrryjcksn: Yep

pkg/services/gitopswriter/writer_remove_test.go
```diff
@@ -0,0 +1,255 @@
...
+	case kube.GVRGitRepository:
+		return automation.ResourceKindGitRepository
+	case kube.GVRHelmRepository:
+		return automation.ResourceKindHelmRepository
+	case kube.GVRHelmRelease:
+		return automation.ResourceKindHelmRelease
+	default:
+		return automation.ResourceKindKustomization
+	}
+}
```
palemtnrider, line 255:
```suggestion

```
Not being used

pkg/services/gitopswriter/writer_remove_test.go
```diff
@@ -0,0 +1,255 @@
...
+	. "github.com/onsi/gomega"
+	wego "github.com/weaveworks/weave-gitops/api/v1alpha1"
+	"github.com/weaveworks/weave-gitops/pkg/flux"
+	"github.com/weaveworks/weave-gitops/pkg/kube"
+	"github.com/weaveworks/weave-gitops/pkg/models"
+	"github.com/weaveworks/weave-gitops/pkg/services/automation"
+	"github.com/weaveworks/weave-gitops/pkg/services/gitrepo"
+	"github.com/weaveworks/weave-gitops/pkg/testutils"
+
+	"k8s.io/apimachinery/pkg/runtime/schema"
```
palemtnrider, line 21:
```suggestion
```
Used by a function which isn't used

pkg/services/gitopswriter/writer_remove_test.go
```diff
@@ -0,0 +1,255 @@
...
+	"io/fs"
+	"os"
+	"path/filepath"
+	"strings"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	wego "github.com/weaveworks/weave-gitops/api/v1alpha1"
+	"github.com/weaveworks/weave-gitops/pkg/flux"
+	"github.com/weaveworks/weave-gitops/pkg/kube"
```
palemtnrider, line 15:
```suggestion
```
Used by a function which isn't used

pkg/services/gitopswriter/writer_add_test.go
```diff
@@ -0,0 +1,305 @@
...
+				path, content = gitClient.WriteArgsForCall(1)
+				Expect(path).To(Equal(".weave-gitops/apps/bar/bar-gitops-deploy.yaml"))
+				Expect(content).To(Equal([]byte("kustomization")))
+
+				path, content = gitClient.WriteArgsForCall(2)
+				Expect(path).To(Equal(".weave-gitops/apps/bar/bar-gitops-source.yaml"))
+
+				augmented, err := automation.AddWegoIgnore(dummyGitSource)
+				Expect(err).To(BeNil())
+				Expect(content).To(Equal(augmented))
```
palemtnrider, line 146:
```suggestion
				Expect(automation.AddWegoIgnore(dummyGitSource)).To(Equal(content))
```
This pattern can be simplified down to a single line - this says the matcher To() should be applied to the first value returned and all other values should be nil or the empty value.

palemtnrider (approved):
Just a few additional comments combined with my previous suggestions.

A couple of items I'd like to discuss further is if we can have the new service tests target the public interfaces i.e., be in package gitopswriter_test instead of gitopswriter. 

Also, in a different PR or as we continue enhancing the code, we need to add golang doc comments for all exported functions and types in public packages to make it easier for people to write against these packages.

This was a huge amount of work - well done 👏 👍

palemtnrider (approved):
Just a few additional comments combined with my previous suggestions.

A couple of items I'd like to discuss further is if we can have the new service tests target the public interfaces i.e., be in package gitopswriter_test instead of gitopswriter. 

Also, in a different PR or as we continue enhancing the code, we need to add golang doc comments for all exported functions and types in public packages to make it easier for people to write against these packages.

This was a huge amount of work - well done 👏 👍

Issue #985, opened 2021-10-28
Add GitLab auth token retrieval to the UI


jpellizzari:
According to the ADR, the Gitlab auth flow for the frontend code/experience will need to go something like this: Note that this differs from the current GitHub auth strategy. It is still OAuth, but GitLab is a bit more involved.

1. Detect the provider from a URL that the user types in
2. (if Gitlab) Present the user with a button that will take them to the Gitlab OAuth page
3. After authorizing, Gitlab will send the user back to our callback page (JS or Go HTTP handler?)
4. A backend route will exchange the challenge code on the backend and covert it into a JWT that gets returned to the UI

Acceptance Criteria:
* A user should be able to complete the GitLab auth flow during adding/removing an app and listing commits
* Unit test coverage with sad path cases handled
* The GitLab UI auth flow should be re-usable by Enterprise (just like the GitHub one is today)

josecordaz:
We saw this message in the UI when trying to authenticate using the button on the commits section. Not sure if this should be expected in this PR. 

![image.png](https://images.zenhubusercontent.com/606389311ead0cc4788a9a72/e6d60b05-1df2-4d60-963f-e950f4424c63)

And also seen this message after clicking on Remove App button
![image.png](https://images.zenhubusercontent.com/606389311ead0cc4788a9a72/8a6d2193-8dcf-4e6f-bde8-ed29bc09b4e4)

@jpellizzari Would you mind helping us to clarify if this is the expected behavior.

palemtnrider:
While testing with the UI running from within the cluster, we are seeing an expired token error which prohibits adding an app. 

@jpellizzari suspects we are requesting a token with an invalid expiration value. 

CC @josecordaz 

Log message from the container:

```
2021-11-29T23:07:22.344Z        INFO    middleware/middleware.go:93     invalid token format
[]
2021-11-29T23:07:23.445Z        INFO    middleware/middleware.go:93     invalid token format
2021-11-29T23:07:23.775Z        INFO    middleware/middleware.go:93     invalid token format
[]
2021-11-29T23:07:27.398Z        INFO    middleware/middleware.go:105    could not parse claims: token is expired by 4s: unauthorized token
2021-11-29T23:07:27.399Z        ERROR   middleware/middleware.go:40     server error    {"error": "rpc error: code = Unauthenticated desc = token error: could not get token from context"}
github.com/weaveworks/weave-gitops/pkg/server/middleware.WithGrpcErrorLogging.func1
[...]

PR #1046, opened 2021-11-08
Add GitLab OAuth flow the UI

jpellizzari:
Closes: #985 

Adds Gitlab token retrieval to the UI:
https://drive.google.com/file/d/1w4zGgM_V1FCn4khD0Tp5DPzCO4kreEpa/view?usp=sharing

Also adds:
* API endpoint around Repo URL parsing
* API endpoints for initiating and completing the token flow
* A new `RepoInputWithAuth` component that seeks to encapsulate the auto-detection behavior.

Known limitations:
* Does not seek to successfully add an `Application` to GitLab via the UI; completing this work would conflict with the #1024 PR that is being worked on
* Does not detect a `Git Provider` on the config repo. Specifying conflicting providers will result in an error.
* Does NOT account for `Git Provider` in a secret name stored on the cluster(!!!). Mixing and matching providers on the same cluster will result in errors. Again, this wasn't fixed as it would conflict with #1024.

@JamWils Could you focus a review on the GitLab auth stuff

@joshri and @simon Could you check out the React stuff?


ui/pages/OAuthCallback.tsx
```diff
@@ -0,0 +1,78 @@
...
+    const state = getCallbackState();
+
+    if (state?.page) {
+      history.push(linkResolver(state.page));
+      return;
+    }
+  }, [res]);
+
+  return (
+    <div className={className}>
```
joshri, line 61: Do we need this wrapping div? We could pass the className and loading state down to `<Page />` right? I'm just noticing that `<LoadingPage />` component (as part of Page) for the first time - do we use that?

ui/pages/ApplicationAdd.tsx
```diff
@@ -231,22 +246,36 @@ function AddApplication({ className }: Props) {
...
+                    state: formState,
+                  });
+                }
+              }}
               required
               id="url"
               label="Source Repo URL"
               variant="standard"
               value={formState.url}
+              helperText="The git repository URL where the application YAML files are stored"
```
joshri, line 277: Are we going to have to put all the new auth stuff in this component into every component where we do some sort of action? It makes me want to put all this auth stuff at the highest level right? Will we know what provider the user is with as a part of the application object?

joshri (approved):
Obviously this is 🔥 super hot fire 🔥. Nice!

ui/pages/ApplicationAdd.tsx
jpellizzari: At the moment, the GitHub auth stuff isn't encapsulated anywhere. The GitLab button is pretty well encapsulated, as anywhere you drop it in It Should Work:tm:.

I would like to get to the point where anywhere you drop in the`GitHubAuthButton` it opens the modal and does the whole thing. This would have required a larger change, however, and made this PR tougher to review.

>  Will we know what provider the user is with as a part of the application object?

We can check the URL(s) of the `Application` to figure out which provider it is.

JamWils (approved):
Nice and clean code structure.  I only reviewed the Go code.  LGTM

jpellizzari:
Added a new `React.Context` provider to better encapsulate the Gitlab auth button behavior. That provider will need to be used wherever a `<GitlabAuthButton />` is used.

PR #1047, opened 2021-11-08
Fix selenium for macOS on nightly

rokshana-b:
**What changed?**
Changed the way `webDriver` is initialized for MacOS

**Why?**
MacOS now comes with selenium 4 by default in github actions. To have successful test runs, we need to initialize `webDriver` instance differently — depending on the OS in which the test is running. This only affects the UI tests.


jrryjcksn (approved):
LGTM

PR #1048, opened 2021-11-08
Updated error message when hit "not found error" in gitprovider

josecordaz:
Closes: #1048 

**What changed?**
Updated error message.
From:
```
Error: failed to get git clients: error setting up deploy keys: error uploading deploy key: error uploading deploy key multiple errors occurred:
- POST https://api.github.com/repos/stefanprodan/podinfo/keys: 404 Not Found []
- the requested resource was not found
```
To: 
```
Error: failed to get git clients: error setting up deploy keys: error uploading deploy key: no permissions to access this repository or repository doesn't exists
```

**Why?**
There was confusion when distinguishing between if there were not enough permissions and if the repository does not exist.

**How did you test it?**
Added unit tests

**Release notes**

**Documentation Changes**

### DIFF
--- a/pkg/gitproviders/provider.go
+++ b/pkg/gitproviders/provider.go
@@ -76,10 +76,19 @@ func New(config Config, owner string, getAccountType AccountTypeGetter) (GitProv
 }
 
 func deployKeyExists(ctx context.Context, repo gitprovider.UserRepository) (bool, error) {
-	_, err := repo.DeployKeys().Get(ctx, DeployKeyName)
-	if err != nil && !strings.Contains(err.Error(), "key is already in use") {
+	_, err := repo.DeployKeys().List(ctx)
+	if err != nil {
 		if errors.Is(err, gitprovider.ErrNotFound) {
 			return false, RepositoryNoPermissionsOrDoesNotExistError
+		} else {
+			return false, fmt.Errorf("error getting deploy key : %s", err)
+		}
+	}
+
+	_, err = repo.DeployKeys().Get(ctx, DeployKeyName)
+	if err != nil && !strings.Contains(err.Error(), "key is already in use") {
+		if errors.Is(err, gitprovider.ErrNotFound) {
+			return false, nil
 		} else {
 			return false, fmt.Errorf("error getting deploy key %s: %s", DeployKeyName, err)
 		}
--- a/pkg/gitproviders/provider.go
+++ b/pkg/gitproviders/provider.go
@@ -25,6 +25,8 @@ const (
 	defaultTimeout = time.Second * 30
 )
 
+var RepositoryNoPermissionsOrDoesNotExistError = errors.New("no permissions to access this repository or repository doesn't exists")
+
 // GitProvider Handler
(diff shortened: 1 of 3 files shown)


pkg/gitproviders/provider_org_test.go
```diff
@@ -137,6 +137,13 @@ var _ = Describe("Org Provider", func() {
 			Expect(err.Error()).Should(ContainSubstring("error uploading deploy key"))
 		})
 
+		It("return error when it fails to upload a deploy key", func() {
```
jpellizzari, line 140: :+1:

pkg/gitproviders/provider_user_test.go
```diff
@@ -94,7 +94,7 @@ var _ = Describe("User Provider", func() {
 			deployKeyClient.GetReturns(nil, gitprovider.ErrNotFound)
 
 			res, err := userProvider.DeployKeyExists(ctx, repoUrl)
-			Expect(err).ToNot(HaveOccurred())
+			Expect(err).ShouldNot(HaveOccurred())
```
jpellizzari, line 97: Why the change to this?

jpellizzari (approved):

pkg/gitproviders/provider_user_test.go
josecordaz: I'll revert this, I modified it twice and when I tried to put it back to the previous state forgot it had `ToNot` instead of `ShouldNot`.

pkg/gitproviders/provider.go
```diff
@@ -89,7 +91,11 @@ func deployKeyExists(ctx context.Context, repo gitprovider.UserRepository) (bool
 func uploadDeployKey(ctx context.Context, repo gitprovider.UserRepository, deployKeyInfo gitprovider.DeployKeyInfo) error {
 	_, err := repo.DeployKeys().Create(ctx, deployKeyInfo)
 	if err != nil {
-		return fmt.Errorf("error uploading deploy key %s", err)
+		if errors.Is(err, gitprovider.ErrNotFound) {
```
luizbafilho, line 94: in general in Go we should use else clauses unless strictly necessary, which is not the case here since we call `return` in the `if`, so a suggestion would be changing to:
```go
if errors.Is(err, gitprovider.ErrNotFound) {
	return RepositoryNoPermissionsOrDoesNotExistError
}

return fmt.Errorf("error uploading deploy key %s", err)
```

pkg/gitproviders/provider.go
```diff
@@ -25,6 +25,8 @@ const (
 	defaultTimeout = time.Second * 30
 )
 
+var RepositoryNoPermissionsOrDoesNotExistError = errors.New("no permissions to access this repository or repository doesn't exists")
```
luizbafilho, line 28: major: errors are prefixed with the `Err` like in `gitprovider.ErrNotFound` and the error types are suffixed with the `Error`

luizbafilho (changes requested):

luizbafilho (approved):

Issue #1017, opened 2021-11-03
Application graph node styling and positioning


bia:
Relates to #876

- Each node should have a more subtle drop shadow, more of an outline like `drop-shadow(rgb(189,189,189) 0px 0px 1px)`.
 
- Each circle should also be larger to be able to display more text and to improve #899 
<img width="868" alt="Screenshot 2021-10-13 at 15 05 15" src="https://user-images.githubusercontent.com/92439/137138209-1405d149-56ac-47c3-8bb2-db78e092d0a2.png">

- Nodes should be radially distributed to save space and to be less similar to Argo. Weave Scope's radial rendering on focus, but [was built in d3](https://github.com/weaveworks/scope/blob/a3597c9639bc32c06b05e149c7e09ab3ec958f42/client/app/scripts/charts/nodes-chart.js).
Maybe we can use [d3 for npm](https://www.npmjs.com/package/c3) to build something similar to the following design

![20211013app-graph](https://user-images.githubusercontent.com/92439/137141089-bf1afd26-437a-4250-9b71-23c1d4ce588c.png)

- When I scroll down on the page, I am most often accidentally zooming out the graph instead. The graph should be completely zoomed out (such that all nodes are visible) on page load.

joshri:
@bia is it possible for me to have the concentric circles relate to their actual depth in the hierarchal tree? That way I can just distribute all the nodes at that depth evenly across the circle. My pod-info repo has six direct children right now in the graph, and I'm just thinking about how to space them all in a way that looks right - the three children of the root in your pic have three different edge lengths, and I'm not sure how to code that in a way that will look correct for any number of children besides dividing up those cool circles you have.  Am I making any sense at all?

bia:
You mean something like this? Sounds like a good start!
![20211108app-graph](https://user-images.githubusercontent.com/92439/140769883-d5959c20-b543-4fff-b4da-5f07fabff71c.png)

jpellizzari:
> Nodes should be radially distributed to save space and to be less similar to Argo. Weave Scope's radial rendering on focus, but was built in d3.
Maybe we can use d3 for npm to build something similar to the following design

Let's split this into a separate ticket. I will create it. Let's leave it out of scope for now.

PR #1051, opened 2021-11-09
Increase graph node size, and decrease graph node drop shadow

joshri:
Related to #1017 

Changing the node size messes with a lot of other styles set on the graph, and doesn't ultimately add too many text characters back from overflow (making the nodes larger than I did just seemed too weird with how the graph is currently zoomed). At least the drop shadow looks good? 

image


ui/components/ReconciliationGraph.tsx
```diff
@@ -117,17 +117,21 @@ export default styled(ReconciliationGraph)`
 
   .node {
     position: relative;
-    font-size: 14px;
+    font-size: 16px;
```
bia, line 120: Why did you bump up the font size? Is that why more text cannot be displayed in the larger node size?

bia (approved):
Drop shadow looks good to me.
I don't care about the node sizing if larger nodes don't allow for more text to fit in them.

ui/components/ReconciliationGraph.tsx
joshri: I think that the issue has more to do with the zoom scaling than anything else - if I leave the font size the same, as soon as you zoom out at all you can't read it. It's a difference of a couple characters unfortunately...I know we're talking about how we're going to change up the graph, so I'm hoping we can address this then.

jpellizzari (approved):

Issue #971, opened 2021-10-27
`Authorization` header should be used for authentication against an origin server


yiannistri:
**Describe the bug**
Currently, the `Authorization` header is used to accept GH tokens so that HTTP handlers can interact with GH on behalf of a user. However, a different header should be used instead in order to distinguish between API endpoint authorization (i.e. listing clusters or templates) and endpoints that require interaction with GH or other git providers (i.e. create a cluster PR). I suggest the use of an alternative header name instead i.e. `Git-Provider-Token` and reserve the `Authorization` header for securing the endpoints.

**Environment**
gitops: v0.3.3

**Affects versions**
v0.3.3

**Expected behavior**
Use a header other than `Authorization` to check for the GH token

https://datatracker.ietf.org/doc/html/rfc7235#section-4.2

PR #965, opened 2021-10-26
Pass user token to API when creating/deleting CAPI clusters

yiannistri:
Closes: weaveworks/weave-gitops-enterprise#203

**What changed?**
- Extract `GetToken` out of `InitGitProvider`
- Use `GetToken` to pass git provider token to API as a header

**Why?**
This change allows cluster creation/deletion PRs to be created under the user account.

**How did you test it?**
No new tests were added but tested locally.

**Release notes**
N/A

**Documentation Changes**
Follow up PRs for docs should be created after the back end change to use the tokens.

PR #1052, opened 2021-11-09
Renaming Git Provider Token header

luizbafilho:
Closes: #971 

**What changed?**
- A different header should be used instead in order to distinguish between API endpoint authorization (i.e. listing clusters or templates) and endpoints that require interaction with GH or other git providers

**Why?**

**How did you test it?**

**Release notes**

**Documentation Changes**


J-Thompson12 (approved):
LGTM

pkg/server/middleware/middleware.go
```diff
@@ -74,14 +74,17 @@ type contextVals struct {
 
 type key int
 
-const tokenKey key = iota
-const GRPCAuthMetadataKey = "grpc-auth"
+const (
+	tokenKey               key = iota
+	GRPCAuthMetadataKey        = "grpc-auth"
+	GitProviderTokenHeader     = "X-Git-Provider-Token"
```
bigkevmcd, line 80: https://datatracker.ietf.org/doc/html/rfc6648

ui/contexts/AppContext.tsx
```diff
@@ -52,7 +52,7 @@ function wrapClient<T>(client: any, tokenGetter: () => string): T {
         ...options,
         headers: new Headers({
           ...(options.headers || {}),
-          Authorization: `token ${tokenGetter()}`,
+          "X-Git-Provider-Token": `token ${tokenGetter()}`,
```
jpellizzari, line 55: Can we make this a `const` as well? For computed keys in JS objects:
```suggestion
          [gitProviderTokenHeader]: `token ${tokenGetter()}`,
```

jpellizzari (approved):

bigkevmcd:
This is for this PR https://github.com/weaveworks/weave-gitops/pull/965/files#diff-142329b4654445c09cf14ebc4afba7677e064ede74cb549a81e935b414156dd2R248 and the header in there is `Git-Provider-Token`.

pkg/server/middleware/middleware.go
luizbafilho: sounds good, I'll remove the `X-` prefix

Issue #1050, opened 2021-11-09
add app --dry-run breaks add app by creating the deploy key


sympatheticmoose:
**Describe the bug**
If you run add app twice, first with the dry-run flag to see what should happen, a deploy key will be added to the cluster and then the second add will fail as unable to store secret with the same name. 

**What is the severity of the bug**
Setting as minor for now given expected usage of dry-run... may raise to major.

**Environment**
 - gitops: [e.g. v0.4.1]
 - kubernetes:  [e.g. 1.20.4]
    - [x ] KinD - _version_]
    - [ ] k3s - _version_
    - [ ] cloud [e.g., EKS, AKS]  _version_
    - [ ] other - _name_ _version_
 - Browser + version: [e.g. chrome 74, safari 12, firefox 87]
 -
**Affects versions**

**To Reproduce**
1. `gitops add app . --dry-run`
2. see: `A deploy key named wego-<kubecontext>-<appname> was found on the git provider, but not in the cluster.
Deploy key generated and uploaded to git provider`
3. `gitops add app .`
4. See error: `Error: failed to create app service: error getting git clients: error setting up deploy keys: error storing deploy key: could not store secret: secrets "wego-<kubecontext>-<appname>" already exists`

**Expected behavior**
A drier dry run

**Config and Logs**
If applicable, add logs to help explain your problem. _please compress the output before attaching_
- [ ] gitops config
- [ ] `gitops` cli call output (and parameters)
- [ ] gitops-controller logs
- [ ] Events from wego-* namespaces
- [ ] `kubectl cluster-info dump`
- [ ] Prometheus alerts
- [ ] Flux logs

**Screenshots**

**Additional context**

PR #1054, opened 2021-11-10
Create empty ssh key for git client to avoid creating real deploy keys when using dry-run.

josecordaz:
Closes: #1050 

**What changed?**
Avoid creating deploy key when running `add app` using `--dry-run` flag

**Why?**
To avoid the error "deploy key already exists" when adding the app without `dry-run` flag

**How did you test it?**
Manually and adding a unit test.

**Release notes**

**Documentation Changes**

### DIFF
--- a/pkg/services/auth/auth.go
+++ b/pkg/services/auth/auth.go
@@ -63,7 +63,7 @@ func (sn SecretName) NamespacedName() types.NamespacedName {
 }
 
 type AuthService interface {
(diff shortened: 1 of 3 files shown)


J-Thompson12 (approved):
LGTM. It would be nice to handle the dry run just inside factory.go so you dont need to pass it around but I think its more unclean to move makePublicKey somewhere that factory can access it.

PR #1064, opened 2021-11-10
Fix cluster pull request statuses

sarataha:
Closes: weaveworks/weave-gitops-enterprise#250

**What changed?**

Updated Cluster struct to get the correct PR type

**Why?**

To differentiate between different PR types whether it's creation or deletion

**How did you test it?**

Tested locally + unit tests

**Release notes**

**Documentation Changes**


foot (approved):
Works great!

It would be _really_ neat show the URL there too. Maybe another column called `STATUS_MESSAGE` for now? (sounds a little bit awkward). We could do this in a follow up PR too

pkg/clusters/clusters.go
```diff
@@ -29,16 +34,16 @@ func GetClusters(r ClustersRetriever, w io.Writer) error {
...
-			if c.PullRequestType == "create" {
+			if c.PullRequest.Type == "create" {
 				c.Status = "Creation PR"
-			} else if c.PullRequestType == "delete" {
+			} else if c.PullRequest.Type == "delete" {
 				c.Status = "Deletion PR"
 			}
 
-			fmt.Fprintf(w, "%s\t%s", c.Name, c.Status)
+			fmt.Fprintf(w, "%s\t%s\t%s", c.Name, c.Status, c.PullRequest.Url)
```
foot, line 46: 🙌 , maybe only if c.status is creation/deletion PR. e.g. I guess leave it blank for "clusterFound" etc.

sarataha: Sure! 👌 Thank you Simon 🙏

Issue #1059, opened 2021-11-10
gitops beta install --app-config-url git@github.com:weavegitops/prod1-repo.git Fails


LutzLange:
**Describe the bug**
I'm trying to install with a management cluster. I'm using the new --app-config-url. I can't make it work. I did try the http and ssh variants. Both fail. 

For the SSH variant : 
gitops beta install --app-config-url git@github.com:weavegitops/prod1-repo.git
....
Error: failed to store cluster manifests: failed to create source manifest: could not create git source for repo "git@github.com:weavegitops/prod1-repo.git" : failed to create source git: failed to run flux with output: ✗ git URL parse failed: parse "git@github.com:weavegitops/prod1-repo.git": **first path segment in URL cannot contain colon
 and error: exit status 1**

And here the http variant as well : 
$ gitops beta install --app-config-url https://github.com/weavegitops/prod1-repo.git
...
Error: failed to store cluster manifests: failed to clone configuration repo: failed cloning user repo: https://github.com/weavegitops/prod1-repo.git: **invalid auth method**

**Affects versions**
gitops version
Current Version: 0.4.1
GitCommit: c0d130b40ff6996c409973aa69573aea5d0751ea
BuildTime: 2021-11-03T20:02:35Z
Branch: HEAD
Flux Version: v0.21.0

jpellizzari:
Looks like the `beta install` command does not use a `gitproviders.RepoURL` that would have converted the input. Changing this at the `InstallParams` could be a good way to fix it:

https://github.com/weaveworks/weave-gitops/blob/d63d9ff2a50766db69b9320402a7530cefb3acae/pkg/services/gitops/install.go#L23

LutzLange:
The workaround is to use an URL with like "ssh://git@github.com/weavegitops/prod1-repo"

josecordaz:
@jpellizzari This doesn't seem to be an issue in the main branch at this moment. I tested both types of URLs and they worked fine. `gitprovider.RepoURL` is being used here already https://github.com/weaveworks/weave-gitops/blob/main/pkg/services/factory.go#L120. That is why it is not failing in the current main. Want me to change it to `gitprovider.RepoURL` anyways?  I wonder if we should change these two URL fields to `gitprovider.RepoURL` as well?

jpellizzari:
@josecordaz Could you double check that we have acceptance test coverage for the format of URL that was originally mentioned? If not, please add a test for it. I just want to be sure it is covered.

[...]

PR #1065, opened 2021-11-10
Changed URL type `git@` for one of the acceptance tests

josecordaz:
Closes: #1059 

**What changed?**
Including URL type `git@` in acceptance tests.

**Why?**
To cover those cases as well.

**How did you test it?**
Updated integration test

**Release notes**

**Documentation Changes**


test/integration/server/helpers/helpers.go
```diff
@@ -47,22 +47,32 @@ func CreateRepo(ctx context.Context, gp gitprovider.Client, url string) (gitprov
...
+		DefaultBranch: gitprovider.StringVar(defaultBranch),
 	}, &gitprovider.RepositoryCreateOptions{AutoInit: gitprovider.BoolVar(true)})
 
 	if err != nil {
 		return nil, nil, fmt.Errorf("could not reconcile org repo: %w", err)
 	}
 
-	err = utils.WaitUntil(bytes.NewBuffer([]byte{}), 3*time.Second, 5*time.Second, func() error {
-		_, err := gp.OrgRepositories().Get(ctx, *ref)
+	err = utils.WaitUntil(os.Stdout, 3*time.Second, 9*time.Second, func() error {
```
jpellizzari, line 61: :question: Why do we need to log this to `stdout`?

test/integration/server/helpers/helpers.go
```diff
@@ -47,22 +47,32 @@ func CreateRepo(ctx context.Context, gp gitprovider.Client, url string) (gitprov
...
+		r, err := gp.OrgRepositories().Get(ctx, *ref)
 		if err != nil {
 			return err
 		}
 
+		commits, err := r.Commits().ListPage(ctx, defaultBranch, 1, 0)
```
jpellizzari, line 67: :question: Why are we waiting for commits to occur on an operation that creates a repo?

jpellizzari (commented):
@josecordaz Why were integration test behaviors changed in this PR?

test/integration/server/helpers/helpers.go
josecordaz: Because we are pushing changes to the repo and we want the repo to be ready to receive commits. When we run `gp.OrgRepositories().Get(ctx, *ref)` it doesn't mean necessarily that the repo is ready to receive commits. You can see the error this fixes here https://github.com/weaveworks/weave-gitops/runs/4172766374?check_suite_focus=true#step:13:125.

test/integration/server/helpers/helpers.go
josecordaz: This is going to help us identify the exact errors this function is facing before every retry and make sure it is retrying enough times. For instance, I didn't realize that we were retrying only once as a result of poll-time being 3sec and timeout being 5sec. I just noticed it until I made the change to `stdout`. As a result, I increased the timeout to 9 seconds so at least it retries 3 times. Another aspect I've seen on previous occasions is that it might hit different errors in each of the retries. Each of those errors could give us more information when debugging.

jpellizzari: Cool makes sense to me

jpellizzari (approved):

PR #1069, opened 2021-11-12
Upgrade uses new helm repo

foot:
This switches from installing the weave-gitops-enterprise profile from a git-repository using `pctl` to installing from a helmrepository!

- Adds support for `--dry-run`
- Adds support for `--set capi.cluster.name=foo`
- Configures an interval for the helmrelease so changes are re-applied
- Auto set clusterName from env
- Auto set repoURL from `--app-config-url`


foot:
I think we can remote this GitRepo stuff now!

pkg/upgrade/upgrade.go
```diff
@@ -56,200 +63,203 @@ const CredentialsSecretName string = "weave-gitops-enterprise-credentials"
...
-	return manager.Install(catalog.InstallConfig{
-		Clients: catalog.Clients{Installer: install.NewInstaller(installConfig)},
-		Profile: profile,
-	})
-}
+	if uv.GitRepository == "" {
+		gitRepositoryNameNamespace := fmt.Sprintf("%s/%s", uv.Namespace, strings.TrimSuffix(filepath.Base(repoUrlString), ".git"))
+		fmt.Fprintf(w, "Deriving name of GitRepository Resource as %v\n", gitRepositoryNameNamespace)
+		uv.GitRepository = gitRepositoryNameNamespace
+	}
```
foot, line 234: I think we can remove this now as it was only important for profiles.

+ also rm the `ensureGitRepositoryResource` and `getGitRepositoryNamespaceAndName` too

cmd/gitops/upgrade/cmd.go
```diff
@@ -37,17 +43,37 @@ func init() {
...
 		upgradeCmdFlags.Namespace = namespace
 
+		log := internal.NewCLILogger(os.Stdout)
+		fluxClient := flux.New(osys.New(), &runner.CLIRunner{})
+		factory := services.NewFactory(fluxClient, log)
+
+		providerClient := internal.NewGitProviderClient(os.Stdout, os.LookupEnv, auth.NewAuthCLIHandler, log)
+
+		gitClient, gitProvider, err := factory.GetGitClients(ctx, providerClient, services.GitConfigParams{
+			URL:       upgradeCmdFlags.RepoURL,
```
foot, line 63: Nice, I think this is simpler to do it as a flag instead of reading from the gitconfig.
Maybe we rename this flag to `--app-config-url` to be consistent w/ `gitops install --app-config-url`.

And we can remove the code that tries to automatically figure this out (`	repoUrlString, err := gitClient.GetRemoteUrl(".", uv.Remote)`).

I don't think we need `RepoOrgAndName` anymore either so we can get rid of the whole `buildUpgradeConfigs` probably. 🤔

foot (commented):
Looking 🔥 !

Excited about all the cleanup.

Some polish that would be really neat
- [x] ask for `--profile-version=v0.0.14` flag that we could set the helmrelease version to
- [ ] Set the `values.config.cluster.name` in the helm release to `cname` var.
- [ ] Set the `values.config.capi.repositoryURL` to --app-config-url

🤔 

🙌

pkg/upgrade/upgrade.go
sarataha: 👌 makes sense 🙏

cmd/gitops/upgrade/cmd.go
sarataha: Thank you very much Simon 🙏 I will add the suggest changes 👍

pkg/upgrade/upgrade.go
```diff
@@ -56,243 +63,183 @@ const CredentialsSecretName string = "weave-gitops-enterprise-credentials"
...
+	valuesTemplate := `{
+		"config": {
+			"cluster": {
+				"name": "%s"
+			},
+			"capi": {
+				"repositoryURL": "%s"
+			}
+		}
+	}`
```
foot, line 177: We can do a literal here to avoid the decode w/ something like this (made a little demo 🎉 )

https://play.golang.org/p/iFTT9sE9FIx

sarataha: Thank you very much Simon, that looks much better than using strings too 👍. I tried to do it with nested `map[string]interface{}` as well but it doesn't look very readable 😅

foot (commented):
Looking great! So nice to have `--set` here. 🔥  well done :)

We had some feedback around flags. Could we remove all the ones we don't use anymore..
- Remote
- ConfigMap
- Out (?)
- Maybe others?

and make sure the other ones are the similar to `add app` / `install` as possible?

sarataha (approved):
Great work! 🥇 🙌

Issue #1032, opened 2021-11-04
Add standard table to Applications page


bia:
Relates to #876

The Applications page should use the same table style as the Clusters page.

The title of that table should be something more descriptive like 'Installed GitOps Applications,' and each item will have a status/health metric displayed. ('Healthiness' being discussed in #1004)

![20211013applications](https://user-images.githubusercontent.com/92439/137141677-e004c366-26d8-48b4-a841-9f9e6642601b.png)

sympatheticmoose:
@foot when you show a pending cluster is that when the PR has been raised but not yet merged? (Thinking similar behaviour would be nice to steal for here)

joshri:
@jpellizzari Blocked by the current state of the Application object - we're missing the status, version, latest commit, and repo link. Is the checkbox functionality tied to the remove app button, and if so, does that mean that we're moving the button out of the AppDetail page, and adding the functionality to delete multiple apps at once?

bia:
@rokshana-b and I also talked about showing whether an app is suspended or not here.

sympatheticmoose:
@bia what about last successful reconciliation? Latest commit is on the repo... would we want to show latest fetched/applied?

JamWils:
There is no context for this table to be run in Core today.  When we have those options we can revisit this issue.

Issue #876, opened 2021-10-13
UI improvements to application flow


bia:
1. #1016
so that the application's state is apparent once the page loads. The 'Application / podinfo-deploy' title is also not needed because of the breadcrumbs in the navbar above.
<img width="821" alt="Screenshot 2021-10-13 at 15 32 56" src="https://user-images.githubusercontent.com/92439/137142838-4891f88c-b550-4549-8881-5192a049cdf6.png">

2. #1017
Each node should have a more subtle drop shadow, more of an outline like `drop-shadow(rgb(189,189,189) 0px 0px 1px)`. Each circle should also be larger to be able to display more text.
<img width="868" alt="Screenshot 2021-10-13 at 15 05 15" src="https://user-images.githubusercontent.com/92439/137138209-1405d149-56ac-47c3-8bb2-db78e092d0a2.png">

How are the graph nodes being distributed? Could we move to a radial distribution? That would save space and remind me less of Argo. 
When I scroll down on the page, I am most often accidentally zooming out the graph instead. It would make sense if the graph was completely zoomed out (such that all nodes are visible) on page load.
![20211013app-graph](https://user-images.githubusercontent.com/92439/137141089-bf1afd26-437a-4250-9b71-23c1d4ce588c.png)

3. Are there ever going to be more rows on this table? If not, this information should be included in the status section instead of formatted as a table.
<img width="1137" alt="Screenshot 2021-10-13 at 15 20 27" src="https://user-images.githubusercontent.com/92439/137140735-b33c0f75-c04b-4378-bfce-3f88064db8a1.png">

4. #1032
The Applications page should use the same table style as the Clusters page. The title of that table should be something more descriptive like 'Installed GitOps Applications,' and each item will have a status/health metric displayed.
![20211013applications](https://user-images.githubusercontent.com/92439/137141677-e004c366-26d8-48b4-a841-9f9e6642601b.png)

5.  #1033
Implement counts for Applications and Clusters in the menu bar navigation as is important for maintaining context.

Let me know your thoughts and suggestions!
cc @sympatheticmoose @jpellizzari

jpellizzari:
> We need a top level status section on each application's page so that the application's state is apparent once the page loads. The 'Application / podinfo-deploy' title is also not needed because of the breadcrumbs in the navbar above.

We don't have those status metrics and need to discuss what an aggregated "Healthy" would mean. We can make space for them in the mean time.

[...]

PR #1071, opened 2021-11-12
Create Styling Skeleton for Applications Table

joshri:
Relates to #1032 and #876

Adds initial code for the future functionality of the Applications page, currently blocked by missing fields from the application object. This includes the `DataTable`, `SortableLabel`, and a bunch of different classes and styling tweaks to get things closer to the design. 

We need to decide on a couple things: 
1.  Is the option to remove an app supposed to be here in addition to in `ApplicationDetail`? I haven't attached the functionality to the button here but I can if we want it.
2. Is the functionality of the checkboxes to remove multiple apps at once? What else do I need to add in, or do we need them at all?
3. I didn't put the two buttons in the top right of the design in because I'm not sure of their functionality.

Check out what I have compared to the design below:

Design -
image

1032-app-page branch -
image


ui/components/Button.tsx
```diff
@@ -11,4 +11,24 @@ function Button(props: Props) {
   return <MaterialButton {...props} />;
 }
 
-export default styled(Button)``;
+export default styled(Button)`
+  &.applications {
```
jpellizzari, line 15: This seems a little backwards. As I understand this, when the `.applications` class is applied to the `Button`, it changes the border properties.

I feel like this should be inverted. The `Applications` page or some page that uses a `<Button />` should apply some overrides. This is a little un-semantic as is.

Maybe if we has a class that was `.square-border` or something, it would be closer to what we are trying to do.

ui/components/DataTable.tsx
```diff
@@ -43,11 +42,13 @@ function DataTable({
     sorted.reverse();
   }
 
+  console.log(sorted);
```
jpellizzari, line 45:
```suggestion
```

ui/components/Flex.tsx
```diff
@@ -22,6 +22,10 @@ const Styled = (component) => styled(component)`
   ${({ wide }) => wide && "width: 100%"};
   ${({ wrap }) => wrap && "flex-wrap: wrap"};
   ${({ end }) => end && "justify-content: flex-end"};
+  &.application-actions {
```
jpellizzari, line 25: Similar to my comment about the `<Button />`, the `<Flex />` component doesn't know about `applications`. It exists in a chaotic universe where there are only misaligned `divs` that need to be evenly distributed :wink:.

ui/pages/Applications.tsx
```diff
@@ -1,50 +1,132 @@
...
       <Flex align between>
         <TitleBar>
-          <h2>Applications</h2>
+          <h2>Installed GitOps Applications</h2>
         </TitleBar>
-        <Link to={PageRoute.ApplicationAdd} className="title-bar-button">
-          <Button variant="contained" color="primary" type="button">
-            Add Application
+      </Flex>
+      <Flex className="application-actions">
```
jpellizzari, line 66: How about splitting this into an `<ActionBar />` component? I feel like we will use this pattern a lot in the future.

I can imagine there will be lots of places where we want a bunch of buttons at the top of a page.

ui/pages/Applications.tsx
```diff
@@ -1,50 +1,132 @@
...
+  const [sort, setSort] = React.useState("Name");
+  const [reverseSort, setReverseSort] = React.useState(false);
   const { listApplications, loading } = useApplications();
 
   React.useEffect(() => {
     listApplications().then((res) => setApplications(res as Application[]));
   }, []);
 
+  type labelProps = { label: string };
+  function SortableLabel({ label }: labelProps) {
```
jpellizzari, line 29: I would expect we would use this on every `<DataTable />`. What about putting this functionality there?

jpellizzari (commented):
I think we are trying to bake too many edge cases into our base components. I often make one-off styled components to handle the weird edge cases, like so:

```js
const MyWhackyButton = styled(Button)`
  color: palevioletred;
`

export function MyCoolPage() {
  return <MyWhackyButton />
}
```
This keeps the base button very generic and flexible. It also avoid having to come up with clever class names that don't get stepped on by other CSS libraries. This might not be a big factor for us, but back in the day, every CSS framework had a `.bold` class and they all stepped on each other (old man rant over).

ui/components/Button.tsx
joshri: Agreed - looking at the Figma doc I'm only seeing buttons that look like this one - maybe we want to make it the default?

ui/components/Flex.tsx
joshri: Looking at it again, and yes it is a bad and ugly class that will be destroyed 🔥

ui/pages/Applications.tsx
joshri: Should I split the buttons into separate components and pass those down, or have them living in `<ActionBar />` somehow?

jpellizzari: I would say make them `children` of the `<ActionBar />`

ui/components/Button.tsx
jpellizzari: > Agreed - looking at the Figma doc I'm only seeing buttons that look like this one - maybe we want to make it the default?

Maybe, but that is a bit out of scope for this PR. This highlights the need for a component library (the project we have been talking about).

bia:
Nice progress!

The 'Add A New App' type buttons are [Material buttons](https://material.io/components/buttons#outlined-button) - are we not importing those?

The two icons on the top right are placeholders for a [filter bar](https://material.io/components/data-tables#anatomy). I am not sure if that can be right aligned or if we'll have to add a new line vertically.
We need to have a discussion about what goes in there and how to implement it. Is it too soon to create a github issue for that?

Issue #945, opened 2021-10-22
GitOps deletes apps during uninstallation, thereby preventing idempotency


rokshana-b:
**Describe the bug**

When uninstalling gitops components, gitops deletes the app deployed in the cluster. This does not adhere to the idempotency requirement of the product. 

**Environment**
 - gitops: [e.g. v0.3.1-rc1]

**To Reproduce**

1. Run `gitops install`
2. Run `gitops app add . --auto-merge=true`
3. Run `gitops app list` and see app listed
4. Run `gitops uninstall` and see all finalizers get deleted
5. Run `gitops app list` and see error

**Screenshots**

![image.png](https://images.zenhubusercontent.com/606389351ead0c83798a9a73/38cfda73-c1cc-4675-a8c4-67064aed9018)

![image.png](https://images.zenhubusercontent.com/606389351ead0c83798a9a73/a01a761f-7033-4e49-9356-a0c0f6e9e793)

![image.png](https://images.zenhubusercontent.com/606389351ead0c83798a9a73/1a246d73-feb7-49fd-b948-fdb5404fe6d2)

jpellizzari:
Working as designed. We will be enhancing this in #890.

PR #1073, opened 2021-11-13
Ask for confirmation when running `gitops uninstall`.

josecordaz:
Closes: #945 

**What changed?**
- Ask for confirmation before running the command `gitops uninstall`
- Added optional `force` flag to skip the confirmation

**Why?**
- To explicitly inform the user that if continuing the execution, it will remove all applications and related cluster resources.

**How did you test it?**
- Manually and by modifying acceptance tests to include changes

**Release notes**

**Documentation Changes**
Yes

Example:
image

### DIFF
--- a/cmd/gitops/uninstall/cmd.go
+++ b/cmd/gitops/uninstall/cmd.go
@@ -8,7 +8,6 @@ import (
 	"fmt"
 	"io"
 	"os"
-	"strings"
 
 	"github.com/weaveworks/weave-gitops/pkg/flux"
 	"github.com/weaveworks/weave-gitops/pkg/kube"
@@ -52,7 +51,7 @@ func init() {
 
 func uninstallRunCmd(cmd *cobra.Command, args []string) error {
 	if !forceUninstall {
-		fmt.Print("Uninstall will remove all your Applications and any related cluster resources. Are you sure you want to uninstall? (y/n) ")
+		fmt.Print("Uninstall will remove all your Applications and any related cluster resources. Are you sure you want to uninstall? [y/N] ")
 
 		reader := bufio.NewReader(os.Stdin)
 
@@ -61,7 +60,7 @@ func uninstallRunCmd(cmd *cobra.Command, args []string) error {
 			return fmt.Errorf("error reading from stdin %w", err)
 		}
 
-		if !strings.Contains(userInput, "y") {
+		if userInput != "y\n" {
 			return nil
 		}
 	}
--- a/cmd/gitops/uninstall/cmd.go
+++ b/cmd/gitops/uninstall/cmd.go
@@ -4,8 +4,11 @@ package uninstall
 // wego installed, the user will be prompted to install wego and then the repository will be added.
 
 import (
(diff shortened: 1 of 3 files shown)


test/acceptance/test/testsuite_common_test.go
```diff
@@ -34,7 +33,7 @@ func TestAcceptance(t *testing.T) {
 	}
 
 	RegisterFailHandler(Fail)
-	gomega.RegisterFailHandler(GomegaFail)
+	//gomega.RegisterFailHandler(GomegaFail)
```
jrryjcksn, line 36: Is this intended to still be commented out?

jrryjcksn (approved):
LGTM

test/acceptance/test/testsuite_common_test.go
josecordaz: nice catch, thanks

cmd/gitops/uninstall/cmd.go
```diff
@@ -42,10 +46,26 @@ var Cmd = &cobra.Command{
...
+		fmt.Print("Uninstall will remove all your Applications and any related cluster resources. Are you sure you want to uninstall? (y/n) ")
+
+		reader := bufio.NewReader(os.Stdin)
+
+		userInput, err := reader.ReadString('\n')
+		if err != nil && err != io.EOF {
+			return fmt.Errorf("error reading from stdin %w", err)
+		}
+
+		if !strings.Contains(userInput, "y") {
```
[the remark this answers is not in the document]
josecordaz: You are right. I'll follow the logic docker uses to commands like `docker system prune --all` for now.

test/acceptance/test/install_tests.go
```diff
@@ -103,8 +105,22 @@ var _ = Describe("Weave GitOps Install Tests", func() {
...
+			c.Stdout = outputStream
+			c.Stdin = inputUser
+			c.Stderr = os.Stderr
+			err := c.Start()
+			Expect(err).ShouldNot(HaveOccurred())
+
+			Eventually(outputStream).Should(gbytes.Say(`Uninstall will remove all your Applications and any related cluster resources\. Are you sure you want to uninstall\? \[y\/N\]`))
+
+			err = c.Wait()
+			Expect(err).ShouldNot(HaveOccurred())
```
rokshana-b, line 123: @josecordaz we're not verifying any actions _after_ the user provides an input `(y/n)`. Is that on purpose?

rokshana-b (approved):

test/acceptance/test/install_tests.go
josecordaz: Actually, we are. Particularly with this test we are validating that the namespace gets removed after running the uninstall command, this can only happen if the "user" types `y` and then enter which is being emulated in this line https://github.com/weaveworks/weave-gitops/pull/1073/files#diff-93d1ec3b2fa24c2149cf72486d97a7dfae21bad613096ee91d767aa070bfe58fR111.

PR #1078, opened 2021-11-16
Add missed changes to goreleaser for brew

J-Thompson12:
**What changed?**
I was able to get a working version of this code tested. The issue I was having was that the releases were binaries and not tar.gz. This was making it so that brew could not find the binary named gitops.

**Why?**

**How did you test it?**
Put this code on a test repo and got the brew install working.

**Release notes**

**Documentation Changes**

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -86,6 +86,6 @@ brews:
       bash_output = Utils.safe_popen_read(bin/"gitops", "completion", "bash")
       (bash_completion/"gitops").write bash_output
       zsh_output = Utils.safe_popen_read(bin/"gitops", "completion", "zsh")
-      (zsh_completion/"gitops").write zsh_output
+      (zsh_completion/"_gitops").write zsh_output
     test: |
-      system "#{bin}/gitops --version"
+      system "#{bin}/gitops version"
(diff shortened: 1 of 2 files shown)


.goreleaser.yml
```diff
@@ -9,13 +9,13 @@ checksum:
...
   prerelease: auto
 archives:
-  - format: binary
+  - format: tar.gz
     replacements:
       amd64: x86_64
     name_template: "gitops-{{.Os}}-{{.Arch}}"
 builds:
   - <<: &build_defaults
-      binary: "gitops-{{.Os}}-{{.Arch}}"
```
J-Thompson12, line 18: This line plus the format not being tar.gz was where the problem was.

jrryjcksn (approved):
:+1:

josecordaz:
I think these changes are going to break this installation guide, right? https://docs.gitops.weave.works/docs/getting-started/#install-the-weave-gitops-cli.

Can we ask David just to make sure this is the desired behavior?

J-Thompson12:
They will break the install guide. I have a different PR for that. These changes wont go into effect until after we do another release. So I have to wait until the release to change the docs.

josecordaz (approved):
LGTM

PR #1081, opened 2021-11-16
Add leaf-cluster auth ADR

jpellizzari:
Adds an ADR around leaf-cluster auth


doc/adr/0010-leaf-cluster-k8s-auth.md
```diff
@@ -0,0 +1,63 @@
...
+
+This approach won't work for most use cases, as we cannot expect every user of the WeGO UI (and subsequently API) to have access or authorization to view objects on all clusters.
+
+## Design
+
+To allow a user to query data about a Leaf Cluster, we must do (roughly) these steps during the "onboarding" CLI command:
+
+1. Create a `ServiceAccount` on the Leaf Cluster with cluster admin priveleges
+2. Add the `kubeconfig` for that `ServiceAccount` as a `Secret` on the Management Cluster.
+3. Add a `Cluster` Custom Resource to the Management Cluster to persist the `Cluster` record for future retreival
```
bigkevmcd, line 21: If I'm reading this correctly, the `Cluster` CR is added to the management cluster, before it's been approved via PR?

How do you detect rejection of the PR and remove the secret and `Cluster`?

doc/adr/0010-leaf-cluster-k8s-auth.md
```diff
@@ -0,0 +1,63 @@
...
+metadata:
+  name: my-cluster
+  namespace: wego-system
+spec:
+  kubeconfig_secret_name: my-cluster-kubeconfig
+  kubeconfig_secret_namespace: wego-system
+  config_repo_url: ssh://git@gitlab.com/jpellizzari1/my-repo.git
+```
+
+## Consequences
```
bigkevmcd, line 57: This should probably also talk about the security implications?

Also, the lack of authorisation of adding clusters?

doc/adr/0010-leaf-cluster-k8s-auth.md
```diff
@@ -0,0 +1,63 @@
...
+apiVersion: wego.weave.works/v1alpha1
+kind: Cluster
+metadata:
+  name: my-cluster
+  namespace: wego-system
+spec:
+  kubeconfig_secret_name: my-cluster-kubeconfig
+  kubeconfig_secret_namespace: wego-system
+  config_repo_url: ssh://git@gitlab.com/jpellizzari1/my-repo.git
+```
```
bigkevmcd, line 55: I like the CR model.

I suspect the keys should use the "camelcase" convention, rather than "snake", for example from K8s...

```yaml
apiVersion: v1
kind: Pod
metadata:
  name: nginx-demo
spec:
  containers:
  - name: nginx
    image: nginx:1.14.2
    ports:
    - containerPort: 80
```

Both `apiVersion` and `containerPort` are camel-case?

bigkevmcd:
I also think this should include a small section on the alternatives that were considered, and why they were discarded.

doc/adr/0010-leaf-cluster-k8s-auth.md
JamWils: That's a valid question.  Doesn't this issue extend to the current implementation of `gitops install` in general?  If you run the gitops install it will apply the flux CRDs and everything else to the cluster today.  The user would have to run `gitops uninstall` in that scenario to remove everything that was applied to the cluster.

jpellizzari:
> I also think this should include a small section on the alternatives that were considered, and why they were considered not suitable.

@bigkevmcd I don't have any info on which alternatives were considered. Can you or @palemtnrider elaborate on what other options were considered in a comment and I will add it to the doc?

JamWils:
@jpellizzari I would just put at the bottom of the ADR that "We have not considered any alternatives".

jpellizzari:
Closing as no longer relevant

PR #1086, opened 2021-11-17
Gitops add weave-gitops

rokshana-b:
Added yamls for weave-gitops


rokshana-b:
I was trying to generate the help text and forgot to type the `-h` flag - this generated a PR on the weave-gitops repo

PR #1097, opened 2021-11-18
Add custom ssh key setup to fix actions

J-Thompson12:
**What changed?**
Copied over the ssh known_hosts fix from https://github.com/weaveworks/weave-gitops-enterprise/pull/312. This fix is in response to https://github.blog/2021-09-01-improving-git-protocol-security-github/. 

**How did you test it?**
If the pipeline passes its good.

### DIFF
--- a/.github/actions/setup-ssh/action.yml
+++ b/.github/actions/setup-ssh/action.yml
@@ -1,4 +1,3 @@
-
 name: 'Setup SSH'
 description: 'Setup SSH keys and known hosts'
 inputs:
@@ -9,7 +8,7 @@ runs:
   using: "composite"
   steps:
     - name: Set up ssh agent
-      uses: webfactory/ssh-agent@v0.5.3
+      uses: webfactory/ssh-agent@v0.5.2
       with:
         ssh-private-key: ${{ inputs.ssh-private-key }}
     - name: Add known_hosts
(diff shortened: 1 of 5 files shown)


foot:
So it looks like `test` also needs the known_hosts adjustment. We copied the whole:

```
- name: Set up ssh
    uses: ./.github/actions/setup-ssh
    with:
      ssh-private-key: ${{ inputs.github-key }}
```

Into one of our tests, even though it doesn't need the private key. 🤷 . If thats not an option you could untangle the ssh-agent and known hosts bits.

jpellizzari (approved):
LGTM if the pipeline passes

foot:
Hmmm, its a bit weird, looks like its stuck.. 👀

J-Thompson12:
> Hmmm, its a bit weird, looks like its stuck.. 👀

Yeah I was just looking at that but couldnt see why. I just restarted the tests to see if it happens again.

.github/actions/setup-ssh/action.yml
```diff
@@ -0,0 +1,28 @@
...
+description: 'Setup SSH keys and known hosts'
+inputs:
+  ssh-private-key:
+    description: 'ssh keys to load'
+    required: true
+runs:
+  using: "composite"
+  steps:
+    - name: Set up ssh
+      uses: ./.github/actions/setup-ssh
```
foot, line 12: ♾️ !

foot, line 12:
```suggestion
    - name: Set up ssh agent
      uses: webfactory/ssh-agent@v0.5.3
```

J-Thompson12: Ohhh nice catch. I guess I did just find all. Whoops

foot:
Copied the wge one over to see if whitespace issues..

Issue #1098, opened 2021-11-18
After `gitops install` the kustomization is failing because of missing namespaces


foot:
**Describe the bug**
kustomization fails to apply completely?

**What is the severity of the bug**
not sure, stops additional changes to `.weave-gitops` being applied

**Environment**
 - gitops: `main` (ish, probably effects `2f4c8359df0826faf526ab1f861c24e646392d4a`?)
 - kubernetes:  [e.g. 1.20.4]
    - [x] KinD - _version_

**To Reproduce**
1. `gitops install --app-config-url ...`
2. `kubectl get kustomization -A`

See errors:
```
NAMESPACE     NAME               READY   STATUS                                                                                                      AGE
wego-system   kind-kind-system   False   Role/resources-reader namespace not specified, error: the server could not find the requested resource...   35m
wego-system   kind-kind-user     True    Applied revision: main/9838e888e864212b5b918bcd297c4d33d0f9f078                                             35m
```
details:
```
status:
  conditions:
  - lastTransitionTime: "2021-11-18T18:06:48Z"
    message: |
      Role/resources-reader namespace not specified, error: the server could not find the requested resource
    reason: ReconciliationFailed
    status: "False"
    type: Ready
  lastAttemptedRevision: main/9838e888e864212b5b918bcd297c4d33d0f9f078
  observedGeneration: 1
```

**Expected behavior**
No errors

**Workaround**

```patch
diff --git a/.weave-gitops/clusters/kind-kind/system/wego-app.yaml b/.weave-gitops/clusters/kind-kind/system/wego-app.yaml
index 78b514c..b56f75d 100644
--- a/.weave-gitops/clusters/kind-kind/system/wego-app.yaml
+++ b/.weave-gitops/clusters/kind-kind/system/wego-app.yaml
@@ -27,6 +27,7 @@ apiVersion: rbac.authorization.k8s.io/v1
 kind: RoleBinding
 metadata:
   name: read-resources
+  namespace: wego-system
 subjects:
   - kind: ServiceAccount
     name: wego-app-service-account
@@ -40,6 +41,7 @@ apiVersion: rbac.authorization.k8s.io/v1
 kind: Role
 metadata:
   name: resources-reader
+  namespace: wego-system
 rules:
   - apiGroups: [""]
     resources: ["secrets"]
```

PR #1099, opened 2021-11-18
Set missing namespace on role/rolebinding

foot:
- Makes kustomization-controller happy

Closes: #1098 

**What changed?**

Added some namespaces

**Why?**

To make kustomization-controller happy. I don't know if these are the correct namespaces though (could be default).

**How did you test it?**

Locally.


foot:
> The roleRef field doesn't support namespace. You can use roleRef either with ClusterRoles which are not namespaced or with Roles which always has to be in the same namespace as the RoleBinding.

So should probably be in the `wego-system` ns. 🤔

josecordaz (approved):
LGTM

jpellizzari (approved):

PR #1110, opened 2021-11-19
Refactor tests to prevent nightly failures

rokshana-b:
**What changed?**

Refactored gitlab tests to ignore cluster-context so nightly workflow could pass.

**Why?**

A recent set of gitlab test additions were verifying `app deletion message` which includes the name of the cluster. This does not cause any problem in workflows other than the nightly. The cluster name for the nightly tests causes failures since the cluster-context for those clusters are different than what the existing function actually captures.

image

**How did you test it?**
Ran nightly workflow on this branch


jrryjcksn (approved):
LGTM

PR #1112, opened 2021-11-19
Pass user token to API via an HTTP header

yiannistri:
Closes: weaveworks/weave-gitops-enterprise#203

**What changed?**
- Refactored out code that returns the git provider token (either via an env var or via the git provider auth flow)
- Pass the git provider token to the API via an HTTP header

**Why?**
This is so that the Weave GitOps Enterprise API can make calls to the git provider on behalf of the end user.

**How did you test it?**
Tested locally + unit tests

**Release notes**
N/A

**Documentation Changes**
N/A

### DIFF
--- a/cmd/gitops/add/clusters/cmd_test.go
+++ b/cmd/gitops/add/clusters/cmd_test.go
@@ -3,7 +3,6 @@ package clusters_test
 import (
 	"encoding/json"
 	"net/http"
-	"os"
 	"testing"
 
 	"github.com/go-resty/resty/v2"
@@ -12,6 +11,7 @@ import (
 	"github.com/weaveworks/weave-gitops/cmd/gitops/cmderrors"
 	"github.com/weaveworks/weave-gitops/cmd/gitops/root"
 	"github.com/weaveworks/weave-gitops/pkg/adapters"
+	"github.com/weaveworks/weave-gitops/pkg/testutils"
 )
 
 func TestSetSeparateValues(t *testing.T) {
@@ -145,8 +145,7 @@ func TestEndpointNotSet(t *testing.T) {
 }
 
 func TestGitProviderToken(t *testing.T) {
-	os.Setenv("GITHUB_TOKEN", "test-token")
-	defer os.Unsetenv("GITHUB_TOKEN")
+	t.Cleanup(testutils.Setenv("GITHUB_TOKEN", "test-token"))
 
 	client := resty.New()
 
--- a/cmd/gitops/add/clusters/cmd_test.go
+++ b/cmd/gitops/add/clusters/cmd_test.go
@@ -3,11 +3,13 @@ package clusters_test
 import (
 	"encoding/json"
 	"net/http"
+	"os"
 	"testing"
 
 	"github.com/go-resty/resty/v2"
 	"github.com/jarcoal/httpmock"
 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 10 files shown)


cmd/gitops/delete/clusters/cmd_test.go
```diff
@@ -24,6 +26,9 @@ func TestEndpointNotSet(t *testing.T) {
 }
 
 func TestPayload(t *testing.T) {
+	os.Setenv("GITHUB_TOKEN", "test-token")
```
jpellizzari, line 29: Same as above with saving the previous value.

cmd/internal/provider.go
```diff
@@ -30,30 +31,9 @@ func NewGitProviderClient(stdout *os.File, lookupEnvFunc func(key string) (strin
 // GetProvider returns a GitProvider containing either the token stored in the <git provider>_TOKEN env var
 // or a token retrieved via the CLI auth flow
 func (c *gitProviderClient) GetProvider(repoUrl gitproviders.RepoURL, getAccountType gitproviders.AccountTypeGetter) (gitproviders.GitProvider, error) {
-	tokenVarName, err := getTokenVarName(repoUrl.Provider())
+	token, err := GetToken(repoUrl, c.stdout, c.lookupEnvFunc, c.authHandlerFunc, c.log)
```
jpellizzari, line 34: :tada: Much cleaner

pkg/adapters/http.go
```diff
@@ -245,6 +245,7 @@ func (c *HTTPClient) CreatePullRequestFromTemplate(params capi.CreatePullRequest
 
 	res, err := c.client.R().
 		SetHeader("Accept", "application/json").
+		SetHeader("Git-Provider-Token", params.GitProviderToken).
```
jpellizzari, line 248: Should this reference a `const` that was set somewhere? I think there is a `GitProviderTokenHeaderName` or something...somewhere.

jpellizzari (commented):
Looks good. The only thing holding me up is the unsetting of an env variable.

cmd/gitops/add/clusters/cmd_test.go
```diff
@@ -141,3 +143,77 @@ func TestEndpointNotSet(t *testing.T) {
 	err := cmd.Execute()
 	assert.EqualError(t, err, "the Weave GitOps Enterprise HTTP API endpoint flag (--endpoint) has not been set")
 }
+
+func TestGitProviderToken(t *testing.T) {
+	os.Setenv("GITHUB_TOKEN", "test-token")
```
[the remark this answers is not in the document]
yiannistri: So I've tested this with my own GITHUB_TOKEN and it seemed unaffected after running these tests. Also in the next line I'm resetting that env var.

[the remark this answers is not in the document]
jpellizzari: So long as `os.Unsetenv()` reset it to the previous value, or at least doesn't change the user's environment :+1:

jpellizzari (approved):
LGTM with the caveat that the user's environment doesn't get changed.

yiannistri:
Thanks @jpellizzari  for the review. While I haven't had any issues locally, I'll change this to make it more robust. Bonus [test helper](https://pkg.go.dev/testing#T.Setenv) for when we switch to Go 1.17 🎉

Issue #1075, opened 2021-11-16
A user can break Weave GitOps with a poorly chosen app name


jrryjcksn:
**Describe the bug**
We generate GitRepository "sources" for both the general directory structure management and for individual applications. As currently implemented, a user can select an application name that will produce a GitRepository which overwrites the generic directory structure one. We don't adequately segregate the cluster and application source namespaces (not to be confused with Kubernetes namespaces). 

**Environment**
 - gitops: Unreleased

**Affects versions**
Only yet to be released versions

**To Reproduce**
1. Do a `gitops install --app-config-url=<a github url>`
2. Add an application specifying the name: `wego-github-<base name of url>`

**Expected behavior**
The expected behavior is the application and the GitOps runtime behave normally.

PR #1113, opened 2021-11-19
Add set text to source to avoid same name errors

J-Thompson12:
Closes: #1075 

**What changed?**
Add auto to flux sourceRef and app to application sourceRef. This forces the user from being able to use the same name for the two sources. Changes the ref name in flux-user-kustomization-resource.yaml, flux-system-kustomization-resource.yaml, nginx-gitops-deploy.yaml and nginx-gitops-source.yaml.

I do want to point out that the get status output has changed a bit with these changes. It uses the source name so it looks like:
```
NAME                            READY   MESSAGE                                                         REVISION                                        SUSPENDED 
gitrepository/wego-app-nginx    True    Fetched revision: main/f69237519da2d3f242f978c424591d2e1810a023 main/f69237519da2d3f242f978c424591d2e1810a023   False    

NAME                            READY   MESSAGE                                                         REVISION                                        SUSPENDED 
kustomization/wego-app-nginx    True    Applied revision: main/f69237519da2d3f242f978c424591d2e1810a023 main/f69237519da2d3f242f978c424591d2e1810a023   False    
```

**How did you test it?**
Manually got an app deployed. Also updated the unit tests. The existing acceptance tests will cover these changes.

**Release notes**

**Documentation Changes**

### DIFF
--- a/pkg/services/automation/generator.go
+++ b/pkg/services/automation/generator.go
@@ -139,9 +139,11 @@ func (a *AutomationGen) generateAppSource(ctx context.Context, app models.Applic
 		return AutomationManifest{}, err
 	}
 
+	sourceName := createAppSourceName(app.Name)
+
 	switch app.SourceType {
 	case models.SourceTypeGit:
-		source, err = a.Flux.CreateSourceGit(app.Name, app.GitSourceURL, app.Branch, appSecretRef.String(), app.Namespace)
+		source, err = a.Flux.CreateSourceGit(sourceName, app.GitSourceURL, app.Branch, appSecretRef.String(), app.Namespace)
 		if err == nil {
 			source, err = AddWegoIgnore(source)
 		}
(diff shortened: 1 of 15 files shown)


pkg/services/automation/generator.go
```diff
@@ -399,6 +403,19 @@ func CreateRepoSecretName(gitSourceURL gitproviders.RepoURL) GeneratedSecretName
 	return GeneratedSecretName(lengthConstrainedName)
 }
```
jrryjcksn, line 405: I think it would be nice to call out in a comment here that the two source names need to be distinct (maybe with a reference to the bug) so that it won't get inadvertently undone later by someone without context.

jrryjcksn (approved):
:+1:

pkg/services/automation/generator.go
J-Thompson12: Added the comment.

jpellizzari (approved):
Nice work

J-Thompson12:
Closing this in favor of https://github.com/weaveworks/weave-gitops/pull/1171

PR #1115, opened 2021-11-19
Allow removing app via PR

josecordaz:
Closes: 

**What changed?**

**Why?**

**How did you test it?**

**Release notes**

**Documentation Changes**

### DIFF
--- a/pkg/services/applicationv2/remover_test.go
+++ b/pkg/services/applicationv2/remover_test.go
@@ -1,0 +1,44 @@
+package applicationv2
+
+import (
+	"context"
+
+	"github.com/weaveworks/weave-gitops/pkg/gitproviders/gitprovidersfakes"
+
+	"github.com/weaveworks/weave-gitops/pkg/git"
+	"github.com/weaveworks/weave-gitops/pkg/git/wrapper/wrapperfakes"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("Remover", func() {
+	Describe(".Remove()", func() {
+		It("removes application auto-merge true", func() {
+			ctx := context.Background()
+
+			fakeGit := &wrapperfakes.FakeGit{}
+			fakeGitClient := git.New(nil, fakeGit)
+
+			orgProvider := &gitprovidersfakes.FakeGitProvider{}
+
+			remover := NewRemover(fakeGitClient, orgProvider)
+
+			err := remover.Remove(ctx, "my-app", "wego-system", true)
+			Expect(err).NotTo(HaveOccurred())
+		})
+		It("removes application auto-merge false", func() {
+			ctx := context.Background()
+
+			fakeGit := &wrapperfakes.FakeGit{}
+			fakeGitClient := git.New(nil, fakeGit)
+
+			orgProvider := &gitprovidersfakes.FakeGitProvider{}
+
+			remover := NewRemover(fakeGitClient, orgProvider)
+
+			err := remover.Remove(ctx, "my-app", "wego-system", false)
+			Expect(err).NotTo(HaveOccurred())
+		})
+	})
+})
--- a/pkg/flux/flux.go
+++ b/pkg/flux/flux.go
@@ -163,7 +163,7 @@ func (f *FluxClient) CreateKustomization(name string, source string, path string
 		"--path", path,
 		"--source", source,
 		"--namespace", namespace,
-		"--prune", "true",
+		"--prune", "false",
 		"--interval", "1m",
 		"--export",
 	}
--- a/pkg/flux/flux.go
+++ b/pkg/flux/flux.go
@@ -163,7 +163,7 @@ func (f *FluxClient) CreateKustomization(name string, source string, path string
 		"--path", path,
 		"--source", source,
 		"--namespace", namespace,
-		"--prune", "false",
+		"--prune", "true",
 		"--interval", "1m",
 		"--export",
 	}
--- a/pkg/services/applicationv2/remover.go
+++ b/pkg/services/applicationv2/remover.go
@@ -1,0 +1,32 @@
+package applicationv2
+
+import (
+	"context"
+
+	"github.com/weaveworks/weave-gitops/pkg/gitproviders"
+
+	"github.com/weaveworks/weave-gitops/pkg/git"
+)
+
+//go:generate go run github.com/maxbrunsfeld/counterfeiter/v6 -generate
+
+//counterfeiter:generate . Fetcher
+type Remover interface {
+	Remove(ctx context.Context, appName string, namespace string, autoMerge bool) error
+}
+
+func NewRemover(gitClient git.Git, gitProvider gitproviders.GitProvider) Remover {
+	return remover{
+		gitClient:   gitClient,
+		gitProvider: gitProvider,
+	}
+}
+
+type remover struct {
+	gitClient   git.Git
+	gitProvider gitproviders.GitProvider
+}
+
+func (r remover) Remove(ctx context.Context, appName string, namespace string, autoMerge bool) error {
+	return nil
+}


pkg/services/applicationv2/remover_test.go
```diff
@@ -0,0 +1,44 @@
...
+
+			fakeGit := &wrapperfakes.FakeGit{}
+			fakeGitClient := git.New(nil, fakeGit)
+
+			orgProvider := &gitprovidersfakes.FakeGitProvider{}
+
+			remover := NewRemover(fakeGitClient, orgProvider)
+
+			err := remover.Remove(ctx, "my-app", "wego-system", true)
+			Expect(err).NotTo(HaveOccurred())
```
jpellizzari, line 28: We need to check on what other side effects we think should or shouldn't happen. Since this is `auto-merge` can we check the local git repository to make sure it equals our expected state?

pkg/services/applicationv2/remover_test.go
```diff
@@ -0,0 +1,44 @@
...
+			ctx := context.Background()
+
+			fakeGit := &wrapperfakes.FakeGit{}
+			fakeGitClient := git.New(nil, fakeGit)
+
+			orgProvider := &gitprovidersfakes.FakeGitProvider{}
+
+			remover := NewRemover(fakeGitClient, orgProvider)
+
+			err := remover.Remove(ctx, "my-app", "wego-system", false)
```
jpellizzari, line 40: Same as above, lets see if the PR that was created matches some expected outcome.

jpellizzari (commented):
The `interface` looks good. I would say we can test `auto-merge` with a real git repo (created locally) and see if it does what we expect. Rather than giving it a `FakeGit{}`. could we give it a real git repo in a local temp dir?

Testing the pull request case in a meaningful way will be harder, but happy to help on that.

josecordaz:
@jpellizzari I like the idea of using real repositories locally, I think it would add more accuracy to our tests. However the functionality we will be implementing, at some point, is pushing those changes to the app config repository. It seems like we currently don't create repositories out of the acceptance tests anywhere. So I see two possible solutions:

1) Bring inside remover_test.go the functionality to create repositories we have in the acceptance tests 
2) Use the fake git interface (where you could check that the commit and/or push methods are called with the appropriate information). 

What do you think would be ok ?

jpellizzari:
@josecordaz When I say "create an actual git repo", I don't mean one on GitHub. Can we create a local repo via `go-git`? I don't think unit tests should rely on a third-party service like GitHub.

josecordaz:
@jpellizzari The functionality we will be implementing will **push** changes to a remote repository at some point, that is the real use case. If the intention is unit test the `Remove` method, using a local repository, how could we prevent the application from failing at that point of execution? Maybe you are referring to creating a fake git client that does real commits but when we push it, it executes some "fake code"? I'm not sure what kind of git client we can use to do real commits and fake push operations. Maybe we need a new implementation?

josecordaz:
@jpellizzari Maybe this is what you are referring to? https://github.com/weaveworks/weave-gitops/blob/main/pkg/services/app/add_test.go#L111

jpellizzari:
@josecordaz How about something like this: https://github.com/weaveworks/weave-gitops/pull/1125/files

I have no idea if this will work, but this is sort of what I am getting at. We want to validated that the business logic of the code is working as expected. We cannot just check for errors.

LMK if you want to hop on a call to discuss.

jpellizzari:
We may need to look at this interface as well as an argument to our constructor:

https://github.com/weaveworks/weave-gitops/blob/main/pkg/services/gitopswriter/gitopswriter.go#L29

josecordaz:
> @josecordaz How about something like this: https://github.com/weaveworks/weave-gitops/pull/1125/files
> 
> I have no idea if this will work, but this is sort of what I am getting at. We want to validated that the business logic of the code is working as expected. We cannot just check for errors.
> 
> LMK if you want to hop on a call to discuss.

I agree that we can't rely on just checking errors. But we needed a way to balance fake and real git operations.

This looks similar to what I was thinking about.

jpellizzari:
> But we needed a way to balance fake and real git operations.

They are all "real" operations. You can use `git` without a remote, so we can check everything up to the point where the repo gets pushed somewhere.

In the case of testing operations done via pull request, we have this at our disposal: https://github.com/weaveworks/weave-gitops/blob/main/pkg/vendorfakes/fakehttp/round_tripper.go#L27

josecordaz:
> > But we needed a way to balance fake and real git operations.
> 
> They are all "real" operations. You can use `git` without a remote, so we can check everything up to the point where the repo gets pushed somewhere.

We can't do push operations here, that would fall under a third-party service use, which we don't want. We can't push without a remote, unless we fake the remote as well. But seems easier to fake one of the interfaces.

> 
> In the case of testing operations done via pull request, we have this at our disposal: https://github.com/weaveworks/weave-gitops/blob/main/pkg/vendorfakes/fakehttp/round_tripper.go#L27

I like this RoundTripper, it seems helpful for http operations. Probably one way to fake push operations. I would argue that it is easier if we just fake the push operation somehow using one of the interfaces we already have. Something similar to this:
```go
ctx := context.Background()

dir := os.TempDir()

realGitClient := git.New(nil, wrapper.NewGoGit())

fakeGit := &gitfakes.FakeGit{}
// We can use all the real git operation in each method
fakeGit.CloneCalls(func(ctx context.Context, s string, s2 string, s3 string) (bool, error) {
    return realGitClient.Clone(ctx, s, s2, s3)
})
fakeGit.CommitCalls(func(commit git.Commit, f ...func(string) bool) (string, error) {
    return realGitClient.Commit(commit, f...)
})
fakeGit.PushCalls(func(ctx context.Context) error {
    // Here we can skip the real push to avoid failures
    // and still validate the real changes made through commits
    // in the local repository
    return nil
})

// we need this client as well to get the config repo url from the schema of the app (it needs more adjustments still)
k8sClient := fake.NewClientBuilder().WithScheme(kube.CreateScheme()).Build()

orgProvider := &gitprovidersfakes.FakeGitProvider{}

remover := NewRemover(fakeGit, orgProvider, k8sClient)

err := remover.Remove(ctx, "my-app", "wego-system", true)

// After this we can still use stat to validate the changes
appPath := fmt.Sprintf("%s/.wego-system/apps/my-app.yaml", dir)
_, err = os.Stat(appPath)
Expect(err).NotTo(BeNil(), "error should not be nil")
Expect(os.IsNotExist(err)).To(BeTrue(), "the app yaml should not exist in the repo")
```

I might think that this approach would also make it easier to use `RepoWritter` and/or `GitOpsDirectoryWritter`. GitOpsDirectoryWritter already has the method `Remove`. I'll keep investigating to see if it will be helpful using it here.

For PRs interaction, I guess we could use the fake client for `gitprovider.PullRequestClient` instead. This is a place where we are already using it https://github.com/weaveworks/weave-gitops/blob/main/pkg/gitproviders/provider_org_test.go#L32.

jpellizzari:
@josecordaz I think we are on the same page as far as the approach goes. Feel free to start on implementation; please be sure to utilize the existing `GitOpsRepoWriter` and/or `RepoWriter` where appropriate.

Regarding your approach to testing, I think having a `fakeGitClient` that then does `realGitClient` under the hood is a little confusing. Imagine someone reading this code in a few months and trying to work out what/why this is happening. Please take another look at my approach that uses some quick composition to achieve the same thing.

josecordaz:
After further analysis, I think maybe I could be doing duplicate code.

I am aware that you are wanting to take this opportunity for me to familiarize myself with the new architecture but I believe that the current logic to eliminate an application already has the new architecture implemented. I have a feeling that maybe all I need to do is implement the necessary logic for the `auto-merge` flag as well as add/modify the corresponding unit tests. I mention this because the `gitOpsDirectoryWriterSvc` service already contains the` Remove` method and even already has the `auto-merge` parameter, although not yet implemented. 

What would be your thoughts on this @jpellizzari.

PR #1116, opened 2021-11-22
Adds capi-app when upgrading

foot:
**What changed?**

- We remove the `wego-app.yaml` as that functionality is provided by EE.
- Add a new kustomization that syncs the capi path

**Why?**

- Last pieces of the directory structure ADR that specifies capi paths

**How did you test it?**

- Manually


sarataha (approved):
LGTM! 🥇 🌟

PR #1146, opened 2021-11-30
Fixed adding a Gitlab app via UI

josecordaz:
Closes: 

**What changed?**
Interpret integer as seconds

**Why?**
It was interpreting the integer as milliseconds instead of seconds, causing unauthorized errors as the token only had a few seconds of validity. 

**How did you test it?**
I manually added an app via UI, getting the next message afterwards:
image

**Release notes**

**Documentation Changes**

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -633,7 +633,7 @@ func (s *applicationServer) AuthorizeGitlab(ctx context.Context, msg *pb.Authori
 		return nil, fmt.Errorf("could not exchange code: %w", err)
 	}
 
-	token, err := s.jwtClient.GenerateJWT(time.Duration(tokenState.ExpiresIn), gitproviders.GitProviderGitLab, tokenState.AccessToken)
+	token, err := s.jwtClient.GenerateJWT(time.Duration(tokenState.ExpiresIn)*time.Second, gitproviders.GitProviderGitLab, tokenState.AccessToken)
 	if err != nil {
 		return nil, fmt.Errorf("could not generate token: %w", err)
 	}


jpellizzari (approved):
Nice find!

Edit: Nit: can we add some test coverage? Like maybe we should validate that the expiration is what we expect it to be

pkg/server/server.go
```diff
@@ -633,7 +633,7 @@ func (s *applicationServer) AuthorizeGitlab(ctx context.Context, msg *pb.Authori
 		return nil, fmt.Errorf("could not exchange code: %w", err)
 	}
 
-	token, err := s.jwtClient.GenerateJWT(time.Duration(tokenState.ExpiresIn), gitproviders.GitProviderGitLab, tokenState.AccessToken)
+	token, err := s.jwtClient.GenerateJWT(time.Duration(tokenState.ExpiresIn)*time.Second, gitproviders.GitProviderGitLab, tokenState.AccessToken)
```
palemtnrider, line 636: Nit but multiplying by seconds here makes the code a bit obscure.  From looking at the code it appears that only GitLab (today) uses the ExpiresIn field.  However, others might.  The code is more readable (to me) if we add a function on tokenState to return the expires value in seconds - or another field ExpiresInSeconds. Then the individual providers can provide the conversion - if necessary.

palemtnrider (approved):
Good catch!  Minor nit on the fix.
+1 on adding a test(s)

pkg/server/server.go
bigkevmcd: You don't need the type assertion `time.Duration` either, because multiplying an integer by time.Duration will get you a time.Duration.

Issue #1079, opened 2021-11-16
Profiles API


Himangini:
### Description
Based on the profiles improvements suggested [here](https://gist.github.com/bigkevmcd/0f39d65b6bb31e9a1a6a69db6d5c2cbc) , we need to move the two profiles endpoints from `cluster` service in `weave-gitops-enterprise` to profiles API in `weave-gitops`

```
service ProfilesService {
  // GetProfiles returns a list of profiles
  // from the cluster.
  rpc GetProfiles(GetProfilesRequest)
    returns (GetProfilesResponse) {
      option (google.api.http) = {
        get: "/v1/profiles"
      };
    }

  // GetProfileValues returns the contents of the `values.yaml` for a Profile
  // with a specific version.
  rpc GetProfileValues(GetProfileValuesRequest)
    returns (google.api.HttpBody) {
      option (google.api.http) = {
        get: "/v1/profiles/{profile_name}/{profile_version}/values"
      };
    }
}
```

### Scope
The code is already written, we need to move it from WGE to wego-core. Make sure it works with WGE. If there are tests written, they should be moved too, if they are missing and then we'll have to add them.

### Acceptance Criteria
As a developer
I want to move profile endpoints from `cluster` service in WGE to profiles api in `weave-gitops`
So that I can call the endpoints from Profiles API

Scenario: `GetProfiles` endpoint called from weave-gitops
Given that I am a developer
When I send a `Get` request to `/v1/profiles`
Then the endpoint should return the list of profiles

Scenario: `GetProfilesValues` endpoint called from weave-gitops
Given that I am a developer
When I send a `Get` request to `/v1/profiles/{profile_name}/{profile_version}/values`
Then the endpoint should return the contents of the `values.yaml` for a Profile with a specific version

Scenario: `GetProfiles` endpoint called from weave-gitops-enterprise
Given that I am a UI developer
When I send a `Get` request to `/v1/profiles`
Then the endpoint should return the list of profiles

Scenario: `GetProfilesValues` endpoint called from weave-gitops-enterprise
Given that I am a UI developer
When I send a `Get` request to `/v1/profiles/{profile_name}/{profile_version}/values`
Then the endpoint should return the contents of the `values.yaml` for a Profile with a specific version

### Technical Notes

- Find the current implementation of two endpoints [here](https://github.com/weaveworks/weave-gitops-enterprise/blob/main/cmd/clusters-service/api/capi_server.proto)
[...]

Issue #1144, opened 2021-11-29
Scanning Helm Repositories


Himangini:
### Description
As suggested [here](https://gist.github.com/bigkevmcd/0ff7197e0a902c3dcde4a1e7bfc3984c#scanning-helmrepositories), we need to create a service which watches HelmRepositories scans for Profiles and caches the result and services the result via the defined APIs.

### Scope
- This applies to Profiles only.
- The Cache needs not persist beyond the lifecycle of the container unless configured otherwise
- Allow the user to specify the cache dir 
- We need to cache the `values.yaml` file 

### Acceptance Criteria
As a developer
I want to scan for profiles and 
I want to cache the result on the file system 
So that I prevent the same profile downloaded multiple times 

### Techincal Notes
- Nice discussion [here](https://weaveworks.slack.com/archives/C01C7FH7EPQ/p1638202268087500)

davidstauffer:
LGTM.

aclevername:
As part of caching we can also consider reducing the cost of interacting with the charts- https://github.com/weaveworks/weave-gitops/pull/1148#discussion_r759540336

Himangini:
Atm, we don't watch ~multiple~ any Helm repository.  The caching mechanism should be implemented keeping in mind that in the future multiple repositories will be watched and we'd cache all of them.

Skarlso:
Caching details:

Some detail about the cache, how it works, and what it caches including the watcher's job in all of this.

# The Watcher ( who watches the watcher? )

The watcher runs as a controller similar to helm-controller. It watches HelmRepositories for changes or creation events. The difference between this and the helm-controller is that we will use the cache and this watcher to notify people about a new version being available for a profile. And we cache the results from the watch. The Watcher is the one fetching the index file and downloading the values file. It does that only when there is a change in the Revision section of the HelmRepository's `Artifact` field. This will prevent us from continuously downloading the same thing over and over again on each request. ( In the tone of Borat ): Is very nice.

# What we cache

We only cache profile related information. That is this thing:

```go
type Profile struct {
	state         protoimpl.MessageState
	sizeCache     protoimpl.SizeCache
	unknownFields protoimpl.UnknownFields

	// The name of the Profile
	Name string `protobuf:"bytes,1,opt,name=name,proto3" json:"name,omitempty"`
	// The URL to a relevant project page, git repo, or contact person
	Home string `protobuf:"bytes,2,opt,name=home,proto3" json:"home,omitempty"`
[...]

PR #1148, opened 2021-11-30
Add profiles API to query for available profiles

aclevername:
Closes: https://github.com/weaveworks/weave-gitops/issues/1079

Paired with @nikimanoledaki (please direct any questions to the pair of us :smile: )

As outlined in https://github.com/weaveworks/weave-gitops/issues/1079 this introduced two new endpoints:
- `/v1/profiles` to get a list of profiles available
- `/v1/profiles/<name>/<version>/values` to get the values file for a profile

q: What is a profile?
a: A helm chart with the `weave.works/profile: <value>` annotation. [Example](https://github.com/weaveworks/profiles-examples/blob/gh-pages/index.yaml)

q: Which helm repository does it look at to find the profiles?
a: You configure which `HelmRepository` resource it looks at through arguments to the `profiles-server` binary. It defaults to `default/weaveworks-charts`. We've updated the manifests so that it will change `default` to whatever namespace weave-gitops is installed in.

## Example

with port forwarding setup to the wego-app pod
```
curl localhost:8000/v1/profiles | jq .profiles
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   424  100   424    0     0  20498      0 --:--:-- --:--:-- --:--:-- 21200
[
  {
    "name": "podinfo",
    "home": "https://github.com/stefanprodan/podinfo",
    "sources": [
      "https://github.com/stefanprodan/podinfo"
    ],
    "description": "Podinfo Helm chart for Kubernetes",
    "keywords": [],
    "maintainers": [
      {
        "name": "stefanprodan",
        "email": "stefanprodan@users.noreply.github.com",
        "url": ""
      }
    ],
    "icon": "",
    "annotations": {},
    "kubeVersion": ">=1.19.0-0",
    "helmRepository": null,
    "availableVersions": [
      "6.0.0",
      "6.0.1"
    ]
  }
]
```

```
curl localhost:8000/v1/profiles/podinfo/6.0.1/values | jq -r .values | base64 --decode
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  4041    0  4041    0     0  10896      0 --:--:-- --:--:-- --:--:-- 10892
# Default values for podinfo.

replicaCount: 1
logLevel: info
host: #0.0.0.0
backend: #http://backend-podinfo:9898/echo
backends: []

image:
  repository: ghcr.io/stefanprodan/podinfo
  tag: 6.0.1
  pullPolicy: IfNotPresent

ui:
  color: "#34577c"
  message: ""
.... 
```

**Why?**
So that profiles can be consumed by other weave-gitops components 

**How did you test it?**
unit & acceptance test added

**Release notes**
- Introduce new `profiles-server` deployment to provide ability to query information about profiles

**Documentation Changes**
Docs already exist for using this in the context of enterprise- would folks like more docs to be added regarding this? Let us know :smile: 

## Help needed!
This is our first time contributing to weave-gitops, so please let us know if we've done anything wrong, weather it be the commit language, code style, testing etc

We have a couple of open questions:
- We've added a new profile deployments instead of having the profiles API be part of the wego-app deployment. This was just to ensure a clean separation, but we understand this increases the footprint of installing weave-gitops. Let us know what you think and if you'd like it to change :smile: 
- Docker images! We've currently manually built the new docker image, and pushed it to our own personal registries. Do we have a process for onboarding new images? Where should we push them to? Do we need to setup some extra automation somewhere so they are automatically built?


Dockerfile.profiles
```diff
@@ -0,0 +1,10 @@
+# Go build
```
aclevername, line 1: We weren't sure where folks would want the Dockerfile to live. How would folks feel about us creating a `build/` directory, containing the two docker files? Or any other suggestions?

pkg/helm/charts.go
```diff
@@ -0,0 +1,289 @@
...
+
+type ChartPredicate func(*repo.ChartVersion) bool
+
+// Profiles is a predicate for scanning charts with the ProfileAnnotation.
+var Profiles = func(v *repo.ChartVersion) bool {
+	return hasAnnotation(v.Metadata, ProfileAnnotation)
+}
+
+// ScanCharts filters charts using the provided predicate.
+// TODO: Add caching based on the Status Artifact Revision.
```
aclevername, line 70: #1144

pkg/helm/charts.go
```diff
@@ -0,0 +1,289 @@
...
+	var secret corev1.Secret
+	if err := kc.Get(ctx, types.NamespacedName{Name: hr.Spec.SecretRef.Name, Namespace: hr.Namespace}, &secret); err != nil {
+		return "", "", fmt.Errorf("repository authentication: %w", err)
+	}
+
+	return string(secret.Data["username"]), string(secret.Data["password"]), nil
+}
+
+func (h *RepoManager) envSettings() *cli.EnvSettings {
+	conf := cli.New()
```
hiddeco, line 225: Not sure you want to do this as it has quite a lot of side effects and default assumptions that you likely not want to have in a service, even if they seem harmless at present: https://github.com/helm/helm/blob/main/pkg/cli/environment.go#L73

manifests/profiles-server/deployment.yaml.tpl
```diff
@@ -0,0 +1,30 @@
...
+  template:
+    metadata:
+      labels:
+        app: profiles-server
+    spec:
+      serviceAccountName: profiles-server-service-account
+      containers:
+        - name: profiles-server
+          # TODO: change
+          image: aclevernameww/profiles-server:{{.ProfilesVersion}}
```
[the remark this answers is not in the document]
jpellizzari:
```suggestion
          # TODO: change
          image: ghrc.io/weaveworks/profiles-server::{{.ProfilesVersion}}
```

pkg/helm/charts_test.go
```diff
@@ -0,0 +1,301 @@
...
+				Expect(profiles).To(BeEmpty())
+			})
+		})
+
+		When("server isn't a valid helm repository", func() {
+			It("errors", func() {
+				testServer := httptest.NewServer(http.FileServer(http.Dir("testdata")))
+				_, err := repoManager.GetCharts(context.TODO(), makeTestHelmRepository(testServer.URL), helm.Profiles)
+				Expect(err).To(MatchError(ContainSubstring("fetching profiles from HelmRepository testing/test-ns")))
+				Expect(err).To(MatchError(ContainSubstring("404")))
```
jpellizzari, line 85: Checking for a substring here might be a little problematic. Can we instead check the status code of the response?

pkg/helm/charts_test.go
```diff
@@ -0,0 +1,301 @@
...
+				Expect(err).NotTo(HaveOccurred())
+				Expect(profiles).To(BeEmpty())
+			})
+		})
+
+		When("server isn't a valid helm repository", func() {
+			It("errors", func() {
+				testServer := httptest.NewServer(http.FileServer(http.Dir("testdata")))
+				_, err := repoManager.GetCharts(context.TODO(), makeTestHelmRepository(testServer.URL), helm.Profiles)
+				Expect(err).To(MatchError(ContainSubstring("fetching profiles from HelmRepository testing/test-ns")))
```
jpellizzari, line 84: Will this test break if someone changes the error message or wraps the error slightly differently?

pkg/helm/charts_test.go
```diff
@@ -0,0 +1,301 @@
...
+				Expect(err).To(MatchError(ContainSubstring("404")))
+			})
+		})
+
+		When("the URL is invalid", func() {
+			It("errors", func() {
+				_, err := repoManager.GetCharts(context.TODO(), makeTestHelmRepository("http://[::1]:namedport/index.yaml"), helm.Profiles)
+				Expect(err).To(MatchError(ContainSubstring("invalid port")))
```
jpellizzari, line 92: Same as my above comments. I would feel better about checking for an error type here.

pkg/server/profiles_test.go
```diff
@@ -0,0 +1,182 @@
...
+			})
+		})
+
+		When("the HelmRepository doesn't exist", func() {
+			It("errors", func() {
+				_, err := s.GetProfileValues(context.TODO(), &pb.GetProfileValuesRequest{})
+				Expect(err).To(MatchError("HelmRepository \"default\"/\"helmrepo\" does not exist"))
+				Expect(err).To(BeAssignableToTypeOf(&grpcruntime.HTTPStatusError{}))
+				//TODO why do we return 200 when the HelmRepository doesn't exist
+				Expect(err.(*grpcruntime.HTTPStatusError).HTTPStatus).To(Equal(http.StatusOK))
```
jpellizzari, line 178: We have this thing that might be of use: https://github.com/weaveworks/weave-gitops/blob/ee12fdf5957a9070994295e70ec5c2320cc3106f/pkg/server/grpc_error_matcher.go#L25

test/acceptance/test/profiles_test.go
```diff
@@ -0,0 +1,171 @@
...
+		_ = initAndCreateEmptyRepo(tip.appRepoName, gitproviders.GitProviderGitHub, private, GITHUB_ORG)
+
+		clientSet, kClient = buildKubernetesClients()
+	})
+
+	AfterEach(func() {
+		deleteRepo(tip.appRepoName, gitproviders.GitProviderGitHub, GITHUB_ORG)
+	})
+
+	It("Works", func() {
```
jpellizzari, line 58: Maybe a little more description on this test case? What are we trying to prove/validate with this test?

test/acceptance/test/profiles_test.go
```diff
@@ -0,0 +1,171 @@
...
+	})
+})
+
+func buildKubernetesClients() (*kubernetes.Clientset, client.Client) {
+	config, err := clientcmd.BuildConfigFromFlags("", filepath.Join(homedir.HomeDir(), ".kube", "config"))
+	Expect(err).NotTo(HaveOccurred())
+	clientSet, err := kubernetes.NewForConfig(config)
+	Expect(err).NotTo(HaveOccurred())
+
+	scheme := runtime.NewScheme()
```
jpellizzari, line 107: Not sure if this applies, but we do have this:
https://github.com/weaveworks/weave-gitops/blob/ee12fdf5957a9070994295e70ec5c2320cc3106f/pkg/kube/kubehttp.go#L35

jpellizzari (commented):
Did a quick read through, mostly looking at tests. I won't be able to validate the logic in the HTTP handlers.

I want to discuss with the team/product if we need to "mux" profiles in the existing `wego-app` runtime, or if we need a separate deployment like we have here.

As it is now, the UI will not be able to query the profiles service (if I understand it correctly).

Dockerfile.profiles
aclevername: moving it to `cmd/profiles` sounds good. If its in there I guess just `Dockerfile` would be okay?

Also how would you feel if we also moved the top level `Dockerfile` into `cmd/gitops`, just for consistency?

api/applications/applications.swagger.json
```diff
@@ -137,6 +137,7 @@
     },
     "/v1/applications/auth_providers/gitlab": {
       "get": {
+        "summary": "GetGitlabAuthURL returns the URL to initiate a GitLab OAuth PKCE flow.\nThe user must browse to the returned URL to authorize the OAuth callback to the GitOps UI.\nSee the GitLab OAuth docs for more more information:\nhttps://docs.gitlab.com/ee/api/oauth2.html#supported-oauth-20-flows",
```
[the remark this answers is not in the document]
aclevername: We merged `main` in a little while ago and it was still there :thinking: I think perhaps its something that got auto-generated when we re-ran the proto command

aclevername:
> Very well laid out and organized, and lots of test coverage +1 .

Thanks!

> I want to discuss with the team/product if we need to "mux" profiles in the existing `wego-app` runtime, or if we need a separate deployment like we have here.

:+1: 

> As it is now, the UI will not be able to query the profiles service (if I understand it correctly).

:thinking: could you expand upon this? The profiles API is accessible via the service, assuming the UI is running in cluster it should have access (does it not run in cluster?).

Dockerfile.profiles
aclevername: https://github.com/weaveworks/weave-gitops/pull/1148/commits/01b38b1b0e5cae11d9d9c2b3a37c62a95942711b

pkg/helm/charts.go
```diff
@@ -0,0 +1,289 @@
...
+// GetValuesFile fetches the value file from a chart.
+func (h *RepoManager) GetValuesFile(ctx context.Context, helmRepo *sourcev1beta1.HelmRepository, c *ChartReference, filename string) ([]byte, error) {
+	if err := h.updateCache(ctx, helmRepo); err != nil {
+		return nil, err
+	}
+
+	chart, err := h.loadChart(ctx, helmRepo, c)
+	if err != nil {
+		return nil, fmt.Errorf("loading %s from chart: %w", filename, err)
+	}
```
hiddeco, line 129: Given you do not seem to validate the chart (`Validate` method on the object), and the minimal validation Helm does while loading can be easily replicated and is not likely to change, I think it would be much more memory efficient to not load the whole data into memory but rather scan through the tarball for the `chartutil.ValuesfileName` in the root of the archive.

We do something similar in a part of the source-controller for the metadata file: https://github.com/fluxcd/source-controller/blob/59dc602b2d0a5b0d4625a4a848303fa2d2117e84/internal/helm/chart/metadata.go#L142

pkg/server/profiles.go
```diff
@@ -0,0 +1,187 @@
...
+		if apierrors.IsNotFound(err) {
+			errMsg := fmt.Sprintf("HelmRepository %q/%q does not exist", s.HelmRepoNamespace, s.HelmRepoName)
+			s.Log.Error(err, errMsg)
+
+			return &httpbody.HttpBody{
+					ContentType: "application/json",
+					Data:        []byte{},
+				}, &grpcruntime.HTTPStatusError{
+					Err:        errors.New(errMsg),
+					HTTPStatus: http.StatusOK,
```
Himangini, line 140: shouldn't this be `http.StatusNotFound` or `http. StatusBadRequest` ? 🤔

aclevername: We thought similar when we first looked at this, it appears to be a very intentional - https://github.com/weaveworks/weave-gitops-enterprise/pull/245/commits/aaccb2dc4183bef9c7aee3ba391077c931d1be90

I want to change it as it feels wrong to me, but I think we need to change the behaviour in the UI first to handle this change gracefully before making the change in the API.

pkg/helm/charts_test.go
aclevername: refactored the error message testing to be more explicit

pkg/helm/charts_test.go
aclevername: refactored the error message testing to be more explicit

pkg/helm/charts_test.go
aclevername: refactored the error message testing to be more explicit

pkg/server/profiles_test.go
aclevername: that seems to be testing the grpc status codes, as opposed to http status codes. I hadn't seen the grpc codes before which is interesting :thinking:

test/acceptance/test/profiles_test.go
aclevername: updated

test/acceptance/test/profiles_test.go
aclevername: very nice! using this :smile:

test/acceptance/test/profiles_test.go
```diff
@@ -0,0 +1,171 @@
...
+	err = rawClient.Create(context.TODO(), weaveworksChartsHelmRepository)
+	Expect(err).NotTo(HaveOccurred())
+	Eventually(func() error {
+		helmRepo := sourcev1beta1.HelmRepository{}
+		err := rawClient.Get(context.TODO(), client.ObjectKeyFromObject(weaveworksChartsHelmRepository), &helmRepo)
+		if err != nil {
+			return err
+		}
+
+		if len(helmRepo.Status.Conditions) == 1 && helmRepo.Status.Conditions[0].Type == "Ready" && helmRepo.Status.Conditions[0].Status == "True" {
```
hiddeco, line 146: This check is very prone to errors due to minimal changes in contract (e.g. the introduction of other conditions), and it would be much more resilient if `apimeta.FindStatusCondition` would be used: https://github.com/kubernetes/apimachinery/blob/master/pkg/api/meta/conditions.go#L74

aclevername: I like this! Will update

pkg/helm/charts.go
aclevername: :thinking: that change sounds reasonable, but I quite like that we are delegating a lot of the complexity of interacting with the chart to the Helm library. I think for the moment this change might be a bit overkill, but in this [story](https://github.com/weaveworks/weave-gitops/issues/1144) we are getting into caching (and in-general being more optimised), I'm going to flag your comment in that issue :smile:

pkg/helm/charts.go
aclevername: @hiddeco can you expand? I'm not really familiar with what it does (Niki and I inherrited this code, so we don't understand all of it completely). what would you suggest as an alternative?

Dockerfile.profiles
aclevername: as per discussions in slack I'm going to remove this dockerimage and move the profiles API into the app api

pkg/server/profiles.go
Himangini: yeah, we should definitely change this 👍🏻  why would we wait for tho, this is in core and the UI for EE is not using this. The graceful handling of UI can be part of https://github.com/weaveworks/weave-gitops-enterprise/issues/311  when we switch over? 🤔

aclevername: I sort of assumed we wouldn't be making the UI changes (I have no idea even how to :smile: ). So we would need to raise it with them, let them make the changes (e.g. when a 404 is returned show an empty list if thats the behaviour they want) and then change it in core

aclevername: if your happy with ^ I'll create an issue in EE and raise it with them

pkg/helm/charts.go
hiddeco: The more detailed explanation for this advise is Helm umbrella charts. When Helm loads such a chart, it also includes all subcharts because their inclusion for the apply is determined based on the values used during release. This means that you load a whole lot tree of nonsense while you are after very specific data.

pkg/helm/charts.go
hiddeco: The `cli` package in Helm contains configuration defaults with an assumption that it runs as a CLI on a machine. This means that e.g. cache path defaults try to create absolute paths for configuration fields relative to the user directory, and in addition, that a whole set of environment variables are taken into account as global configuration flags.

This means that unless you are very aware of what the action does with the configured defaults, you may end up with data magically stored somewhere, or things being unexpectedly taken into account.

The simplest alternative is to pass the `cli.EnvSettings` construct with no default values and limited config, the place where the returned object is used doesn't seem to use many of them: https://github.com/helm/helm/blob/main/pkg/action/install.go#L675

aclevername:
> > could you expand upon this? The profiles API is accessible via the service, assuming the UI is running in cluster it should have access (does it not run in cluster?).
> 
> The network path currently is like this: `browser requests <cluster address>:9001/v1/applications --> wego-app UI server --> /v1/applications`
> 
> ^^ `wego-app` doesn't know about `/v1/profiles` and will 404

Merged the profiles API into the app api

manifests/wego-app/deployment.yaml.tpl
```diff
@@ -13,12 +13,17 @@ spec:
       serviceAccountName: wego-app-service-account
       containers:
         - name: wego-app
-          image: ghcr.io/weaveworks/wego-app:{{.Version}}
-          args: ["ui", "run", "-l"]
+          image: ghcr.io/weaveworks/wego-app:{{.AppVersion}}
+          args: ["ui", "run", "-l", "--helm-repo-namespace", "$(RUNTIME_NAMESPACE)"]
```
jpellizzari, line 17: Should this be a CLI arg if it is already an env variable?

pkg/server/applications.go
```diff
@@ -0,0 +1,777 @@
+package server
```
jpellizzari, line 1: This diff is a little problematic. This will blow away any git history on the original `server.go` file. Is there a way we can get `git` to recognize this as a file rename instead of a new file?

pkg/server/server.go
```diff
@@ -2,793 +2,38 @@ package server
```
jpellizzari, line 2: Now that I see this, the reason for the diff I commented on before makes more sense.

I would still like to preserve the git history of the applications `server.go` file. Maybe we can have a new file called `combined_server.go` (terrible name I know) that would represent the new mux'd apps+profiles server?

pkg/server/server.go
```diff
@@ -2,793 +2,38 @@ package server
...
-// NewApplicationsHandler allow for other applications to embed the Weave GitOps HTTP API.
-// This handler can be muxed with other services or used as a standalone service.
-func NewApplicationsHandler(ctx context.Context, cfg *ApplicationsConfig, opts ...runtime.ServeMuxOption) (http.Handler, error) {
-	appsSrv := NewApplicationsServer(cfg)
-
-	mux := runtime.NewServeMux(middleware.WithGrpcErrorLogging(cfg.Logger))
-	httpHandler := middleware.WithLogging(cfg.Logger, mux)
-	httpHandler = middleware.WithProviderToken(cfg.JwtClient, httpHandler, cfg.Logger)
+func NewApplicationsAndProfilesHandler(ctx context.Context, cfg *Config, opts ...runtime.ServeMuxOption) (http.Handler, error) {
+	mux := runtime.NewServeMux(middleware.WithGrpcErrorLogging(cfg.AppConfig.Logger))
```
jpellizzari, line 20: I had imagined apps and profiles being two separate muxes. Not sure if that is possible or not, however.

jpellizzari (commented):
Code looks good. The mux'ing comments I left can be left for later if we ever add a third service.

I would really like to try and preserve the git history for the `pkg/server/server{_test}.go` files. Here is an idea on how we might be able to resolve it:

1. Move the contents of `pkg/server/server.go` in this PR to `pkg/server/handler.go`
2. `$ git checkout origin/main pkg/server/server.go`
2. `$ git checkout origin/main pkg/server/server_test.go`
3. Make the contents of `pkg/server/applications.go` the contents of `pkg/server/server.go` so that all the relevant changes are present.

We can decide to rename `server.go` if we want, but at that point git should pick up the rename and preserve the history.

@aclevername @nikimanoledaki LMK if that makes sense. Happy to hop on a call to pair on this.

manifests/wego-app/deployment.yaml.tpl
aclevername: I think it makes sense to have it as a CLI flag since we want it to be configurable & have sensible defaults (that someone calling the binary can discover). I'm using the env var as a way of fetching the value. I think its okay? Open to alternatives :smile:

pkg/server/server.go
aclevername: Yeah I wasn't 100% sure the approach I went was the best. This approach means they share the same middle ware which is nice. Open to alternatives

cmd/gitops/ui/run/cmd.go
```diff
@@ -54,21 +60,23 @@ func runCmd(cmd *cobra.Command, args []string) error {
...
+	profilesConfig := server.NewProfilesConfig(helmRepoNamespace, helmRepoName)
+
+	appAndProfilesHandlers, err := server.NewApplicationsAndProfilesHandler(context.Background(), &server.Config{AppConfig: appConfig, ProfilesConfig: profilesConfig})
 	if err != nil {
-		return fmt.Errorf("could not create applications handler: %w", err)
+		return fmt.Errorf("could not create handler: %w", err)
 	}
 
-	mux.Handle("/v1/", appsHandler)
+	mux.Handle("/v1/", appAndProfilesHandlers)
```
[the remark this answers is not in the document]
aclevername: yeah definitely somethings we can refactor into when we add a third API

manifests/wego-app/deployment.yaml.tpl
aclevername: An alternative @nikimanoledaki pointed out would be changing:
```
          args: ["ui", "run", "-l", "--helm-repo-namespace", "$(RUNTIME_NAMESPACE)"]
          ports:
            - containerPort: 9001
              protocol: TCP
          imagePullPolicy: IfNotPresent
          env:
            - name: RUNTIME_NAMESPACE
              valueFrom:
                fieldRef:
                  fieldPath: metadata.namespace
```

to:
```
          args: ["ui", "run", "-l", "--helm-repo-namespace", "{{.Namespace}}"]
```

using the templating to insert the namespace instead. I felt env var approach of fetching it from `metadata.namespace` was a tiny bit cleaner, but if the env var adds too much noise to the file then I'm happy to change :smile:

api/profiles/profiles.proto
```diff
@@ -0,0 +1,87 @@
+syntax = "proto3";
```
Skarlso, line 1: For the record, if we want a nice documentation of these using `buf doc` this will need to be commented nicely. All fields with appropriate comments, otherwise they just show up blank.

aclevername:
> I would really like to try and preserve the git history for the `pkg/server/server{_test}.go` files. 

Should be resolved now :smile: 

> 1. Move the contents of `pkg/server/server.go` in this PR to `pkg/server/handler.go`
> 2. `$ git checkout origin/main pkg/server/server.go`
> 3. `$ git checkout origin/main pkg/server/server_test.go`
> 4. Make the contents of `pkg/server/applications.go` the contents of `pkg/server/server.go` so that all the relevant changes are present.

Steps `2.` and `3.` aren't actually needed. Just doing stop `1.` and `4.` was enough to preserve it :smile:

pkg/server/applications.go
```diff
@@ -0,0 +1,777 @@
...
+
+	"github.com/weaveworks/weave-gitops/pkg/server/middleware"
+
+	"github.com/benbjohnson/clock"
+	"github.com/weaveworks/weave-gitops/pkg/flux"
+	"github.com/weaveworks/weave-gitops/pkg/osys"
+	"github.com/weaveworks/weave-gitops/pkg/runner"
+	"github.com/weaveworks/weave-gitops/pkg/server/internal"
+	"github.com/weaveworks/weave-gitops/pkg/services"
+
```
Skarlso: Can we please sort these imports? :))

aclevername: my linter doesn't complain :ahhhhh: what do you want to have changed? I always forget what the desired approach is :smile:

api/applications/applications.swagger.json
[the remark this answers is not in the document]
aclevername: Looks like https://github.com/weaveworks/weave-gitops/pull/1155/files will address this

pkg/server/applications.go
Skarlso: :DDDD

stdlib
thridparty
organization
package-local

:D

aclevername: my linter doesn't agree with that completely, moving `flux` imports to below third party gets auto-moved

manifests/wego-app/deployment.yaml.tpl
aclevername: Decided to go with the templating. Resolved https://github.com/weaveworks/weave-gitops/pull/1148/commits/149d528bafca07e0581d75f89249f9d52ef7adca

pkg/helm/charts.go
aclevername: Thanks! We've resolved it in https://github.com/weaveworks/weave-gitops/pull/1148/commits/c37413a9d2b91d05a6e2cb857fd848524998e966

lmk if that works :smile:

pkg/server/applications.go
aclevername: https://github.com/weaveworks/weave-gitops/pull/1148/commits/0ae0b6c02cd02869f341bb009c0f5b9725fc6cb1

pkg/helm/charts.go
hiddeco: If someone without context now tries to expand on the configuration for another Helm action, it should crash or obviously fail and require them to investigate instead of things seemingly working without them knowing about any hidden side effects, so this looks all good to me :100:

Think that the CI wants a `go fmt` to agree on the indentation of the fields however :grimacing:

pkg/server/profiles.go
Himangini: yeah, I didn't mean we should do the UI changes, let's open a ticket in EE for UI to handle this change 👍🏻

aclevername:
The acceptance tests are failing. Do the workflow runs re-build the docker image? Or are we supposed to do something manually to get them to pass? (e.g. push the image?) I think the docker image not containing our changes is whats causing the failures

api/profiles/profiles.proto
nikimanoledaki: that's true, wego core's `Applications` and enterprise's `ClustersService` ([here](https://github.com/weaveworks/weave-gitops-enterprise/blob/main/cmd/clusters-service/api/capi_server.proto)) services are also missing doc comments. we could open a separate issue for this :) also, couldn't find `buf doc`, are you referring to `proto-gen-doc`? 🤔

hiddeco (approved):
I did not review this PR in full, but looked at common missteps of working with the Helm SDK, and all those have either been addressed or noted for another round. Within this context, it looks good to me. :1st_place_medal:

Thanks for following through on my comments @aclevername (and you work @nikimanoledaki) :bow: 

:spiral_notepad: One last note for a future iteration, not all possible getter configuration options as supported by the source-controller are taken into account (TLS), and in addition, you may need to be aware of `--pass-credentials` introduced in https://github.com/helm/helm/releases/tag/v3.6.1 to be able to cope with certain configurations / servers.

pkg/server/profiles.go
aclevername: https://github.com/weaveworks/weave-gitops-enterprise/issues/366

pkg/server/handler.go
```diff
@@ -0,0 +1,39 @@
...
+	pbprofiles "github.com/weaveworks/weave-gitops/pkg/api/profiles"
+	"github.com/weaveworks/weave-gitops/pkg/server/middleware"
+)
+
+type Config struct {
+	AppConfig      *ApplicationsConfig
+	ProfilesConfig ProfilesConfig
+}
+
+func NewApplicationsAndProfilesHandler(ctx context.Context, cfg *Config, opts ...runtime.ServeMuxOption) (http.Handler, error) {
```
Skarlso, line 19: `opts` is never used

cmd/gitops-server/main.go
```diff
@@ -35,17 +37,29 @@ func NewAPIServerCommand() *cobra.Command {
...
+			if err != nil {
+				return fmt.Errorf("could not create client config: %w", err)
+			}
+
+			_, rawClient, err := kube.NewKubeHTTPClientWithConfig(rest, clusterName)
+			if err != nil {
+				return fmt.Errorf("could not create kube http client: %w", err)
+			}
+
+			profilesConfig := server.NewProfilesConfig(rawClient, "default", "weaveworks-charts")
```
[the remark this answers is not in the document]
aclevername: this binary is only used for integration testing. The code responsible for bringing up the profiles API is in `cmd/gitops/`. I was confused at first too :smile:

pkg/server/handler.go
aclevername: good spot

cmd/gitops-server/main.go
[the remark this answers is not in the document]
Skarlso: It's....unexpected, yeah.

manifests/manifests.go
```diff
@@ -21,24 +21,50 @@ var (
...
+
+	err = template.Execute(yaml, params)
+	if err != nil {
+		return nil, fmt.Errorf("error injecting values to template: %w", err)
+	}
+
+	return yaml.Bytes(), nil
+}
+
+func readTemplateDirectory(params Params, templateFiles embed.FS, templatestDir string) ([][]byte, error) {
```
Skarlso, line 56: `templatestDir` -> `templateDir` ?

Skarlso: The name of the function is a bit misleading. It doesn't just read the template directory.

It gathers, executes and returns manifest data.

Can we call this `executeTemplates` or `renderManifestFiles`?

manifests/manifests_test.go
```diff
@@ -1,23 +1,60 @@
...
+kind: Role
+metadata:
+  name: resources-reader`))
+
+			By("containing a Role Binding manifest")
+			Expect(manifests).To(ContainSubstring(`
+kind: RoleBinding
+metadata:
+  name: read-resources`))
+		})
```
Skarlso, line 58: Can we add some tests for invalid template data and some error scenarios?

pkg/helm/charts.go
```diff
@@ -0,0 +1,283 @@
...
+
+// ChartReference is a Helm chart reference, the SourceRef is a Flux
+// SourceReference for the Helm chart.
+type ChartReference struct {
+	Chart     string
+	Version   string
+	SourceRef helmv2beta1.CrossNamespaceObjectReference
+}
+
+// DefaultChartGetter provides default ways to get a chart index.yaml based on
```
Skarlso, line 59:
```suggestion
// DefaultChartGetters provides default ways to get a chart index.yaml based on
```

Skarlso: On second thought. Why is this exported?

aclevername: indeed, fixing

manifests/manifests_test.go
aclevername: do you mean when the template params aren't set?

pkg/helm/charts.go
```diff
@@ -0,0 +1,283 @@
...
+		}
+	}
+
+	return nil, fmt.Errorf("failed to find file: %s", filename)
+}
+
+func (h *RepoManager) updateCache(ctx context.Context, helmRepo *sourcev1beta1.HelmRepository) error {
+	entry, err := h.entryForRepository(ctx, helmRepo)
+	if err != nil {
+		return err
```
Skarlso, line 149: We wrap the other error, might as well be consistent and wrap this one and the one for Download as well. :)

manifests/manifests.go
aclevername: renamed

cmd/gitops/ui/run/cmd.go
```diff
@@ -54,21 +61,33 @@ func runCmd(cmd *cobra.Command, args []string) error {
...
+
+	_, rawClient, err := kube.NewKubeHTTPClientWithConfig(rest, clusterName)
+	if err != nil {
+		return fmt.Errorf("could not create kube http client: %w", err)
 	}
 
-	appsHandler, err := server.NewApplicationsHandler(context.Background(), cfg)
+	profilesConfig := server.NewProfilesConfig(rawClient, helmRepoNamespace, helmRepoName)
+
+	appAndProfilesHandlers, err := server.NewApplicationsAndProfilesHandler(context.Background(), &server.Config{AppConfig: appConfig, ProfilesConfig: profilesConfig})
```
Skarlso, line 85: Yeah, I know what this is doing, but it's a weird thing bringing these together.

Usually an `And` in a function doesn't mean a good thing. :D It means it isn't doing just one thing. :)

pkg/helm/charts.go
aclevername: its wrapped in the parent call to `updateCache`

cmd/gitops/ui/run/cmd.go
aclevername: I mean it could just be `NewHandlers` I guess?

pkg/helm/charts.go
```diff
@@ -0,0 +1,283 @@
...
+}
+
+type ChartPredicate func(*repo.ChartVersion) bool
+
+// Profiles is a predicate for scanning charts with the ProfileAnnotation.
+var Profiles = func(v *repo.ChartVersion) bool {
+	return hasAnnotation(v.Metadata, ProfileAnnotation)
+}
+
+// ScanCharts filters charts using the provided predicate.
```
Skarlso, line 75:
```suggestion
// GetCharts returns charts using the provided predicate.
```

?

pkg/helm/charts.go
```diff
@@ -0,0 +1,283 @@
...
+	var secret corev1.Secret
+	if err := kc.Get(ctx, types.NamespacedName{Name: hr.Spec.SecretRef.Name, Namespace: hr.Namespace}, &secret); err != nil {
+		return "", "", fmt.Errorf("repository authentication: %w", err)
+	}
+
+	return string(secret.Data["username"]), string(secret.Data["password"]), nil
+}
+
+func fetchIndexFile(chartURL string) (*repo.IndexFile, error) {
+	if hostname := os.Getenv("SOURCE_CONTROLLER_LOCALHOST"); hostname != "" {
```
Skarlso, line 231: What is the purpose of this environment property? I'm only seeing it here and not anywhere other and there is no documentation of it.

Skarlso: If this is for testing purposes only to change the hostname to a local test service, we should either hide it better or make it injectable.

pkg/helm/charts.go
```diff
@@ -0,0 +1,283 @@
...
+						p.Maintainers = append(p.Maintainers, &pb.Maintainer{
+							Name:  m.Name,
+							Email: m.Email,
+							Url:   m.URL,
+						})
+					}
+					p.AvailableVersions = append(p.AvailableVersions, v.Version)
+					ps[name] = p
+				}
+			}
```
Skarlso, line 112:
```suggestion
			if !pred(v) {
				continue
			}
			// if already added, update the versions array
			if p, ok := ps[name]; ok {
				p.AvailableVersions = append(p.AvailableVersions, v.Version)
			} else { // otherwise create a new profile and add to map
				p = &pb.Profile{
					Name:        name,
					Home:        v.Home,
					Sources:     v.Sources,
					Description: v.Description,
					Keywords:    v.Keywords,
					Icon:        v.Icon,
					KubeVersion: v.KubeVersion,
				}
				for _, m := range v.Maintainers {
					p.Maintainers = append(p.Maintainers, &pb.Maintainer{
						Name:  m.Name,
						Email: m.Email,
						Url:   m.URL,
					})
				}
				p.AvailableVersions = append(p.AvailableVersions, v.Version)
				ps[name] = p
			}
```

api/profiles/profiles.proto
jpellizzari: We have pretty good coverage on doc coverage in `Applications`: 
https://github.com/weaveworks/weave-gitops/blob/60848106b264b6b56b1733d8949fc919966336ac/api/applications/applications.proto#L180

jpellizzari (approved):
Thanks for making the rename changes :+1:

jpellizzari:
> The acceptance tests are failing. Do the workflow runs re-build the docker image? Or are we supposed to do something manually to get them to pass? (e.g. push the image?) I think the docker image not containing our changes is whats causing the failures
> 
> https://github.com/weaveworks/weave-gitops/runs/4397466184?check_suite_focus=true

App is crashing in the container: https://github.com/weaveworks/weave-gitops/runs/4397466184?check_suite_focus=true#step:3:9102

api/profiles/profiles.proto
nikimanoledaki: you're right, we'll add some doc comments for profiles! 👍

manifests/manifests_test.go
Skarlso: Yep

pkg/helm/charts.go
Skarlso: But so are the others. :D

cmd/gitops/ui/run/cmd.go
Skarlso: That makes sense, yes.

manifests/manifests_test.go
Skarlso: This is actually fine. The template engine doesn't care and we won't test embed FS.

cmd/gitops/ui/run/cmd.go
nikimanoledaki: I also prefer not having an `And` in the func. `NewHandler` works. :)

nikimanoledaki: I changed it to `NewHandlers` since it's a bit better than having `And` in the function

pkg/helm/charts.go
aclevername: removing it since we aren't using it

pkg/helm/charts.go
aclevername: I am not really a big fan of this :/ I don't mind the nested if. The `continue` is a bit easier to accidentally not see

Skarlso (approved):

Issue #1124, opened 2021-11-22
upgrade using GitLab repo fails with invalid memory address


palemtnrider:
**Describe the bug**
When upgrading a Weave GitOps installation, I get an Invalid memory address error

**What is the severity of the bug**

| Label | Description |
| --- | --- |
| severity/Minor | Weave Gitops functionality is broken, but there is a workaround |

**Environment**
 - gitops [0.5.0-rc2]
Current Version: v0.0.0
GitCommit: 4555d5f8
BuildTime: 2021-11-22_16:50:20
Branch: main
Flux Version: v0.21.0

 - kubernetes:  [e.g. 1.20.4]
    - [ ] KinD - _version_]
    - [ ] k3s - _version_
    - [ ] cloud [e.g., EKS, AKS]  _version_
    - [ ] other - _name_ _version_
 - Browser + version: [e.g. chrome 74, safari 12, firefox 87]
 -
**Affects versions**
v0.4.1, v0.5.0-rc2

**To Reproduce**
1. gitops install using GL
2. Install you entitlement into the cluster
3. Issue the upgrade command `GITLAB_TOKEN=your_token ~/src/weaveworks/weave-gitops/bin/gitops upgrade --app-config-url ssh://git@gitlab.com/mark.emeis/hxg4.git --branch test-ee-upgrade-11 --profile-version v0.0.15`
4. see the following error
```panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x4809ba7]

goroutine 1 [running]:
github.com/fluxcd/go-git-providers/gitlab.(*CommitClient).Create(0xc004d96558, 0x9149d30, 0xc0000540d0, 0x7ffeefbff844, 0x12, 0x5dde58a, 0xe, 0xc004d7d2f0, 0x3, 0x3, ...)
        /Users/me/go/pkg/mod/github.com/fluxcd/go-git-providers@v0.2.1-0.20211006191809-83344459c0dc/gitlab/client_repository_commit.go:78 +0x1e7
github.com/weaveworks/weave-gitops/pkg/gitproviders.createPullRequest(0x9149d30, 0xc0000540d0, 0x9166a00, 0xc004d3b180, 0x5dddfa2, 0xe, 0x5e2eb08, 0x32, 0x5dde58a, 0xe, ...)
[...]

PR #1149, opened 2021-11-30
Fixes `gitops upgrade` with gitlab

foot:
- Don't try and delete a file for now
- This is okay as we sorted out https://github.com/weaveworks/weave-gitops/pull/1099

Closes: https://github.com/weaveworks/weave-gitops/issues/1124

**What changed?**

Don't try and delete the wego-app.yaml for now. Its not causing any issues and now no longer raises reconciliation errors.

**Why?**

go-git-provider gitlab doesn't support deleting files w/ `nil` (maybe it works with an empty string?)

**How did you test it?**

manually


yiannistri (approved):
LGTM 👍

PR #1153, opened 2021-12-01
Fixes `gitops install` wego-app.yaml deployment when running in container

foot:
- Container builds _in the container_ and doesn't use go-releaser and so
  doesn't have a version.Version

**What changed?**

`version.Version` is set when building `gitops` using `make bin` (used for local dev and inside the published container).

**Why?**

So that capi cluster bootstrapping doesn't show an imagepullerror as the version is used to generate the deployment.image when running `gitops install`

**How did you test it?**

`docker build -t...`


jpellizzari (approved):

palemtnrider (approved):
LGTM 
Thanks for fixing this.

Issue #942, opened 2021-10-21
Wego-app UI auto-refresh application list


palemtnrider:
As an app developer, when I add applications either through the UI or CLI I want the application list to show the application in the list once it is available in the cluster without having to perform any action, e.g., browser refresh, click the application link.

---
_The original ticket_

**Describe the bug**
I opened the ui from the wego-app pod running in the cluster.  When I added an app via the CLI, the app wasn't displayed in the UI until refreshing the page.  I waited several minutes.  I also tried clicking the applications link on the nav.

**Environment**
 - gitops: v0.3.2-rc1
 - kubernetes:  1.21.1
    - [ ] KinD - 0.11.1]
  - Browser + version: chrome Version 94.0.4606.61 (Official Build) (x86_64)
 -
**Affects versions**
* v0.3.2-rc1

**To Reproduce**
1. install weave gitops
2. do a port forward to the weave gitops service
3. stay on the applications page
4. add the app via `gitops app add .`
5. watch the UI page

**Expected behavior**
After a short period of time, the list of applications shows my new app

jpellizzari:
Scope for this: add polling to the app list and app detail pages.

Acceptance Criteria:
* Poll every n seconds
* Indicate to the user that the page is polling

PR #1154, opened 2021-12-01
Add polling to Applications Page

joshri:
Closes: #942 

The Applications page now runs ListApplications every 30 seconds, with a timestamp + CircularProgress as indication in a new component, `<PollingIndicator />`.

https://user-images.githubusercontent.com/65822698/144296624-862a416b-1e0e-4fce-b9ff-09d876abc03b.mov

### DIFF
--- a/ui/components/PollingIndicator.tsx
+++ b/ui/components/PollingIndicator.tsx
@@ -16,15 +16,15 @@ type Props = {
 function PollingIndicator({ className, loading, interval }: Props) {
   const date = new Date();
   const [updated, setUpdated] = React.useState(date.toTimeString().slice(0, 8));
-  const [indicator, setIndicator] = React.useState(false);
+  const [visible, setVisible] = React.useState(false);
 
   React.useEffect(() => {
     loading
-      ? setIndicator(true)
+      ? setVisible(true)
       : setTimeout(() => {
           const date = new Date();
           setUpdated(date.toTimeString().slice(0, 8));
-          setIndicator(false);
+          setVisible(false);
         }, 1000);
   }, [loading]);
 
@@ -34,10 +34,12 @@ function PollingIndicator({ className, loading, interval }: Props) {
         <p className="timestamp-p">last updated {interval} seconds ago</p>
       </Tooltip>
       <Spacer padding="base">
-        <CircularProgress
-          size={theme.spacing.base}
-          className={indicator ? "in" : "out"}
-        />
+        <Flex>
+          <CircularProgress
+            size={theme.spacing.base}
+            className={visible ? "in" : "out"}
+          />
+        </Flex>
       </Spacer>
     </Flex>
   );
@@ -50,6 +52,7 @@ export default styled(PollingIndicator)`
   }
   .MuiCircularProgress-root {
     transition: opacity 2s;
+    margin-bottom: 0;
     &.in {
       opacity: 1;
     }
--- a/ui/components/PollingIndicator.tsx
+++ b/ui/components/PollingIndicator.tsx
@@ -2,6 +2,7 @@ import { CircularProgress, Tooltip } from "@material-ui/core";
 import * as React from "react";
 import styled from "styled-components";
 import { theme } from "..";
+import { muiTheme } from "../lib/theme";
 import Flex from "./Flex";
 import Spacer from "./Spacer";
 
@@ -30,7 +31,7 @@ function PollingIndicator({ className, loading, interval }: Props) {
   return (
     <Flex align className={className}>
       <Tooltip title={updated} arrow placement="top">
-        <p>last updated {interval} seconds ago</p>
+        <p className="timestamp-p">last updated {interval} seconds ago</p>
       </Tooltip>
       <Spacer padding="base">
         <CircularProgress
@@ -43,6 +44,10 @@ function PollingIndicator({ className, loading, interval }: Props) {
 }
 
 export default styled(PollingIndicator)`
+  .timestamp-p {
+    color: ${muiTheme.palette.text.secondary};
+    margin-bottom: 12px;
+  }
   .MuiCircularProgress-root {
     transition: opacity 2s;
     &.in {
--- a/ui/components/PollingIndicator.tsx
+++ b/ui/components/PollingIndicator.tsx
@@ -1,4 +1,4 @@
-import { CircularProgress } from "@material-ui/core";
+import { CircularProgress, Tooltip } from "@material-ui/core";
 import * as React from "react";
 import styled from "styled-components";
 import { theme } from "..";
@@ -28,7 +28,9 @@ function PollingIndicator({ className, loading }: Props) {
 
   return (
     <Flex align className={className}>
-      last updated at {updated}
+      <Tooltip title={updated} arrow placement="top">
+        <p>last updated 30 seconds ago</p>
+      </Tooltip>
       <Spacer padding="base">
         <CircularProgress
           size={theme.spacing.base}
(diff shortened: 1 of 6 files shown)


ui/components/PollingIndicator.tsx
```diff
@@ -0,0 +1,52 @@
...
+import Flex from "./Flex";
+import Spacer from "./Spacer";
+
+type Props = {
+  /** whether `<CircularProgress />` should be shown */
+  loading: boolean;
+  className?: string;
+};
+
+function PollingIndicator({ className, loading }: Props) {
```
jpellizzari, line 14: Nice work on a re-usable component!

ui/pages/Applications.tsx
```diff
@@ -16,24 +18,47 @@ type Props = {
...
+  const getApps = () => {
+    setLoading(true);
+    applicationsClient
+      .ListApplications({ namespace: "wego-system" })
+      .then((res) => setApplications(res.applications))
+      //how deep are we going with these error messages? Should I have one for 500 and one for 400?
+      .catch((err) =>
+        doAsyncError(
+          "Unable to retrieve applications",
+          "Something went wrong - please try again later"
```
jpellizzari, line 33: We want to give the user some breadcrumbs to follow:

```suggestion
          err.message
```

jpellizzari (changes requested):
Code LGTM, but the timestamp is way too prominent. It is not as important as the name of the page (which it is displayed next to). Can we move this somewhere?

Also, let's use a relative timestamp string: `last updated 30s ago`.

ui/components/PollingIndicator.tsx
joshri: 🌞

ui/pages/Applications.tsx
joshri: So I can use err.message and err.detail? Do I have to check if it's a 500 first and handle that case (if there's no message?)

jpellizzari: A 500 will still have a message. Every error will probably have a message.

joshri:
> Code LGTM, but the timestamp is way too prominent. It is not as important as the name of the page (which it is displayed next to). Can we move this somewhere?
> 
> Also, let's use a relative timestamp string: `last updated 30s ago`.

Are you envisioning that as like a live counter, or just always saying "last updated 30 seconds ago"? I think it's nice to give some context on why the app list might not appear like the user expects it to, but I definitely agree that the timestamp is a little extra right now

ui/pages/Applications.tsx
joshri: Awesome I'll change it 🌞

jpellizzari:
> > Code LGTM, but the timestamp is way too prominent. It is not as important as the name of the page (which it is displayed next to). Can we move this somewhere?
> > Also, let's use a relative timestamp string: `last updated 30s ago`.
> 
> Are you envisioning that as like a live counter, or just always saying "last updated 30 seconds ago"? I think it's nice to give some context on why the app list might not appear like the user expects it to, but I definitely agree that the timestamp is a little extra right now

It's more about what should be important to the user. When we give the user a timestamp like `13:38:22`, we are making them do math to figure out "when was that?". Just give them the info they are going to have to figure out for themselves.

In the case of "I need to find events that correspond to an incident", THEN we can show explicit timestamps, since that's their "search" modality.

joshri:
> > > Code LGTM, but the timestamp is way too prominent. It is not as important as the name of the page (which it is displayed next to). Can we move this somewhere?
> > > Also, let's use a relative timestamp string: `last updated 30s ago`.
> > 
> > 
> > Are you envisioning that as like a live counter, or just always saying "last updated 30 seconds ago"? I think it's nice to give some context on why the app list might not appear like the user expects it to, but I definitely agree that the timestamp is a little extra right now
> 
> It's more about what should be important to the user. When we give the user a timestamp like `13:38:22`, we are making them do math to figure out "when was that in relation to now?". Just give them the info they are going to have to figure out for themselves.
> 
> In the case of "I need to find events that correspond to an incident", THEN we can show explicit timestamps, since that's their "search" modality.

Is having a live counter for the thirty seconds bad in terms of the constant state updates? I don't like that it'd be kind of distracting, but I also don't really like that "last updated 30 seconds ago" will only actually be accurate for one second you know? Maybe it could just be "updates every 30 seconds"?

jpellizzari:
> > > > Code LGTM, but the timestamp is way too prominent. It is not as important as the name of the page (which it is displayed next to). Can we move this somewhere?
> > > > Also, let's use a relative timestamp string: `last updated 30s ago`.
> > > 
> > > 
> > > Are you envisioning that as like a live counter, or just always saying "last updated 30 seconds ago"? I think it's nice to give some context on why the app list might not appear like the user expects it to, but I definitely agree that the timestamp is a little extra right now
> > 
> > 
> > It's more about what should be important to the user. When we give the user a timestamp like `13:38:22`, we are making them do math to figure out "when was that in relation to now?". Just give them the info they are going to have to figure out for themselves.
> > In the case of "I need to find events that correspond to an incident", THEN we can show explicit timestamps, since that's their "search" modality.
> 
> Is having a live counter for the thirty seconds bad in terms of the constant state updates? I don't like that it'd be kind of distracting, but I also don't really like that "last updated 30 seconds ago" will only actually be accurate for one second you know? Maybe it could just be "updates every 30 seconds"?

Nah the relative timestamp text is the way to go if we want to have any text at all IMO. The periodic spinner might be enough to signal that the page will update automatically.

bigkevmcd:
The common pattern I've seen is to show the "last updated 30 seconds ago" and if you hover over, to popup the actual timestamp?

jpellizzari:
> The common pattern I've seen is to show the "last updated 30 seconds ago" and if you hover over, to popup the actual timestamp?

https://c.tenor.com/2jgbEi-vCuQAAAAC/mando-way-this-is-the-way.gif

bia:
WFM

Could you move down 'last updated' (horizontally align the two headings) and change the color to `#737373`?
Screenshot 2021-12-07 at 18 58 34

bia (approved):

jpellizzari (approved):
Approving to remove the merge blocking

ui/pages/Applications.tsx
```diff
@@ -16,24 +18,41 @@ type Props = {
...
-    <Page loading={loading} className={className}>
-      <Flex align between>
-        <TitleBar>
+    <Page className={className}>
+      <Flex between align>
+        <Flex align>
           <h2>Applications</h2>
-        </TitleBar>
+          <Spacer padding="small" />
+          <PollingIndicator loading={loading} interval="30" />
```
AlinaGoaga, line 47: This looks a bit misaligned I think, there's a margin-bottom in the css that might be the culprit?

ui/components/PollingIndicator.tsx
```diff
@@ -0,0 +1,60 @@
...
+  /** whether `<CircularProgress />` should be shown */
+  loading: boolean;
+  interval: string;
+  className?: string;
+};
+
+function PollingIndicator({ className, loading, interval }: Props) {
+  const date = new Date();
+  const [updated, setUpdated] = React.useState(date.toTimeString().slice(0, 8));
+  const [indicator, setIndicator] = React.useState(false);
```
AlinaGoaga, line 19: I would use visible, setVisible here since we're in the PollingIndicator component so we know what we're referencing by that attribute

AlinaGoaga (approved):
Hey Josh! This looks v good to me, I can see the polling in the network tab working well also. I've left some tiny comments, no blockers though. Thanks :)

PR #1155, opened 2021-12-01
update protobuf files

palemtnrider:
**What changed?**
Update tools.go with used protobuf plugins.  Update the Makefile to use new version of `buf mod update` and change the googleapis to the non-beta version.

**Why?**
While building protobuf files from a clean environment I received a message about older version of googleapis being used.  I also received a message about missing plugins and noticed some were defined in the tools.go file but not all.

**How did you test it?**
Manually via make proto and make proto-deps.  Ran the unit-tests to ensure they still passed.

**Release notes**
No.

**Documentation Changes**
None.


jpellizzari (approved):
:+1:

luizbafilho (approved):

PR #1162, opened 2021-12-02
Set explicit node version

jpellizzari:
Release job was using some node version other than 14.x

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -23,6 +23,10 @@ jobs:
         uses: actions/setup-go@v2
         with:
           go-version: 1.16.x
+      - name: Use Node.js
+        uses: actions/setup-node@v1
+        with:
+          node-version: 14.x
       - name: Set env var
         run: |
           make -B dependencies


josecordaz (approved):

.github/workflows/release.yaml
```diff
@@ -21,6 +21,10 @@ jobs:
           git fetch --tags -f
       - name: Setup Go
         uses: actions/setup-go@v2
+      - name: Use Node.js ${{ matrix.node-version }}
```
palemtnrider, line 24: Not certain if it matters, but looking at the file and I think this block should be before the next `name:` The 
```
with:
  go-version:...
```
Seems like it should go with the Setup Go entry

palemtnrider (approved):
Not certain if the position of the changes makes a difference but calling it out just in case.  LGTM otherwise

.github/workflows/release.yaml
jpellizzari: Oh wow good catch. That would not have worked.

PR #1167, opened 2021-12-03
Validate namespaces in Gitops Install and Uninstall

luizbafilho:
Closes: 

**What changed?**

**Why?**

**How did you test it?**

**Release notes**

**Documentation Changes**


pkg/kube/kubehttp.go
```diff
@@ -376,6 +390,91 @@ func (k *KubeHTTP) SetResource(ctx context.Context, resource Resource) error {
 	return nil
 }
 
+func (k *KubeHTTP) SetWegoConfig(ctx context.Context, config WegoConfig, namespace string) error {
+	configBytes, err := kyaml.Marshal(config)
+	if err != nil {
+		return fmt.Errorf("failed marshalling wego config: %w", err)
+	}
+
+	name := types.NamespacedName{Name: "weave-gitops-config", Namespace: namespace}
```
J-Thompson12, line 399: NIT: You might as well use the WegoConfigMapName var you set at the top here.

pkg/kube/kubehttp.go
```diff
@@ -376,6 +390,91 @@ func (k *KubeHTTP) SetResource(ctx context.Context, resource Resource) error {
...
+			Kind:       "ConfigMap",
+			APIVersion: corev1.SchemeGroupVersion.String(),
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      name.Name,
+			Namespace: name.Namespace,
+		},
+	}
+
+	if err := k.Client.Get(ctx, name, cm); err != nil {
```
J-Thompson12, line 412: Out of curiosity why not use the GetWegoConfig function here?

pkg/kube/kubehttp.go
```diff
@@ -58,6 +60,14 @@ var (
 	GVRHelmRelease    schema.GroupVersionResource = helmv2.GroupVersion.WithResource("helmreleases")
 )
 
+var (
+	WegoConfigMapName = "weave-gitops-config"
```
J-Thompson12, line 64: This should be a const.

pkg/kube/kubehttp.go
```diff
@@ -376,6 +390,91 @@ func (k *KubeHTTP) SetResource(ctx context.Context, resource Resource) error {
...
+		return fmt.Errorf("failed getting weave-gitops configmap: %w", err)
+	}
+
+	cm.Data["config"] = string(configBytes)
+
+	return k.Client.Update(ctx, cm)
+}
+
+func (k *KubeHTTP) GetWegoConfig(ctx context.Context, namespace string) (*WegoConfig, error) {
+	name := types.NamespacedName{Name: "weave-gitops-config", Namespace: namespace}
```
J-Thompson12, line 434: Same as above. Use the global var.

pkg/kube/kubehttp_test.go
```diff
@@ -422,6 +422,71 @@ metadata:
...
+			ctx := context.Background()
+
+			err := k.SetWegoConfig(ctx, kube.WegoConfig{FluxNamespace: "foo"}, namespace.Name)
+			Expect(err).NotTo(HaveOccurred())
+
+			wegoConfig, err := k.GetWegoConfig(ctx, namespace.Name)
+			Expect(err).NotTo(HaveOccurred())
+			Expect(wegoConfig.FluxNamespace).To(Equal("foo"))
+		})
+	})
```
J-Thompson12, line 488: There needs to be some fail tests here.

pkg/kube/kubehttp.go
luizbafilho: I just wanna update the data field in the ConfigMap, the `GetWegoConfig` returns the parsed data, not the configmap itself

pkg/kube/kubehttp_test.go
luizbafilho: just added one for getting the config, but since we use actually calls to a kube API server, we can't mock whatever we want there, there is no way to force a generic error for those calls

J-Thompson12 (approved):
LGTM

pkg/services/gitops/install.go
```diff
@@ -229,3 +259,47 @@ func (g *Gitops) applyManifestsToK8s(ctx context.Context, namespace string, mani
...
+		LabelSelector: selector,
+	}
+
+	nsl := &corev1.NamespaceList{}
+	if err := g.kube.Raw().List(ctx, nsl, &options); err != nil {
+		return "", fmt.Errorf("error setting resource: %w", err)
+	}
+
+	if len(nsl.Items) == 0 {
+		return "", ErrNamespaceNotFound
```
JamWils, line 287: 💯 on the error variable here.

pkg/services/gitops/install.go
```diff
@@ -229,3 +259,47 @@ func (g *Gitops) applyManifestsToK8s(ctx context.Context, namespace string, mani
...
+	}
+
+	return nsl.Items[0].Name, nil
+}
+
+func (g *Gitops) saveWegoConfig(ctx context.Context, params InstallParams) error {
+	fluxNamespace, err := g.fetchNamespaceWithLabel(ctx, LabelPartOf, "flux")
+	if err != nil {
+		if !errors.Is(err, ErrNamespaceNotFound) {
```
JamWils, line 296: It looks like you are dropping the `ErrNamespaceNotFound` error here.  What about?

```
if errors.Is(err, ErrNamespaceNotFound) {
  return err
else if err != nil {
 return fmt.Errorf("failed fetching flux namespace: %w", err)
}
```
Also can you write some unit tests capturing this scenario when `Install` is called?

pkg/kube/kubehttp.go
```diff
@@ -376,6 +390,91 @@ func (k *KubeHTTP) SetResource(ctx context.Context, resource Resource) error {
...
+
+			return nil
+		}
+
+		return fmt.Errorf("failed getting weave-gitops configmap: %w", err)
+	}
+
+	cm.Data["config"] = string(configBytes)
+
+	return k.Client.Update(ctx, cm)
```
palemtnrider, line 430: Nit: Code is more symmetric if you also error check the Update() call similar to Create() i.e., adding "failed updating weave-gitops configmap"

pkg/kube/kubehttp.go
```diff
@@ -376,6 +390,91 @@ func (k *KubeHTTP) SetResource(ctx context.Context, resource Resource) error {
...
+	if namespace != "" {
+		if err := k.Client.Get(ctx, name, cm); err != nil {
+			if apierrors.IsNotFound(err) {
+				return &WegoConfig{}, ErrWegoConfigNotFound
+			}
+
+			return nil, fmt.Errorf("failed getting weave-gitops configmap: %w", err)
+		}
+	} else {
+		configMap, err := k.getWegoConfigMapFromAllNamespaces(ctx)
```
palemtnrider, line 447: This is a bit confusing to me.  Typically no namespace implies "default" namespace.  This code pulls all configmaps and returns the first that has a matching name.  I think this behavior might be confusing if there are many configmaps across multiple namespaces - unlikely but likely causes additional issues.  I don't think we should propagate them here.  I think the code should either 
* if no namespace, use the default weave-gitops namespace 
* or return an error and require the caller to set the namespace

pkg/kube/kubehttp.go
```diff
@@ -376,6 +390,91 @@ func (k *KubeHTTP) SetResource(ctx context.Context, resource Resource) error {
...
+		if err != nil {
+			return &WegoConfig{}, err
+		}
+
+		cm = configMap
+	}
+
+	wegoConfig := &WegoConfig{}
+	if err := kyaml.Unmarshal([]byte(cm.Data["config"]), wegoConfig); err != nil {
+		return nil, nil
```
palemtnrider, line 457: Any particular reason we are not returning this error to the caller?

pkg/kube/kubehttp.go
palemtnrider: If you do keep this behavior, it should be documented for the exported function.

pkg/kube/kubehttp_test.go
```diff
@@ -422,6 +422,76 @@ metadata:
...
+			Expect(wegoConfig.FluxNamespace).To(Equal("flux-system"))
+		})
+
+		It("fails getting config map", func() {
+			_, err := k.GetWegoConfig(context.Background(), "foo")
+			Expect(err.Error()).To(ContainSubstring("Wego Config not found"))
+		})
+	})
+
+	Describe("SetWegoConfig", func() {
```
palemtnrider, line 482: Please add a test for an empty namespace value.

pkg/services/gitops/install.go
```diff
@@ -229,3 +259,47 @@ func (g *Gitops) applyManifestsToK8s(ctx context.Context, namespace string, mani
...
+	nsl := &corev1.NamespaceList{}
+	if err := g.kube.Raw().List(ctx, nsl, &options); err != nil {
+		return "", fmt.Errorf("error setting resource: %w", err)
+	}
+
+	if len(nsl.Items) == 0 {
+		return "", ErrNamespaceNotFound
+	}
+
+	return nsl.Items[0].Name, nil
```
palemtnrider, line 290: We need to either error out of the list of namespaces > 1 or at least print a warning message with the NS you selected to use.  Assuming the first one is the right one will cause issues later.

pkg/services/gitops/install_test.go
```diff
@@ -56,18 +76,46 @@ var _ = Describe("Install", func() {
 		Expect(os.RemoveAll(dir)).To(Succeed())
 	})
 
-	It("checks cluster status", func() {
-		kubeClient.GetClusterStatusStub = func(c context.Context) kube.ClusterStatus {
-			return kube.FluxInstalled
-		}
-		_, err := gitopsSrv.Install(installParams)
-		Expect(err).Should(MatchError("Weave GitOps does not yet support installation onto a cluster that is using Flux.\nPlease uninstall flux before proceeding:\n  $ flux uninstall"))
+	Context("validate install", func() {
```
palemtnrider, line 79: Please add a test for multiple namespaces with the label

pkg/services/gitops/install.go
```diff
@@ -78,13 +80,17 @@ func (g *Gitops) Install(params InstallParams) (map[string][]byte, error) {
...
 			}
 		}
 
 		systemManifests["wego-app.yaml"] = bytes.Join(wegoAppManifests, []byte("---\n"))
 	}
 
+	if err := g.saveWegoConfig(ctx, params); err != nil {
+		return nil, err
+	}
+
```
palemtnrider, line 93: The wego config should be part of the manifests returned.  That way they show up in the dryrun and will be persisted.

pkg/services/gitops/uninstall.go
```diff
@@ -27,22 +29,42 @@ func (g *Gitops) Uninstall(params UninstallParams) error {
...
-	if err := g.flux.Uninstall(params.Namespace, params.DryRun); err != nil {
-		g.logger.Printf("received error uninstalling flux: %q, continuing with uninstall", err)
+	wegoConfig, err := g.kube.GetWegoConfig(ctx, params.Namespace)
+	if err != nil {
+		return fmt.Errorf("failed getting wego config in namespace=%s: %w", params.Namespace, err)
+	}
+
+	uninstallFlux := true
+	if wegoConfig.FluxNamespace != params.Namespace {
+		uninstallFlux = false
```
palemtnrider, line 59: This confused me at first. Please add a comment that we only remove flux if it matches what our wego config is.  Or hide the flux uninstall in a function with a name like removeFluxIfMatchingWegoConfig()

pkg/services/gitops/uninstall_test.go
```diff
@@ -173,10 +178,26 @@ var _ = Describe("Uninstall", func() {
...
 		Expect(err).ShouldNot(HaveOccurred())
 
-		Expect(kubeClient.DeleteCallCount()).To(Equal(1))
+		wegoAppManifests, err := manifests.GenerateWegoAppManifests(manifests.WegoAppParams{Version: version.Version, Namespace: "default"})
+		Expect(err).ShouldNot(HaveOccurred())
+
+		Expect(kubeClient.DeleteCallCount()).To(Equal(len(wegoAppManifests)+1), "deletes all wego app manifests plus the app crd")
+	})
+
+	It("fails if cant fetch wego config", func() {
```
palemtnrider, line 187:
```suggestion
	It("fails if we can't fetch the wego config", func() {
```

palemtnrider (changes requested):
Just a few changes here and here.

pkg/kube/kubehttp.go
luizbafilho: I totally agree is confusing, that came out from me trying to find a good way to search if there is a Wego Config in the cluster, since having just one by cluster is mandatory, the validation has been added.

pkg/services/gitops/install_test.go
luizbafilho: that case doesn't exist the goal of this PR is to guarantee that.

pkg/services/gitops/install.go
luizbafilho: as mentioned before, there shouldn't be more than one, that's guaranteed in the installation, since this is not an external method for other parts of the code to consume, and belongs entirely to this business logic,  it's ok to assume the first one is the correct one.

pkg/kube/kubehttp.go
luizbafilho: good catch, it should

pkg/kube/kube.go
```diff
@@ -55,6 +61,9 @@ type Kube interface {
 	GetResource(ctx context.Context, name types.NamespacedName, resource Resource) error
 	SetResource(ctx context.Context, resource Resource) error
 	GetSecret(ctx context.Context, name types.NamespacedName) (*corev1.Secret, error)
+	SetWegoConfig(ctx context.Context, config WegoConfig, namespace string) error
+	GetWegoConfig(ctx context.Context, namespace string) (*WegoConfig, error)
```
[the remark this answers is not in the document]
luizbafilho: I disagree, the doc linked refers to getter and setter on structs, while in service clients we often see `Get` or `Fetch` prefixes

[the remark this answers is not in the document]
palemtnrider: You're right.  It is specific about structs.  Just for my edification, do we have documentation around when we should prefix with Get and when we shouldn't?

pkg/services/gitops/install.go
palemtnrider: I'd still like to see a log message here that we are selecting the first NS in the list if there is >1 found.  :+1 to the go doc on the calling function but having a log message, in the unlikely event there is >1 NS with the label, will help us debug.  Optionally, if we are guaranteed to only have at most 1, then if we encounter more than one we are dealing with a badly configured cluster and will experience unknown issues so erroring out would be a good thing.

palemtnrider (approved):
Just a follow-up question and would still like reporting (or erroring) on >1 NS found - but I'll let you decide.  Approved.

pkg/kube/kube.go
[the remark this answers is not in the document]
luizbafilho: I don't know any docs stating that

pkg/services/gitops/install.go
luizbafilho: fair enough, adding the error

pkg/services/gitops/install.go
```diff
@@ -26,13 +31,11 @@ type InstallParams struct {
...
 func (g *Gitops) Install(params InstallParams) (map[string][]byte, error) {
 	ctx := context.Background()
-	status := g.kube.GetClusterStatus(ctx)
 
-	switch status {
-	case kube.FluxInstalled:
-		return nil, errors.New("Weave GitOps does not yet support installation onto a cluster that is using Flux.\nPlease uninstall flux before proceeding:\n  $ flux uninstall")
-	case kube.Unknown:
-		return nil, errors.New("Weave GitOps cannot talk to the cluster")
+	g.logger.Actionf("Validating Weave Gitops installation")
```
Skarlso, line 35: This Log message here, breaks the output of validateWegoInstall.

https://github.com/weaveworks/aws-marketplace/runs/4500418327?check_suite_focus=true

Results in a generated output in case dry run is used such as this:

```
► Validating Weave Gitops installation
---
# Flux version: v0.21.0
# Components: source-controller,kustomize-controller,helm-controller,notification-controller,image-reflector-controller,image-automation-controller
apiVersion: v1
kind: Namespace
metadata:
  labels:
    app.kubernetes.io/instance: wego-system
    app.kubernetes.io/part-of: flux
    app.kubernetes.io/version: v0.21.0
  name: wego-system
---
```

PR #1171, opened 2021-12-03
Fail when app is added with prefix wego

J-Thompson12:
Closes: #1075 

**What changed?**
Added a check on add to not allow a user to add and app with the prefix "wego-". This prevents the app name from matching the flux cluster name.

**Why?**

**How did you test it?**
Manually and with a unit test.

**Release notes**

**Documentation Changes**

### DIFF
--- a/pkg/services/app/add.go
+++ b/pkg/services/app/add.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"strings"
 
 	"github.com/weaveworks/weave-gitops/pkg/git"
 	"github.com/weaveworks/weave-gitops/pkg/gitproviders"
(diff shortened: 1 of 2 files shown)


rokshana-b (approved):

pkg/services/app/add.go
```diff
@@ -73,6 +74,10 @@ func (a *AppSvc) Add(configGit git.Git, gitProvider gitproviders.GitProvider, pa
 		return err
 	}
 
+	if strings.HasPrefix(params.Name, "wego-") {
```
palemtnrider, line 77: The check doesn't match the error message.  I was able to create an app with just `wego` - it doesn't cause an issue but the check needs to match the error.

palemtnrider (approved):
Just a minor change on the error message (or the prefix check) other than that LGTM.  Marking approved assuming the message/check is updated.

PR #1180, opened 2021-12-06
Invert boldness in font definitions

ozamosi:
The bold was listed as regular weight, and vice versa, which made the
UI bold everywhere except for in headers.

Closes: 

**What changed?**
 Fix which version of proxima-nova is considered bold and which one isn't

**Why?**
The bold font was listed as regular, while the regular font was listed as bold - that made the whole UI bold, except for in headers and other places of emphasis.

**How did you test it?**
 Looking at the UI

**Release notes**
No

**Documentation Changes**
No


joshri (approved):
LGTM Thanks!!

PR #1186, opened 2021-12-06
Small Storybook Fixes

### LINKED ISSUE #1185: Storybook is empty when running it
**Describe the bug**
When I run `npm run storybook` I am presented with a blank page.  Fix storybook so I can see the existing components.

joshri:
Closes #1185 

Storybook took a second to load, but it looks like it does appear after a second or two - I added some Typescript config options in `.storybook/main.js` that could help, as well as formatted components so their auto-importing for docs actually works. Removed some Button styling that was made redundant by new defaults as well.


AlinaGoaga (approved):
Hey Josh! This looks good to me! I've run `make node_modules` followed by `npm run storybook` and was able to access the Storybook without problems. Let's continue the convo on how we want to organise this stuff in the new year, this looks v promising! Thanks!

PR #1196, opened 2021-12-08
Add gitlab exports

AlinaGoaga:
**What changed?**
- Added new exports to support Gitlab Auth in EE. 
- Added a new prop to RepoInputWithAuth.tsx so we can make its input disabled for EE
- Allow configuration of the gitlab app via env vars


foot (approved):
Molto bene!

Issue #1203, opened 2021-12-09
Support on-prem gitlab


davidstauffer:
As a User, I want to use my on-prem GitLab that uses an URL like (gitlab.mycustomdomain.com/myorg/myrepo.git)  so that I can use Weave GitOps.

We need to support onprem gitlab for engagements we have with current customers. 
Currently commands like `gitops install --app-config-url https://gitlab.mycustomdomain.com/myorg/myrepo.git` with a onprem gitlab URL will fail.
 
`Error: error creating git clients: error normalizing url: could not get provider name from URL https://gitlab.mycustomdomain.com/myorg/myrepo.git: no git providers found for "https://gitlab.mycustomdomain.com/myorg/myrepo.git"`

We need to make sure a custom URL can be provided.

LutzLange:
We are running a Trial with Amgen and this blocks them. They are currently trying to work around by using https://gitlab.com.

foot:
- We derive provider from the URL right now (in core), needs some work to have extra flags / env to override auto-detection!
  - Update the `getProvider(url string)` fn to look for env var override? (cheap and easy)
  - EE we have an env var for PROVIDER right now (and can be set in values.yaml)
- known_hosts is generated at build time for EE (and derived from users terminal in core)
  - Need a way to add additional entries to known_hosts.
  - Update core: Grab the known_hosts from the GitRepository CR when add apping etc.

foot:
Also need to pass a flag through to flux boostrap (`--hostname`) , maybe for known_hosts?

PR #1219, opened 2021-12-13
Allow specifying git-host-to-provider-type mapping for on-prem gitlab support

foot:
Adds support for:
- `gitops install` for on-prem gitlab
- oauth for one on-prem host _or_ gitlab.com at a time (globally set via `GITLAB_HOST`)

On oauth with multiple gitlab hosts (requires more work):
  - We must configure a gitlab-app for each gitlab host
  - UI tokens saved in localStorage should be keyed by host, not provider.

On github:
- `gitops install` for on-prem github _might_ work (untested)
- oauth support for on-prem github probably doesn't work?.. (untested)

Closes: #1203 

**What changed?**

- New flag / env var `git-host-types` / `GITOPS_GIT_HOST_TYPES` to specify a mapping
  - `--git-host-types="gl.acme.org=gitlab,gh.acme.org=github"`

**Why?**

Support `gitops install` on-prem non-gitlab.com hosts

**How did you test it?**

- Manually against on-prem gitlab

**Release notes**

Not yet..

**Documentation Changes**

Not yet..


pkg/gitproviders/factory.go
```diff
@@ -41,28 +41,40 @@ func buildGitProvider(config Config) (gitprovider.Client, string, error) {
...
 		}
-		if config.Hostname != "" {
-			opts = append(opts, gitprovider.WithDomain(config.Hostname))
+
+		// Quirk, see above
+		hostname := gitlab.DefaultDomain
+		if config.Hostname != "" && config.Hostname != gitlab.DefaultDomain {
+			// Quirk, see above
+			hostname = "https://" + config.Hostname
+			opts = append(opts, gitprovider.WithDomain(hostname))
```
foot, line 71: Factor this out?

pkg/services/auth/internal/gitlab.go
```diff
@@ -38,6 +38,10 @@ func getEnvDefault(key, defaultValue string) string {
 	return val
 }
 
+func getGitlabHost() string {
+	return getEnvDefault("GITLAB_HOSTNAME", gitlabHost)
+}
```
foot, line 43: Allow configuring where to send oauth requests..

pkg/gitproviders/repo_url.go
```diff
@@ -139,18 +150,33 @@ func normalizeRepoURLString(url string, providerName GitProviderName) string {
...
-		trimmed = strings.TrimPrefix(url, sshPrefix)
-	} else if strings.HasPrefix(url, httpsPrefix) {
-		trimmed = strings.TrimPrefix(url, httpsPrefix)
+// ViperGetStringMapString looks up a command line flag or env var in the format "foo=1,bar=2"
+// GetStringMapString tries to JSON decode the env var
+// If that fails (silently), try and decode the classic "foo=1,bar=2" form.
+// https://github.com/spf13/viper/issues/911
+func ViperGetStringMapString(key string) map[string]string {
+	sms := viper.GetStringMapString(key)
+	if len(sms) > 1 {
```
foot, line 167:
```suggestion
	if len(sms) > 0 {
```

cmd/gitops/root/cmd.go
```diff
@@ -134,3 +143,12 @@ func configureLogger() {
 		log.SetLevel(log.DebugLevel)
 	}
 }
+
+func initViper(cmd *cobra.Command) error {
+	replacer := strings.NewReplacer("-", "_")
+	viper.SetEnvKeyReplacer(replacer)
+	viper.SetEnvPrefix("GITOPS")
+	viper.AutomaticEnv()
```
bigkevmcd, line 152: You don't need to do this bit for each command you register?

cmd/gitops/root/cmd.go
```diff
@@ -85,6 +88,11 @@ func RootCmd(client *resty.Client) *cobra.Command {
 `, wego.DefaultNamespace),
 		PersistentPreRun: func(cmd *cobra.Command, args []string) {
 			configureLogger()
+			err := initViper(cmd)
+			if err != nil {
+				fmt.Fprintf(os.Stderr, "Error initing viper: %v\n", err)
+				os.Exit(1)
+			}
```
bigkevmcd, line 95: You could use [log.Fatalf()](https://pkg.go.dev/log#Fatalf) here, it writes to Stderr, and exits with code 1.

pkg/gitproviders/factory.go
```diff
@@ -41,28 +41,40 @@ func buildGitProvider(config Config) (gitprovider.Client, string, error) {
 		opts := []gitprovider.ClientOption{
 			gitprovider.WithOAuth2Token(config.Token),
 		}
-		if config.Hostname != "" {
-			opts = append(opts, gitprovider.WithDomain(config.Hostname))
+
+		// Quirk of ggp, if using github.com or gitlab.com and you prepend
+		// that with https:// you end up with https://https//github.com !!!
+		// What is that...
+		hostname := github.DefaultDomain
```
bigkevmcd, line 48: How about opening an issue in GGP?

pkg/services/auth/internal/gitlab.go
bigkevmcd: Should we be documenting this somewhere for users?

pkg/gitproviders/repo_url.go
```diff
@@ -139,18 +150,33 @@ func normalizeRepoURLString(url string, providerName GitProviderName) string {
...
+func ViperGetStringMapString(key string) map[string]string {
+	sms := viper.GetStringMapString(key)
+	if len(sms) > 0 {
+		return sms
 	}
 
-	if trimmed != "" {
-		return fmt.Sprintf("ssh://git@%s.com/%s", providerName, trimmed)
+	ss := viper.GetStringSlice(key)
+	out := make(map[string]string, len(ss))
```
bigkevmcd, line 172: I am not sure there's any benefit in providing the `cap` here?

bigkevmcd (changes requested):

pkg/gitproviders/repo_url.go
foot: Nope, brought the code over from https://github.com/spf13/pflag/blob/master/string_to_string.go#L90-L98 which might need an attribution actually?

pkg/services/auth/internal/gitlab.go
foot: Not yet it doesn't feel like a very complete feature. See PR description, it becomes the global HOST for all gitlab oauth requests. I could make some follow up issues though.

cmd/gitops/root/cmd.go
foot: This is the root command and seems to work for `install` and `ui run` at least so assume it works for others.

cmd/gitops/root/cmd.go
foot: Nice! I copied the existing pattern in the file to make minimal changes, will test this out too!

luizbafilho (approved):
Thanks, that looks good.

cmd/gitops/root/cmd.go
```diff
@@ -33,6 +35,18 @@ var options struct {
 	endpoint          string
 	overrideInCluster bool
 	verbose           bool
+	gitHostTypes      map[string]string
+}
+
+// Only want AutomaticEnv to be called once!
+func init() {
+	// Setup flag to env mapping:
+	//   config-repo => GITOPS_CONFIG_REPO
```
bigkevmcd, line 44: Thanks for adding this!

bigkevmcd (approved):
Looks good!

efenex:
We have a setup with an onprem GitLab server running SSH on a non-standard port (2222 as opposed to 22) and I believe it is currently not possible to get this to work with the code in this PR.

I try the installation with something like:

```
gitops install --config-repo=git@gitlab.acme.org:2222/deployments/gitops-weaveworks.git --namespace gitops --git-host-types="gitlab.acme.org:2222=gitlab"
```

And it fails with:

```
Error: error creating git clients: error normalizing url: could not get provider name from URL git@gitlab.acme.org:2222/deployments/gitops-weaveworks.git: no git providers found for "git@gitlab.acme.org:2222/deployments/gitops-weaveworks.git"
```

Would be nice if there was a way to specify how SSH authentication should take place, as I'm not sure how to specify an SSH key or a PAT for this URL at all.

foot:
Hi @efenex ! Thanks for testing this out.

Hmmm, just checking you've tried `--git-host-types="gitlab.acme.org=gitlab"` without the port?

Custom ports is indeed something we overlooked in this PR 🤦 .

foot:
This code is a bit suspect https://github.com/weaveworks/weave-gitops/blob/main/pkg/gitproviders/repo_url.go#L131 and should probably be replaced with something more robust like https://pkg.go.dev/github.com/whilp/git-urls

To workaround, things to try:

Use the full ssh:// form:
```
gitops install --config-repo=ssh://git@gitlab.acme.org:2222/deployments/gitops-weaveworks.git --namespace gitops --git-host-types="gitlab.acme.org:2222=gitlab"
```

Issue #1211, opened 2021-12-10
GitOps UI tries to authenticate user with wrong git host provider


rokshana-b:
**Describe the bug**

Given I've added an app from a gitlab repo to my cluster, if I try to authenticate myself to view `app commit list`, I'm taken to `GitHub` auth page. Since my app repo is originating from `GitLab`, the auth process fails and I'm not able to see the app commit list. 

**What is the severity of the bug**

| severity/Major | Weave Gitops functionality is broken, there is a workaround, and the workaround requires significant effort |
| --- | --- |

**Environment**
 - gitops: [v0.5.0]

**To Reproduce**

1. Run `gitops install --config-repo=<git@gitlab>
2. Run `gitops add app --config-repo=<git@gitlab> --url=<git@gitlab>`
3. Run `gitops ui run`
4. Go to UI page and see app listed
5. Click on app link and see app status page
6. At the bottom of the page, see button `Authenticate with GitHub`
7. As the button title states, auth flow will be from github which is irrelevant for a gitlab workflow

**Expected behavior**

Ui should detect which git host provider is being used and provide relevant authentication method.

**Screenshots**

![image.png](https://images.zenhubusercontent.com/606389351ead0c83798a9a73/e6993e92-da26-459d-ae47-7e4182bdce2e)

JamWils:
@rokshana-b and @jmickey, I won't both github and gitlab manually tested before we merge this story.

PR #1221, opened 2021-12-13
Add GitLab auth flow to application detail and remove

joshri:
Closes: #1211 

GitLab auth flow was not implemented past the Application Add form - I added the context wrapper in ApplicationDetail and ApplicationRemove as well as the button in the AuthAlert component itself.

### DIFF
--- a/ui/components/CommitsTable.tsx
+++ b/ui/components/CommitsTable.tsx
@@ -20,6 +20,7 @@ type Props = {
   app: Application;
   authSuccess: boolean;
   onAuthClick?: () => void;
+  provider: string;
 };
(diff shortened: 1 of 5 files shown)


JamWils (approved):
It would be nice if we could reduce some of this duplicative code from the `CallbackStateContextProvider`.  In addition, we'll need to retroactively go back and add unit tests.

JamWils:
## Testing Notes

There are issues with removing an app via Gitlab.  We tested adding  an app, viewing the commit list, and removing an app via Gitlab.  The user is now taken through the correct git flow.

PR #1232, opened 2021-12-15
Revert field name from config_repo to config_url in wego app CRD

josecordaz:
Closes: 

**What changed?**
- Revert field name from config_repo to config_url in wego app CRD

**Why?**
- Avoid breaking changes

**How did you test it?**
- I used `gitops` version 0.5.0 to create a config repo and then I used that repo in install with this branch changes.
- I added an app using the UI with a config repo.
Both scenarios result in app deployments.

**Release notes**

**Documentation Changes**

### DIFF
--- a/api/v1alpha1/application_types.go
+++ b/api/v1alpha1/application_types.go
@@ -27,7 +27,7 @@ const ApplicationKind = "Application"
 // ApplicationSpec defines the desired state of Application
 type ApplicationSpec struct {
 	// ConfigRepo is the address of the git repository containing the automation for this application
-	ConfigRepo string `json:"config_repo,omitempty"`
+	ConfigRepo string `json:"config_url,omitempty"`
 	// URL is the address of the git repository for this application
 	URL string `json:"url,omitempty"`
 	// Path is the path in the repository where the k8s yaml files for this application are stored.
--- a/manifests/crds/wego.weave.works_apps.yaml
+++ b/manifests/crds/wego.weave.works_apps.yaml
@@ -40,7 +40,7 @@ spec:
                 description: Branch is the branch in the repository where the k8s
                   yaml files for this application are stored.
                 type: string
-              config_repo:
+              config_url:
                 description: ConfigRepo is the address of the git repository containing
                   the automation for this application
                 type: string


JamWils (approved):

api/v1alpha1/application_types.go
```diff
@@ -27,7 +27,7 @@ const ApplicationKind = "Application"
 // ApplicationSpec defines the desired state of Application
 type ApplicationSpec struct {
 	// ConfigRepo is the address of the git repository containing the automation for this application
-	ConfigRepo string `json:"config_repo,omitempty"`
+	ConfigRepo string `json:"config_url,omitempty"`
```
luizbafilho, line 30: I think we should rename the Field itself too, having it in the yaml only will make it even more confusing.

luizbafilho (changes requested):

api/v1alpha1/application_types.go
josecordaz: I think it would be more confusing having more places with ConfigURL than just this single one

josecordaz: do you mind if we merge this now? and discuss it later?. I think we can come up with an even better solution for all cases. Renaming this won't be the final solution for all the cases involved.

josecordaz: @luizbafilho ^

manifests/crds/wego.weave.works_apps.yaml
```diff
@@ -40,7 +40,7 @@ spec:
                 description: Branch is the branch in the repository where the k8s
                   yaml files for this application are stored.
                 type: string
-              config_repo:
+              config_url:
                 description: ConfigRepo is the address of the git repository containing
```
palemtnrider, line 44:
```suggestion
                description: ConfigUrl is the address of the git repository containing
```

luizbafilho (approved):

palemtnrider (approved):
LGTM.  Please add a follow-up issue around the mixed use for config_url and config_repo

manifests/crds/wego.weave.works_apps.yaml
josecordaz: done https://github.com/weaveworks/weave-gitops/issues/1253

PR #1243, opened 2021-12-16
Forcing kube apply

luizbafilho:
**What changed?**
Forces kube manifest apply to avoid conflict with Flux due service side apply 

for reference: https://kubernetes.io/docs/reference/using-api/server-side-apply/#conflicts

**Why?**

**How did you test it?**

**Release notes**

**Documentation Changes**

### DIFF
--- a/pkg/kube/kubehttp.go
+++ b/pkg/kube/kubehttp.go
@@ -213,9 +213,12 @@ func (k *KubeHTTP) Apply(ctx context.Context, manifest []byte, namespace string)
 		return fmt.Errorf("failed to dynamic resource interface: %w", err)
 	}
 
+	force := true
 	_, err = dr.Patch(ctx, name, types.ApplyPatchType, data, metav1.PatchOptions{
 		FieldManager: "wego",
+		Force:        &force,
 	})
+
 	if err != nil {
 		return fmt.Errorf("failed applying %s: %w", string(data), err)
 	}


pkg/kube/kubehttp.go
```diff
@@ -213,9 +213,12 @@ func (k *KubeHTTP) Apply(ctx context.Context, manifest []byte, namespace string)
 		return fmt.Errorf("failed to dynamic resource interface: %w", err)
 	}
 
+	force := true
 	_, err = dr.Patch(ctx, name, types.ApplyPatchType, data, metav1.PatchOptions{
 		FieldManager: "wego",
```
luizbafilho, line 218: @bigkevmcd I think we could actually avoid setting this `FieldManager` right? that might avoid the conflict too.

pkg/kube/kubehttp.go
```diff
@@ -213,9 +213,12 @@ func (k *KubeHTTP) Apply(ctx context.Context, manifest []byte, namespace string)
 		return fmt.Errorf("failed to dynamic resource interface: %w", err)
 	}
 
+	force := true
```
palemtnrider, line 216: Always forcing this will cause all the manifests to be force patched on install.  I'd prefer that we only force the install for the application CRD.  I don't know if there would be issues force installing other manifests from flux.

palemtnrider: After discussing this a bit more, we are going to proceed this way (after some more testing) and look at a more complete upgrade process.

PR #1245, opened 2021-12-17
Run acceptance tests in both github and gitlab

rokshana-b:
**What changed?**
Added a new env GIT_PROVIDER

**Why?**

**How did you test it?**

### DIFF
--- a/.github/actions/run-acceptance-test/action.yml
+++ b/.github/actions/run-acceptance-test/action.yml
@@ -29,6 +29,9 @@ inputs:
 
 runs:
   using: "composite"
+  strategy:
+    matrix:
+      git-provider: [github, gitlab]
   steps:
   - name: Install Go
     uses: actions/setup-go@v2
@@ -71,7 +74,7 @@ runs:
     run: |
       chmod +x bin/gitops
       ls -la bin
-  - name: Download dependecies
+  - name: Download dependencies
     shell: bash
     run: |
       make dependencies
@@ -99,6 +102,7 @@ runs:
         GITLAB_SUBGROUP: weave-gitops-sub
         GITLAB_TOKEN: ${{ inputs.gitlab-token }}
         GITLAB_KEY: ${{ inputs.gitlab-key }}
+        GIT_PROVIDER: ${{ matrix.git-provider }}
         ARTIFACTS_BASE_DIR: ${{ inputs.artifacts-base-dir }}
     run: |
       export PATH=${PATH}:`go env GOPATH`/bin
--- a/test/acceptance/test/add_tests.go
+++ b/test/acceptance/test/add_tests.go
@@ -23,7 +23,7 @@ var clusterName string
 
 var clusterContext string
 
-var _ = Describe("Weave GitOps Add App Tests", func() {
+var _ = XDescribe("Weave GitOps Add App Tests", func() {
 
 	deleteWegoRuntime := false
 	if os.Getenv("DELETE_WEGO_RUNTIME_ON_EACH_TEST") == "true" {
@@ -1639,7 +1639,7 @@ var _ = Describe("Weave GitOps Add App Tests", func() {
 	})
 })
 
-var _ = Describe("Weave GitOps Add Tests With Long Cluster Name", func() {
+var _ = XDescribe("Weave GitOps Add Tests With Long Cluster Name", func() {
 	deleteWegoRuntime := false
 	if os.Getenv("DELETE_WEGO_RUNTIME_ON_EACH_TEST") == "true" {
 		deleteWegoRuntime = true
--- a/test/acceptance/test/utils.go
+++ b/test/acceptance/test/utils.go
@@ -845,3 +845,17 @@ func getGitProvider(org string, repo string, providerName gitproviders.GitProvid
 
 	return gitProvider, orgRef, err
 }
+
+func setGitProvider() (gitproviders.GitProviderName, string, string) {
+	gitlab := "gitlab"
+	github := "github"
+	gitProvider := os.Getenv("GIT_PROVIDER")
+
+	if gitProvider == "gitlab" {
+		log.Infof("Using git provider: %s", gitlab)
+		return gitproviders.GitProviderGitLab, gitlabOrg, gitlab
+	}
+
+	log.Infof("Using git provider: %s", github)
+	return gitproviders.GitProviderGitHub, githubOrg, github
+}
--- a/test/acceptance/test/install_tests.go
+++ b/test/acceptance/test/install_tests.go
@@ -24,9 +24,14 @@ import (
 
 var _ = Describe("Weave GitOps Install Tests", func() {
 
+	var gitProvider gitproviders.GitProviderName
+	var gitOrg string
+	var gitURLName string
 	var sessionOutput *gexec.Session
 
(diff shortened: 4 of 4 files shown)


iahmad9 (approved):
LGTM

test/acceptance/test/utils.go
```diff
@@ -845,3 +846,18 @@ func getGitProvider(org string, repo string, providerName gitproviders.GitProvid
 
 	return gitProvider, orgRef, err
 }
+
+func setGitProvider() (gitproviders.GitProviderName, string, string) {
```
josecordaz, line 850: Let's use this logic only once here https://github.com/weaveworks/weave-gitops/blob/main/test/acceptance/test/testsuite_common_test.go#L52 instead of calling it multiple time in each `BeforeEach`.  If you want to keep this function and call it once in `BeforeSuite`  it's fine I would just suggest changing the function name to something more align to its functionality, maybe something like `getGitProviderInfo`,`getGitProviderData` or something similar. Usually methods with `set` in their name are meant to receive parameters and do not return values.

test/acceptance/test/add_tests.go
```diff
@@ -49,19 +55,19 @@ var _ = Describe("Weave GitOps Add App Tests", func() {
 		var exitCode int
 		private := true
 		tip := generateTestInputs()
-		appRepoRemoteURL := "ssh://git@github.com/" + githubOrg + "/" + tip.appRepoName + ".git"
+		appRepoRemoteURL := "ssh://git@" + gitURLName + ".com/" + gitOrg + "/" + tip.appRepoName + ".git"
```
josecordaz, line 59: Please rename this var to `gitProviderName` or something similar just to align more with the value it contains. It would help for readability.

.github/actions/run-acceptance-test/action.yml
```diff
@@ -71,7 +74,7 @@ runs:
     run: |
       chmod +x bin/gitops
       ls -la bin
-  - name: Download dependecies
+  - name: Download dependencies
```
josecordaz, line 77: Nice catch

.github/workflows/test.yml
```diff
@@ -138,6 +138,14 @@ jobs:
     runs-on: ubuntu-latest
     needs: build
     timeout-minutes: 60
+    strategy:
+      matrix:
+        git-provider: [github, gitlab]
+        include:
+        - git-provider: github
+          GIT_PROVIDER: github
```
josecordaz, line 146: It seems `GIT_PROVIDER` is set in `actions/run-accepance-test` using the parameters `git-provider` in the previous line. Having this two variables with the same value makes me wonder  if we need it twice here for a particular reason?

josecordaz (commented):
Quite a lot of changes here, I have to recongince the effort you put on doing all this modifications. Great work.

test/acceptance/test/utils.go
```diff
@@ -845,3 +846,18 @@ func getGitProvider(org string, repo string, providerName gitproviders.GitProvid
...
+	gitlab := "gitlab"
+	github := "github"
+	gitProvider := os.Getenv("GIT_PROVIDER")
```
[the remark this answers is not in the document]
rokshana-b: But the tests won't fail even if the env is not set. The function returns `github` by default even if the env is missing. Si the tests would run on github by default. The only change this function brings is if we want to run all the github tests in gitlab repos. I don't want this variable set as a requirement because I avoided using this env for specific gitlab tests.

.github/workflows/test.yml
rokshana-b: It was the only way I could make it work. I'm open to suggestions ...

test/acceptance/test/utils.go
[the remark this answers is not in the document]
josecordaz: That makes sense. I would argue then that we should have some doc explaining this behavior and also how to run all this cases, like:
- How to run github tests in gitlab
- How to run gitlab tests
- How to run github tests

BTW if you need to have a different value for specific variable in one of the tests you can add it to that specific test only. Like here for instance https://github.com/weaveworks/weave-gitops/blob/main/test/integration/server/remove_test.go#L254. This `AutoMerge` variable is not needed in all cases only on that specific one. If you ever need to use something specific in more than one test, and you want to avoid writting it twice , you could add a  new Context and set that variable's value in a BeforeEach within that context. Like this for instance https://github.com/weaveworks/weave-gitops/blob/main/test/integration/server/remove_test.go#L223. `gitlabGroup ` was written once but it runs for each test within that context.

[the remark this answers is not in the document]
rokshana-b: I can definitely update the ReadME file for clarification

.github/workflows/test.yml
josecordaz: where is GIT_PROVIDER used ?

rokshana-b: in .github/actions/run-acceptance-test/action.yml

josecordaz: do you have the error output when this is not set?

rokshana-b: No, this env only applies to the `smoke-test-long` job which is running tests in parallel. The only way I could tell it failed was when I manually checked if the parallel jobs were running on both gitlab and github repos

.github/actions/run-acceptance-test/action.yml
```diff
@@ -99,13 +102,17 @@ runs:
...
+        GIT_PROVIDER: ${{ inputs.git-provider }}
         ARTIFACTS_BASE_DIR: ${{ inputs.artifacts-base-dir }}
     run: |
       export PATH=${PATH}:`go env GOPATH`/bin
       export WEGO_BIN_PATH=$(pwd)/bin/gitops
       export CLUSTER_PROVIDER=kubectl
       go get github.com/onsi/ginkgo/ginkgo
       go get github.com/onsi/gomega/...
+      if [ "${{ github.job }}" == "smoke-tests-long" ]; then
+        export GIT_PROVIDER=${{ matrix.GIT_PROVIDER }}
```
josecordaz, line 114: is it possible to use `inputs.git-provider` here to avoid having GIT_PROVIDER here https://github.com/weaveworks/weave-gitops/pull/1245/files#diff-faff1af3d8ff408964a57b2e475f69a6b7c7b71c9978cccc8f471798caac2c88R146

josecordaz: If this is set here maybe we don't need this https://github.com/weaveworks/weave-gitops/pull/1245/files#diff-8552c4eceb0d8c09cef35ae019c005254bfe2042358e5acaac36dd9846c739e4R105 ?

rokshana-b: I tried that previously - it didn't work

test/acceptance/test/utils.go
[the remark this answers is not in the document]
rokshana-b: I'll update the ReadME in a follow up PR and try to see if setting git-provider variables at `BeforeSuite` works. For now, I'm merging this PR.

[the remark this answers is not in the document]
rokshana-b: NVM, I updated the readme and set the variables at suite level.

Issue #1111, opened 2021-11-19
Can't add public repo that is not under the ownership of the user


josecordaz:
**Describe the bug**
Trying to add an app coming from a repo that is public fails with error `Error: failed to get git clients: error setting up deploy keys: error uploading deploy key: no permissions to access this repository or repository doesn't exists`

**Environment**
 - gitops: v0.4.1
 - kubernetes: 
    - [x] KinD - v0.11.1
    - [ ] k3s - _version_
    - [ ] cloud [e.g., EKS, AKS]  _version_
    - [ ] other - _name_ _version_
 - Browser + version: [e.g. chrome 74, safari 12, firefox 87]
 -
**Affects versions**
v0.4.1

**To Reproduce**
1. Having wego installed
2. Run `gitops add app --url git@github.com:stefanprodan/podinfo.git --auto-merge --app-config-url ssh://git@github.com/foo/test-config-repo`

**Expected behavior**
It should not try to create a deploy key and succeed.

jpellizzari:
Probably just need to check the repo visibility value from `go-git-providers`.

PR #1249, opened 2021-12-20
Avoid uploading key in public repos

luizbafilho:
Closes: #1111 

**What changed?**
- Avoids uploading key in public repos

**Why?**
- gitops was erroring out when using public repos

**How did you test it?**
 - manual and unit tests

**Release notes**

**Documentation Changes**

### DIFF
--- a/test/acceptance/test/add_tests.go
+++ b/test/acceptance/test/add_tests.go
@@ -227,7 +227,115 @@ var _ = Describe("Weave GitOps Add App Tests", func() {
 		})
 	})
 
-	It("Test1 - Verify that gitops can deploy and delete an app with separate config-repo & app-repo", func() {
+	It("Test2 - Verify that gitops can deploy a public gitlab app", func() {
+		var repoAbsolutePath string
+		private := false
+		tip := generateTestInputs()
+		appName := tip.appRepoName
+		configRepoName := "config-" + appName
+
+		configRepoURL := "ssh://git@gitlab.com/" + gitlabPublicGroup + "/" + configRepoName + ".git"
+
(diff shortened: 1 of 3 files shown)


test/acceptance/test/add_tests.go
```diff
@@ -283,11 +283,14 @@ var _ = Describe("Weave GitOps Add App Tests", func() {
 		private := false
 		tip := generateTestInputs()
 		appName := tip.appRepoName
-		appRepoRemoteURL := "ssh://git@gitlab.com/" + gitlabPublicGroup + "/" + tip.appRepoName + ".git"
+		configRepoName := "config-" + appName
 
-		addCommand := "add app . --auto-merge=true"
+		configRepoURL := "ssh://git@gitlab.com/" + gitlabPublicGroup + "/" + configRepoName + ".git"
```
josecordaz, line 288: I would suggest to use `fmt.Sprintf` to build this string for readability

josecordaz (approved):
LGTM, just a small nit

test/acceptance/test/add_tests.go
luizbafilho: I agree, but I decided t keep it this way because that's how URLs are being built across the whole acceptance suite, I'd rather keep consistency over something I prefer.

palemtnrider (commented):
LGTM - however, I didn't see any acceptance tests using a public app repo - did I just miss it?

luizbafilho:
> LGTM - however, I didn't see any acceptance tests using a public app repo - did I just miss it?

yep, they are there but didn't break because they use a private repo as config url

Issue #1164, opened 2021-12-03
Acceptance tests do not specify the precise line of the error


josecordaz:
**Describe the bug**
When acceptance tests fail in CI, ginkgo doesn't provide the exact line of the test failure.

Failure in CI:
![image.png](https://images.zenhubusercontent.com/606389311ead0cc4788a9a72/c2a2d421-3ea2-4388-bbf4-bb63fea83665)

**To Reproduce**
1. This is a link in CI where this is happening already https://github.com/weaveworks/weave-gitops/runs/4402760349?check_suite_focus=true#step:3:8886

**Expected behavior**
It should fail like this:
![image.png](https://images.zenhubusercontent.com/606389311ead0cc4788a9a72/3e02ee36-1b90-49eb-ace9-4091bd780dd9)
https://github.com/weaveworks/weave-gitops/runs/4402991107?check_suite_focus=true#step:3:9011

aclevername:
I've noticed this too. I think something must be going wrong with how the suite it setup for the lines to be incorrect

PR #1252, opened 2021-12-20
Removed duplicate fail register

josecordaz:
Closes:  
- #1164 

**What changed?**
- Removed duplicate ginkgo failure registration

**Why?**
- When tests fail it doesn't show the exact line where the error occurred.

**How did you test it?**
- CI

**Release notes**

**Documentation Changes**

### DIFF
--- a/test/acceptance/test/testsuite_common_test.go
+++ b/test/acceptance/test/testsuite_common_test.go
@@ -8,9 +8,7 @@ import (
 	"os"
 	"testing"
 
-	"github.com/onsi/ginkgo"
 	. "github.com/onsi/ginkgo"
-	"github.com/onsi/gomega"
 	. "github.com/onsi/gomega"
 	log "github.com/sirupsen/logrus"
 )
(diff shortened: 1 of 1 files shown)


josecordaz:
@rokshana-b @iahmad9 Any reason why we have this duplicate register (`RegisterFailHandler `)  ?

palemtnrider:
Do you have some example output before/after this change? The change LGTM but seeing the output would help.

Also, you may have been able to provide a skip in the call to ginko.Fail()

josecordaz:
> Do you have some example output before/after this change? The change LGTM but seeing the output would help.

You right, an example of the output would be really helpful. The actual difference is in the description of the ticket. https://github.com/weaveworks/weave-gitops/issues/1164
> 
> Also, you may have been able to provide a skip in the call to ginko.Fail()

You mean, like a parameter?

rokshana-b (approved):

palemtnrider:
> You mean, like a parameter?

Yes.  There is a callerSkip argument that should change who is reporting the error.

palemtnrider (approved):
LGTM

Issue #1120, opened 2021-11-22
`tag-release` script gives an incorrect release candidate increment


jpellizzari:
**Describe the bug**
When doing a release, the release tagging script does not work as expected

**What is the severity of the bug**
Please add a label for this ticket indicating the severity of this bug.  

**Affects versions**

**To Reproduce**
1. `$ ./tools/tag-release.sh -m -c`

**Expected behavior**
A tag that looks like this gets created: `v0.5.0-rc1`

**Actual behavior**
` * [new tag]           v0.5.0-rc0.4.1 -> v0.5.0-rc0.4.1`

palemtnrider:
We need to fix this as we ended up creating lightweight tags manually instead of annotated tags because we didn't use this script.  Lightweight tags cause commands like `git describe --always --match "v*"` which is used for setting the version in the code, to report an older version (the last annotated tag)

palemtnrider:
@JamWils

PR #1266, opened 2022-01-04
Fix release script to avoid invalid tags

J-Thompson12:
Closes: #1120 

**What changed?**
- Removed forcing an rc candidate to be created first
- Fixed bug in above issue
- Added check for not changing version ie doing a two releases in a row without incrementing the number

**Why?**

**How did you test it?**
Manually

**Release notes**

**Documentation Changes**


jpellizzari (approved):
Code LGTM; did not pull it down to test.

joshri (approved):
🌞 LGTM🌞

Issue #1191, opened 2021-12-07
500 Errors Don't Stop Polling in Remove App Page


joshri:
**Describe the bug**
If something crazy happens while removing an app (like the app doesn't exist), a 500 error is thrown on the remove request, but the polling to check if the application has been removed continues. The Alert that is shown is also the success variant, instead of error

**What is the severity of the bug**
| severity/Minor | Weave Gitops functionality is broken, but there is a workaround |

**Expected behavior**
500 errors stop polling and a proper alert is shown

**Screenshots**
![image](https://user-images.githubusercontent.com/65822698/145098929-1c8d7823-9118-4870-8aef-0953601bc18b.png)

Issue #1190, opened 2021-12-07
Polling Bug in Remove App Page


joshri:
**Describe the bug**
Polling immediately begins on page load - can it only begin when there is a result for the remove request?

| severity/Low | Doesn’t affect primary flow/functionality but would be good to fix |

**To Reproduce**
Navigate to the Remove App page by pressing the Remove App button in Application Detail. Then check your Network devtools tab

**Expected behavior**
UI checks if the app is removed only after the backend communicates to us that it has been removed.

**Screenshots**
![image](https://user-images.githubusercontent.com/65822698/145099174-effb0a79-a691-4ae3-b6a9-e6bde1be804f.png)

PR #1267, opened 2022-01-04
Fixes Polling Errors on Remove Page

joshri:
Closes: #1190
Closes: #1191 

Fixes display of 500 errors in `RemoveApplication` and only initiates polling on remove request instead of on page load.

### DIFF
--- a/ui/pages/ApplicationRemove.tsx
+++ b/ui/pages/ApplicationRemove.tsx
@@ -154,16 +154,19 @@ function ApplicationRemove({ className, name }: Props) {
           state: { authSuccess: false },
         }}
       >
-        {error &&
-          (error.code === GrpcErrorCodes.Unauthenticated ? (
+        {!authSuccess &&
+          error &&
+          error.code === GrpcErrorCodes.Unauthenticated && (
             <AuthAlert
               title="Error"
               provider={repoInfo.provider}
               onClick={() => setAuthOpen(true)}
             />
-          ) : (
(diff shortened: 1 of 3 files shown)


ui/pages/ApplicationRemove.tsx
```diff
@@ -109,17 +109,15 @@ function ApplicationRemove({ className, name }: Props) {
...
     const poll = poller(() => {
       applicationsClient
         .GetApplication({ name, namespace: "wego-system" })
-        .catch(() => {
-          // Once we get a 404, the app is gone for good
+        .catch((e) => {
           clearInterval(poll);
-          setRemovedFromCluster(true);
+          // Once we get a 404, the app is gone for good
+          if (e.code === GrpcErrorCodes.NotFound)
```
jpellizzari, line 119: Are you sure the backend will return this status code when it doesn't find the app? Might need to change it here:

https://github.com/weaveworks/weave-gitops/blob/1517a32c7806d3d62fc54e29c28d7e8e6cdb7a8a/pkg/server/server.go#L163

Something like:

```golang
		if apierrors.IsNotFound(err) {
                    return nil, grpcStatus.Errorf(codes.NotFound, "not found: %s", err.Error())
		}

```

joshri: Was able to successfully remove app from cluster using this code - the app recognizes this error code and then shows the removed from cluster notification!

jpellizzari (approved):
LGTM :+1:

Issue #1092, opened 2021-11-18
As a User I want to authenticate using my organisation's OIDC provider so that I can take advantage of SSO and get a streamlined experience


yiannistri:
- [x] Document how to configure Weave GitOps to integrate with an existing OIDC setup
- [x] Add feature flag to enable/disable authentication flow
- [x] Add middleware handler to redirect unauthenticated users to the OIDC provider
- [x] Add callback handler that will be called by the OIDC provider after user authentication attempt
- [x] Add middleware that will impersonate authenticated user when making calls to the Kubernetes API
  - [x] GET /v1/applications
  - [x] GET /v1/applications/{name}
  - [x] POST /v1/applications
  - [x] DELETE /v1/applications/{name}
  - [x] POST /v1/applications/{automationName}/reconciled_objects
  - [x] POST /v1/applications/child_objects
  - [x] POST /v1/applications/{name}/sync
  - [X] GET /v1/applications/{name}/commits
  - [x] GET /v1/applications/parse_repo_url
  - [x] POST /v1/applications/validate_token/pp
  - [x] GET /v1/profiles
  - [x] GET /v1/profiles/{profile_name}/{profile_version}/values

Cookie handling
- [x] Secure attribute set to ensure cookies are included in HTTPS sessions (not set if redirect URL does not use HTTPS)
- [x] HttpOnly attribute set to ensure cookie cannot be read by JS
- [ ] Consider setting SameSite attribute to mitigate CSRF attacks
- [x] Cookie expires after 1h by default although this is configurable

Feature flags
- [x] Ensure the OIDC CLI flags match Kubernetes api-server OIDC CLI flags

Scopes
- [x] `openid` scope supplied by default
- [x] `offline_access` scope is necessary for issuing refresh tokens and may be configurable
- [x] `email` scope is necessary to get back user's email address

PR #1268, opened 2022-01-05
Add WithAuth middleware

yiannistri:
Part of: #1092  

**What changed?**
- Added feature flag to enable authentication flow
- Added middleware that inspects cookie or auth header for an ID token
  - Unauthenticated API requests will 401
  - Unauthenticated web/html requests will redirect to OIDC issuer

**Why?**

**How did you test it?**
Unit tests and manually.

**Release notes**

**Documentation Changes**

Note that more changes (follow-up PRs) will be needed to actually use the identity of the authenticated user in subsequent calls to the Kubernetes API.


cmd/gitops/ui/run/cmd.go
```diff
@@ -80,9 +109,47 @@ func runCmd(cmd *cobra.Command, args []string) error {
...
+		if redirectURL.Scheme == "https" {
+			oidcIssueSecureCookies = true
+		}
+
+		oidcCookieDuration, err := time.ParseDuration(options.OIDC.CookieDuration)
+		if err != nil {
+			return fmt.Errorf("invalid cookie duration: %w", err)
+		}
+
+		cfg, err := auth.NewAuthConfig(cmd.Context(), options.OIDC.IssuerURL,
```
jpellizzari, line 137: Lots of arguments here; might be time to turn them into a `struct`?

pkg/server/auth/jwt.go
```diff
@@ -0,0 +1,123 @@
...
+
+// JWTCookiePrincipalGetter inspects a cookie for a JWT token
+// and returns a principal object.
+type JWTCookiePrincipalGetter struct {
+	log        logr.Logger
+	verifier   *oidc.IDTokenVerifier
+	cookieName string
+}
+
+func NewJWTCookiePrincipalGetter(log logr.Logger, verifier *oidc.IDTokenVerifier, cookieName string) PrincipalGetter {
```
jpellizzari, line 29: I am assuming our existing `jwt` machinery wasn't  quite right for this, hence the new implementation?

pkg/server/handler.go
```diff
@@ -4,23 +4,38 @@ import (
...
 	ProfilesConfig ProfilesConfig
+	AuthConfig     *auth.AuthConfig
 }
 
 func NewHandlers(ctx context.Context, cfg *Config) (http.Handler, error) {
 	mux := runtime.NewServeMux(middleware.WithGrpcErrorLogging(cfg.AppConfig.Logger))
 	httpHandler := middleware.WithLogging(cfg.AppConfig.Logger, mux)
 	httpHandler = middleware.WithProviderToken(cfg.AppConfig.JwtClient, httpHandler, cfg.AppConfig.Logger)
 
+	if AuthEnabled() {
```
jpellizzari, line 35: :+1:

jpellizzari (approved):
Looks very elegant and well thought out :+1: .

I don't fully grasp all of the OIDC stuff, but this is implemented in a low-risk way. Nice work.

cmd/gitops/ui/run/cmd.go
bigkevmcd: Yes, this is https://martinfowler.com/bliki/DataClump.html

cmd/gitops/ui/run/cmd.go
```diff
@@ -19,29 +20,57 @@ import (
...
+	cmd.Flags().StringVar(&options.Path, "path", "", "Path url")
+	cmd.Flags().StringVar(&options.HelmRepoNamespace, "helm-repo-namespace", "default", "the namespace of the Helm Repository resource to scan for profiles")
+	cmd.Flags().StringVar(&options.HelmRepoName, "helm-repo-name", "weaveworks-charts", "the name of the Helm Repository resource to scan for profiles")
+
+	if server.AuthEnabled() {
+		cmd.Flags().StringVar(&options.OIDC.IssuerURL, "oidc-issuer-url", "", "The URL of the OpenID Connect issuer")
+		cmd.Flags().StringVar(&options.OIDC.ClientID, "oidc-client-id", "", "The client ID for the OpenID Connect client")
+		cmd.Flags().StringVar(&options.OIDC.ClientSecret, "oidc-client-secret", "", "The client secret to use with OpenID Connect issuer")
+		cmd.Flags().StringVar(&options.OIDC.RedirectURL, "oidc-redirect-url", "", "The OAuth2 redirect URL")
+		cmd.Flags().StringVar(&options.OIDC.CookieDuration, "oidc-cookie-duration", "1h", "The duration of the ID token cookie. It should be set in the format: number + time unit (s,m,h) e.g., 20m")
```
bigkevmcd, line 70: How about using https://pkg.go.dev/github.com/spf13/pflag#DurationVar

cmd/gitops/ui/run/cmd.go
```diff
@@ -80,9 +109,47 @@ func runCmd(cmd *cobra.Command, args []string) error {
...
+		if err != nil {
+			return fmt.Errorf("invalid redirect URL: %w", err)
+		}
+
+		var oidcIssueSecureCookies bool
+		if redirectURL.Scheme == "https" {
+			oidcIssueSecureCookies = true
+		}
+
+		oidcCookieDuration, err := time.ParseDuration(options.OIDC.CookieDuration)
```
bigkevmcd, line 132: You could simplify this by using https://pkg.go.dev/github.com/spf13/pflag#DurationVar above

cmd/gitops/ui/run/cmd.go
```diff
@@ -96,20 +163,24 @@ func runCmd(cmd *cobra.Command, args []string) error {
...
+			} else {
+				redirector(w, req)
+			}
 			return
 		}
 		assetHandler.ServeHTTP(w, req)
 	}))
 
-	addr := "0.0.0.0:" + port
+	addr := "0.0.0.0:" + options.Port
```
bigkevmcd, line 176: https://pkg.go.dev/net#JoinHostPort

pkg/server/auth/auth.go
```diff
@@ -0,0 +1,130 @@
...
+	// IDTokenCookieName is the name of the cookie that holds the ID Token once
+	// the user has authenticated successfully with the OIDC Provider.
+	IDTokenCookieName = "id_token"
+	// RefreshTokenCookieName is the name of the cookie that holds the refresh
+	// token.
+	RefreshTokenCookieName = "refresh_token"
+	// ScopeProfile is the "profile" scope
+	ScopeProfile = "profile"
+	// ScopeEmail is the "email" scope
+	ScopeEmail = "email"
```
bigkevmcd, line 24: Do all these need to be exported?

pkg/server/auth/auth.go
```diff
@@ -0,0 +1,130 @@
...
+// Unauthorized requests will be denied with a 401 status code.
+func WithAPIAuth(next http.Handler, cfg *AuthConfig) http.Handler {
+	cookieAuth := NewJWTCookiePrincipalGetter(cfg.logger,
+		cfg.verifier(), IDTokenCookieName)
+	headerAuth := NewJWTAuthorizationHeaderPrincipalGetter(cfg.logger, cfg.verifier())
+
+	return http.HandlerFunc(func(rw http.ResponseWriter, r *http.Request) {
+		principal, err := MultiAuthPrincipal{cookieAuth, headerAuth}.Principal(r)
+		if err != nil || principal == nil {
+			cfg.logger.Error(err, "failed to get principal")
```
bigkevmcd, line 58: err might be nil, because principal could be nil?

Also, you could create the `MultiAuthPrincipal` in the closure, rather than on each request?

pkg/server/auth/jwt.go
```diff
@@ -0,0 +1,123 @@
...
+		return nil, fmt.Errorf("failed to parse claims from the JWT token: %w", err)
+	}
+
+	return &UserPrincipal{ID: claims.Email, Groups: claims.Groups}, nil
+}
+
+// MultiAuth combines the JWTCookie and JWTAuthorizationHeader
+// principal getters together to look for a principal first in
+// a cookie and then in an Authorization header.
+type MultiAuthPrincipal []PrincipalGetter
```
bigkevmcd, line 108: This comment doesn't seem quite right?

It doesn't seem to care about cookies or authorization ?

It loops over the provided authorizers and if it finds an error or a principal, it returns, otherwise it returns `nil, nil`.

Also, no tests for this?

pkg/server/auth/auth.go
```diff
@@ -0,0 +1,130 @@
...
+
+			rw.Header().Set("WWW-Authenticate", `Bearer realm="Weave GitOps"`)
```
bigkevmcd, line 60: We don't accept Basic Auth?

bigkevmcd (changes requested):
Nothing major in here.

cmd/gitops/ui/run/cmd.go
yiannistri: Good idea, I'll change this.

pkg/server/auth/jwt.go
yiannistri: Indeed, there is some overlap but they mostly serve different needs.
As I understand it, the existing `jwt` implementation is written to wrap/extend tokens that are issued from git providers and are used to interact with their APIs. 
This PR adds support for parsing ID tokens from OIDC issuers (which may or may not be git providers) in order to extract information about the logged-in user (email/groups) and make calls to the Kubernetes API (but not the git provider API).

bigkevmcd (approved):
Thanks for the work to make this better!

Issue #1269, opened 2022-01-05
Return profile layer in ListProfiles response


foot:
As a new field in the response

details TBD

PR #1277, opened 2022-01-06
Adds a layer field to the profiles response

foot:
- Layers allow you to specify the order in which profiles are installed

Closes: #1269 

**What changed?**

Layer field added to the profiles response

**Why?**

We want to show the layer in the UI

**How did you test it?**

manually and w/ unit tests


pkg/helm/charts.go
```diff
@@ -26,6 +27,7 @@ import (
 // ProfileAnnotation is the annotation that Helm charts must have to indicate
 // that they provide a Profile.
 const ProfileAnnotation = "weave.works/profile"
+const LayerAnnotation = "weave.works/layer"
```
yiannistri, line 30: Shall we add a comment here describing how this is used?

yiannistri (approved):

Issue #1236, opened 2021-12-15
Refresh application details graph when sync button is pressed.


J-Thompson12:
Currently if a user presses sync the graph in the application details page will not update. 
Steps:
- Load application details page
- Update number of replicas for application
- Click sync button
- Graph does not change
- Reload page
- Graph is now correct

The graph needs to update when a user clicks the sync button.

sympatheticmoose:
Perhaps regular polling similar to the applications summary page?

sympatheticmoose:
Related comment from an internal exercise:
- Had to page reload the UI to see the updated commits (the list was cached showing older commit as the last synced)
  - Pressing the sync button doesn’t seems to update the source/kustomization commits list

PR #1282, opened 2022-01-07
Refresh AppDetail on Sync

joshri:
Closes: #1236 

The result of an application sync is now tied to refreshing the App details, remaking graph, commits table, etc.

### DIFF
--- a/ui/pages/ApplicationDetail.tsx
+++ b/ui/pages/ApplicationDetail.tsx
@@ -68,7 +68,7 @@ function ApplicationDetail({ className, name }: Props) {
     };
 
     req(p());
-  }, [name]);
+  }, [name, syncRes]);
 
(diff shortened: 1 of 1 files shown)


jpellizzari (changes requested):
We need to redo the hook below this as well so we re-get children/reconciled objects

joshri:
> We need to redo the hook below this as well so we re-get children/reconciled objects

Won't `res` change triggering that useEffect as well?

jpellizzari:
> > We need to redo the hook below this as well so we re-get children/reconciled objects
> 
> Won't `res` change triggering that useEffect as well?

Ah yes that is correct, and also very confusing, we probably need a `useReducer` here.

jpellizzari (approved):

PR #1286, opened 2022-01-11
Fix nightly tests and refactor add tests

rokshana-b:
**What changed?**
- running nightly tests 6 days a week to leave Sunday free for the new upcoming weekly tests
- refactor add tests to prevent nightly failures

**How did you test it?**
- ran nightly on ci, it passed


.github/workflows/nightly.yaml
```diff
@@ -129,18 +129,29 @@ jobs:
         fi
     - name: Install Ginkgo binary
       run: go get github.com/onsi/ginkgo/ginkgo
-    - name: User Acceptance Tests (${{matrix.os}})
+    - name: User Acceptance Tests for GitHub (${{matrix.os}})
       run: |
         export WEGO_BIN_PATH=$(pwd)/bin/gitops-${{matrix.os}}-nightly
         export CLUSTER_PROVIDER=kubectl
         export DELETE_WEGO_RUNTIME_ON_EACH_TEST=true
-        export K8S_VERSION=1.19.1
```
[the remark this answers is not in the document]
iahmad9: It's fine tests are not using this env variable anymore

test/acceptance/test/add_tests.go
```diff
@@ -1211,7 +1221,9 @@ var _ = Describe("Weave GitOps Add App Tests", func() {
 		})
 
 		By("And repo created has private visibility", func() {
-			Eventually(getGitRepoVisibility(githubOrg, appRepoName, gitproviders.GitProviderGitHub)).Should(ContainSubstring("private"))
+			if os.Getenv("GIT_PROVIDER") != "gitlab" {
```
josecordaz, line 1224: I wonder if we need else cases in all the places we have this validation. I ask because there is an else in line 1728.

test/acceptance/test/add_tests.go
```diff
@@ -658,12 +637,26 @@ var _ = Describe("Weave GitOps Add App Tests", func() {
...
+			selectCluster(cluster2Context)
+			uninstallWegoRuntime(WEGO_DEFAULT_NAMESPACE)
+		})
+
+		By("And gitops check pre kubernetes version is compatible and flux is not installed in Cluster2", func() {
+			selectCluster(cluster2Context)
+			c := exec.Command(gitopsBinaryPath, "check", "--pre")
+			output, err := c.CombinedOutput()
+			Expect(err).NotTo(HaveOccurred())
+			expectedOutput := fmt.Sprintf(`✔ Kubernetes %s.* >=[0-9]+.[0-9]+\.[0-9]+-[0-9]+
```
josecordaz, line 655: I wonder if there is any particular reason for `.*` after `%s`. I ask because there is no `.*` in the test `And gitops check pre validates kubernetes and flux compatibility for Cluster1` where check pre is called as well.

test/acceptance/test/add_tests.go
rokshana-b: We need this validation when we run tests on gitlab repos; this validation fails for gitlab

josecordaz:
@rokshana-b  I see you removed env var `K8S_VERSION` from some config files. Do we need to rewrite it? or is it present already?
```
func getK8sVersion() string {
	k8sVersion, found := os.LookupEnv("K8S_VERSION")
	if found {
		return k8sVersion
	}

	return DEFAULT_K8S_VERSION
}
```

test/acceptance/test/add_tests.go
rokshana-b: good catch on that - i need to get that removed

rokshana-b:
> @rokshana-b I see you removed env var `K8S_VERSION` from some config files. Do we need to rewrite it? or is it present already?
> 
> ```
> func getK8sVersion() string {
> 	k8sVersion, found := os.LookupEnv("K8S_VERSION")
> 	if found {
> 		return k8sVersion
> 	}
> 
> 	return DEFAULT_K8S_VERSION
> }
> ```

@josecordaz I thought I left this function in the utils file. i just moved this validation to a test where we have a skip tag for nightly. Can you pinpoint where it needs to be re-added?

test/acceptance/test/add_tests.go
rokshana-b: ### Context: 
This PR _was_ originally opened to enable test runs in gitlab repos. However, the tests couldn't be run on parallel since we have standing clusters and that would create conflict(and failures). So I tried running the tests one after another. First the acceptance tests would run on github, then on gitlab. This worked but it took ridiculously long to run: https://github.com/weaveworks/weave-gitops/actions/runs/1721328971 (almost 4 hrs!!)
Nightly tests are famously hard to fix because of the long runtime and adding more time is not ideal.

### Fix:
So, for now, I've refactored stuff to get the nightly passing and I've configured the nightly to run Mon-Sat. I have a plan to write up another job that will run every week on Sun that will cover the entire github/gitlab workflow. This seemed like a reasonable bargain — run time-consuming tests weekly instead of daily.

@josecordaz Hope this answers your question

josecordaz (approved):

test/acceptance/test/add_tests.go
josecordaz: @rokshana-b  Sorry probably I'm not making myself understood. It's just that it seems weird to me that we have a `By` clause saying `And repo created has private visibility`, and when I see this `if os.Getenv("GIT_PROVIDER") != "gitlab" {` it tells that we are not going to check the visibility of the repo for gitlab. But what is still not clear to me is "why". You mentioned `we only need this to work when GIT_PROVIDER is NOT gitlab` but this doesn't explain why.  What is even stranger for me is that we are using go-git-providers in function `getGitRepoVisibility` which in theory should handle both types of providers. At this point I wonder if the "why" is a bug in the go-git-providers repo. If this fails for gitlab what would be the error?

rokshana-b: @josecordaz I wanted to leave a msg for you in this PR so its documented. There's no real bug here. The problem is for github, the repo visibility is either `private` or `public` — simple as that. For gitlab, however, the repo visibility depends on the `group`. So, everytime we create a private repo for github, it should work for gitlab. But if we create a public repo for github, the same will not be true for gitlab. If the gitlab-repo-group is private, the repo visibility will remain private. 

The failures I saw were related to github being a `public` repo and gitlab remaining as `private`. It was probably overkill for me to disable repo visibility checks for _all_ gitlab tests. I will re-add the checks wherever the github repos are `private` in another PR.

PR #1288, opened 2022-01-11
Watch and cache helmcharts

Skarlso:
Closes https://github.com/weaveworks/weave-gitops/issues/1144.

Watching HelmRepository resources and caching the profile data from them including the values.

## Question:

Right now, the watcher is started when `gitops ui run` is executed or when the server starts up. This could be a completely separate project running as a completely separate pod as well. I believe it's fine for now like this, but still. Let's discuss what others think.

Also, I'm not a 100% sure if that's the right place to start this service from although it's as good as any. :)

## What is the current architecture?

Cache replaced whatever the profiles endpoint was calling. Which is things in under `helm` the `charts.*` methods. Those methods have been all shifted to the Watcher. The watcher is notified on creation and update events regarding ALL HelmRepository objects. Through the Predicate, it's determined if it's a create or an update event. In case of an update if the revision doesn't match, we reconcile.

In case of create we only run the the reconcile if the helmrepository has an artifact. In which case, we do the same thing that has been done before. We call GetCharts with the profiles predicate and the values.

Once that's done, this information is cached. The cache is always recent. It might be that there is a brief period where some information for a profile can't be found if there is no cache for it yet, but in that case the user just refreshes. Chances are that the cache is filled up earlier than the user requesting information about said object.

The following folder structure is used to cache items:

```
➜  helm-cache tree
.
├── cache.lock
└── wego-system
    └── podinfo
        ├── podinfo
        │   └── 6.0.1
        │       └── values.yaml
        └── profiles.yaml

4 directories, 3 files
```

The cache.lock is global and not helmnamespace/helmname because there is a scenario where if we are checking that the folder exists before creating, another process creates it, it returns false, and tries to create it and overwrites whatever has been put there. To ensure that it's recent, we always lock. We could use optimistic locking and versioning, but I doubt that that is needed here.

The top folder is helm-namespace and under it is helm-name. Then followed by a `profiles.yaml` file which is the yaml marshalled profile information:

```yaml
- available_versions:
  - 6.0.0
  - 6.0.1
  description: Podinfo Helm chart for Kubernetes
  home: https://github.com/stefanprodan/podinfo
  kube_version: '>=1.19.0-0'
  maintainers:
  - email: stefanprodan@users.noreply.github.com
    name: stefanprodan
  name: podinfo
  sources:
  - https://github.com/stefanprodan/podinfo
```

This information is usually small data, so it can sit in a single file. The values is the bigger fish. We save values according to the name of the profile ( which should be unique per helmRepository ( I HOPE ) ) and under that we have folders of versions. And in those versions we'll have a `values.yaml` file if there are values for that version.

The values file contains whatever was returned before. Which is the full raw value including comments. We already parse that correctly.

## What does the cache return?

The cache returns the exact data as requested by `GetProfiles` and `GetProfileValues` so there is very little code change in those files. I literally just exchanged RepoManger with the cache.


pkg/server/profiles.go
```diff
@@ -146,15 +140,40 @@ func (s *ProfilesServer) GetProfileValues(ctx context.Context, msg *pb.GetProfil
 		return nil, fmt.Errorf("failed to get HelmRepository %q/%q", s.HelmRepoNamespace, s.HelmRepoName)
 	}
 
-	sourceRef := helmv2beta1.CrossNamespaceObjectReference{
```
Skarlso, line 149: This was actually never used. And has no function in any case anyways because only the profile name and version matter.

pkg/server/profiles.go
```diff
@@ -146,15 +140,40 @@ func (s *ProfilesServer) GetProfileValues(ctx context.Context, msg *pb.GetProfil
...
-		APIVersion: helmRepo.TypeMeta.APIVersion,
-		Kind:       helmRepo.TypeMeta.Kind,
-		Name:       helmRepo.ObjectMeta.Name,
-		Namespace:  helmRepo.ObjectMeta.Namespace,
+	data, err := s.HelmCache.Get(s.HelmCache.Key(helmRepo.Namespace, helmRepo.Name))
+	if data == nil || err != nil {
```
Skarlso, line 144: TODO better handling once I'm done

cmd/gitops-server/main.go
```diff
@@ -52,7 +55,34 @@ func NewAPIServerCommand() *cobra.Command {
...
+
+			go func() {
+				if err := helmWatcher.StartWatcher(); err != nil {
+					appConfig.Logger.Error(err, "failed to start the watcher")
+					os.Exit(1)
+				}
+			}()
+
+			// Create the cache here as well and pass it in through the profiles Config thing.
+			runtimeNamespace := os.Getenv("RUNTIME_NAMESPACE")
```
Skarlso, line 81: This is less than ideal. But I'm not sure how to pass in anything here.

cmd/gitops-server/main.go
```diff
@@ -52,7 +55,34 @@ func NewAPIServerCommand() *cobra.Command {
...
+			if runtimeNamespace == "" {
+				runtimeNamespace = "default"
+			}
+			profilesConfig := server.NewProfilesConfig(rawClient, helmCache, runtimeNamespace, "weaveworks-charts")
```
Skarlso, line 85: This should probably also be configurable from somewhere?

cmd/gitops/ui/run/cmd.go
```diff
@@ -62,6 +66,7 @@ func NewCommand() *cobra.Command {
 	cmd.Flags().StringVar(&options.Path, "path", "", "Path url")
 	cmd.Flags().StringVar(&options.HelmRepoNamespace, "helm-repo-namespace", "default", "the namespace of the Helm Repository resource to scan for profiles")
 	cmd.Flags().StringVar(&options.HelmRepoName, "helm-repo-name", "weaveworks-charts", "the name of the Helm Repository resource to scan for profiles")
+	cmd.Flags().StringVar(&options.HelmCacheDir, "helm-cache-dir", "/tmp/helm-cache", "the folder where to cache Profile data")
```
Skarlso, line 69: Maybe it would be better renamed to `profile-cache`?

pkg/helm/watcher/server.go
```diff
@@ -0,0 +1,94 @@
...
+func (w *Watcher) StartWatcher() error {
+	utilruntime.Must(clientgoscheme.AddToScheme(scheme))
+	utilruntime.Must(sourcev1.AddToScheme(scheme))
+	ctrl.SetLogger(zap.New(zap.UseFlagOptions(&zap.Options{
+		Development: true,
+	})))
+
+	mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
+		Scheme:                 scheme,
+		MetricsBindAddress:     ":9980",
```
Skarlso, line 54: These should probably be configured from a CLI. Or maybe we don't care for now. I'm not sure. :) Thoughts?

jpellizzari (commented):
This looks controller-y and I would think this overlaps quite a bit with `helm-controller`? I may not have the full context, however.

How does this differ from `helm-controller`?

Skarlso:
@jpellizzari It differs in two ways:

1. It uses the cache which is then used by profiles
2. It only uses profile related helmrepositories

This is a stepping stone to the next thing which is to detect if there is a new version available for a profile and report that back to the UI.

pkg/helm/watcher/server.go
stefanprodan: You may want to consider using `fluxcd/pkg/runtime` as it contains helpers for metrics, events, logging, watching sources, sending alerts to notification-service and many more.

Skarlso: Hey @stefanprodan! I actually looked at this package! :) And, please forgive me, I'm probably blind, but which part of this package is supposed to helm with watching sources? Do you mean the predicate for example? I'm not sure how I would have used that. I know that the metrics part is probably interesting for me, but to be honest, this is such a small thing, I'm thinking about completely leaving out that part. WDYT?

stefanprodan: Here is an example on how to setup structured logging, API rate limits, metrics, events, leader election, health checks: https://github.com/fluxcd/kustomize-controller/blob/main/main.go#L83-L107

Skarlso: Thanks!

cmd/gitops-server/main.go
bigkevmcd: How about looking at how this works https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/master/main.go#L86-L90 ?

stefanprodan: The namespace a controller watches is not the same thing as the runtime namespace (the ns where it's been deployed). The runtime namespace is mapped from the Kubernetes Downward API.

Skarlso: That is perfectly true. I added this as a workaround. This definitely is not right and I'm not sure what to do here about it. We had some hardcoded things in here which said, `default` and `weaveworks-charts` or something like that. But that's definitely not true and not even desired for this to work. So I'm trying to find a better way to do this.

stefanprodan: In Flux source-controller, we use the runtime namespace value, then we advertise it for other controllers in the status sub-resource. This way all the other controllers can determine at runtime where the source cache is located.

Skarlso: For example, the thing I'm testing has a helmchart for profile in wego-system with name `podinfo`. And even so I can only use that when I configure the namespace from the command line when I'm running `gitops ui run --helm-repo-namespace wego-system --helm-repo-name podinfo`. So right now, my hands I tied in this. This is how this is designed for the moment, and later we can improve upon that. For now, I guess only a single HelmRepository will be a profile.

pkg/helm/watcher/controller/helm_watcher.go
```diff
@@ -0,0 +1,94 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/weaveworks/weave-gitops/pkg/helm"
+	"github.com/weaveworks/weave-gitops/pkg/helm/watcher/cache"
+)
+
+// HelmWatcherReconciler runs the `reconcile` loop for the watcher.
+type HelmWatcherReconciler struct {
+	client.Client
+	Scheme      *runtime.Scheme
```
bigkevmcd, line 21: I don't think you're using the Scheme?

cmd/gitops/ui/run/cmd.go
Skarlso: Sounds good 👍

pkg/helm/watcher/cache/cache.go
```diff
@@ -0,0 +1,182 @@
...
+				return fmt.Errorf("failed to write out values for version %s: %w", version, err)
+			}
+		}
+	}
+
+	return nil
+}
+
+// GetProfiles gathers all profiles for a helmRepo if found. Returns an error otherwise.
+func (c *HelmCache) GetProfiles(helmRepoNamespace, helmRepoName string) ([]*pb.Profile, error) {
```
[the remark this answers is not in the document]
Skarlso: It doesn't matter where this is called from I believe, right? Why would the cache care about that? It literally just returns data from a disk. Granted, the lock can time out, but that's about it.

pkg/helm/watcher/cache/cache_test.go
```diff
@@ -0,0 +1,231 @@
...
+		"0.0.5": []byte("values-5"),
+	}
+	helmNamespace = "test-namespace"
+	helmName      = "test-name"
+)
+
+func TestCacheGetProfiles(t *testing.T) {
+	t.Log("setup the cache and data for the cache")
+
+	dir, err := os.MkdirTemp("", "cache-get-profiles")
```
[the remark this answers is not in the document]
Skarlso: Oh, I like `t.Cleanup`. I wanted to extract this, but didn't know about t.Cleanup 🤦

pkg/helm/watcher/controller/helm_watcher.go
Skarlso: Good spot. I don't think I am.

pkg/helm/watcher/cache/cache.go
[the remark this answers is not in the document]
bigkevmcd: You are creating a context in the code, if you want to continue doing that, how about providing configuration for how long to wait, i.e. a wait duration as a parameter?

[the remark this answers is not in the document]
Skarlso: Fair point

pkg/helm/watcher/server.go
```diff
@@ -0,0 +1,94 @@
...
+
+	return &Watcher{
+		cache:       cache,
+		repoManager: helm.NewRepoManager(kubeClient, tempDir),
+	}, nil
+}
+
+func (w *Watcher) StartWatcher() error {
+	utilruntime.Must(clientgoscheme.AddToScheme(scheme))
+	utilruntime.Must(sourcev1.AddToScheme(scheme))
```
aclevername, line 47: worth moving to the `watcher` constructer, to prevent the code panicing after the go routine has started? seems a bit cleaner to fail earlier

bigkevmcd: (I'd move it to the `main.go:main()` function which is what most other things do.)

pkg/helm/watcher/controller/helm_watcher.go
```diff
@@ -0,0 +1,94 @@
...
+				log.Error(err, "failed to get values for chart and version, skipping...", "chart", chart.Name, "version", v)
+				// log and skip version
+				continue
+			}
+
+			values[chart.Name] = map[string][]byte{
+				v: valueBytes,
+			}
+		}
+	}
```
aclevername, line 74: do we need to scan the values file for EVERY chart? isn't it only the charts that have the profiles tag?

pkg/helm/watcher/controller/helm_watcher.go
```diff
@@ -0,0 +1,94 @@
...
+
+	data := cache.Data{
+		Profiles: charts,
+		Values:   values,
```
aclevername, line 78: might be easier to map charts & values together, rather than having them separated?

pkg/helm/watcher/server.go
Skarlso: Good idea! 👍 Also, I'm actually gonna remove the `Must` call and return an error rather than panicking.

pkg/helm/watcher/cache/cache.go
```diff
@@ -0,0 +1,182 @@
...
+type profileVersion = string
+
+// ValueMap contains easy access for a profile name and version based values file.
+type ValueMap map[profileName]map[profileVersion][]byte
+
+// Cache defines an interface to work with the profile data cacher.
+//go:generate go run github.com/maxbrunsfeld/counterfeiter/v6 -generate
+//counterfeiter:generate . Cache
+type Cache interface {
+	Update(helmRepoNamespace, helmRepoName string, value Data) error
```
aclevername, line 37: bit nitty but isn;t this `CreateOrUpdate`? Or some better word

pkg/helm/watcher/cache/cache.go
```diff
@@ -0,0 +1,182 @@
...
+
+var _ Cache = &HelmCache{}
+
+// NewCache initialises the cache and returns it.
+func NewCache(cacheDir string) (*HelmCache, error) {
+	if _, err := os.Stat(cacheDir); os.IsNotExist(err) {
+		if err := os.MkdirAll(cacheDir, 0700); err != nil {
+			return nil, fmt.Errorf("failed to create helm cache dir: %w", err)
+		}
+	}
```
aclevername, line 65:
```suggestion
	if err := os.MkdirAll(cacheDir, 0700); err != nil {
		return nil, fmt.Errorf("failed to create helm cache dir: %w", err)
	}
```

```
// If path is already a directory, MkdirAll does nothing
// and returns nil.
func MkdirAll(path string, perm FileMode) error {
	// Fast path: if we can tell whether path is a directory or file, stop with success or error.
	dir, err := Stat(path)
	if err == nil {
		if dir.IsDir() {
			return nil
		}
		return &PathError{Op: "mkdir", Path: path, Err: syscall.ENOTDIR}
	}
```

pkg/helm/watcher/cache/cache.go
```diff
@@ -0,0 +1,182 @@
...
+
+	return &HelmCache{
+		cacheDir: cacheDir,
+	}, nil
+}
+
+// Update adds a new entry to the cache for the current helmRepository.
```
aclevername, line 72:
```suggestion
// Update adds a new entry or updates an existing entry in the cache for the helmRepository.
```

pkg/helm/watcher/cache/cache.go
Skarlso: Sure, why not. :) Technically, you update nothing to be something. :D

pkg/helm/watcher/cache/cache.go
Skarlso: Ah thanks for reminding me. :D I had some other logic in there, but removed that.

pkg/helm/watcher/controller/helm_watcher.go
```diff
@@ -0,0 +1,94 @@
...
+	RepoManager helm.HelmRepoManager
+}
+
+// +kubebuilder:rbac:groups=helm.watcher,resources=helmrepositories,verbs=get;list;watch
+// +kubebuilder:rbac:groups=helm.watcher,resources=helmrepositories/status,verbs=get
+
+// Reconcile is either called when there is a new HelmRepository or, when there is an update to a HelmRepository.
+// Because the watcher watches all helmrepositories, it will update data for all of them.
+func (r *HelmWatcherReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	log := logr.FromContextOrDiscard(ctx)
```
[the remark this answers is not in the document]
Skarlso: Does flux remove HelmRepository objects? I think they just stay there. In any case, sure, why not.

pkg/helm/watcher/controller/helm_watcher.go
Skarlso: I found this way to be stupid easy to handle. And returning the right values is simpler and faster without having to rummage through profiles. What easier handling would you get out of bringing them together and in what way would you do that? Also, they are separate because I'm writing out profiles by yaml serialising them.

pkg/helm/watcher/cache/cache.go
```diff
@@ -0,0 +1,182 @@
...
+	}()
+
+	// release on timeout
+	ctx, cancel := context.WithTimeout(context.TODO(), time.Second)
+	defer cancel()
+
+	ok, err := lock.TryRLockContext(ctx, 250*time.Millisecond) // try to lock every 1/4 second
+	if !ok {
+		return fmt.Errorf("unable to read lock file %s: %w", lockFilename, err)
+	}
```
aclevername, line 89: so if it can't acquire the lock after 1 second, it exits with an error, which will then cause the reconciliation loop to repeat (I believe if the reconciliation loop errors it requeues the resource again, you can configure a delay I believe). Given that repeated reconciliation loops for the same resource can be quite noisy in the logs, and in this case also involves a lot of API calls to fetch the index.yaml, is it worth drastically increasing this timeout? Maybe 1 minute?

pkg/helm/watcher/controller/helm_watcher.go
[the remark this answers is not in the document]
aclevername: its a resource that is manually created atm, so in theory manually deleted too

pkg/helm/watcher/cache/cache.go
Skarlso: Yeah, makes sense to increase the timeout.

pkg/helm/watcher/controller/helm_watcher.go
aclevername: is it easier?

Its currentyl
```
		Profiles: []*pb.Profile{
                 &pb.Profile{
		  Name:              "test-profiles-1",
		  Home:              "home",
		  Description:       "description",
		  Icon:              "icon",
		  KubeVersion:       "1.21",
		  AvailableVersions: []string{"0.0.1"},
	        }, 
                 &pb.Profile{
		  Name:              "test-profiles-2",
		  Home:              "home",
		  Description:       "description",
		  Icon:              "icon",
		  KubeVersion:       "1.21",
		  AvailableVersions: []string{"0.0.4"},
	        }, 
		Values: map[string]map[string][]byte{
			profile1.Name: {
				"0.0.1": []byte("value"),
			},
			profile2.Name: {
				"0.0.4": []byte("value"),
			},
		},
```

Woudn't:
```
Profiles: []Profile{
               {
                   &pb.Profile{
		    Name:              "test-profiles-1",
		    Home:              "home",
		    Description:       "description",
		    Icon:              "icon",
		    KubeVersion:       "1.21",
		    AvailableVersions: []string{"0.0.1"},
	          }, 
                  Values:
		    "0.0.1": []byte("value"),
		  },
	       },
               {
                   &pb.Profile{
		    Name:              "test-profiles-1",
		    Home:              "home",
		    Description:       "description",
		    Icon:              "icon",
		    KubeVersion:       "1.21",
		    AvailableVersions: []string{"0.0.1"},
	          }, 
                  Values:
		    "0.0.1": []byte("value"),
		  },
	       },
	     }    
```

be an easier mental model?

pkg/helm/watcher/cache/cache.go
bigkevmcd: Quite often Caches are `Put` (or `Set`) and `Get`

Skarlso: Put is good too

aclevername: I like `Set`

pkg/helm/watcher/cache/cache.go
aclevername: :+1: sorry didn't intend to come across so odd, was just thinking aloud as I typed. Glad it made sense :joy:

pkg/helm/watcher/controller/helm_watcher.go
aclevername: wheres my coffee ffs

pkg/helm/watcher/cache/cache.go
Skarlso: No worries!! It did. :D I could literally read your mental process while thinking this over and I really appreciate that! :)

pkg/helm/watcher/controller/helm_watcher.go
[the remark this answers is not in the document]
Skarlso: @Skarlso Going to add this tomorrow.

stefanprodan:
You may want to address:
```
✗ High severity vulnerability found in golang.org/x/crypto/ssh
  Description: NULL Pointer Dereference
```

Like this https://github.com/fluxcd/flux2/blob/main/go.mod
```
// Fix for CVE-2020-29652: https://github.com/golang/crypto/commit/8b5274cf687fd9316b4108863654cc57385531e8
// Fix for CVE-2021-43565: https://github.com/golang/crypto/commit/5770296d904e90f15f38f77dfc2e43fdf5efc083
replace golang.org/x/crypto => golang.org/x/crypto v0.0.0-20211215153901-e495a2d5b3d3
```

pkg/helm/watcher/controller/helm_watcher.go
```diff
@@ -0,0 +1,103 @@
...
+	Cache       cache.Cache
+	RepoManager helm.HelmRepoManager
+}
+
+// +kubebuilder:rbac:groups=helm.watcher,resources=helmrepositories,verbs=get;list;watch
+// +kubebuilder:rbac:groups=helm.watcher,resources=helmrepositories/status,verbs=get
+
+// Reconcile is either called when there is a new HelmRepository or, when there is an update to a HelmRepository.
+// Because the watcher watches all helmrepositories, it will update data for all of them.
+func (r *HelmWatcherReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
```
aclevername, line 29: does this mean we scan every HelmRepository resource now? Before it was just the 1 that was configured

pkg/helm/watcher/cache/cache.go
```diff
@@ -0,0 +1,193 @@
...
+
+	return &ProfileCache{
+		cacheLocation: cacheLocation,
+	}, nil
+}
+
+// Put adds a new entry or updates an existing entry in the cache for the helmRepository.
+func (c *ProfileCache) Put(ctx context.Context, helmRepoNamespace, helmRepoName string, value Data) error {
+	op := func() error {
```
aclevername, line 74: "op" :thinking: can we be more descriptive?

pkg/helm/watcher/controller/helm_watcher.go
bigkevmcd: We should be...

There's an argument to be made to only scan ones with specific labels, but that's probably an optimisation :-)

pkg/helm/watcher/cache/cache.go
Skarlso: short for operation :) but we can, sure

pkg/helm/watcher/controller/helm_watcher.go
Skarlso: Yep, we scan everything. This will be a good stepping stone to work from later when profiles supports multiple HelmRepositories.

pkg/helm/watcher/cache/cache.go
aclevername: operation to do what though :D

Skarlso: Well, anything. :D It's a blanket term for a function that is going to be executed with a lock so I don't have to repeat the locking logic everywhere. :D so I don't know. It's a generic operation. Maybeeeee ummm.... lockOperation or cacheOperation or cacheFunc?

pkg/helm/watcher/controller/helm_watcher.go
Skarlso: The what from the where why what? :D

Skarlso: Oh, no that shit is still needed otherwise the GetProfiles commands will not find the helmRepository. That's a different thing separate from this implementation.

Skarlso: Sadly, that stuff will still only work with a single HelmRepository until fixed.

Skarlso: Opened https://github.com/weaveworks/weave-gitops/issues/1306 to track to switch to tracking and using multiple HelmRepositories.

pkg/helm/watcher/controller/helm_watcher.go
```diff
@@ -0,0 +1,103 @@
...
+	log := logr.FromContextOrDiscard(ctx)
+
+	// get source object
+	var repository sourcev1.HelmRepository
+	if err := r.Get(ctx, req.NamespacedName, &repository); err != nil {
+		return ctrl.Result{}, client.IgnoreNotFound(err)
+	}
+
+	// handle deleted event
+	if repository.DeletionTimestamp != nil {
```
Skarlso, line 39: Hmpf. This doesn't work for things deleted by hand which don't have the curtesy of adding a deletion timestamp. This might be a problem. I can't definitively detect if something is deleted or not.

Skarlso: Hmm.. I might be able to do it in the Get actually. Assume that it couldn't find it, because it was deleted and that's why the recon loop was triggered. That's good enough :)

Skarlso: Hum.... upon delete the repository isn't given, meaning, it's empty. so I can't remove the cache for it.

Sorry @aclevername, deleting is not something I can support from the reconciler. And now I remember what finalisers are there for. :D But we don't have that. :)

bigkevmcd: Can you say more about this, what does "deleted by hand" mean here?

Skarlso: `kubectl delete`

Skarlso: I'm not a 100% sure if flux, when removing something, sets a delete timestamp either. Do you know something, Kevin?

Skarlso: Actually, I can just test that. :)

bigkevmcd: Ummm...`kubectl delete` should set the timestamp https://kubernetes.io/docs/reference/using-api/api-concepts/#resource-deletion

Skarlso: If it does, the reconcile does not get it. I did a kubectl delete and it didn't have a timestamp.

Skarlso: No, I know what's going on. The reconcile needs to get the object in order to see anything about it. But by the time the reconcilre is called the object is no longer there.

```go

	// get source object
	var repository sourcev1.HelmRepository
	if err := r.Get(ctx, req.NamespacedName, &repository); err != nil {
		return ctrl.Result{}, client.IgnoreNotFound(err)
	}

```

This runs and returns, because the object is not found.

Skarlso: This is why there are finalisers. You can do your cleanup in a finaliser.

Skarlso: https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#finalizers

Which I'm sure you know about. :)

bigkevmcd: Maybe have a look at https://github.com/fluxcd/source-controller/blob/main/controllers/helmrepository_controller.go#L94-L107

Skarlso: So this registers a finaliser with:

```
// +kubebuilder:rbac:groups=source.toolkit.fluxcd.io,resources=helmrepositories/finalizers,verbs=get;create;update;patch;delete
```

And somewhere else I can't find. :) I can do that too, but I'm not sure I should.

Skarlso: Ok. Gonna add a finaliser. 👍

cmd/gitops/ui/run/cmd.go
```diff
@@ -110,7 +121,30 @@ func runCmd(cmd *cobra.Command, args []string) error {
...
+		HealthzBindAddress: options.WatcherHealthzBindAddress,
+		WatcherPort:        options.WatcherPort,
+	})
+	if err != nil {
+		return fmt.Errorf("failed to start the watcher: %w", err)
+	}
+
+	go func() {
+		if err := profileWatcher.StartWatcher(); err != nil {
+			log.Error(err, "failed to start profile watcher")
```
aclevername, line 142: this might not be from starting but just the controller running in general right?

aclevername (approved):
LGTM, excellent stuff @Skarlso :heavy_check_mark: could maybe do with some more logging in the cache/controller, to help with debugging

pkg/helm/watcher/controller/generation_update_predicate.go
```diff
@@ -0,0 +1,20 @@
...
+type GenerationUpdatePredicate struct {
+	predicate.Funcs
+}
+
+func (GenerationUpdatePredicate) Update(e event.UpdateEvent) bool {
+	if e.ObjectOld == nil || e.ObjectNew == nil {
+		return false
+	}
+
+	return e.ObjectOld.GetGeneration() < e.ObjectNew.GetGeneration()
```
aclevername, line 19: when would this not be the case :thinking:

Skarlso: Yes.

Because...

How this works is as follows. The HelmRepository is created. At that point, it's generation 1. This is the create event, we don't care about that since it doesn't have an artifact. Then flux reconciles it, making it generation two, but artifact is still not updated. Then comes a patch event the artifact is added, but the generation doesn't change. To illustrate: 

This is an update event where the artifact is added. Called with old and called with new are the two artifact that come in. Here, the observed generation stays two, but for the first artifact is `nil` for the second, artifact is there.
```
called with old: &{TypeMeta:{Kind: APIVersion:} ObjectMeta:{Name:podinfo GenerateName: Namespace:wego-system SelfLink: UID:78d8620a-f19c-4fa0-a8e3-33c62775d050 ResourceVersion:41538 Generation:2 CreationTimestamp:2022-01-14 17:20:51 +0100 CET DeletionTimestamp:<nil> DeletionGracePeriodSeconds:<nil> Labels:map[kustomize.toolkit.fluxcd.io/name:kind-kind-user kustomize.toolkit.fluxcd.io/namespace:wego-system] Annotations:map[kubectl.kubernetes.io/last-applied-configuration:{"apiVersion":"source.toolkit.fluxcd.io/v1beta1","kind":"HelmRepository","metadata":{"annotations":{"kustomize.toolkit.fluxcd.io/checksum":"57c8b5346547f7eba33632994f4b57dd5cf744db"},"labels":{"kustomize.toolkit.fluxcd.io/name":"kind-kind-user","kustomize.toolkit.fluxcd.io/namespace":"wego-system"},"name":"podinfo","namespace":"wego-system"},"spec":{"interval":"30s","url":"https://weaveworks.github.io/profiles-examples"}}
 kustomize.toolkit.fluxcd.io/checksum:57c8b5346547f7eba33632994f4b57dd5cf744db] OwnerReferences:[] Finalizers:[finalizers.fluxcd.io finalizers.helm.watcher] ClusterName: ManagedFields:[{Manager:gitops Operation:Update APIVersion:source.toolkit.fluxcd.io/v1beta1 Time:2022-01-14 17:20:51 +0100 CET FieldsType:FieldsV1 FieldsV1:{"f:metadata":{"f:finalizers":{"v:\"finalizers.helm.watcher\"":{}}}} Subresource:} {Manager:kustomize-controller Operation:Update APIVersion:source.toolkit.fluxcd.io/v1beta1 Time:2022-01-14 17:20:51 +0100 CET FieldsType:FieldsV1 FieldsV1:{"f:metadata":{"f:annotations":{".":{},"f:kubectl.kubernetes.io/last-applied-configuration":{},"f:kustomize.toolkit.fluxcd.io/checksum":{}},"f:labels":{".":{},"f:kustomize.toolkit.fluxcd.io/name":{},"f:kustomize.toolkit.fluxcd.io/namespace":{}}},"f:spec":{".":{},"f:interval":{},"f:url":{}}} Subresource:} {Manager:source-controller Operation:Update APIVersion:source.toolkit.fluxcd.io/v1beta1 Time:2022-01-14 17:20:51 +0100 CET FieldsType:FieldsV1 FieldsV1:{"f:metadata":{"f:finalizers":{".":{},"v:\"finalizers.fluxcd.io\"":{}}},"f:spec":{"f:timeout":{}},"f:status":{".":{},"f:conditions":{},"f:observedGeneration":{}}} Subresource:}]} Spec:{URL:https://weaveworks.github.io/profiles-examples SecretRef:<nil> PassCredentials:false Interval:{Duration:30s} Timeout:&Duration{Duration:1m0s,} Suspend:false AccessFrom:<nil>} Status:{HERE =>>>>>>>>>>ObservedGeneration:2 Conditions:[{Type:Ready Status:Unknown ObservedGeneration:0 LastTransitionTime:2022-01-14 17:20:51 +0100 CET Reason:Progressing Message:reconciliation in progress}] URL: HERE >>>>>>>>>>>Artifact:<nil> ReconcileRequestStatus:{LastHandledReconcileAt:}}}



called with new: &{TypeMeta:{Kind: APIVersion:} ObjectMeta:{Name:podinfo GenerateName: Namespace:wego-system SelfLink: UID:78d8620a-f19c-4fa0-a8e3-33c62775d050 ResourceVersion:41540 Generation:2 CreationTimestamp:2022-01-14 17:20:51 +0100 CET DeletionTimestamp:<nil> DeletionGracePeriodSeconds:<nil> Labels:map[kustomize.toolkit.fluxcd.io/name:kind-kind-user kustomize.toolkit.fluxcd.io/namespace:wego-system] Annotations:map[kubectl.kubernetes.io/last-applied-configuration:{"apiVersion":"source.toolkit.fluxcd.io/v1beta1","kind":"HelmRepository","metadata":{"annotations":{"kustomize.toolkit.fluxcd.io/checksum":"57c8b5346547f7eba33632994f4b57dd5cf744db"},"labels":{"kustomize.toolkit.fluxcd.io/name":"kind-kind-user","kustomize.toolkit.fluxcd.io/namespace":"wego-system"},"name":"podinfo","namespace":"wego-system"},"spec":{"interval":"30s","url":"https://weaveworks.github.io/profiles-examples"}}
 kustomize.toolkit.fluxcd.io/checksum:57c8b5346547f7eba33632994f4b57dd5cf744db] OwnerReferences:[] Finalizers:[finalizers.fluxcd.io finalizers.helm.watcher] ClusterName: ManagedFields:[{Manager:gitops Operation:Update APIVersion:source.toolkit.fluxcd.io/v1beta1 Time:2022-01-14 17:20:51 +0100 CET FieldsType:FieldsV1 FieldsV1:{"f:metadata":{"f:finalizers":{"v:\"finalizers.helm.watcher\"":{}}}} Subresource:} {Manager:kustomize-controller Operation:Update APIVersion:source.toolkit.fluxcd.io/v1beta1 Time:2022-01-14 17:20:51 +0100 CET FieldsType:FieldsV1 FieldsV1:{"f:metadata":{"f:annotations":{".":{},"f:kubectl.kubernetes.io/last-applied-configuration":{},"f:kustomize.toolkit.fluxcd.io/checksum":{}},"f:labels":{".":{},"f:kustomize.toolkit.fluxcd.io/name":{},"f:kustomize.toolkit.fluxcd.io/namespace":{}}},"f:spec":{".":{},"f:interval":{},"f:url":{}}} Subresource:} {Manager:source-controller Operation:Update APIVersion:source.toolkit.fluxcd.io/v1beta1 Time:2022-01-14 17:20:51 +0100 CET FieldsType:FieldsV1 FieldsV1:{"f:metadata":{"f:finalizers":{".":{},"v:\"finalizers.fluxcd.io\"":{}}},"f:spec":{"f:timeout":{}},"f:status":{".":{},"f:artifact":{".":{},"f:checksum":{},"f:lastUpdateTime":{},"f:path":{},"f:revision":{},"f:url":{}},"f:conditions":{},"f:observedGeneration":{},"f:url":{}}} Subresource:}]} Spec:{URL:https://weaveworks.github.io/profiles-examples SecretRef:<nil> PassCredentials:false Interval:{Duration:30s} Timeout:&Duration{Duration:1m0s,} Suspend:false AccessFrom:<nil>} Status:{HERE =>>>>>>>>> ObservedGeneration:2 Conditions:[{Type:Ready Status:True ObservedGeneration:0 LastTransitionTime:2022-01-14 17:20:51 +0100 CET Reason:IndexationSucceed Message:Fetched revision: eca4147a15d2ac4a5eeca1cf247ac25505ae831c}] URL:http://source-controller.wego-system.svc.cluster.local./helmrepository/wego-system/podinfo/index.yaml HERE =>>>>>>>>>>>>>>>Artifact:0xc0048d1440 ReconcileRequestStatus:{LastHandledReconcileAt:}}}
```

You can see that generation stays, but artifact has been changed. So we need two predicates. One observing artifacts which are update events. And one observing the generation so we also catch and deal with Delete events.

Skarlso: Ok, discussion outcome:

- First, I was wrong. :) The req contains the name and namespace which would be enough to delete the cache. No idea why I didn't spot that. :D
- Second, we still opted to keep the finaliser because we want to make sure that a resource only gets deleted once we successfully deleted the cache as well, otherwise we would be showing stale data on the UI. Still saying profile exists even though the helmrepository was deleted
- Third, there is an optimization possible later where we also check the revision of the new Generation of the resource so we don't update the cache if only the interval is changed for example. ( Interval and spec change increase the generation of the object ) ( I'll create a ticket about that to further improve the inner working of the reconcile loop )

pkg/server/profiles.go
```diff
@@ -112,7 +99,7 @@ func (s *ProfilesServer) GetProfiles(ctx context.Context, msg *pb.GetProfilesReq
 		return nil, fmt.Errorf("failed to get HelmRepository %q/%q: %w", s.HelmRepoNamespace, s.HelmRepoName, err)
 	}
 
-	ps, err := s.HelmChartManager.GetCharts(ctx, helmRepo, helm.Profiles)
+	ps, err := s.HelmCache.ListProfiles(ctx, helmRepo.Namespace, helmRepo.Name)
 	if err != nil {
```
aclevername, line 103: do we wanna delete the get request above? line 80

Skarlso:
This PR now also adds now a missing HelmRepository setting for the returning profiles:

```
curl http://localhost:9001/v1/profiles
{"profiles":[{"name":"podinfo", "home":"https://github.com/stefanprodan/podinfo", "sources":["https://github.com/stefanprodan/podinfo"], "description":"Podinfo Helm chart for Kubernetes", "keywords":[], "maintainers":[{"name":"stefanprodan", "email":"stefanprodan@users.noreply.github.com", "url":""}], "icon":"", "annotations":{}, "kubeVersion":">=1.19.0-0", "helmRepository":{"name":"podinfo", "namespace":"wego-system"}, "availableVersions":["6.0.0", "6.0.1"], "layer":""}]}
```

`"helmRepository":{"name":"podinfo", "namespace":"wego-system"}` was not set before.

pkg/server/profiles.go
Skarlso: I think that should be part of the next PR where we eliminate the use of a single helm repo, not?

Just because it would change how we return things. It would remove at least one of the errors. Either the failed to scan or the not found being returned as a 200. Which is fine by me to get it nuked. But I don't know what the UI expect.

Skarlso: Same for `GetProfileValues`.

pkg/helm/watcher/controller/helm_watcher.go
```diff
@@ -125,13 +126,17 @@ func (r *HelmWatcherReconciler) reconcileDelete(ctx context.Context, repository
...
 
+	log.Info("deleted repository cache", "namespace", repository.Namespace, "name", repository.Name)
 	// Remove our finalizer from the list and update it
 	controllerutil.RemoveFinalizer(&repository, watcherFinalizer)
 
 	if err := r.Update(ctx, &repository); err != nil {
-		return ctrl.Result{}, err
+		log.Error(err, "failed to update repository to remove the finalizer, re-queuing after 5 seconds", "namespace", repository.Namespace, "name", repository.Name)
+		// note that we don't return the error if we want the 5 second requeue to work. with the error the Result is ignored.
+		return ctrl.Result{RequeueAfter: 5 * time.Second}, nil
```
bigkevmcd, line 136: Don't do this...return an error and it will retry and automatically backoff.

Skarlso: Huh nice, that's true. It's using a rate limiter by default. Neat.

pkg/server/profiles.go
aclevername: sure

pkg/helm/watcher/controller/helm_watcher.go
```diff
@@ -0,0 +1,140 @@
...
+		return r.reconcileDelete(ctx, repository)
+	}
+
+	if repository.Status.Artifact == nil {
+		// This should not occur because the predicate already checks for artifact's existence, but we do this as a
+		// precaution in case that was circumvented.
+		return ctrl.Result{}, nil
+	}
+
+	log.Info("found the repository: ", "name", repository.Name)
```
bigkevmcd, line 66: Can I suggest doing one of two things...

Either add a function like:

```go
func namespacedName(o client.Object) types.NamespacedName {
       return types.NamespacedName{
               Name:      o.GetName(),
               Namespace: o.GetNamespace(),
       }
}
```
Which would let you simplify things to:

```go
log.Info("deleting repository cache", "repository", namespacedName(repository))
```

Because `types.NamespacedName` implements `String()`.

OR alternatively...

you could do...

```go
log := logr.FromContextOrDiscard(ctx).WithValues("repository", req.NamespacedName)
```
And then pass that into the other functions

Feel free to create a sublogger from the one if you prefer that...

Skarlso: The idea was to be somewhat consistent because not all of the functions are called with a logger context. But yeah I can do that.

Skarlso: For the record, what I wanted to say is that `req` is from the watcher, `Put` is the only operation called through the watcher where req is available. For the other two, these will have to be constructed from namespace and name and then passed in from GetProfiles and GetProfileValues where they are called.

luizbafilho (approved):
Looks great, thanks.

PR #1289, opened 2022-01-11
Disable checkpoint for all CI jobs

rokshana-b:
**What changed?**
Added `CHECKPOINT_DISABLE` env on all test jobs

**Why?**
To avoid interfering with the statistics that keep track of the number of downloads customers are making for `Weave-GitOps`

### DIFF
--- a/.github/actions/run-acceptance-test/action.yml
+++ b/.github/actions/run-acceptance-test/action.yml
@@ -112,6 +112,7 @@ runs:
       if [ "${{ github.job }}" == "smoke-tests-long" ]; then
         export GIT_PROVIDER="${{ inputs.git-provider }}"
       fi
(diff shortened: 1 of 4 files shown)


palemtnrider (changes requested):
NIT I think the code is easier to read if you add this override in the `env:`.  The coverage tests in the deploy action is missing the override.

rokshana-b:
> NIT I think the code is easier to read if you add this override in the `env:`. The coverage tests in the deploy action is missing the override.

So just adding the `CHECKPOINT_DISABLE` in the list of other envs should work, right? I'll do that then

rokshana-b:
@palemtnrider is there a way to test if the changes in this PR have the desired effect?

iahmad9 (approved):
LGTM

Issue #1240, opened 2021-12-15
UI should indicate that a PR was created on app removal 


palemtnrider:
**Short description**
As an app developer, I want to be informed that a PR has been created during the removal of my app so that I can go to Git(lab,hub) and accept the changes.

**Acceptance criteria**
* UI presents a message (similar to add) that informs me a PR has been created 
* UI gives me a link that I can use to go to the PR page
* 

**Design materials**
- See the add path and PR creation notifications

Currently, the UI shows the following which doesn't inform the user that a PR is waiting for review. This message is a bit misleading
![Screen Shot 2021-12-15 at 3 48 42 PM](https://user-images.githubusercontent.com/7633548/146280806-6cd4de8c-1d7d-449e-9591-bfe7d7613c4e.png)

rokshana-b:
This is working for me on both - github and gitlab

PR #1299, opened 2022-01-13
Link to PR on App Remove

joshri:
Closes: #1240

Auto-merge functionality added to app remove. If not auto-merged, a button will appear linking the user to the PR! 

image

image


jpellizzari (changes requested):
When removing via PR, I don't think we want to check for things to be removed on the cluster. Things won't get removed from the cluster until the PR is merged.

jpellizzari (approved):
LGTM!

Issue #1210, opened 2021-12-09
commits should open in new tab


sympatheticmoose:
Clicking a commit in the application details view takes you away from the UI and means the user may struggle to navigate back.
<img width="437" alt="Screenshot 2021-12-09 at 22 15 30" src="https://user-images.githubusercontent.com/19860021/145484326-fd543ef6-e950-45f2-853a-89e4692eeff6.png">

sympatheticmoose:
@jpellizzari interested if there's any other reasons beyond the fact that it overrides default behaviour? I raised the issue after witnessing this (slightly) trip up a demo, and personally I would expect that _any_ external links should open in a new tab by default - that feels fairly standard behaviour for web UIs and helps ensure that the dashboard isn't lost.

PR #1300, opened 2022-01-13
Links to commits open new tab

joshri:
Closes: #1210

Added `newTab` prop to the Link component which allows toggling of the` target='_blank'` attribute of the <a> element

### DIFF
--- a/ui/components/CommitsTable.tsx
+++ b/ui/components/CommitsTable.tsx
@@ -85,7 +85,11 @@ function CommitsTable({
         fields={[
           {
             label: "SHA",
-            value: (row: Commit) => <Link href={row.url}>{row.hash}</Link>,
+            value: (row: Commit) => (
+              <Link newTab href={row.url}>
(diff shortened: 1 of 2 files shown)


dhwthompson:
Given these are links to external sites, do we need to worry about adding a [`rel="noopener"`](https://developer.mozilla.org/en-US/docs/Web/HTML/Link_types/noopener) for a bit of added security?

jpellizzari (approved):

joshri:
> Given these are links to external sites, do we need to worry about adding a [`rel="noopener"`](https://developer.mozilla.org/en-US/docs/Web/HTML/Link_types/noopener) for a bit of added security?

I'm pretty sure browsers now have that functionality built in (https://stackoverflow.com/questions/50709625/link-with-target-blank-and-rel-noopener-noreferrer-still-vulnerable/50709760) - but I'll throw it in there cause it can't hurt right?

Issue #1278, opened 2022-01-06
Unable to add gitlab app


J-Thompson12:
**Describe the bug**
Unable to add a gitlab app. 
Error for public repo:
```
"error":"unable to clone 'ssh://git@gitlab.com/gogittest/sub-group/test-app-t75anmiy.git': ssh: handshake failed: ssh: unable to authenticate, attempted methods [none publickey] │
│ , no supported methods remain"}  
```

When adding for a private repo running gitops add works but then the source controller shows:
```
Error: failed to add the app : failed to clone repo: failed cloning user repo: ssh://git@gitlab.com/J-Thompson12/nginxlab.git: ssh: handshake failed: ssh: unable to authenticate, attempted methods [none publickey], no supported methods remain
```

**Environment**
 - gitops: [e.g. v0.1.0]
 - kubernetes:  [e.g. 1.20.4]
    - [ ] KinD - _version_]
    - [ ] k3s - _version_
    - [ ] cloud [e.g., EKS, AKS]  _version_
    - [ ] other - _name_ _version_
 - Browser + version: [e.g. chrome 74, safari 12, firefox 87]
 -
**Affects versions**

**To Reproduce**
1. Start with fresh env
2. Install gitops
3. Run `gitops add app . --config-repo=<repo>`

**Expected behavior**
The app to add and reconcile properly.

**Additional context**
I commented out the change made in this pr https://github.com/weaveworks/weave-gitops/pull/1249/files and ran an add for gitlab. This allowed me to add successfully.

J-Thompson12:
The changes made in https://github.com/weaveworks/weave-gitops/pull/1249 made it so that without a deploy key we are unable to clone the repo. Install without auto-merge does not clone it just creates a PR and then add always clones so it fails. Install with auto-merge clones the repo so it fails to install. We either need to add deploy keys on install because we are editing that repo or not allow a public repo to be the config repo.

J-Thompson12:
Possible solutions:
1. Force users to use private repos for config repo
2. Use go-git-providers for auto-merge to avoid cloning.
3. Remove changes to only using deploy key on private.
4. A solution I dont know of that magically fixes all our problems.

PR #1301, opened 2022-01-13
Add deploy key for public repo

J-Thompson12:
Closes: #1278 

**What changed?**
Revert changes made in https://github.com/weaveworks/weave-gitops/pull/1249. I left in the changes for the acceptance tests because they were good tests to keep in. 

**Why?**

**How did you test it?**
Manually and unit tests

**Release notes**

**Documentation Changes**

### DIFF
--- a/cmd/gitops/install/cmd.go
+++ b/cmd/gitops/install/cmd.go
@@ -68,7 +68,7 @@ func installRunCmd(cmd *cobra.Command, args []string) error {
 	ctx := context.Background()
 	namespace, _ := cmd.Parent().Flags().GetString("namespace")
 
-	configURL, err := gitproviders.NewRepoURL(installParams.ConfigRepo)
+	configURL, err := gitproviders.NewRepoURL(installParams.ConfigRepo, true)
 	if err != nil {
 		return err
 	}
@@ -122,9 +122,11 @@ func installRunCmd(cmd *cobra.Command, args []string) error {
 		providerClient := internal.NewGitProviderClient(osysClient.Stdout(), osysClient.LookupEnv, auth.NewAuthCLIHandler, log)
 
 		gitClient, gitProvider, err = factory.GetGitClients(context.Background(), providerClient, services.GitConfigParams{
-			URL:       installParams.ConfigRepo,
-			Namespace: namespace,
-			DryRun:    installParams.DryRun,
+			// We need to set URL and ConfigRepo to the same value so a deploy key is created for public config repos
+			URL:        installParams.ConfigRepo,
+			ConfigRepo: installParams.ConfigRepo,
+			Namespace:  namespace,
+			DryRun:     installParams.DryRun,
 		})
 
 		if err != nil {
--- a/pkg/services/auth/auth.go
+++ b/pkg/services/auth/auth.go
@@ -6,7 +6,6 @@ import (
 	"io"
 	"net/http"
 
-	"github.com/fluxcd/go-git-providers/gitprovider"
 	"github.com/weaveworks/weave-gitops/pkg/services/auth/internal"
 
 	"github.com/go-git/go-git/v5/plumbing/transport/ssh"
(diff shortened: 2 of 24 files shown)


jpellizzari (commented):
@J-Thompson12 Can you elaborate on what is actually going on here? This seems like a band-aid over a bigger problem.

Passing a single bool value all the way down into the stack is a code smell for me.

J-Thompson12:
> @J-Thompson12 Can you elaborate on what is actually going on here? This seems like a band-aid over a bigger problem.
> 
> Passing a single bool value all the way down into the stack is a code smell for me.

Oh I absolutely agree. This is definitely a patch over a bigger issue. There was some discussion with @palemtnrider about if we should always do a PR using go-git-providers and just merge the PR if auto-merge is true. The decision was that goes against the choice to use go-git whenever possible.

pkg/services/auth/auth.go
```diff
@@ -157,16 +157,16 @@ func (a *authSvc) setupDeployKey(ctx context.Context, name SecretName, targetNam
...
 
-func (a *authSvc) provisionDeployKey(ctx context.Context, targetName string, name SecretName, repo gitproviders.RepoURL) (*ssh.PublicKeys, error) {
+func (a *authSvc) provisionDeployKey(ctx context.Context, targetName string, name SecretName, repo gitproviders.RepoURL, isInstall bool) (*ssh.PublicKeys, error) {
 	visibility, err := a.gitProvider.GetRepoVisibility(ctx, repo)
 	if err != nil {
 		return nil, fmt.Errorf("error getting repo visibility: %w", err)
 	}
 
-	if *visibility == gitprovider.RepositoryVisibilityPublic {
+	if *visibility == gitprovider.RepositoryVisibilityPublic && !isInstall {
```
jpellizzari, line 169: @J-Thompson12 Wouldn't removing this ` *visibility == gitprovider.RepositoryVisibilityPublic` logic give us the result we are looking for?

We need a deploy key any time we are writing to a repo (regardless of it's visibility), correct?

J-Thompson12: We only need the deploy on install. We dont need it when we are adding a repo because we dont make changes to those repos.

J-Thompson12: This samw code is used for add and install so i needed some way to say which one was using it. Since this is a temp fix till a decision is made i went the easy route with the variable.

luizbafilho: The statement that we don't need to deploy keys on app repos is temporarily valid since we'll need those when implementing automated image updates, which was in the wish list in the retro we had a couple of days ago.

So, probably removing the visibility validation seems to me as a better long-term solution.

jpellizzari (approved):

luizbafilho:
The main reason for adding that check was that using an app public repo failed.

I think a good solution for handling that could be adding a new `IsConfigRepo` field to the `RepoURL` and calling it when initializing a new url of that type, for instance here:https://github.com/weaveworks/weave-gitops/blob/main/cmd/gitops/install/cmd.go#L71

With that in place we can check for it and just upload the deploy key when it's a configrepo. We endup with the same problem when in the future we decide to have the image updater feature, but at least with this new field the change is self contained and doesn't created special flows like `IsInstall`. What do you think @jpellizzari ?

J-Thompson12:
@luizbafilho I agree with that solution but dont we end up with the same problem in the future when we want to implement automated image updates?

luizbafilho:
> Sorry @luizbafilho what I was saying is that when we implement that feature wont we just start adding deploy keys to everything again. So we might as well just add them now.

I think when that happens we'll have some sort of flag indicating that we have the image updater enabled for that repo and then we can use that to indicate we _need_ to push the deploy key, but pushing it _always_ creates a usability issue.

J-Thompson12:
@luizbafilho I agree with that solution. I think it makes sense for now and the future.

cmd/gitops/install/cmd.go
```diff
@@ -122,9 +122,11 @@ func installRunCmd(cmd *cobra.Command, args []string) error {
 		providerClient := internal.NewGitProviderClient(osysClient.Stdout(), osysClient.LookupEnv, auth.NewAuthCLIHandler, log)
 
 		gitClient, gitProvider, err = factory.GetGitClients(context.Background(), providerClient, services.GitConfigParams{
-			URL:       installParams.ConfigRepo,
-			Namespace: namespace,
-			DryRun:    installParams.DryRun,
+			// We need to set URL and ConfigRepo to the same value so a deploy key is created for public config repos
+			URL:        installParams.ConfigRepo,
+			ConfigRepo: installParams.ConfigRepo,
```
J-Thompson12, line 127: I dont love having to set these two values but this is the only solution short of rewriting the entire GetGitClients functionality and all the places that use it.

jpellizzari (changes requested):
This latest change to pass in a bool to `NewRepoURL` is not acceptable. Let's discuss on a call.

jpellizzari (approved):
Just spoke with @luizbafilho and @J-Thompson12; approving this to unblock it for a release, but a follow up issue needs to be created and worked on ASAP to address the underlying issue and remove this tech debt.

PR #1307, opened 2022-01-14
Refactor application endpoints to get `client` at the time of the request

yiannistri:
Part of: #1092  

**What changed?**
Instead of creating a Kubernetes `client.Client` when an `applicationsServer` is created, we create it when the HTTP request is handled so that we can use existing middleware to determine the identity of the user making the request. Following that, we can assume the identity of that user when making calls to the Kubernetes API.

**Why?**

**How did you test it?**
Updated existing unit tests + manually

**Release notes**

**Documentation Changes**


sarataha (approved):
LGTM! 🌟

PR #1309, opened 2022-01-17
Upgrades to a ggp w/ the on-prem-gitlab fix

foot:
**What changed?**

- #1297 introduced a version of ggp that breaks on-prem support
- This PR upgrades to ggp that includes https://github.com/fluxcd/go-git-providers/pull/141 which fixes on-prem-gitlab support. 
- It also adds support for older wg installations that might not have a kustomization.yaml file present

**Why?**

For on-prem things!

**How did you test it?**

`gitops upgrade` before:

```
Error: failed to get git clients: error setting up deploy keys: failed check for existing deploy key: error getting user repo reference for owner XXXX, repo demo-01, error getting user repository domai
n "https://XXXX" not supported by this client: the client doesn't support handling requests for this domain
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -7,7 +7,7 @@ require (
 	github.com/benbjohnson/clock v1.1.0
 	github.com/coreos/go-oidc/v3 v3.1.0
 	github.com/deepmap/oapi-codegen v1.8.1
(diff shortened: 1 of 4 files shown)


yiannistri (approved):
LGTM 👍

PR #1312, opened 2022-01-18
Make docker builds faster

ozamosi:
**What changed?**
Change download-deps to not interact with git - it doesn't need git,
and docker builds shouldn't contain git.

Add a bunch more entries to dockerignore - this cuts out hundreds of megs
of build context.

Remove `make clean` from the Dockerfile. We shouldn't copy cleanable
state into the docker image in the first place. In this case, it seems
we run make clean to run git clean to remove the flux binary - so I
just added the flux binary to the dockerignore instead.

Reorder the Dockerfile and be more explicit, in order to cache more -
go changes now only makes the go code recompile.

**Why?**
Before this change, almost every change caused `make docker` to run
for 90-100s on my machine, because almost all caches were always
invalidated.

After this change, go changes only take 30s to re-build, whereas
javascript changes takes 40s (because they still trigger a go build),
and only initial builds or dependency changes take longer.

**How did you test it?**
Local, manual testing

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,2 +1,8 @@
 node_modules
 .parcel-cache
+.deps
+.git
+dist
+bin
+pkg/flux/bin/flux
+cmd/gitops/ui/run/dist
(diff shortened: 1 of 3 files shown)


jpellizzari (approved):
LGTM. Thanks!

luizbafilho (approved):

PR #1317, opened 2022-01-19
Add notifications if a new profile version is available

### LINKED ISSUE #1143: Profile Notifications
### Description
As suggested [here](https://gist.github.com/bigkevmcd/0ff7197e0a902c3dcde4a1e7bfc3984c#notifications), we want to generate notifications when there is a new profile version is available

### Scope
- We will use [Flux Notifications API](https://fluxcd.io/docs/components/notification/api/) to achieve this 
- The notifications should be sent to Flux notifications, we don’t care how Flux notifications then forward the information
- This will only track updates to Profiles, this does not solve the problem of out-of-date Helm charts that are dependencies of a Profile.

### Acceptance Criteria
Given that I am a developer
When I encounter a new version of the profile available on a watched helm repository
Then a notification should be sent to the Flux Notifications API 
So that it can be consumed later by different providers

Skarlso:
Closes #1143

# Testing

- Have your own Slack space and create a webhook URL outlined here: https://api.slack.com/messaging/webhooks
- Create a secret with the webhook URL
```
kubectl -n wego-system create secret generic slack-url \
--from-literal=address=https://hooks.slack.com/services/<token>
```
- Setup Slack notification for flux using the following notification provider:
```yaml
apiVersion: notification.toolkit.fluxcd.io/v1beta1
kind: Provider
metadata:
  name: slack
  namespace: wego-system
spec:
  type: slack
  channel: general
  secretRef:
    name: slack-url
```
- Port forward all the things which is the `notification-controller` and the `source-controller`
- Start UI
```
./bin/gitops ui run --helm-repo-namespace wego-system --helm-repo-name podinfo --notification-controller-address http://localhost:8081
```
- Install main components
```
./gitops install --config-repo=ssh://git@github.com/Skarlso/gitops-training.git
```
- Create a profile app
```
./bin/gitops add app --name podinfo --url https://weaveworks.github.io/profiles-examples --config-repo ssh://git@github.com/Skarlso/gitops-training.git --chart podinfo  --deployment-type helm --auto-merge
```
- Update the profile yaml file to include a new version and update the code in your gitops repo
- Observe incoming messages to your slack channel

Screenshot 2022-01-19 at 17 23 31

Interesting.. where is my message?

_Update_:

Works!

Screenshot 2022-01-20 at 13 45 51


pkg/helm/watcher/cache/cache.go
```diff
@@ -158,6 +160,41 @@ func (c *ProfileCache) ListProfiles(ctx context.Context, helmRepoNamespace, helm
...
+// GetAvailableVersionsForProfile returns all stored available versions for a profile.
+func (c *ProfileCache) GetAvailableVersionsForProfile(ctx context.Context, helmRepoNamespace, helmRepoName, profileName string) ([]string, error) {
+	// Because the folders of versions are only stored when there are values for a version
+	// we, instead, look in the profiles.yaml file for available versions.
+	var result []string
+
+	getAllAvailableVersionsOp := func() error {
+		var profiles []*pb.Profile
+
+		content, err := os.ReadFile(filepath.Join(c.cacheLocation, helmRepoNamespace, helmRepoName, profileFilename))
```
Skarlso, line 172: ignore if not found

cmd/gitops/ui/run/cmd.go
```diff
@@ -72,6 +73,7 @@ func NewCommand() *cobra.Command {
 	cmd.Flags().StringVar(&options.ProfileCacheLocation, "profile-cache-location", "/tmp/helm-cache", "the location where the cache Profile data lives")
 	cmd.Flags().StringVar(&options.WatcherHealthzBindAddress, "watcher-healthz-bind-address", ":9981", "bind address for the healthz service of the watcher")
 	cmd.Flags().StringVar(&options.WatcherMetricsBindAddress, "watcher-metrics-bind-address", ":9980", "bind address for the metrics service of the watcher")
+	cmd.Flags().StringVar(&options.NotificationControllerAddress, "notification-controller-address", "http://notification-controller", "the address of the notification-controller running in the cluster")
```
bigkevmcd, line 76: How about defaulting this to something like...

`http://notification-controller.flux-system.svc.cluster.local./` ?

pkg/helm/watcher/cache/cache.go
```diff
@@ -43,6 +43,8 @@ type Cache interface {
 	// GetProfileValues will try and find a specific values file for the given profileName and profileVersion. Returns an
 	// error if said version is not found.
 	GetProfileValues(ctx context.Context, helmRepoNamespace, helmRepoName, profileName, profileVersion string) ([]byte, error)
+	// GetAvailableVersionsForProfile returns all stored available versions for a profile.
+	GetAvailableVersionsForProfile(ctx context.Context, helmRepoNamespace, helmRepoName, profileName string) ([]string, error)
```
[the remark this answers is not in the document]
Skarlso: I don't have a strong opinion on this, but `get pod` `get pods`...

cmd/gitops/ui/run/cmd.go
bigkevmcd: You probably want to read https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/ 

That'll help you work out what the right answer is :-)

pkg/helm/watcher/cache/cache.go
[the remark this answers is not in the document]
Skarlso: Which just shows that there are multiple opinions on naming this thing. :) I think the GetVersions checks out grammatically. But I can rename it.

cmd/gitops/ui/run/cmd.go
Skarlso: I'm aware of that document, thanks. :) But doesn't flux set up dedicated service accounts for each controller?

Skarlso: In any case, I can set that, flux has probably some custom coredns rules to get that to work anyways.

Skarlso: `http://notification-controller./`

pkg/helm/watcher/controller/helm_watcher.go
```diff
@@ -75,16 +89,22 @@ func (r *HelmWatcherReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 	values := make(cache.ValueMap)
 
 	for _, chart := range charts {
+		if v, err := r.checkForNewVersion(ctx, chart); err != nil {
+			log.Error(err, "checking for new versions failed")
+		} else if v != "" {
+			log.Info("sending notification event for new version", "version", v)
+			r.event(ctx, &repository, repository.Status.Artifact.Revision, "info", fmt.Sprintf("New version available for profile %s with version %s", chart.Name, v))
+		}
+
```
aclevername, line 98: do we want to alert for every profile? Or just when a profile that is installed has a new version? cc @bigkevmcd 

I don't know how useful it is to know a new versions of profile X is available, if I'm not evening using X :thinking: or perhaps adjust it to only alert `majors`(and maybe `minors`?)? Just thinking about how noisy this is

pkg/helm/watcher/controller/helm_watcher.go
```diff
@@ -138,3 +158,84 @@ func (r *HelmWatcherReconciler) reconcileDelete(ctx context.Context, repository
 	// Stop reconciliation as the object is being deleted
 	return ctrl.Result{}, nil
 }
+
+// event emits a Kubernetes event and forwards the event to notification controller if configured.
+func (r *HelmWatcherReconciler) event(ctx context.Context, hr *sourcev1.HelmRepository, revision, severity, msg string) {
```
aclevername, line 163:
```suggestion
// sendEvent sends a Kubernetes event and forwards the event to notification controller if configured.
func (r *HelmWatcherReconciler) sendEvent(ctx context.Context, hr *sourcev1.HelmRepository, revision, severity, msg string) {
```

bigkevmcd: This is a really good question, and actually, a tricker one to answer than it seems...

I'd say "only installed ones", but determining that is harder than it looks, let's get it going for all, and then work out to filter for "only installed ones".

pkg/helm/watcher/controller/helm_watcher.go
Skarlso: So, yeah I had the same question. I'll create some tickets about this when I get there. :D But in general, yes for now. because it's not quite easy to get what's installed or not.

bigkevmcd: This is a really good question, and actually, a tricker one to answer than it seems...

I'd say "only installed ones", but determining that is harder than it looks, let's get it going for all, and then work out how to filter for "only installed ones".

aclevername: > So, yeah I had the same question. I'll create some tickets about this when I get there. :D But in general, yes for now. Because it's not quite easy to get what's installed or not.

yeah leaving it as is and adding tickets sounds good. @bigkevmcd pointed out on slack:
```
But, what does it mean to filter for installed?
4:06
Installed in this cluster ?
4:06
Installed in a repo ?
4:06
Installed in any of a set of clusters ?
```

lets capture this and we can decide how to proceed

aclevername: cc @Himangini

pkg/helm/watcher/cache/cache.go
```diff
@@ -158,6 +160,46 @@ func (c *ProfileCache) ListProfiles(ctx context.Context, helmRepoNamespace, helm
...
+			if os.IsNotExist(err) {
+				return nil
+			}
+
+			return fmt.Errorf("failed to read profiles data for helm repo: %w", err)
+		}
+
+		if err := yaml.Unmarshal(content, &profiles); err != nil {
+			return fmt.Errorf("failed to unmarshal profiles data: %w", err)
+		}
```
aclevername, line 184: this part is probably repeated in other funcs too right? worth extracting out into a `getProfilesFromFile` func or something?

pkg/helm/watcher/controller/helm_watcher.go
```diff
@@ -138,3 +158,84 @@ func (r *HelmWatcherReconciler) reconcileDelete(ctx context.Context, repository
 	// Stop reconciliation as the object is being deleted
 	return ctrl.Result{}, nil
 }
+
+// sendEvent emits a Kubernetes event and forwards the event to notification controller if configured.
+func (r *HelmWatcherReconciler) sendEvent(ctx context.Context, hr *sourcev1.HelmRepository, revision, severity, msg string) {
+	log := logr.FromContextOrDiscard(ctx)
```
aclevername, line 164: are we not already doing this in the reconciliation loop? can we not just pass the `log` in? Its also nice that the `log` in the reconciliation loop has the helm repo details attached

pkg/helm/watcher/controller/helm_watcher.go
```diff
@@ -138,3 +158,84 @@ func (r *HelmWatcherReconciler) reconcileDelete(ctx context.Context, repository
...
+
+	if r.ExternalEventRecorder == nil {
+		return
+	}
```
aclevername, line 168: Is this because notifications is optional, or because this could be unset? If its the latter this is smelly... can we make `HelmWatcherReconciler` fields private, have a constructor instead and then do any validation of fields (or defaulting) done in there?

Skarlso: Optional. If not set, we don't send an event. It's only set if the address is set. If set specifically to empty, then we don't send any events.

pkg/helm/watcher/cache/cache.go
Skarlso: Yeah, now it is. It wasn't previously. So yes, maybe it's worth extracting it now.

Skarlso: Though, it varies slightly. Here, we ignore if the file is not found.

pkg/helm/watcher/controller/helm_watcher.go
aclevername: isn't the address defaulting to a value, so its always set?

pkg/helm/watcher/controller/helm_watcher.go
```diff
@@ -140,3 +159,82 @@ func (r *HelmWatcherReconciler) reconcileDelete(ctx context.Context, repository
...
+}
+
+func (r *HelmWatcherReconciler) convertStringListToSemanticVersionList(versions []string) ([]*semver.Version, error) {
+	var result []*semver.Version
+
+	for _, v := range versions {
+		ver, err := semver.NewVersion(v)
+		if err != nil {
+			return nil, err
+		}
```
aclevername, line 234: felt this pain in eksctl too but it feels so frustrating to blow up (error) if 1 out of 100 versions can't be parsed. Perhaps worth just skipping the version and logging a warning instead?

bigkevmcd: Sometimes I'd say yes, but in this case, there are a number of things upstream that should prevent this, so if a chart has evaded these checks, I'd say it's reasonable to blow up :-)

pkg/helm/watcher/controller/helm_watcher.go
Skarlso: Yes, hence, if you don't want it you have to explicitly set it to empty.

pkg/helm/watcher/controller/helm_watcher.go
aclevername: Fair point! Sounds good

pkg/helm/watcher/cache/cache.go
```diff
@@ -158,6 +155,39 @@ func (c *ProfileCache) ListProfiles(ctx context.Context, helmRepoNamespace, helm
...
+
+		return fmt.Errorf("profile with name %s not found in cached profiles", profileName)
+	}
+
+	if err := c.tryWithLock(ctx, getAllAvailableVersionsOp); err != nil {
+		return nil, err
+	}
+
+	return result, nil
```
aclevername, line 188:
```suggestion
	return result, c.tryWithLock(ctx, getAllAvailableVersionsOp)
```

this would work right? Not saying its cleaner, just thinking

pkg/helm/watcher/controller/helm_watcher.go
```diff
@@ -75,16 +88,22 @@ func (r *HelmWatcherReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 	values := make(cache.ValueMap)
 
 	for _, chart := range charts {
+		if v, err := r.checkForNewVersion(ctx, chart); err != nil {
+			log.Error(err, "checking for new versions failed")
+		} else if v != "" {
+			log.Info("sending notification event for new version", "version", v)
+			r.sendEvent(log, &repository, repository.Status.Artifact.Revision, "info", chart.Name, v)
```
aclevername, line 95: some duplication here, passing in `repository` and `repository.Status.Artifact.Revision`? could this be condensed to just `repository`?

aclevername (approved):
couple nits- I'd wait for 2nd reviewer before merging :smile:

pkg/helm/watcher/cache/cache.go
Skarlso: technically... but I don't really like that. :D

cmd/gitops-server/main.go
```diff
@@ -71,11 +72,12 @@ func NewAPIServerCommand() *cobra.Command {
...
-				Cache:              profileCache,
-				MetricsBindAddress: metricsBindAddress,
-				HealthzBindAddress: healthzBindAddress,
-				WatcherPort:        watcherPort,
+				KubeClient:                    rawClient,
+				Cache:                         profileCache,
+				MetricsBindAddress:            metricsBindAddress,
+				HealthzBindAddress:            healthzBindAddress,
+				NotificationControllerAddress: notificationBindAddress,
+				WatcherPort:                   watcherPort,
```
bigkevmcd, line 80: Why are these not cmd-line flags?

pkg/helm/watcher/cache/cache.go
```diff
@@ -139,15 +141,10 @@ func (c *ProfileCache) ListProfiles(ctx context.Context, helmRepoNamespace, helm
 	var result []*pb.Profile
 
 	listOperation := func() error {
-		content, err := os.ReadFile(filepath.Join(c.cacheLocation, helmRepoNamespace, helmRepoName, profileFilename))
-		if err != nil {
+		if err := c.getProfilesFromFile(helmRepoNamespace, helmRepoName, &result); err != nil {
 			return fmt.Errorf("failed to read profiles data for helm repo: %w", err)
```
bigkevmcd, line 145: Could this include the details of the Helm repo it couldn't read profiles data from?

pkg/helm/watcher/controller/helm_watcher.go
```diff
@@ -140,3 +159,82 @@ func (r *HelmWatcherReconciler) reconcileDelete(ctx context.Context, repository
 	// Stop reconciliation as the object is being deleted
 	return ctrl.Result{}, nil
 }
+
+// sendEvent emits a Kubernetes event and forwards the event to notification controller if configured.
```
bigkevmcd, line 163: I don't see where in this code it is emitting a Kubernetes event?

pkg/helm/watcher/controller/helm_watcher.go
```diff
@@ -140,3 +159,82 @@ func (r *HelmWatcherReconciler) reconcileDelete(ctx context.Context, repository
...
+	if err != nil {
+		return "", err
+	}
+
+	oldVersions, err := r.convertStringListToSemanticVersionList(versions)
+	if err != nil {
+		return "", err
+	}
+
+	sort.SliceStable(newVersions, func(i, j int) bool {
```
bigkevmcd, line 205: This could be a local func value...or arguably `func sortVersions(versions []*semver.Version)`

Himangini:
LGTM, 2 ques
- Did you find a place to document this ?
> I'm looking for a place to document how to set up the alerting for the notification-controller to pick up and forward the notifications.
- Are you going to add testing steps ?
> testing steps incoming

cmd/gitops-server/main.go
Skarlso: They are, but not in this section which doesn't have command line flags.

pkg/helm/watcher/controller/helm_watcher.go
Skarlso: Heh, I knew you would say that. :) I was debating to extract it too.

pkg/helm/watcher/controller/helm_watcher.go
Skarlso: Ops, good spot

Skarlso:
@Himangini Ah good questions!

> testing steps

I'll add some, thanks

> documentation

Nope, I have not. I'm hoping someone will point me to some docs I can update.

Himangini (approved):
LGTM 🎉 
Re the documentation, I'll find out where's the best place to put it. Not a blocker for this PR :)

PR #1320, opened 2022-01-19
Print out GitSource from config repository on dry-run

josecordaz:
Closes: 
- #1340 

**What changed?**
Split up git client instantiation on install
Removed old `targetName` field from `getGitClients` function as it wasn't used
Moved `GitSource` creation to `models.BootstrapManifests` as this is the function used in dry-run.

**Why?**
To be able to print out GitSource on dry-run without setting up a deploy key, which is not needed for dry-run.

**How did you test it?**
Manually, CI

**Release notes**

**Documentation Changes**


pkg/services/auth/auth.go
```diff
@@ -124,24 +120,29 @@ func (a *authSvc) CreateGitClient(ctx context.Context, repoUrl gitproviders.Repo
 	return git.New(pubKey, wrapper.NewGoGit()), nil
 }
 
-// setupDeployKey creates a git.Git client instrumented with existing or generated deploy keys.
+// SetupDeployKey creates a git.Git client instrumented with existing or generated deploy keys.
 // This ensures that git operations are done with stored deploy keys instead of a user's local ssh-agent or equivalent.
-func (a *authSvc) setupDeployKey(ctx context.Context, name SecretName, targetName string, repo gitproviders.RepoURL) (*ssh.PublicKeys, error) {
+func (a *authSvc) SetupDeployKey(ctx context.Context, namespace string, repo gitproviders.RepoURL) (*ssh.PublicKeys, error) {
```
jpellizzari, line 125: Wasn't this changed in another branch? I have the feeling that some stuff from other branches made it onto this PR.

josecordaz: I'm can't tell if there are any other branches with this change. I intentionally changed this 🙂

pkg/models/manifest.go
```diff
@@ -92,6 +92,16 @@ func BootstrapManifests(fluxClient flux.Flux, clusterName string, namespace stri
 		return nil, fmt.Errorf("failed marshalling wego config: %w", err)
 	}
 
+	configBranch, err := gitProvider.GetDefaultBranch(ctx, configURL)
+	if err != nil {
+		return nil, err
+	}
+
+	sourceManifest, err := GetSourceManifest(ctx, fluxClient, gitProvider, clusterName, namespace, configURL, configBranch)
```
jpellizzari, line 100: Does this mean that this manifest isn't being applied on `main`?

jpellizzari (approved):
Having a tough time understanding exactly what is happening here, but I assume this was cleanup from the previous refactoring PR @josecordaz ?

Code LGTM

pkg/models/manifest.go
josecordaz: it is applied but not by using this function (which is the ideal scenario).

PR #1321, opened 2022-01-19
Adds Sorting to Data Table

joshri:
Step one in adding filtering to the table is getting the sorting properly implemented. Use `npm run storybook` to check out the component with data. 

image


jpellizzari (commented):
Code looks good. How about some tests to ensure that the sorting is working?

jpellizzari (approved):
Excellent work on the tests!

PR #1328, opened 2022-01-21
Add missing roles for the helm watcher

### LINKED ISSUE #1327: Fix missing cluster roles and service account for the helm watcher
I have some missing cluster roles that the watcher requires and I failed to add.

Skarlso:
Closes #1327 

~~TODO: Add acceptance test for cross namespace access.~~


manifests/helm-watcher/helm_watcher_role.yaml.tpl
```diff
@@ -0,0 +1,30 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: helm-watcher-role
+  namespace: {{ .Namespace }}
+rules:
+  - apiGroups:
+      - helm.watcher
```
aclevername, line 8: shouldn't this be `source.toolkit.fluxcd.io`?

manifests/helm-watcher/helm_watcher_service_account.yaml.tpl
```diff
@@ -0,0 +1,5 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: helm-watcher
+  namespace: {{ .Namespace }}
```
aclevername, line 5: do we need another service account? Can we not just bind to the existing one?

manifests/helm-watcher/helm_watcher_role.yaml.tpl
Skarlso: nope

manifests/helm-watcher/helm_watcher_service_account.yaml.tpl
Skarlso: yes, nope

Skarlso: :trollface:

aclevername: wait what :joy:

manifests/manifests.go
```diff
@@ -25,9 +28,28 @@ type Params struct {
...
+// GenerateHelmWatcherManifests generates helm-watcher manifests from embedded template files.
+func GenerateHelmWatcherManifests(params Params) ([][]byte, error) {
+	hwTemplates, err := processTemplates(params, helmWatcherTemplates, helmWatcherManifestsDir)
+	if err != nil {
+		return nil, fmt.Errorf("failed reading templates directory: %w", err)
+	}
+
+	return hwTemplates, nil
+}
+
```
aclevername, line 50: is it worth separating out if they are still 1 deployment in the end? might just be adding more confusion

manifests/helm-watcher/helm_watcher_service_account.yaml.tpl
Skarlso: Yeah, we do, because the other service account is different with different roles attached. Also, I don't really want to bind to it, it belongs to the running ui thing. And I don't think everyone wants that running maybe, or I don't know. I was reluctant to re-use that.

manifests/helm-watcher/helm_watcher_role.yaml.tpl
Skarlso: This is the apiGroup of the helm watcher which is defined as helm.watcher.... uh, somewhere. :D I remember defining it.

aclevername: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.22/#policyrule-v1-rbac-authorization-k8s-io
```
apiGroups
string array	APIGroups is the name of the APIGroup that contains the resources. If multiple API groups are specified, any action requested against one of the enumerated resources in any API group will be allowed.
```

doesn't the resource belong to the `source.toolkit.fluxcd.io` API group?

manifests/helm-watcher/helm_watcher_service_account.yaml.tpl
aclevername: but its the same affect in the end? The UI, app API and profiles API all run in the same deployment. Adding the extra service account just adds more yaml, has no change on the outcome?

manifests/manifests.go
Skarlso: I didn't want to do it together in case the ui is ever pulled out and things are separated into their own proper services. I'm not sure if, for example, the ui will be the only one starting this service. It feels like the wrong place to be honest in the first place.

aclevername (dismissed):
can we get the acceptance test updated to create the HelmRepo in a different namespace to wego, to help cover this issue?

manifests/helm-watcher/helm_watcher_role.yaml.tpl
Skarlso: The resource in this case is the watcher. It's the `resources` that is the important one.

aclevername: isn't the resource the `HelmRepository`? The same as we do for the regular `Role` for wego-app:
```
    resources: [ "helmreleases" ]
    verbs: [ "*" ]
```

https://github.com/weaveworks/weave-gitops/blob/main/manifests/wego-app/role.yaml.tpl

manifests/helm-watcher/helm_watcher_service_account.yaml.tpl
Skarlso: Technically, yes. I could even modify the wego-app yamls themselves and add extra permissions there.

But it felt like a sin. :D I'm reluctant, because what if we decide that the helm-watcher should be in its own deployment since it's almost a separate entity, unlike the profiles service?

manifests/helm-watcher/helm_watcher_role.yaml.tpl
Skarlso: nope, it's plural, `helmrepositories`. 

However, the API group might be different though. For now, that should be the wego-app, that's true.

manifests/helm-watcher/helm_watcher_service_account.yaml.tpl
aclevername: > But it felt like a sin. :D I'm reluctant, because what if we decide that the helm-watcher should be in its own deployment since it's almost a separate entity, unlike the profiles service?

then we still have to update the resources of `wego-app` to remove stuff in that case as the Binding/Deployment-spec will change anyway

Skarlso: Also! The helm-watcher needs cross namespace access. So I can't modify the existing role.

manifests/helm-watcher/helm_watcher_role.yaml.tpl
aclevername: sorry I mean't `helmrepositories`. I'm still pretty adamant that the API group isn't `helm.watcher` though. But I could be wrong still
```
  - apiGroups: ["source.toolkit.fluxcd.io"]
    resources: [ "helmrepositories" ]
    verbs: [ "*" ]
```

https://github.com/weaveworks/weave-gitops/blob/1fd04684d2f076e4068670d42ff11535e7db9f0e/manifests/wego-app/role.yaml.tpl#L19-L21

Skarlso: Ah. Right. Yes. True

manifests/helm-watcher/helm_watcher_service_account.yaml.tpl
aclevername: yeah. I think just adding the `role` and `rolebinding` to `wego-app` is nicer than a new `helm-watcher` dir with `role`, `rolebinding` and `serviceaccount`. Also baring in mind that you still have to update the `deployment` in `wego-apps` to reference the `serviceaccount`, so it becomes overlapping already

manifests/helm-watcher/helm_watcher_role.yaml.tpl
bigkevmcd: @aclevername is right, the watcher requires permissions on objects to watch...so, you're saying that it has permissions to watch watch objects in the "helm.watcher" API group...which isn't a thing...

(Even if it was, it needs permissions to watch flux objects)

manifests/helm-watcher/helm_watcher_service_account.yaml.tpl
Skarlso: Discussion outcome: I'll pull these together into the `wego-app` folder so I don't have to have separate functions.

manifests/helm-watcher/helm_watcher_role.yaml.tpl
Skarlso: @bigkevmcd yeah we discussed this. I was still in the land where I generated my sample project using kubebuilder. :) I had the wrong context in my head.

manifests/wego-app/helm_watcher_role.yaml.tpl
```diff
@@ -0,0 +1,30 @@
...
+      - delete
+      - get
+      - patch
+      - update
+  - apiGroups:
+      - source.toolkit.fluxcd.io
+    resources:
+      - helmrepositories/status
+    verbs:
+      - get
```
aclevername, line 30: is this covered by the below?
```
  - apiGroups:
      - source.toolkit.fluxcd.io
    resources:
      - helmrepositories
    verbs:
      - get
      - list
      - watch
```

aclevername: like `helmrepositories/status` resource is a sub resource of `helmrepositories`

Skarlso: As far as I remember it doesn't propagate. Subresources are used with the exact API. If you don't specify them, you don't have the same access. I mean, you can use something like `*` to grant access to all resources, but I don't want that. I only need get and only for specific resources.

Skarlso:
Uh, this started two wego-apps. :D

Skarlso:
# Testing

```log
{"level":"info","ts":1642777817.7781508,"logger":"controller.helmrepository","msg":"starting put operation","reconciler group":"source.toolkit.fluxcd.io","reconciler kind":"HelmRepository","name":"podinfo","namespace":"wego-system","repository":"wego-system/podinfo"}
{"level":"info","ts":1642777817.7794151,"logger":"controller.helmrepository","msg":"finished put operation","reconciler group":"source.toolkit.fluxcd.io","reconciler kind":"HelmRepository","name":"podinfo","namespace":"wego-system","repository":"wego-system/podinfo"}
{"level":"info","ts":1642777817.7794402,"logger":"controller.helmrepository","msg":"cached data from repository","reconciler group":"source.toolkit.fluxcd.io","reconciler kind":"HelmRepository","name":"podinfo","namespace":"wego-system","repository":"wego-system/podinfo","url":"http://source-controller.wego-system.svc.cluster.local./helmrepository/wego-system/podinfo/index.yaml","name":"podinfo","number of profiles":1}
{"level":"info","ts":1642777817.789897,"logger":"controller.helmrepository","msg":"found the repository: ","reconciler group":"source.toolkit.fluxcd.io","reconciler kind":"HelmRepository","name":"cross-test","namespace":"cross-test","repository":"cross-test/cross-test","name":"cross-test"}
{"level":"info","ts":1642777818.185261,"logger":"controller.helmrepository","msg":"starting put operation","reconciler group":"source.toolkit.fluxcd.io","reconciler kind":"HelmRepository","name":"cross-test","namespace":"cross-test","repository":"cross-test/cross-test"}
{"level":"info","ts":1642777818.1867452,"logger":"controller.helmrepository","msg":"finished put operation","reconciler group":"source.toolkit.fluxcd.io","reconciler kind":"HelmRepository","name":"cross-test","namespace":"cross-test","repository":"cross-test/cross-test"}
{"level":"info","ts":1642777818.186769,"logger":"controller.helmrepository","msg":"cached data from repository","reconciler group":"source.toolkit.fluxcd.io","reconciler kind":"HelmRepository","name":"cross-test","namespace":"cross-test","repository":"cross-test/cross-test","url":"http://source-controller.wego-system.svc.cluster.local./helmrepository/cross-test/cross-test/index.yaml","name":"cross-test","number of profiles":1}

```

Seems to work fine now... implementing the acceptance test next.

manifests/wego-app/helm_watcher_role.yaml.tpl
aclevername: :ok_hand: makes sense

Skarlso:
Finally tested this in cluster. The above test run was a dev run. It doesn't work yet sadly. 😂

Skarlso:
Fiiixed :) 

```
time="2022-01-22T15:17:17Z" level=info msg="Serving on port 9001"
{"level":"info","ts":1642864638.228255,"logger":"controller-runtime.metrics","msg":"metrics server is starting to listen","addr":":9980"}
{"level":"info","ts":1642864638.2285573,"logger":"setup","msg":"starting manager"}
{"level":"info","ts":1642864638.229215,"msg":"starting metrics server","path":"/metrics"}
{"level":"info","ts":1642864638.2295551,"logger":"controller.helmrepository","msg":"Starting EventSource","reconciler group":"source.toolkit.fluxcd.io","reconciler kind":"HelmRepository","source":"kind source: /, Kind="}
{"level":"info","ts":1642864638.2299635,"logger":"controller.helmrepository","msg":"Starting Controller","reconciler group":"source.toolkit.fluxcd.io","reconciler kind":"HelmRepository"}
{"level":"info","ts":1642864638.3306725,"logger":"controller.helmrepository","msg":"Starting workers","reconciler group":"source.toolkit.fluxcd.io","reconciler kind":"HelmRepository","worker count":1}
{"level":"info","ts":1642864638.3429396,"logger":"controller.helmrepository","msg":"found the repository: ","reconciler group":"source.toolkit.fluxcd.io","reconciler kind":"HelmRepository","name":"podinfo","namespace":"wego-system","repository":"wego-system/podinfo","name":"podinfo"}
{"level":"info","ts":1642864638.90731,"logger":"controller.helmrepository","msg":"starting put operation","reconciler group":"source.toolkit.fluxcd.io","reconciler kind":"HelmRepository","name":"podinfo","namespace":"wego-system","repository":"wego-system/podinfo"}
{"level":"info","ts":1642864638.9083583,"logger":"controller.helmrepository","msg":"finished put operation","reconciler group":"source.toolkit.fluxcd.io","reconciler kind":"HelmRepository","name":"podinfo","namespace":"wego-system","repository":"wego-system/podinfo"}
{"level":"info","ts":1642864638.9084306,"logger":"controller.helmrepository","msg":"cached data from repository","reconciler group":"source.toolkit.fluxcd.io","reconciler kind":"HelmRepository","name":"podinfo","namespace":"wego-system","repository":"wego-system/podinfo","url":"http://source-controller.wego-system.svc.cluster.local./helmrepository/wego-system/podinfo/index.yaml","name":"podinfo","number of profiles":1}
{"level":"info","ts":1642864689.9879515,"logger":"controller.helmrepository","msg":"found the repository: ","reconciler group":"source.toolkit.fluxcd.io","reconciler kind":"HelmRepository","name":"cross-test","namespace":"cross-test","repository":"cross-test/cross-test","name":"cross-test"}
{"level":"info","ts":1642864690.431456,"logger":"controller.helmrepository","msg":"starting put operation","reconciler group":"source.toolkit.fluxcd.io","reconciler kind":"HelmRepository","name":"cross-test","namespace":"cross-test","repository":"cross-test/cross-test"}
{"level":"info","ts":1642864690.4392412,"logger":"controller.helmrepository","msg":"finished put operation","reconciler group":"source.toolkit.fluxcd.io","reconciler kind":"HelmRepository","name":"cross-test","namespace":"cross-test","repository":"cross-test/cross-test"}
{"level":"info","ts":1642864690.4394677,"logger":"controller.helmrepository","msg":"cached data from repository","reconciler group":"source.toolkit.fluxcd.io","reconciler kind":"HelmRepository","name":"cross-test","namespace":"cross-test","repository":"cross-test/cross-test","url":"http://source-controller.wego-system.svc.cluster.local./helmrepository/cross-test/cross-test/index.yaml","name":"cross-test","number of profiles":1}
```

I was missing patch verb. :D

manifests/wego-app/helm_watcher_role.yaml.tpl
```diff
@@ -0,0 +1,31 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: helm-watcher-role
```
aclevername, line 4: is this too ambiguous? Since its not namespaced maybe worth putting `wego-` prefix? Idk I don't feel that strongly but eh :shrug:

pkg/services/gitops/uninstall.go
```diff
@@ -4,10 +4,11 @@ import (
 	"context"
 	"fmt"
 
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+
 	"github.com/weaveworks/weave-gitops/cmd/gitops/version"
 	"github.com/weaveworks/weave-gitops/manifests"
 	"github.com/weaveworks/weave-gitops/pkg/kube"
-	apierrors "k8s.io/apimachinery/pkg/api/errors"
```
aclevername, line 10: gergely footprint :laughing:

test/acceptance/test/profiles_test.go
```diff
@@ -104,6 +103,26 @@ podinfo	Podinfo Helm chart for Kubernetes	6.0.0,6.0.1
 		Expect(err).NotTo(HaveOccurred())
 		Expect(string(values)).To(ContainSubstring("# Default values for podinfo"))
 	})
+	It("profiles are installed into a different namespace", func() {
```
aclevername, line 106: can we just make this the default behaviour of the above test, rather than a new test? We don't need to test same-namespace works if cross-namespace works

aclevername (approved):
lgtm couple nits :smile:

test/acceptance/test/profiles_test.go
Skarlso: Does that work? I mean, I'm not sure, I have to look at the test code, but it installs wego in one namespace, but the install requires the helmRepository in order for getProfiels to work to see the helmrepository. `GetProfiles` still needs a namespace and repo to work, and we aren't providing that as a setup value right now. We haven't yet made it work with all HelmRepositories...

Skarlso: Yeah, that won't work, because the install sets it to that namespace + `weavework-charts` so if I run this with the other namespace it fails since there are no profiles because it's not in the expected namespace.

For now, this is fine, once we make all of this work without the need of specifying the namespace and the name of the helmrepository, we can update this test to no longer care. :)

aclevername: :+1:

PR #1330, opened 2022-01-21
Saving config

luizbafilho:
Closes: 

**What changed?**

**Why?**

**How did you test it?**

**Release notes**

**Documentation Changes**

### DIFF
--- a/cmd/gitops/add/app/cmd.go
+++ b/cmd/gitops/add/app/cmd.go
@@ -112,6 +112,20 @@ func runCmd(cmd *cobra.Command, args []string) error {
 		return fmt.Errorf("failed to create app service: %w", err)
 	}
 
+	kubeClient, err := factory.GetKubeService()
+	if err != nil {
+		return fmt.Errorf("failed getting kube service: %w", err)
+	}
+
+	wegoConfig, err := kubeClient.GetWegoConfig(ctx, params.Namespace)
+	if err != nil {
+		return fmt.Errorf("failed getting wego config")
+	}
+
+	if params.ConfigRepo == "" {
+		params.ConfigRepo = wegoConfig.ConfigRepo
+	}
+
 	gitClient, gitProvider, err := factory.GetGitClients(ctx, providerClient, services.GitConfigParams{
 		URL:              params.Url,
 		ConfigRepo:       params.ConfigRepo,
--- a/cmd/gitops/install/cmd.go
+++ b/cmd/gitops/install/cmd.go
@@ -6,6 +6,7 @@ package install
 import (
 	"context"
 	_ "embed"
+	"errors"
 	"fmt"
 	"os"
 	"strings"
@@ -89,7 +90,19 @@ func installRunCmd(cmd *cobra.Command, args []string) error {
 	}
 
 	if installParams.DryRun {
-		manifests, err := models.BootstrapManifests(fluxClient, clusterName, namespace, configURL)
+		fluxNamespace, err := kubeClient.FetchNamespaceWithLabel(ctx, flux.PartOfLabelKey, flux.PartOfLabelValue)
+		if err != nil {
+			if !errors.Is(err, kube.ErrNamespaceNotFound) {
+				return fmt.Errorf("failed fetching flux namespace: %w", err)
+			}
+		}
+
+		manifests, err := models.BootstrapManifests(fluxClient, models.BootstrapManifestsParams{
+			ClusterName:   clusterName,
+			WegoNamespace: namespace,
+			FluxNamespace: fluxNamespace.Name,
+			ConfigRepo:    configURL,
+		})
 		if err != nil {
 			return fmt.Errorf("failed getting gitops manifests: %w", err)
 		}
--- a/cmd/gitops/upgrade/cmd.go
+++ b/cmd/gitops/upgrade/cmd.go
@@ -36,7 +36,6 @@ var Cmd = &cobra.Command{
 }
 
 func init() {
-	Cmd.PersistentFlags().StringVar(&upgradeCmdFlags.ConfigRepo, "config-repo", "", "URL of external repository that will hold automation manifests")
 	Cmd.PersistentFlags().StringVar(&upgradeCmdFlags.Version, "version", "", "Version of Weave GitOps Enterprise to be installed")
 	Cmd.PersistentFlags().StringVar(&upgradeCmdFlags.BaseBranch, "base", "", "The base branch to open the pull request against")
 	Cmd.PersistentFlags().StringVar(&upgradeCmdFlags.HeadBranch, "branch", "tier-upgrade-enterprise", "The branch to create the pull request from")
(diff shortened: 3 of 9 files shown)


pkg/models/manifest_test.go
```diff
@@ -23,208 +23,221 @@ var _ = Describe("Installer", func() {
...
-						Content: strconv.AppendQuote(nil, "# keep"),
-					},
-				}
-
-				for ind, manifest := range manifestsFiles[6:] {
-					Expect(manifest.Path).Should(Equal(expectedManifests[ind].Path))
-					Expect(string(manifest.Content)).Should(Equal(string(expectedManifests[ind].Content)))
-				}
+			Context("success case", func() {
+				It("should pass successfully", func() {
```
jpellizzari, line 198: What is up with these whitespace diffs?

cmd/gitops/add/app/cmd.go
```diff
@@ -118,6 +118,15 @@ func runCmd(cmd *cobra.Command, args []string) error {
 		return fmt.Errorf("failed to create app service: %w", err)
 	}
 
+	wegoConfig, err := kubeClient.GetWegoConfig(ctx, params.Namespace)
+	if err != nil {
+		return fmt.Errorf("failed getting wego config")
+	}
+
+	if params.ConfigRepo == "" {
+		params.ConfigRepo = wegoConfig.ConfigRepo
```
jpellizzari, line 127: Will this logic need to be repeated on API calls as well? Maybe this should be within the `.Add()` func?

J-Thompson12, line 127: I was under the impression that we were getting rid of params.ConfigRepo in favor of only pulling it from the wego config. Do we have coverage for when the params and the wegoConfig do not match?

jpellizzari (commented):
Only one blocker, then LGTM

pkg/models/manifest.go
```diff
@@ -46,9 +46,16 @@ const (
 	WegoConfigMapName = "weave-gitops-config"
 )
 
+type BootstrapManifestsParams struct {
+	ClusterName   string
+	WegoNamespace string
+	FluxNamespace string
+	ConfigRepo    gitproviders.RepoURL
+}
```
J-Thompson12, line 54: 👍 I like this approach.

J-Thompson12 (commented):
Just one question.

pkg/models/manifest.go
```diff
@@ -94,63 +101,66 @@ func BootstrapManifests(fluxClient flux.Flux, clusterName string, namespace stri
...
+	FluxClient    flux.Flux
+	GitProvider   gitproviders.GitProvider
+	ClusterName   string
+	WegoNamespace string
+	ConfigRepo    gitproviders.RepoURL
+}
 
-	systemKustomization := CreateKustomization(clusterName, namespace, RuntimePath, SourcePath, SystemKustResourcePath, UserKustResourcePath, WegoAppPath)
+// GitopsManifests generates all yaml files that are going to be written in the config repo
+func GitopsManifests(ctx context.Context, bootstrapManifests []Manifest, params GitopsManifestsParams) ([]Manifest, error) {
```
josecordaz, line 139: small suggestion. I think we actually don't need a function that returns all the manifests at this point. I'm thinking that you probably added `bootstrapManifests` field here so we don't call BootstrapManifests twice in the Install function, which is great (Should have seen that 🤔 ). If that was you intention. I would suggest to just remove `bootstrapManifests` parameter here and just return the 3 manifests generated( `Source`, `Kustomization` and `Keep`) (and change the name of the function). Then we can just append the result of this function to the previous result from calling BootstrapManifests in `Install` method.

cmd/gitops/install/cmd.go
```diff
@@ -84,7 +85,25 @@ func installRunCmd(cmd *cobra.Command, args []string) error {
...
-		manifests, err := models.BootstrapManifests(fluxClient, clusterName, namespace, configURL)
+		fluxNamespace, err := kubeClient.FetchNamespaceWithLabel(ctx, flux.PartOfLabelKey, flux.PartOfLabelValue)
+		if err != nil {
+			if !errors.Is(err, kube.ErrNamespaceNotFound) {
+				return fmt.Errorf("failed fetching flux namespace: %w", err)
+			}
+		}
+
+		if fluxNamespace == nil {
+			fluxNamespace = &corev1.Namespace{ObjectMeta: metav1.ObjectMeta{
```
josecordaz, line 96: I wonder if we really need this whole object. I only see we are using `fluxNamespace.Name` after this line 🤔

cmd/gitops/upgrade/cmd.go
```diff
@@ -70,6 +68,15 @@ func upgradeCmdRunE() func(*cobra.Command, []string) error {
 		fluxClient := flux.New(osys.New(), &runner.CLIRunner{})
 		factory := services.NewFactory(fluxClient, log)
 
+		wegoConfig, err := kubeClient.GetWegoConfig(ctx, namespace)
+		if err != nil {
+			return fmt.Errorf("failed getting wego config")
+		}
+
+		if upgradeCmdFlags.ConfigRepo == "" {
```
josecordaz, line 76: Probably we can remove this if statement as there is no config-repo flag for this command.

pkg/services/gitops/uninstall.go
```diff
@@ -8,6 +8,7 @@ import (
 
 	"github.com/weaveworks/weave-gitops/cmd/gitops/version"
 	"github.com/weaveworks/weave-gitops/manifests"
+
```
josecordaz, line 11: Let's group all `weave-gitops` imports together.

pkg/services/install/install.go
```diff
@@ -57,7 +59,25 @@ func (i *install) Install(namespace string, configURL gitproviders.RepoURL, auto
 		return fmt.Errorf("failed installing flux: %w", err)
 	}
 
-	manifests, err := models.BootstrapManifests(i.fluxClient, clusterName, namespace, configURL)
+	fluxNamespace, err := i.kubeClient.FetchNamespaceWithLabel(ctx, flux.PartOfLabelKey, flux.PartOfLabelValue)
```
josecordaz, line 62: This block from line 62 to 80 seems duplicate from install/cmd.go. I wonder if there is a way we could avoid it. Maybe by passing in the kube client as parameter to `BoostrapManifests` ?

pkg/models/manifest.go
josecordaz: No worries to much about this, I have a pending PR where I'm modifying some logic in install/cmd.go. We can address this there, once I start fixing conflicts after you merge this (I know you had a lot of conflicts to fix).

luizbafilho: that was exactly the case, yes please. 😄

cmd/gitops/add/app/cmd.go
luizbafilho: yeah, I think we should just remove that params like in the other methods, it doesn't not make sense

josecordaz (approved):
LGTM

J-Thompson12 (approved):
LGTM as well

jpellizzari (approved):
Assuming we use the same code path on the server and CLI :+1:

PR #1331, opened 2022-01-21
Support HTTPs Installations (copy)

jpellizzari:
Copied and adapted from https://github.com/weaveworks/weave-gitops/pull/1276

I resolved the merge conflicts in the original PR, as well as removed a noisy acceptance test case.

### DIFF
--- a/cmd/gitops/install/cmd.go
+++ b/cmd/gitops/install/cmd.go
@@ -25,18 +25,13 @@ import (
 	"github.com/weaveworks/weave-gitops/pkg/services/applier"
 	"github.com/weaveworks/weave-gitops/pkg/services/auth"
 	"github.com/weaveworks/weave-gitops/pkg/services/automation"
+	"github.com/weaveworks/weave-gitops/pkg/services/gitops"
 	"github.com/weaveworks/weave-gitops/pkg/services/gitopswriter"
 	"github.com/weaveworks/weave-gitops/pkg/services/gitrepo"
 )
 
-type params struct {
-	DryRun     bool
-	AutoMerge  bool
-	ConfigRepo string
-}
-
 var (
-	installParams params
+	gitopsParams gitops.InstallParams
 )
 
 var Cmd = &cobra.Command{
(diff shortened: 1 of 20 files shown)


cmd/gitops/install/cmd.go
```diff
@@ -123,10 +121,12 @@ func installRunCmd(cmd *cobra.Command, args []string) error {
...
-			URL:        installParams.ConfigRepo,
-			ConfigRepo: installParams.ConfigRepo,
-			Namespace:  namespace,
-			DryRun:     installParams.DryRun,
+			URL:               gitopsParams.ConfigRepo,
+			ConfigRepo:        gitopsParams.ConfigRepo,
+			Namespace:         namespace,
+			DryRun:            gitopsParams.DryRun,
+			FluxHTTPSUsername: gitopsParams.FluxHTTPSUsername,
+			FluxHTTPSPassword: gitopsParams.FluxHTTPSPassword,
```
bigkevmcd, line 129: In my branch to address the conflicts, I had shifted this to passing in a `*flux.HTTPSCreds` rather than the two fields.

pkg/services/automation/generator.go
```diff
@@ -132,7 +132,8 @@ func (a *AutomationGen) generateAppSource(ctx context.Context, app models.Applic
 
 	switch app.SourceType {
 	case models.SourceTypeGit:
-		source, err = a.Flux.CreateSourceGit(app.Name, app.GitSourceURL, app.Branch, appSecretRef.String(), app.Namespace)
+		// Passes nil in as the credentials as we don't have anything.
+		source, err = a.Flux.CreateSourceGit(app.Name, app.GitSourceURL, app.Branch, appSecretRef.String(), app.Namespace, nil)
```
bigkevmcd, line 136: I don't think this is correct?

pkg/services/factory.go
```diff
@@ -29,14 +29,20 @@ type Factory interface {
...
-	ConfigRepo       string
-	Namespace        string
-	IsHelmRepository bool
-	DryRun           bool
+	URL               string
+	ConfigRepo        string
+	Namespace         string
+	IsHelmRepository  bool
+	DryRun            bool
+	FluxHTTPSUsername string
```
bigkevmcd, line 40: See above for comment about using a `*flux.HTTPSCreds`

pkg/services/factory.go
```diff
@@ -137,11 +143,16 @@ func (f *defaultFactory) GetGitClients(ctx context.Context, gpClient gitprovider
...
 	}
 
-	var appClient, configClient git.Git
+	var httpsCreds *flux.HTTPSCreds
+	if params.FluxHTTPSUsername != "" && params.FluxHTTPSPassword != "" {
+		httpsCreds = &flux.HTTPSCreds{Username: params.FluxHTTPSUsername, Password: params.FluxHTTPSPassword}
+	}
 
+	var appClient, configClient git.Git
+	// TODO: KEVIN!!!! Do we need to make these flux.HTTPSCreds?
```
bigkevmcd, line 152: I think this comment can come out now :-)

pkg/services/gitops/install.go
```diff
@@ -292,3 +291,15 @@ func (g *Gitops) saveWegoConfig(ctx context.Context, params InstallParams) (*cor
 
 	return cm, nil
 }
+
+func credsFromParams(p InstallParams) *flux.HTTPSCreds {
+	var creds *flux.HTTPSCreds
```
bigkevmcd, line 296: You could simplify this to just return nil if either string is empty, or the pointer if not.

See https://go.dev/doc/effective_go#control-structures for some guidance on this.

bigkevmcd (changes requested):
As far as I can see this code doesn't work.

Since the changes to support cluster, it needs a significant rework to ensure that the cluster setup is authenticated.,

jpellizzari:
Closing per @JamWils

PR #1294, opened 2022-01-12
Change to Snyk monitor.

bigkevmcd:
Closes: 

**What changed?**
Change to push data to Snyk.

**Why?**

**How did you test it?**

**Release notes**

**Documentation Changes**

PR #1333, opened 2022-01-21
Enable code scanning upload on failure

palemtnrider:
Closes: 

**What changed?**
Updating the snyk action to upload results even on failure.  

**Why?**
The only output is the log from the action.  Having it available in the security tab will be useful. This upload enables us to triage the problem and either fix it or mark as a false positive.

**How did you test it?**
Ran the action

### DIFF
--- a/.github/workflows/scan.yaml
+++ b/.github/workflows/scan.yaml
@@ -41,6 +41,7 @@ jobs:
         run: rm package-lock.json && rm package.json && make cmd/gitops/ui/run/dist/index.html
       - name: Run Snyk to check for vulnerabilities
         uses: snyk/actions/golang@master
+        continue-on-error: true
         env:
           SNYK_TOKEN: ${{ secrets.SNYK_API_TOKEN }}
         with:


jpellizzari (approved):
Green :heavy_check_mark: looks good to me! I wish we could get rid of these license warnings, but I suspect that is a different issue.

mnowster:
Why are we uploading to the tab and not back to snyk itself as per https://github.com/weaveworks/weave-gitops/pull/1294

Doesn't this give a false positive of hey snyk says green?

palemtnrider:
> Why are we uploading to the tab and not back to snyk itself as per #1294
> 
> Doesn't this give a false positive of hey snyk says green?

@mnowster Sort of.  The job doesn't fail, but the files tab shows the failed checks giving us the ability to review them here.

palemtnrider:
@mnowster a bit more details. that test created a duplicate weaveworks/weave-gitops project.  I was able to go into snyk and update those checks.

mnowster (changes requested):
🙅🏻‍♀️  I don't agree with this approach. 

I would rather we change the licence severity in the config in snyk, so that it correctly shows fail mark on a PR if something is introduced. 
I think that this would be better if it uses the monitor command so it reports the data back to snyk, so it's able to be emailed to a group of people and reported on.

palemtnrider:
Closing.  The code will fail if new dependencies don't pass the snyk scan.  However, we will still upload the scan results on a successful run - i.e. this PR doesn't address that existing behavior.  That should be addressed in a new PR - (similar #1294 )

Issue #1323, opened 2022-01-20
Add deploy key only when needed


J-Thompson12:
The issue #1278 reported a bug where we are unable to install or add a repo if the config repo is public because there is no deploy key. This causes wego to be unable to clone the repo. #1301 is a temporary fix to allow us to do a release and be able to deploy. That fix is not the desired fix. We need to find a better interface to handle if a deploy key should be added or not. RepoUrl is not the right place.

Acceptance criteria:
- Deploy key is only added when it is needed like on install or for private repos
- The quick fix #1301 has been removed

rokshana-b:
﻿
![Screen Shot 2022-01-27 at 23 00 43](https://user-images.githubusercontent.com/76788247/151841127-cb7d75aa-45fb-4e25-a261-541676811b6f.png)
![Screen Shot 2022-01-27 at 23 03 04](https://user-images.githubusercontent.com/76788247/151841128-a18fe485-219f-4db0-91f7-59780d0f647d.png)
![Screen Shot 2022-01-27 at 23 10 14](https://user-images.githubusercontent.com/76788247/151841129-66717844-2ed5-47b6-a135-b1127cd4c458.png)

rokshana-b:
@J-Thompson12 ^^

PR #1334, opened 2022-01-21
Only add deploy key when needed

J-Thompson12:
Closes: #1323 

**What changed?**
Refactored deploy key code and removed quick fix. This adds a deploy key on install for a config repo and then on add it only adds a deploy key if the repo is private.

**Why?**

**How did you test it?**
Manually and current tests cover it

**Release notes**

**Documentation Changes**

### DIFF
--- a/pkg/services/factory.go
+++ b/pkg/services/factory.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 
+	"github.com/fluxcd/go-git-providers/gitprovider"
 	wego "github.com/weaveworks/weave-gitops/api/v1alpha1"
 	"github.com/weaveworks/weave-gitops/pkg/flux"
 	"github.com/weaveworks/weave-gitops/pkg/git"
@@ -13,6 +14,7 @@ import (
 	"github.com/weaveworks/weave-gitops/pkg/osys"
 	"github.com/weaveworks/weave-gitops/pkg/services/app"
 	"github.com/weaveworks/weave-gitops/pkg/services/auth"
+	"github.com/weaveworks/weave-gitops/pkg/services/automation"
 )
 
 //go:generate go run github.com/maxbrunsfeld/counterfeiter/v6 -generate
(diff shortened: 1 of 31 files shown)


pkg/services/factory.go
```diff
@@ -104,22 +105,14 @@ func (f *defaultFactory) GetKubeService() (kube.Kube, error) {
...
-
-	var providerUrl string
-
-	switch {
-	case !params.IsHelmRepository:
-		providerUrl = params.URL
-	case isExternalConfig:
-		providerUrl = params.ConfigRepo
-	default:
-		return nil, nil, nil
```
J-Thompson12, line 117: I am not really sure when we will ever hit this default option. I believe with the new directory structure we will always have a config-repo. I also deleted the two tests associated with this.

pkg/services/factory.go
```diff
@@ -104,22 +105,14 @@ func (f *defaultFactory) GetKubeService() (kube.Kube, error) {
...
+	// This is temporary. When config repo is always present in the config map this can be removed.
```
jpellizzari, line 108: :+1: for the comment

pkg/services/factory.go
```diff
@@ -132,40 +125,44 @@ func (f *defaultFactory) GetGitClients(ctx context.Context, gpClient gitprovider
...
+		repoVisibility, err := provider.GetRepoVisibility(ctx, normalizedUrl)
 		if err != nil {
-			return nil, nil, fmt.Errorf("error normalizing url: %w", err)
+			return nil, nil, fmt.Errorf("error getting repo visibility: %w", err)
 		}
 
-		configRepoClient, configRepoErr := authSvc.CreateGitClient(ctx, normalizedConfigRepo, targetName, params.Namespace, params.DryRun)
-		if configRepoErr != nil {
-			return nil, nil, configRepoErr
+		if *repoVisibility == gitprovider.RepositoryVisibilityPrivate {
```
jpellizzari, line 147: A comment here about why we are doing this would be helpful. I try to leave comments wherever a non-obvious `if` statement appears.

pkg/services/factory.go
```diff
@@ -104,22 +105,14 @@ func (f *defaultFactory) GetKubeService() (kube.Kube, error) {
 }
 
 func (f *defaultFactory) GetGitClients(ctx context.Context, gpClient gitproviders.Client, params GitConfigParams) (git.Git, gitproviders.GitProvider, error) {
-	isExternalConfig := models.IsExternalConfigRepo(params.ConfigRepo)
```
jpellizzari, line 107: :question: Do we use this factory on the CLI?

jpellizzari (commented):
One question:

pkg/services/factory.go
J-Thompson12: Yeah we do use this on the cli.

pkg/services/factory.go
```diff
@@ -104,22 +105,14 @@ func (f *defaultFactory) GetKubeService() (kube.Kube, error) {
...
-	switch {
-	case !params.IsHelmRepository:
-		providerUrl = params.URL
-	case isExternalConfig:
-		providerUrl = params.ConfigRepo
-	default:
-		return nil, nil, nil
+	// This is temporary. When config repo is always present in the config map this can be removed.
+	if params.ConfigRepo == "" && !params.IsHelmRepository {
+		params.ConfigRepo = params.URL
```
josecordaz, line 110: At this point we might want to get rid of the field `ConfigRepo` and just leave URL.

J-Thompson12: That is only in the case someone has the same config repo as the app they are adding. So running `gitops add app .` would hit this.

pkg/services/factory.go
jpellizzari: @J-Thompson12 Why only for a private repo?

pkg/services/factory.go
```diff
@@ -132,40 +125,44 @@ func (f *defaultFactory) GetGitClients(ctx context.Context, gpClient gitprovider
...
-		configRepoClient, configRepoErr := authSvc.CreateGitClient(ctx, normalizedConfigRepo, targetName, params.Namespace, params.DryRun)
-		if configRepoErr != nil {
-			return nil, nil, configRepoErr
+		if *repoVisibility == gitprovider.RepositoryVisibilityPrivate {
+			secretName := auth.SecretName{
+				Name:      models.CreateRepoSecretName(normalizedUrl),
+				Namespace: params.Namespace,
+			}
+
+			_, err = authSvc.SetupDeployKey(ctx, secretName, targetName, normalizedUrl)
```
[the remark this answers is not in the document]
J-Thompson12: Yes thats exactly what is happening now.

luizbafilho (approved):
lgtm after addressing other folks comments

jpellizzari (approved):

Issue #1057, opened 2021-11-10
Add example using --name to the add app help


sympatheticmoose:
To help mitigate errors such as https://github.com/weaveworks/weave-gitops/issues/1045 I propose we add something like this to the help text for `add app`:

```
Examples:
...

# Add an application from a directory within a repository and provide a unique name
gitops add app --url git@github.com:myorg/podinfo --name "application-a" --path "/folder"

Flags:
...
--name     Name of an application; defaults to application repository name. Must be unique.
```

Would appreciate assistance in describing the extent to which it must be unique.

sympatheticmoose:
(yes there are a ludicrous number of edits due to me second guessing whether `an unique` or `a unique` is correct)

sympatheticmoose:
@palemtnrider - assigned you the docs issue, adding you here as well as probably makes sense to tackle together

sympatheticmoose:
worth noting we have 3 different issues where help text for add app could be improved, others are #1205 and #1207

palemtnrider:
How does this sound:

```
# Add an application from a directory within a repository and provide a unique name
gitops add app --url git@github.com:myorg/podinfo --name "application-a" --path "/folder"

Flags:
...
--name     Name of an application; defaulting to the application repository name. The 
           name must be unique in the config repo, consisting of up to 63 lowercase 
           alphanumeric characters and `-`, starting and ending with an alphanumeric 
           character.
```
Basically taking the text from https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-label-names

sympatheticmoose:
From conversation we will just end at unique in config repo and provide the additional guidance in docs. If we encounter issues as a result of long repo names, we will revisit how we want to handle - error/truncate/hash etc

palemtnrider:
`--name string                            Name of application; defaults to the application repository name if not specified; must be unique in the config repo.`

PR #1339, opened 2022-01-24
Validate app naming using IsDNS1123Label from k8s

palemtnrider:
Closes: #1057 

**What changed?**
Replaced the check for name length with a call to IsDNS1123Label().  Updated the help text for the --name parameter

**Why?**
 Ensure the application name can be used as a label value.

**How did you test it?**
Manually to review the help text.  Added new unit tests.

**Release notes**
Yes

**Documentation Changes**
Yes.  Separate PR for updating the docs.

### DIFF
--- a/pkg/models/manifest.go
+++ b/pkg/models/manifest.go
@@ -11,6 +11,7 @@ import (
 	"strings"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/validation"
 
 	corev1 "k8s.io/api/core/v1"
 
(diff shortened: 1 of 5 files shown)


pkg/models/manifest.go
```diff
@@ -224,6 +225,24 @@ func ApplicationNameTooLong(name string) bool {
 	return len(name) > MaxKubernetesResourceNameLength
```
josecordaz, line 225: Could we get rid of this function? It seems it won't be used anymore

josecordaz (approved):
LGTM, just tiny suggestion

pkg/models/manifest.go
palemtnrider: Great question.  It is still in use by hashNameIfTooLong() so I left it in.

luizbafilho (approved):

Issue #1332, opened 2022-01-21
K8s test environment causes test failures on a clean build


dhwthompson:
**Describe the bug**

There appears to be some kind of race condition that causes test failures in packages that rely on the Kubernetes test environment.

**Environment**
 - gitops: failing on the latest `main` (1fd04684d2f076e4068670d42ff11535e7db9f0e as of my latest checkout)
 - kubernetes:  nothing real, but the test versions report:
   - `etcd` version 3.4.9
   - `kube-apiserver` version 1.19.2
   - `kubectl` version `version.Info{Major:"1", Minor:"22", GitVersion:"v1.22.5", GitCommit:"5c99e2ac2ff9a3c549d9ca665e7bc05a3e18f07e", GitTreeState:"clean", BuildDate:"2021-12-16T08:38:33Z", GoVersion:"go1.16.12", Compiler:"gc", Platform:"darwin/arm64"}`

**Affects versions**

Looks like `0.6.1` is the latest version I’ve got.

**To Reproduce**

1. From the `weave-gitops` directory, run `go clean -cache ./... && go test -tags unittest -v ./pkg/server ./pkg/kube`

**Expected behavior**
The tests pass.

**Observed behavior**
Some of the tests fail with the following error:

```
[BeforeSuite] BeforeSuite
/Users/david/Projects/weaveworks/weave-gitops/pkg/server/suite_test.go:66

  Unexpected error:
      <*fmt.wrapError | 0x140003d3780>: {
          msg: "could not create controller manager: error listening on :8080: listen tcp :8080: bind: address already in use",
          err: <*fmt.wrapError | 0x140003d3760>{
              msg: "error listening on :8080: listen tcp :8080: bind: address already in use",
              err: <*net.OpError | 0x1400037f090>{
                  Op: "listen",
                  Net: "tcp",
                  Source: nil,
                  Addr: <*net.TCPAddr | 0x1400091b6e0>{IP: nil, Port: 8080, Zone: ""},
                  Err: <*os.SyscallError | 0x140003d3740>{
                      Syscall: "bind",
                      Err: <syscall.Errno>0x30,
                  },
              },
          },
      }
      could not create controller manager: error listening on :8080: listen tcp :8080: bind: address already in use
  occurred

  /Users/david/Projects/weaveworks/weave-gitops/pkg/server/suite_test.go:73
------------------------------
Panic [0.000 seconds]
[AfterSuite] AfterSuite
[...]

PR #1341, opened 2022-01-24
Fix flaky test by disabling Prometheus metrics

dhwthompson:
The test suites for some of our packages spin up parts of a Kubernetes cluster using the `envtest` library, so that we have something more solid than mocks to test against. One problem this was causing was that running tests for both the `kube` and `server` packages at once would cause one of those test suites to fail.

Side-note: it's possible to reproduce this problem a little more efficiently by spinning up a fake web server to sit on port 8080, for example using Python's built-in web server:

    python3 -m http.server 8080

After some digging, I found out that the reason for this is that both environments were trying to spin up a Prometheus metrics handler listening on port 8080. If they both tried to do this at once, one would fail, causing the entire test suite for that package to fail too (including some nil-pointer panics during the clean up, which might be worth tidying up in a future pull request).

Setting the metrics bind address to the string "0" disables metrics for the test clusters (the documentation for which is buried in [a comment in the Kubernetes docs][1]), which fixes this particular problem. Testing on my local machine, there didn't seem to be any other fixed port assignments cropping up to take its place.

[1]: https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.10.2/pkg/manager

Closes: #1332

### DIFF
--- a/pkg/testutils/testutils.go
+++ b/pkg/testutils/testutils.go
@@ -87,7 +87,8 @@ func StartK8sTestEnvironment(crdPaths []string) (*K8sTestEnv, error) {
 			&sourcev1.GitRepository{},
 			&v1.CustomResourceDefinition{},
 		},
-		Scheme: scheme,
+		Scheme:             scheme,
+		MetricsBindAddress: "0",
 	})
 	if err != nil {
 		return nil, fmt.Errorf("could not create controller manager: %w", err)


jpellizzari (commented):
:exploding_head:

jpellizzari (approved):
Never would have found this. Nice work!

J-Thompson12 (approved):

PR #1345, opened 2022-01-25
Get kube client during HTTP request

yiannistri:
Part of: #1092 

**What changed?**
To get an instance of `kube.Kube` a `Factory` was used that was initialised during application startup with a `*rest.Config`. The responsibility of getting a `kube.Kube` instance now lies with the `KubeGetter` which reads the `*rest.Config` during an HTTP request and uses it to impersonate the call to the Kubernetes API.

**Why?**

**How did you test it?**
Updated existing unit tests + manually

**Release notes**

**Documentation Changes**

### DIFF
--- a/pkg/server/config_getter.go
+++ b/pkg/server/config_getter.go
@@ -76,3 +76,28 @@ func (g *DefaultClientGetter) Client(ctx context.Context) (client.Client, error)
 
 	return rawClient, nil
 }
+
+// KubeGetter implementations should create a Kube client from a context.
+type KubeGetter interface {
+	Kube(ctx context.Context) (kube.Kube, error)
+}
+
+// DefaultKubeGetter implements the KubeGetter interface and uses a ConfigGetter
+// to get a *rest.Config and create a Kube client.
+type DefaultKubeGetter struct {
+	configGetter ConfigGetter
+	clusterName  string
+}
+
+// Kube creates a new Kube client using the *rest.Config returned from its
+// ConfigGetter.
+func (g *DefaultKubeGetter) Kube(ctx context.Context) (kube.Kube, error) {
+	config := g.configGetter.Config(ctx)
+
+	kube, _, err := kube.NewKubeHTTPClientWithConfig(config, g.clusterName)
+	if err != nil {
+		return nil, fmt.Errorf("could not create kube http client: %w", err)
+	}
+
+	return kube, nil
+}
--- a/pkg/server/server_test.go
+++ b/pkg/server/server_test.go
@@ -605,7 +605,6 @@ var _ = Describe("ApplicationsServer", func() {
 
 	Context("RemoveApplication Tests", func() {
 		var ctx context.Context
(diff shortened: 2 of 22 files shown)


test/integration/server/remove_test.go
```diff
@@ -48,7 +48,7 @@ var _ = Describe("RemoveApplication", func() {
 	Context("Github", func() {
 		var gh *ghAPI.Client
 		var gp gitprovider.Client
-		var githubOrg = "weaveworks-gitops-test"
+		var githubOrg = os.Getenv("GITHUB_ORG")
```
jpellizzari, line 51: :+1:

jpellizzari (approved):
LGTM

pkg/server/config_getter.go
```diff
@@ -76,3 +76,28 @@ func (g *DefaultClientGetter) Client(ctx context.Context) (client.Client, error)
...
+func (g *DefaultKubeGetter) Kube(ctx context.Context) (kube.Kube, error) {
+	config := g.configGetter.Config(ctx)
+
+	kube, _, err := kube.NewKubeHTTPClientWithConfig(config, g.clusterName)
+	if err != nil {
+		return nil, fmt.Errorf("could not create kube http client: %w", err)
+	}
+
+	return kube, nil
+}
```
foot, line 103: I like this idea of keeping them separate  getters👌 . (instead of return kube and rawclient in the same getter), hopefully make it easier to adapt the code later

pkg/services/factory.go
```diff
@@ -63,47 +57,11 @@ func NewFactory(fluxClient flux.Flux, log logger.Logger) Factory {
...
-		kubeClient, _, err := kube.NewKubeHTTPClientWithConfig(f.rest, f.clusterName)
-		if err != nil {
-			return nil, err
-		}
-
-		mainKubeClient = kubeClient
-	}
-
-	return mainKubeClient, nil
-}
```
foot, line 104: ✨ ✨ ✨ ✨ ✨ ✨ 🧹

pkg/server/server_test.go
```diff
@@ -1341,3 +1330,17 @@ func NewFakeClientGetter(client client.Client) ClientGetter {
...
+
+func NewFakeKubeGetter(kube kube.Kube) KubeGetter {
+	return &FakeKubeGetter{
+		kube: kube,
+	}
+}
+
+func (g *FakeKubeGetter) Kube(ctx context.Context) (kube.Kube, error) {
+	return g.kube, nil
+}
```
foot, line 1346: Oh yeah! Can we move all the fake stuff out of test package and into a fakes package so that its importable and usable from go module consumers and their test infra too?

foot (approved):
Wicked this.

pkg/server/server_test.go
yiannistri: I couldn't easily do that because of circular dependencies. I'll attempt to change that in one of the next PRs 🤔

PR #1355, opened 2022-01-27
Only run tests when PRing to main

jpellizzari:
CI running on our `v2` branch was causing issues. We have a separate, simplified CI workflow for that branch.


josecordaz (approved):
LGTM

PR #1356, opened 2022-01-27
Adds some new endpoint test coverage and filtering logic

jpellizzari:
Add more test coverage and filtering logic to prototype endpoints


core/server/types/app.go
```diff
@@ -26,9 +30,9 @@ func AppAddProtoToCustomResource(msg *pb.AddAppRequest) *v1alpha2.Application {
 			Name:      msg.Name,
 			Namespace: msg.Namespace,
 			Labels: map[string]string{
-				"app.kubernetes.io/part-of":    msg.Name,
-				"app.kubernetes.io/managed-by": "weave-gitops",
-				"app.kubernetes.io/created-by": "kustomize-controller",
+				PartOfLabel:    msg.Name,
+				ManagedByLabel: "weave-gitops",
```
josecordaz, line 34: Probably we can replace `"weave-gitops"` with `managedByWeaveGitops`.

core/server/types/app.go
```diff
@@ -26,9 +30,9 @@ func AppAddProtoToCustomResource(msg *pb.AddAppRequest) *v1alpha2.Application {
...
+				CreatedByLabel: "kustomize-controller",
```
josecordaz, line 35: also instead of `"kustomize-controller"` we could use `createdByKustomizeController`

josecordaz (approved):

core/server/kustomization_test.go
```diff
@@ -97,6 +98,108 @@ func TestCreateKustomization(t *testing.T) {
...
+		AppName:   "",
+		SourceRef: &pb.SourceRef{
+			Kind: pb.SourceRef_GitRepository,
+			Name: "othersource",
+		},
+	}
+
+	_, err = c.AddKustomization(ctx, unAssociatedKustomizationReq)
+	g.Expect(err).NotTo(HaveOccurred())
+
```
josecordaz, line 141: might want to add `g.Expect(addRes.Success).To(BeTrue())` here

jpellizzari: I deliberately left if off since that isn't what this test is trying to prove.

core/server/types/app.go
jpellizzari: Fixed

core/server/types/app.go
jpellizzari: Fixed

Issue #1199, opened 2021-12-08
Fix flaky unit test


JamWils:
The unit test below fails intermittently on our CI pipeline.  This is causing a lot of unnecessary delays.  Fix the unit test so it is no longer flaky.

```
• Failure [0.003 seconds]
Github Device Flow
/home/runner/work/***/***/pkg/services/auth/github_test.go:40
  pollAuthStatus
  /home/runner/work/***/***/pkg/services/auth/github_test.go:92
    returns a token after a slow_down [It]
    /home/runner/work/***/***/pkg/services/auth/github_test.go:118

    should have tried the first time
    Expected
        <int>: 0
    to equal
        <int>: 1

    /home/runner/work/***/***/pkg/services/auth/github_test.go:133
```

J-Thompson12:
I am still getting this issue in the pipeline 
```
Github Device Flow
/home/runner/work/***/***/pkg/services/auth/github_test.go:41
  pollAuthStatus
  /home/runner/work/***/***/pkg/services/auth/github_test.go:93
    returns a token after a slow_down [It]
    /home/runner/work/***/***/pkg/services/auth/github_test.go:119

    should have tried the first time
    Expected
        <int>: 0
    to equal
        <int>: 1

    /home/runner/work/***/***/pkg/services/auth/github_test.go:135
```

J-Thompson12:
We need to find a new package to handle this or code up our own solution. The current package isnt working how we want.

PR #1358, opened 2022-01-27
Restructure GitHub auth test to remove flakiness

dhwthompson:
Closes: #1199

As best as I can tell, the flakiness here was coming from two places:

1. Plain ol' data races, where we were sharing a variable between threads, updating it one thread and reading it in another. Sometimes these wouldn't happen in the right order, causing failures.
2. Race conditions when interacting with the `clock` library, where (for example) we would kick off a piece of code that we knew would call `Sleep`, call `runtime.Gosched` to give it a chance to execute, and then tell the `clock` library to advance time. It looks like these, again, weren't always happening in order.

A bunch of the complexity here stems from the fact that we're poking our clock from one end with the `Sleep` function, then poking it from the other end to tell it time has advanced. Since all we're relying on in our non-test code is a `sleep` function, we can fake this out with a small `struct`, then use the `mockRoundTripper` to keep track of what the "current time" is when it receives requests.

Another change is using a channel to communicate when `pollAuthStatus` has completed, so that we can wait for this (and know nothing else is going to change) before testing the state of the system.

Finally, I took the opportunity to restructure the tests a little bit to take advantage of Ginkgo's `Context`s, providing a slightly clearer separation between what scenario we're testing and what behavior we want to see.

**How did you test it?**

I mostly tested this manually (and repeatedly), but also with a bit of help from [Go’s race detector](https://go.dev/doc/articles/race_detector).

### DIFF
--- a/pkg/services/auth/github_test.go
+++ b/pkg/services/auth/github_test.go
@@ -37,11 +37,6 @@ func (t *testServerTransport) RoundTrip(r *http.Request) (*http.Response, error)
 	return t.roundTripper.RoundTrip(r)
 }
 
-type outcome struct {
-	string
-	error
-}
-
 // sleeper is a very lightweight fake sleep timer. Instead of faking out the system
 // clock, we can accept `sleep` calls and keep track of how long we've slept.
 type sleeper struct {
@@ -119,7 +114,6 @@ var _ = Describe("Github Device Flow", func() {
 	Describe("pollAuthStatus", func() {
 		var rt *mockAuthRoundTripper
 		var s *sleeper
-		outcomeChan := make(chan outcome)
 
 		// pollTimes is a convenience function to convert from a series of polling intervals
 		// to their respective polling timestamps, relative to the sleeper type's starting time
@@ -154,12 +148,7 @@ var _ = Describe("Github Device Flow", func() {
 			It("retries with a longer interval", func() {
 				interval := 5 * time.Second
 
-				go func() {
-					resultToken, err := pollAuthStatus(s.sleep, interval, client, "somedevicecode")
-					outcomeChan <- outcome{resultToken, err}
-				}()
-
-				<-outcomeChan
+				_, _ = pollAuthStatus(s.sleep, interval, client, "somedevicecode")
 
 				expectedPollTimes := pollTimes([]time.Duration{
 					interval,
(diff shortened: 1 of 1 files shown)


pkg/services/auth/github_test.go
```diff
@@ -91,57 +111,86 @@ var _ = Describe("Github Device Flow", func() {
...
-			rt := newMockRoundTripper(3, token)
-			client.Transport = &testServerTransport{testServeUrl: ts.URL, roundTripper: rt}
-			interval := 5 * time.Second
+	Describe("pollAuthStatus", func() {
+		var rt *mockAuthRoundTripper
+		var s *sleeper
+
+		// pollTimes is a convenience function to convert from a series of polling intervals
+		// to their respective polling timestamps, relative to the sleeper type's starting time
+		pollTimes := func(intervals []time.Duration) []time.Time {
```
jpellizzari, line 120: :question: Why a `func` literal here? I might be missing it but I don't see what we are closing over (if that is the itnent).

pkg/services/auth/github_test.go
```diff
@@ -91,57 +111,86 @@ var _ = Describe("Github Device Flow", func() {
...
+					times[index] = zero.Add(interval)
+				default:
+					times[index] = times[index-1].Add(interval)
+				}
+			}
+			return times
+		}
 
-			c := clock.NewMock()
+		drainPollTimes := func(pollChan <-chan time.Time) (result []time.Time) {
```
jpellizzari, line 134: We will probably need to decide as a team if we like this "named return variable" pattern. I have always found it an odd choice for the language and I have a feeling there will be mixed opinions on whether or not we use it.

Until then, so long as it passes the linter, :+1:

jpellizzari (approved):
A very smart solution. Nice work. Only questions for me.

pkg/services/auth/github_test.go
dhwthompson: You’re correct that we’re not closing over anything here. The intent is to keep the helper function scoped to the tests where we’re using it, for test readability as much as anything else.

I’m not sure why Go doesn’t allow `func` declarations within functions to use the same syntax as top-level ones; there’s probably some reason it would make the compiler more complicated.

pkg/services/auth/github_test.go
dhwthompson: Fair point. I wouldn’t use named return variables in a longer function; it seemed worth the trade-off here, but I’ll be happy to go with the team consensus if we decide to ditch the pattern entirely.

pkg/services/auth/github_test.go
jpellizzari: So the argument is "it's only ever used in this one place, so I put it right here"? Makes sense to me.

J-Thompson12 (approved):
This is much clearer.

Issue #1109, opened 2021-11-19
Add `gitops add/apply profile/profiles` command


Himangini:
### Description
As a weave gitops user
I want to run a command `gitops add/apply profile/profiles` on the CLI
So that I can install a profile

### Scope

- The command should be added in `weave-gitops` and behave similarly to `pctl add` command
- The command should create a PR by default
- The command should use the [git repository setting documented here](https://docs.gitops.weave.works/docs/next/cluster-management/getting-started/#profiles-and-clusters) as the value of `--git-repository` flag from the `pctl` command
- Testing and documentation should be added for the new command in `weave-gitops` 

### Acceptance Criteria
Given that I am a weave gitops user
When I run CLI command `gitops add/apply profile/profiles <profile-name> --cluster <cluster-name> --version=0.0.1 --set port=80 replicas=3 pr-title=my_pr branch=my_branch commit=my_commit`
Then the command should generate the profile installation for the version provided and
Then it should open a PR against my repository and 
I should get a link to the PR on the terminal `Created pull request: <link>`

When I run CLI command `gitops add/apply profile/profiles <profile-name> --cluster <cluster-name>`
And I don't provide the `--version` flag
Then the profile installation should be generated using the latest version of the profile

When I run CLI command `gitops add/apply profile/profiles <profile-name>`
And I use the  `--skip-pr` flag
Then the PR should not be created as default

When I run CLI command `gitops add/apply profile/profiles <profile-name> --cluster <cluster-name>`
And I do not provide `port, replicas, pr-title, branch, commit` flags
Then I should be prompted for input for all of these before the profile is installed

When I run CLI command `gitops add/apply profile/profiles`
And I provide incorrect `<profile-name>` value
Then I should see the error on the terminal

When I run CLI command `gitops add/apply profile/profiles`
And I provide incorrect `<cluster-name>` value
Then I should see the error on the terminal

When I hit the endpoint to install a profile
And I provide valid profile data as part of the request
Then I should get back the `yaml` as a successful response

### Technical Notes
- Create an API so that we can install a profile using both CLI and UI [ UI will not be implemented as part of this ticket] 
- Some of these flags from the current pctl command will become redundant in `weave-gitops`, investigate which ones we need to keep
```
   --pr-remote value             The remote to push the branch to. (default: origin)
   --pr-base value               The base branch to open a PR against. (default: main)
   --pr-repo value               The repository to open a pr against. Format is: org/repo-name.
[...]

Issue #1311, opened 2022-01-18
[Subtask] Add gitops add command


Himangini:
follow up from #1109

### Description
As a weave gitops user
I want to run a command `gitops add/apply profile/profiles` on the CLI
So that I can install a profile

This subtask adds the remaining flags from the todo 
- [ ]  --pr-title=my_pr
- [ ]  --commit=my_commit
- [ ]  --branch value The branch to merge to. (default: main)

### Scope

- The command should be added in `weave-gitops` and behave similarly to `pctl add` command
- The command should create a PR by default
- The command should use the [git repository setting documented here](https://docs.gitops.weave.works/docs/next/cluster-management/getting-started/#profiles-and-clusters) as the value of `--git-repository` flag from the `pctl` command
- Testing and documentation should be added for the new command in `weave-gitops` 

### Acceptance Criteria
Given that I am a weave gitops user
When I run CLI command `gitops add profile <profile-name> --cluster <cluster-name> --version=0.0.1 --set port=80 pr-title=my_pr branch=my_branch commit=my_commit`
Then the command should generate the profile installation for the version provided and
Then it should open a PR against my repository and 
I should get a link to the PR on the terminal `Created pull request: <link>`

Himangini:
@Himangini check with product/ux what replica is used for?

Himangini:
Use the flags names which are consistent with current implementation since the UX sheet seems to be outdated

PR #1360, opened 2022-01-28
Implement `add profile` to install a profile to a cluster

nikimanoledaki:
Closes: #1109

**What changed? Why?**
This PR adds the ability to install a profile to a cluster. 

#### Usage
Minimum/required flags: 
`gitops add profile --name=podinfo --cluster=prod --config-repo=ssh://git@github.com/owner/config-repo.git`

Optional flags: 
- `--namespace` defaults to `wego-system`
- `--version` defaults to `latest`
- `--port` for the Profiles API port
- `--auto-merge`
- `--kubeconfig`

#### Implementation Notes
1. The `HelmRepository` Flux resource that contains the Chart must already be present and available for installation in the cluster. 
2. `add profile` discovers the `HelmRepository`'s name and namespace through a `Get` query to the Profiles API. While this PR tries to rely on the Git provider API to use the `.weave-gitops` directory as the Source of Truth, `add profile` defers to `get profiles` to discover the `HelmRepository`'s name and namespace, and its workflow still requires access to the cluster through a local kubeconfig. However, this will become optional as part of the following ticket: #1371
3. `add profile` creates a `HelmRelease` that references the `HelmRepository` name and namespace.
```
kind: HelmRelease
metadata:
  name: nginx
  namespace: wego-system
spec:
  releaseName: nginx
  chart: 
    spec:
      chart: nginx
      sourceRef:
        kind: HelmRepository
        name: // discovered by Get Profiles
        namespace: wego-system // discovered by Get Profiles
      Version: 0.0.1
  interval: 1m
status: {}
```
4. If no other `HelmRelease` exists with the same name/version/namespace/cluster combination in the config repository's Profiles manifest located at `/.weave-gitops/clusters/<cluster name>/system/profiles.yaml`, `add profile` appends the new `HelmRelease` to it. Otherwise, it returns an error.
```
.weave-gitops
└── clusters
    	└── prod
    		└── system
        		└── profiles.yaml // kind: HelmRelease
```
5. Flux watches the `profile.yaml` file for new `HelmRelease`s and installs them.
 
#### `GitProvider` interface additions
- This PR implements `gitprovider.GetRepoFiles` to fetch a repo's subdirectory and its files. It's essentially a wrapper around [go-git-providers to use `.Files().Get(ctx, path, branch)`](https://github.com/fluxcd/go-git-providers/pull/80). `add profile` uses this func to get the `/weave-gitops/clusters/<cluster-name>/system/profiles.yaml` file from the config repo to read/append `HelmRelease`s to it.

- This PR also implements `gitprovider.MergePullRequest` to merge a PR through the Git provider API without relying on a local clone of the config repo. `auto-merge` in the acceptance tests uses this function.

**How did you test it?**

#### Manual Test
```
// Create cluster
kind create cluster --name test-01

// Edit `manifests/wego-app/deployment.yaml` to point to own container registry & build/push image
docker build -t niki2401/wego-app:latest -f Dockerfile . && docker push niki2401/wego-app:latest

// Make binary
make bin

// Install gitops
./bin/gitops install --config-repo ssh://git@github.com/nikimanoledaki/fun-times.git --auto-merge

// Create helmrepository 
touch helmrepo.yaml
cat <<EOF >./helmrepo.yaml 
apiVersion: source.toolkit.fluxcd.io/v1beta1
kind: HelmRepository
metadata:
  name: weaveworks-charts
  namespace: wego-system
spec:
  interval: 1m0s
  timeout: 1m0s
  url: https://weaveworks.github.io/profiles-examples
EOF
k apply -f helmrepo.yaml 

// Port-forward wego-app API
k port-forward pod/wego-app-<generated name> 9001 9001 -n wego-system

// Install Profile
./bin/gitops add profile --name podinfo --config-repo ssh://git@github.com/nikimanoledaki/fun-times.git --cluster kind-test-01 --version 6.0.0 --auto-merge
► getting available profiles in kind-test-01/wego-system
► Pull Request created: https://github.com/nikimanoledaki/fun-times/pull/29
► auto-merge=true; merging PR number 29
Adding profile:

Name: podinfo
Version: 6.0.0
Cluster: kind-test-01
Namespace: wego-system

// check profile is running
k get pods -A
wego-system          pod/kind-test-01-podinfo-569494d49d-xch8m           1/1     Running   0          30s
```

#### Acceptance Test 
- Example PR + changes created in a config repo by the acceptance test: https://github.com/weaveworks-gitops-test/test-app-o1yvvchf

**Release notes**

**Documentation Changes**
Not sure where to add docs for this? 🤔


pkg/services/profiles/add.go
```diff
@@ -0,0 +1,172 @@
...
+	} else if !repoExists {
+		return fmt.Errorf("repository '%v' could not be found", configRepoUrl.String())
+	}
+	defaultBranch, err := gitProvider.GetDefaultBranch(ctx, configRepoUrl)
+	if err != nil {
+		return err
+	}
+
+	// TODO: We need a ticket for when the user auto-completes the chart name, helm repository name, and helm repository namespace,
+	// so that a user doesn't only rely on hitting the Kube API if they don't have access to the cluster.
```
nikimanoledaki, line 53: cc @aclevername

aclevername: https://github.com/weaveworks/weave-gitops/issues/1371

cmd/gitops/add/profiles/cmd.go
```diff
@@ -0,0 +1,132 @@
...
+	"k8s.io/client-go/util/homedir"
+)
+
+var opts profiles.AddOptions
+
+// AddCommand provides support for adding a profile to a cluster.
+func AddCommand(client *resty.Client) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:           "profile",
+		Aliases:       []string{"profiles"},
```
aclevername, line 35: do we want the alias since you can only install 1 at a time?

cmd/gitops/add/profiles/cmd.go
```diff
@@ -0,0 +1,132 @@
...
+		gitops add profile --name=podinfo --cluster=prod --version=1.0.0 --config-repo=ssh://git@github.com/owner/config-repo.git
+		`,
+		RunE: addProfileCmdRunE(client),
+	}
+
+	cmd.Flags().StringVar(&opts.Name, "name", "", "Name of the profile")
+	cmd.Flags().StringVar(&opts.Version, "version", "latest", "Version of the profile")
+	cmd.Flags().StringVar(&opts.ConfigRepo, "config-repo", "", "URL of external repository (if any) which will hold automation manifests")
+	cmd.Flags().StringVar(&opts.Cluster, "cluster", "", "Name of the cluster to add the profile to")
+	cmd.Flags().StringVar(&opts.Port, "port", server.DefaultPort, "Port the profiles API is running on")
```
aclevername, line 50:
```suggestion
	cmd.Flags().StringVar(&opts.Port, "profiles-api-port", server.DefaultPort, "Port the profiles API is running on")
```

i wonder if its even possible to just have the code discover this, since its already querying k8s. perhaps an improvement for down the line

cmd/gitops/add/profiles/cmd.go
```diff
@@ -0,0 +1,132 @@
...
+		if err != nil {
+			return fmt.Errorf("error initializing kubernetes client: %w", err)
+		}
+
+		kubeClient, _, err := kube.NewKubeHTTPClient()
+		if err != nil {
+			return fmt.Errorf("failed to create kube client: %w", err)
+		}
+
+		ctx := context.Background()
```
aclevername, line 88: is this what we do in the other commands?

cmd/gitops/add/profiles/cmd.go
```diff
@@ -0,0 +1,132 @@
...
+			Namespace:        opts.Namespace,
+			IsHelmRepository: true,
+			DryRun:           false,
+		})
+		if err != nil {
+			return fmt.Errorf("failed to get git clients: %w", err)
+		}
+
+		profilesService := profiles.NewService(clientSet, log)
+		return profilesService.Add(ctx, gitProvider, opts)
```
aclevername, line 100:
```suggestion
		return profiles.NewService(clientSet, log).Add(ctx, gitProvider, opts)
```

aclevername: personal pref, might not be for everyone

cmd/gitops/add/profiles/cmd.go
```diff
@@ -0,0 +1,132 @@
...
+		SilenceErrors: true,
+		Example: `
+		# Add a profile to a cluster
+		gitops add profile --name=podinfo --cluster=prod --version=1.0.0 --config-repo=ssh://git@github.com/owner/config-repo.git
+		`,
+		RunE: addProfileCmdRunE(client),
+	}
+
+	cmd.Flags().StringVar(&opts.Name, "name", "", "Name of the profile")
+	cmd.Flags().StringVar(&opts.Version, "version", "latest", "Version of the profile")
```
aclevername, line 47:
```suggestion
	cmd.Flags().StringVar(&opts.Version, "version", "latest", "Version of the profile, following semver e.g.: v0.1.0")
```

cmd/gitops/add/profiles/cmd_test.go
```diff
@@ -0,0 +1,118 @@
...
+			})
+
+			It("fails if --name is prefixed by 'wego'", func() {
+				cmd.SetArgs([]string{
+					"add", "profile",
+					"--name", "wego-app",
+				})
+				err := cmd.Execute()
+				Expect(err).To(MatchError("the prefix 'wego' is used by weave gitops and is not allowed for a profile name"))
+			})
```
aclevername, line 68: what prompted this?

cmd/gitops/add/profiles/cmd.go
```diff
@@ -0,0 +1,132 @@
...
+		log := internal.NewCLILogger(os.Stdout)
+		fluxClient := flux.New(osys.New(), &runner.CLIRunner{})
+		factory := services.NewFactory(fluxClient, log)
+		providerClient := internal.NewGitProviderClient(os.Stdout, os.LookupEnv, auth.NewAuthCLIHandler, log)
+
+		if err := validateAddOptions(opts); err != nil {
+			return err
+		}
+
+		ns, err := cmd.Parent().Parent().Flags().GetString("namespace")
```
bigkevmcd, line 67: `namespace` is a [PersistentFlag](https://github.com/weaveworks/weave-gitops/blob/main/cmd/gitops/root/cmd.go#L128) which means you should be able to simplify this to just...

```go
ns, err := cmd.Flags().GetString("namespace")
```

cmd/gitops/add/profiles/cmd.go
```diff
@@ -0,0 +1,132 @@
...
+		if err != nil {
+			return fmt.Errorf("failed to get git clients: %w", err)
+		}
+
+		profilesService := profiles.NewService(clientSet, log)
+		return profilesService.Add(ctx, gitProvider, opts)
+	}
+}
+
+func validateAddOptions(opts profiles.AddOptions) error {
```
bigkevmcd, line 103: This sets up an unfortunate error pattern, where you run it and it tells you something is invalid, so you fix it, and then you run it again, and it tells you something _else_ is invalid.

Maybe a [`multierror`](https://github.com/hashicorp/go-multierror) would be better?

At least, return a slice of strings with all the errors.

cmd/gitops/add/profiles/cmd_test.go
```diff
@@ -0,0 +1,118 @@
...
+
+var _ = Describe("Add Profiles", func() {
+	var (
+		cmd *cobra.Command
+	)
+
+	BeforeEach(func() {
+		client := resty.New()
+		httpmock.ActivateNonDefault(client.GetClient())
+		defer httpmock.DeactivateAndReset()
```
bigkevmcd, line 20: Doesn't this cause it to be deactivated and reset when it leaves the func() in the `BeforeEach` ?

pkg/helm/releases.go
```diff
@@ -0,0 +1,39 @@
...
+	"time"
+
+	helmv2beta1 "github.com/fluxcd/helm-controller/api/v2beta1"
+	sourcev1beta1 "github.com/fluxcd/source-controller/api/v1beta1"
+	pb "github.com/weaveworks/weave-gitops/pkg/api/profiles"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// MakeHelmRelease returns a HelmRelease object given a profile, cluster, namespace, and version.
+func MakeHelmRelease(p *pb.Profile, cluster, namespace string) *helmv2beta1.HelmRelease {
```
aclevername, line 13: given this is relating to profiles only, does it make sense to have it in a different package? ` pkg/services/profiles/helm/`

pkg/helm/releases.go
```diff
@@ -0,0 +1,39 @@
...
+		},
+		TypeMeta: metav1.TypeMeta{
+			APIVersion: helmv2beta1.GroupVersion.Identifier(),
+			Kind:       helmv2beta1.HelmReleaseKind,
+		},
+		Spec: helmv2beta1.HelmReleaseSpec{
+			Chart: helmv2beta1.HelmChartTemplate{
+				Spec: helmv2beta1.HelmChartTemplateSpec{
+					Chart:   p.Name,
+					Version: p.AvailableVersions[0],
```
aclevername, line 27: potential panic?

pkg/helm/releases_test.go
```diff
@@ -0,0 +1,44 @@
...
+	It("creates a helm release", func() {
+		hr := helm.MakeHelmRelease(profile, cluster, ns)
+		Expect(hr.Name).To(Equal(cluster + "-" + name))
+		Expect(hr.Namespace).To(Equal(ns))
+		Expect(hr.TypeMeta.APIVersion).To(Equal(helmv2beta1.GroupVersion.Identifier()))
+		Expect(hr.TypeMeta.Kind).To(Equal(helmv2beta1.HelmReleaseKind))
+		Expect(hr.Spec.Chart.Spec.Chart).To(Equal(name))
+		Expect(hr.Spec.Chart.Spec.Version).To(Equal("6.0.0"))
+		Expect(hr.Spec.Chart.Spec.SourceRef.Name).To(Equal(name))
+		Expect(hr.Spec.Chart.Spec.SourceRef.Namespace).To(Equal(ns))
```
aclevername, line 42: this doesn't test all the fields your not setting, e.g. I could add some other field to the `spec` and no test would fail. could we do a full resource comparison instead? e.g.

```
Expect(hr).To(Equal(helmv1.HelmRelease{
  Metadata:
    ..
  Spec:
    ...
})
```

pkg/models/manifest.go
```diff
@@ -44,6 +44,7 @@ const (
 	SystemKustomizationPath = "kustomization.yaml"
 	WegoAppPath             = "wego-app.yaml"
 	WegoConfigPath          = "wego-config.yaml"
+	WegoProfilePath         = "profiles.yaml"
```
aclevername, line 47: I know we paired on this but we should see if we can remove having to have this here and in `git/git.go`. I think this is probably the better place

pkg/services/profiles/add.go
```diff
@@ -0,0 +1,175 @@
...
+	"github.com/weaveworks/weave-gitops/pkg/helm"
+	"github.com/weaveworks/weave-gitops/pkg/services/automation"
+
+	"github.com/fluxcd/go-git-providers/gitprovider"
+	"github.com/fluxcd/helm-controller/api/v2beta1"
+	goyaml "github.com/go-yaml/yaml"
+	"sigs.k8s.io/yaml"
+)
+
+const AddCommitMessage = "Add Profile manifests"
```
aclevername, line 20:
```suggestion
const AddDefaultCommitMessage = "Add Profile manifests"
```

aclevername: oh wait maybe this isn't configurable yet? If so nevermind :D

pkg/services/profiles/add.go
```diff
@@ -0,0 +1,175 @@
...
+
+// Add installs an available profile in a cluster's namespace by appending a HelmRelease to the profile manifest in the config repo,
+// provided that such a HelmRelease does not exist with the same profile name and version in the same namespace and cluster.
+func (s *ProfilesSvc) Add(ctx context.Context, gitProvider gitproviders.GitProvider, opts AddOptions) error {
+	configRepoUrl, err := gitproviders.NewRepoURL(opts.ConfigRepo, true)
+	if err != nil {
+		return err
+	}
+
+	// TODO: refactor this into 1 func that returns repoRef
```
aclevername, line 40: :eyes:

cmd/gitops/add/profiles/cmd.go
```diff
@@ -0,0 +1,132 @@
...
+		`,
+		RunE: addProfileCmdRunE(client),
+	}
+
+	cmd.Flags().StringVar(&opts.Name, "name", "", "Name of the profile")
+	cmd.Flags().StringVar(&opts.Version, "version", "latest", "Version of the profile")
+	cmd.Flags().StringVar(&opts.ConfigRepo, "config-repo", "", "URL of external repository (if any) which will hold automation manifests")
+	cmd.Flags().StringVar(&opts.Cluster, "cluster", "", "Name of the cluster to add the profile to")
+	cmd.Flags().StringVar(&opts.Port, "port", server.DefaultPort, "Port the profiles API is running on")
+	cmd.Flags().BoolVar(&opts.AutoMerge, "auto-merge", false, "If set, 'gitops add profile' will merge automatically into the set --branch")
```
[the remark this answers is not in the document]
nikimanoledaki: We're following up with leftover params in a separate ticket since the scope of this PR is already pretty huge: https://github.com/weaveworks/weave-gitops/issues/1311

Current implementation only discovers the default branch.

pkg/services/profiles/add.go
```diff
@@ -0,0 +1,175 @@
...
+	repoExists, err := gitProvider.RepositoryExists(ctx, configRepoUrl)
+	if err != nil {
+		return err
```
aclevername, line 43: some error wrapping missing in a few places, would be nice for more context

pkg/services/automation/generator.go
```diff
@@ -410,6 +411,16 @@ func GenerateResourceName(url gitproviders.RepoURL) string {
 	return models.ConstrainResourceName(url.RepositoryName())
 }
 
+// GetRandomName generates a random unique name.
```
bigkevmcd, line 414: This isn't quite what it does, it generates a random unique string, and appends it to the provided string.

This should probably also limit it to a specific length.

pkg/services/profiles/add.go
```diff
@@ -0,0 +1,175 @@
...
+func (s *ProfilesSvc) printAddSummary(opts AddOptions) {
+	s.Logger.Println("Adding profile:\n")
+	s.Logger.Println("Name: %s", opts.Name)
+	s.Logger.Println("Version: %s", opts.Version)
+	s.Logger.Println("Cluster: %s", opts.Cluster)
+	s.Logger.Println("Namespace: %s", opts.Namespace)
+	s.Logger.Println("")
+}
+
+func MakeManifestFile(files []*gitprovider.CommitFile, newRelease *v2beta1.HelmRelease, path string) ([]gitprovider.CommitFile, error) {
```
aclevername, line 112: `ManifestFile` :thinking: is this the list of profiles?

pkg/services/profiles/add_test.go
```diff
@@ -0,0 +1,342 @@
...
+		gitProviders.GetRepoFilesReturns(makeTestFiles(), nil)
+		clientSet.AddProxyReactor("services", func(action testing.Action) (handled bool, ret restclient.ResponseWrapper, err error) {
+			return true, newFakeResponseWrapper(getProfilesResp), nil
+		})
+		fakePR.GetReturns(gitprovider.PullRequestInfo{
+			WebURL: "url",
+		})
+		gitProviders.CreatePullRequestReturns(fakePR, nil)
+		Expect(profilesSvc.Add(context.TODO(), gitProviders, addOptions)).Should(Succeed())
+		Expect(gitProviders.RepositoryExistsCallCount()).To(Equal(1))
```
aclevername, line 64: Are we missing more assertions around what fakes were called and with what arguments? e.g.
```
Expect(gitProviders.RepositoryExistsCallCount()).To(Equal(1)) # already here
Expect(gitProviders.RepositoryExistsArgsForCall(0)).To(Equal(<something>))

Expect(gitProviders.GetDefaultBrancnCallCount()).To(Equal(1))
Expect(gitProviders.GetDefaultBrancnArgsForCall(0)).To(Equal(<something>))



```

aclevername: also what does `latest` get converted too here?

pkg/services/profiles/add_test.go
```diff
@@ -0,0 +1,342 @@
...
+			})
+			gitProviders.CreatePullRequestReturns(nil, fmt.Errorf("err"))
+			err := profilesSvc.Add(context.TODO(), gitProviders, addOptions)
+			Expect(err).NotTo(BeNil())
+			Expect(err).To(MatchError("failed to create the pull request: err"))
+		})
+	})
+})
+
+var _ = Describe("MakeManifestFile", func() {
```
aclevername, line 204:
```suggestion
  Describe("MakeManifestFile", func() {
```

pkg/services/profiles/add.go
```diff
@@ -0,0 +1,175 @@
...
+	return nil
+}
+
+func (s *ProfilesSvc) printAddSummary(opts AddOptions) {
+	s.Logger.Println("Adding profile:\n")
+	s.Logger.Println("Name: %s", opts.Name)
+	s.Logger.Println("Version: %s", opts.Version)
+	s.Logger.Println("Cluster: %s", opts.Cluster)
+	s.Logger.Println("Namespace: %s", opts.Namespace)
+	s.Logger.Println("")
```
bigkevmcd, line 109: For some reason, [`Println`](https://github.com/weaveworks/weave-gitops/blob/main/cmd/internal/logger.go#L21) and `Printf` do effectively the same thing (which is why you can use `%s`), so you could append `\n` to the end of the previous line

pkg/services/profiles/get.go
```diff
@@ -14,34 +15,80 @@ import (
...
 	if err != nil {
-		return fmt.Errorf("failed to unmarshal response: %w", err)
+		return nil, fmt.Errorf("failed to unmarshal response: %w", err)
 	}
+	return profiles, nil
+}
 
-	printProfiles(profiles, opts.Writer)
+// GetAvailableProfile returns a single available profile.
+func (s *ProfilesSvc) GetAvailableProfile(ctx context.Context, opts GetOptions) (*pb.Profile, error) {
```
aclevername, line 55: `Available` seems a bit odd, are you trying to differentiate the fact its not installed?

pkg/services/profiles/add.go
```diff
@@ -0,0 +1,175 @@
...
+	return []gitprovider.CommitFile{
+		{
+			Path:    &path,
+			Content: &content,
+		},
+	}, nil
+}
+
+// splitYAML splits a manifest file that contains multiple YAML resources separated by '---'.
+func splitYAML(resources []byte) ([][]byte, error) {
```
bigkevmcd, line 152: This code also ensures that each of the elements that were split are valid YAML which might not be obvious to a reader.

jpellizzari (commented):
LGTM! Only minor questions for me, save for the `package-lock.json` thing. I will let other folks give the :heavy_check_mark: since I am not super familiar with profiles.

cmd/gitops/add/profiles/cmd.go
Himangini: CC @davidstauffer @sympatheticmoose

cmd/gitops/add/cmd.go
```diff
@@ -21,6 +22,7 @@ gitops add cluster`,
 
 	cmd.AddCommand(clusters.ClusterCommand(endpoint, client))
 	cmd.AddCommand(app.Cmd)
+	cmd.AddCommand(profiles.AddCommand(client))
```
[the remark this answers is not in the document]
Skarlso: While I agree that the format is nice, I would rather be consistent with what's already there. :)

cmd/gitops/add/profiles/cmd.go
Skarlso: Technically, semver does not include `v`. so `v0.1.0` is not following semver. :)

cmd/gitops/add/profiles/cmd.go
```diff
@@ -0,0 +1,131 @@
...
+			return err
+		}
+
+		ns, err := cmd.Parent().Parent().Flags().GetString("namespace")
+		if err != nil {
+			return err
+		}
+		opts.Namespace = ns
+
+		config, err := clientcmd.BuildConfigFromFlags("", filepath.Join(homedir.HomeDir(), ".kube", "config"))
```
cPu1, line 73: This would prevent a user from using a custom kubeconfig location set using the `KUBECONFIG` environment variable. We should also account for this case.

cmd/gitops/add/profiles/cmd.go
Skarlso: Isn't the problem that these depend on each other? If `opts.Name` is not defined, you can't call `ApplicationNameTooLong`. And then you can't call HasPrefix either. Though I agree that ConfigRepo and Cluster could be checked separately.

cmd/gitops/add/profiles/cmd_test.go
Skarlso: Right, I think this needs to be in an AfterEach, right? 🤔

pkg/git/git.go
```diff
@@ -74,6 +77,12 @@ func GetSystemQualifiedPath(clusterName string, relativePath string) string {
 	return filepath.Join(GetSystemPath(clusterName), relativePath)
 }
 
+// GetProfilesPath returns the path of the file containing the manifests of installed Profiles
+// joined with the cluster's system directory
+func GetProfilesPath(clusterName string) string {
```
Skarlso, line 82: This might be a thing that the notification stuff can use in order to notify only for installed profiles 🤔 Nice.

cmd/gitops/add/profiles/cmd.go
```diff
@@ -0,0 +1,131 @@
...
+		return fmt.Errorf("the prefix 'wego' is used by weave gitops and is not allowed for a profile name")
+	}
+
+	if opts.ConfigRepo == "" {
+		return errors.New("--config-repo should be provided")
+	}
+
+	if opts.Cluster == "" {
+		return errors.New("--cluster should be provided")
+	}
```
cPu1, line 123: Required flags is a feature provided by Cobra, so you could do this in the command initialization code:

```golang
	requiredFlags := []string{"name", "config-repo", "cluster"}
	for _, f := range requiredFlags {
		if err := cobra.MarkFlagRequired(cmd.Flags(), f); err != nil {
			panic(fmt.Errorf("unexpected error: %w", err))
		}

	}
```

```suggestion
	if models.ApplicationNameTooLong(opts.Name) {
		return fmt.Errorf("--name value is too long: %s; must be <= %d characters",
			opts.Name, models.MaxKubernetesResourceNameLength)
	}

	if strings.HasPrefix(opts.Name, "wego") {
		return fmt.Errorf("the prefix 'wego' is used by weave gitops and is not allowed for a profile name")
	}
```

cmd/gitops/add/profiles/cmd.go
```diff
@@ -0,0 +1,131 @@
...
+
+	if opts.Version != "latest" {
+		if _, err := semver.StrictNewVersion(opts.Version); err != nil {
+			return fmt.Errorf("error parsing --version=%s: %s", opts.Version, err)
```
cPu1, line 127: Did you mean to use the `%w` verb?

```suggestion
			return fmt.Errorf("error parsing --version=%s: %w", opts.Version, err)
```

pkg/gitproviders/provider_org.go
```diff
@@ -125,3 +125,26 @@ func (p orgGitProvider) GetCommits(ctx context.Context, repoUrl RepoURL, targetB
 func (p orgGitProvider) GetProviderDomain() string {
 	return getProviderDomain(p.provider.ProviderID())
 }
+
+// GetRepoFiles returns the files found in a directory. The targetPath must point to a directory, not a file.
+// Note that the current implementation only gets an end subdirectory. It does not get multiple directories recursively. See https://github.com/fluxcd/go-git-providers/issues/143.
+func (p orgGitProvider) GetRepoFiles(ctx context.Context, repoUrl RepoURL, targetPath, targetBranch string) ([]*gitprovider.CommitFile, error) {
```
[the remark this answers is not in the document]
Skarlso: The change in naming is subtle, but I don't really like it because this, in fact, does not just `get files from a directory`. It fetches things from a URL and then checks for content inside that. It does more and I think the name reflects that nicely that it gets files from a repository. It could be more bespoke, but I think GetFilesInDir would simplify that too much.

pkg/gitproviders/provider_org.go
```diff
@@ -125,3 +125,26 @@ func (p orgGitProvider) GetCommits(ctx context.Context, repoUrl RepoURL, targetB
...
+	}
+	files, err := repo.Files().Get(ctx, targetPath, targetBranch)
+	if err != nil {
+		return nil, err
+	}
+	return files, nil
+}
+
+// MergePullRequest merges a pull request given the repository's URL and the PR's number with a merge method, and a commit message.
+func (p orgGitProvider) MergePullRequest(ctx context.Context, repoUrl RepoURL, pullRequestNumber int, mergeMethod gitprovider.MergeMethod, commitMesage string) error {
```
Skarlso, line 144: What's the point of this parameter `mergeMethod` if it's always Merge?

pkg/gitproviders/provider_org_test.go
```diff
@@ -313,4 +317,52 @@ var _ = Describe("Org Provider", func() {
...
+			Expect(prNumber).To(Equal(1))
+			Expect(mergeMethod).To(Equal(gitprovider.MergeMethodMerge))
+			Expect(message).To(Equal("message"))
+		})
+
+		When("merge the PR fails", func() {
+			It("returns an error", func() {
+				pullRequestsClient.MergeReturns(fmt.Errorf("err"))
+				err := orgProvider.MergePullRequest(context.TODO(), repoUrl, 1, gitprovider.MergeMethodMerge, "message")
+				Expect(err).To(HaveOccurred())
```
Skarlso, line 362: Here and elsewhere, you don't need this check if you have MetchError. :)

cmd/gitops/add/profiles/cmd.go
```diff
@@ -0,0 +1,131 @@
...
+	cmd.Flags().StringVar(&opts.Version, "version", "latest", "Version of the profile")
+	cmd.Flags().StringVar(&opts.ConfigRepo, "config-repo", "", "URL of external repository (if any) which will hold automation manifests")
+	cmd.Flags().StringVar(&opts.Cluster, "cluster", "", "Name of the cluster to add the profile to")
+	cmd.Flags().StringVar(&opts.Port, "port", server.DefaultPort, "Port the profiles API is running on")
+	cmd.Flags().BoolVar(&opts.AutoMerge, "auto-merge", false, "If set, 'gitops add profile' will merge automatically into the set --branch")
+
+	return cmd
+}
+
+func addProfileCmdRunE(client *resty.Client) func(*cobra.Command, []string) error {
```
cPu1, line 56: `client` is unused.
```suggestion
func addProfileCmdRunE() func(*cobra.Command, []string) error {
```

pkg/services/automation/generator.go
Skarlso: Also, you aren't seeding rand. This is a bit nitpicky, and probably fine for these scenarios, but if you ARE using `math/rand` you need to seed it with time.Nano or something like that.

pkg/services/gitopswriter/gitopswriter.go
```diff
@@ -102,7 +102,7 @@ func (dw *gitOpsDirectoryWriterSvc) AddApplication(ctx context.Context, app mode
 		Description:   fmt.Sprintf("Added yamls for %s", app.Name),
 		CommitMessage: AddCommitMessage,
 		TargetBranch:  defaultBranch,
-		NewBranch:     automation.GetAppHash(app),
+		NewBranch:     automation.GetRandomName("wego"),
```
Skarlso, line 105: Why wasn't AppHash enough here? Other than the annoying thing that you needed to remove the branch constantly. :D I guess that's why?

pkg/services/profiles/add.go
```diff
@@ -0,0 +1,175 @@
...
+	s.Logger.Println("Cluster: %s", opts.Cluster)
+	s.Logger.Println("Namespace: %s", opts.Namespace)
+	s.Logger.Println("")
+}
+
+func MakeManifestFile(files []*gitprovider.CommitFile, newRelease *v2beta1.HelmRelease, path string) ([]gitprovider.CommitFile, error) {
+	var content string
+	for _, f := range files {
+		if *f.Path == path {
+			if *f.Content == "" {
```
Skarlso, line 116: to avoid any problems, use `len(f.Content) == 0` here, because that returns 0 even if Content is nil. Which I see that is a required field, but still. :) No need to paint the devil on the wall.

pkg/services/profiles/add.go
```diff
@@ -0,0 +1,175 @@
...
+	s.Logger.Println("Version: %s", opts.Version)
+	s.Logger.Println("Cluster: %s", opts.Cluster)
+	s.Logger.Println("Namespace: %s", opts.Namespace)
+	s.Logger.Println("")
+}
+
+func MakeManifestFile(files []*gitprovider.CommitFile, newRelease *v2beta1.HelmRelease, path string) ([]gitprovider.CommitFile, error) {
+	var content string
+	for _, f := range files {
+		if *f.Path == path {
```
Skarlso, line 115: I'm a bit wary about this dereference here. I know Path is a required field.... But this leaves me of anxious. :D

pkg/services/profiles/add.go
```diff
@@ -0,0 +1,175 @@
...
+				}
+			}
+			content = *f.Content + "\n---\n"
+		}
+	}
+	helmReleaseManifest, err := yaml.Marshal(newRelease)
+	if err != nil {
+		return nil, err
+	}
+	content = content + string(helmReleaseManifest)
```
Skarlso, line 141:
```suggestion
	content += string(helmReleaseManifest)
```

pkg/services/profiles/add.go
```diff
@@ -0,0 +1,175 @@
...
+	var splitResources [][]byte
+	for {
+		var value interface{}
+		err := decoder.Decode(&value)
+		if err == io.EOF {
+			break
+		}
+		if err != nil {
+			return nil, err
+		}
```
Skarlso, line 163:
```suggestion
		if err := decoder.Decode(&value); err != nil {
			if err == io.EOF {
				break
			}
			return nil, err
		}
```

cmd/gitops/add/profiles/cmd_test.go
```diff
@@ -0,0 +1,118 @@
...
+			cmd.SetArgs([]string{
+				"add", "profile",
+				"--name", "podinfo",
+			})
+
+			err := cmd.Execute()
+			Expect(err).To(MatchError("--config-repo should be provided"))
+		})
+
+		It("fails if --config-repo is not provided", func() {
```
cPu1, line 81:
```suggestion
		It("fails if --cluster is not provided", func() {
```

cmd/gitops/add/profiles/cmd_test.go
```diff
@@ -0,0 +1,118 @@
...
+	When("a flag is unknown", func() {
+		It("fails", func() {
+			cmd.SetArgs([]string{
+				"add", "profile",
+				"--unknown", "param",
+			})
+
+			err := cmd.Execute()
+			Expect(err).To(HaveOccurred())
+			Expect(err).To(MatchError("unknown flag: --unknown"))
```
cPu1, line 115: Minor nit:
```suggestion
			Expect(err).To(MatchError("unknown flag: --unknown"))
```

cmd/gitops/add/profiles/cmd.go
Skarlso: Yes, but this is a function to addRunE which needs a specific signature.

cPu1: That's not a good reason to repeat this signature here. Moreover, `RunE` needs this signature: `func(*cobra.Command, []string) error`.

pkg/gitproviders/provider_org.go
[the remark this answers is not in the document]
nikimanoledaki: Agree with Gergely, also this leaves it open to make use of Github's `RepositoryContentGetOptions` to get a repo's entire filesystem recursively. 

I opened an issue about this in go-git-providers: https://github.com/fluxcd/go-git-providers/issues/143

cmd/gitops/add/profiles/cmd.go
nikimanoledaki: Changed this in the commit that addresses review feedback -  the `validateAddOptions` func only has 2 checks in it now since I moved the required flags to Cobra's feature for checking for required flags.

cmd/gitops/add/profiles/cmd_test.go
nikimanoledaki: Done!

pkg/services/automation/generator.go
nikimanoledaki: Thanks both, addressed this in the latest commit :)

pkg/services/gitopswriter/gitopswriter.go
bigkevmcd: If you use the app hash, you can only have one PR for an app at a time...which I guess is fine for some cases...

cmd/gitops/add/profiles/cmd.go
```diff
@@ -0,0 +1,131 @@
...
+		return fmt.Errorf("--name value is too long: %s; must be <= %d characters",
+			opts.Name, models.MaxKubernetesResourceNameLength)
+	}
+
+	if strings.HasPrefix(opts.Name, "wego") {
+		return fmt.Errorf("the prefix 'wego' is used by weave gitops and is not allowed for a profile name")
+	}
+
+	if opts.ConfigRepo == "" {
+		return errors.New("--config-repo should be provided")
```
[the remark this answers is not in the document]
nikimanoledaki: Yes! Changed this :)

cmd/gitops/add/profiles/cmd.go
```diff
@@ -0,0 +1,131 @@
...
+			return err
+		}
+		opts.Namespace = ns
+
+		config, err := clientcmd.BuildConfigFromFlags("", filepath.Join(homedir.HomeDir(), ".kube", "config"))
+		if err != nil {
+			return fmt.Errorf("error initializing kubernetes config: %w", err)
+		}
+
+		clientSet, err := kubernetes.NewForConfig(config)
```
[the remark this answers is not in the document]
nikimanoledaki: They're different types unfortunately!

package-lock.json
```diff
@@ -51,9 +51,9 @@
           }
         },
```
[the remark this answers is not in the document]
nikimanoledaki: Done! Thank you :)

cmd/gitops/add/profiles/cmd.go
nikimanoledaki: Thanks, I added a `--kubeconfig` flag that defaults to `filepath.Join(homedir.HomeDir()` or could be set to `$KUBECONFIG` as an alternative. Other parts of the CLI are also missing this feature so hopefully we will address this in a better way at some point.

cmd/gitops/add/profiles/cmd.go
nikimanoledaki: Thank you for this suggestion, I added this in the latest commit. :)

cmd/gitops/add/profiles/cmd.go
nikimanoledaki: Done!

pkg/gitproviders/provider_org.go
nikimanoledaki: That's true, the alternative would be to `squash` merge but this func is only ever used to just merge 1 commit, so you're right that there's no need for `mergeMethod` - removed it in the latest commit!

cmd/gitops/add/profiles/cmd.go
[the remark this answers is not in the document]
bigkevmcd: Do we need to use the versioned clientset ?

pkg/services/gitopswriter/gitopswriter.go
nikimanoledaki: `GetAppHash`  doesn't generate a string that is actually random + unique! It converts the input using `md5.Sum()`, hence the duplicates: https://github.com/weaveworks/weave-gitops/blob/main/pkg/services/automation/generator.go#L393-L395

cmd/gitops/add/profiles/cmd_test.go
nikimanoledaki: Removed this by using Cobra's required flags feature :)

cmd/gitops/add/profiles/cmd.go
nikimanoledaki: Addressed! Thanks both :)

pkg/helm/releases_test.go
```diff
@@ -0,0 +1,44 @@
...
+				Name:      name,
+				Namespace: ns,
+			},
+			AvailableVersions: []string{"6.0.0", "6.0.1"},
+		}
+	})
+
+	It("creates a helm release", func() {
+		hr := helm.MakeHelmRelease(profile, cluster, ns)
+		Expect(hr.Name).To(Equal(cluster + "-" + name))
```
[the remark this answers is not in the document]
nikimanoledaki: I changed this to use `go-cmp`, it's great! Thanks for the tip 😄  cc @aclevername who left a comment about this too!

cmd/gitops/add/profiles/cmd.go
```diff
@@ -0,0 +1,123 @@
...
+		`,
+		RunE: addProfileCmdRunE(),
+	}
+
+	cmd.Flags().StringVar(&opts.Name, "name", "", "Name of the profile")
+	cmd.Flags().StringVar(&opts.Version, "version", "latest", "Version of the profile specified as semver (e.g.: 0.1.0) or as 'latest'")
+	cmd.Flags().StringVar(&opts.ConfigRepo, "config-repo", "", "URL of external repository (if any) which will hold automation manifests")
+	cmd.Flags().StringVar(&opts.Cluster, "cluster", "", "Name of the cluster to add the profile to")
+	cmd.Flags().StringVar(&opts.ProfilesPort, "profiles-port", server.DefaultPort, "Port the Profiles API is running on")
+	cmd.Flags().BoolVar(&opts.AutoMerge, "auto-merge", false, "If set, 'gitops add profile' will merge automatically into the set --branch")
```
aclevername, line 49:
```suggestion
	cmd.Flags().BoolVar(&opts.AutoMerge, "auto-merge", false, "If set, 'gitops add profile' will merge automatically")
```

doesn't exist yet :P

aclevername: raises question what does it merge to :joy:, `main` or `master`? or is it the default on the repo?

tools/testcrds/helm.toolkit.fluxcd.io_helmreleases.yaml
```diff
@@ -429,10 +429,10 @@ spec:
                       minLength: 1
                       type: string
                     optional:
-                      description: Optional marks this ValuesReference as optional. When set, a not found error for the values reference is ignored, but any ValuesKey, TargetPath or transient error will still result in a reconciliation failure.
+                      description: Optional marks this ValuesReference as optional. When set, a not found error for the values reference is ignored, but any ValuesKey, dirPath or transient error will still result in a reconciliation failure.
                       type: boolean
-                    targetPath:
-                      description: TargetPath is the YAML dot notation path the value should be merged at. When set, the ValuesKey is expected to be a single flat value. Defaults to 'None', which results in the values getting merged at the root.
+                    dirPath:
+                      description: dirPath is the YAML dot notation path the value should be merged at. When set, the ValuesKey is expected to be a single flat value. Defaults to 'None', which results in the values getting merged at the root.
```
aclevername, line 435: @nikimanoledaki any idea what this is :thinking:

aclevername (approved):

nikimanoledaki:
Hey folks 👋 @aclevername @Skarlso @cPu1 @Himangini @bigkevmcd @jpellizzari 

Thank you all for the reviews! I've responded to and/or addressed all feedback. I didn't want to spam everyone by responding with `Done!` 50 times so I marked each comment that was addressed and completed with a "👍". The commit that addresses all review feedback (not linking it since it'll change anyway) also contains a list of all the requested changes for this round of reviews. :)

Feel free to mark any leftover comment as resolved and/or review and comment more!

PS: integration tests + linter currently pass ✅

pkg/services/gitopswriter/gitopswriter.go
Himangini: can we not use UUID then instead? 
`NewBranch:     uuid.New().String(),` 🤔

Himangini (commented):
Well done Niki 👏🏻 , I have gone through all the files and have a few minor comments, otherwise the PR looks pretty good to me 🎉

Himangini:
For Documentation, adding something under https://docs.gitops.weave.works/docs/cluster-management/profiles section should be sufficient I think. cc @sympatheticmoose @josefaworks wdyt?

pkg/services/profiles/add.go
```diff
@@ -0,0 +1,195 @@
...
+				if err := yaml.Unmarshal(manifestBytes, &r); err != nil {
+					return gitprovider.CommitFile{}, fmt.Errorf("error unmarshaling %s: %w", models.WegoProfilesPath, err)
+				}
+
+				if profileIsInstalled(r, *newRelease) {
+					return gitprovider.CommitFile{}, fmt.Errorf("version %s of profile '%s' already exists in namespace %s", r.Spec.Chart.Spec.Version, r.Name, r.Namespace)
+				}
+			}
+
+			content = *f.Content
```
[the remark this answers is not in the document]
Skarlso: Yes, however, it's a good point that if we only care about a single path, we should break the loop at this point, since we found it, there is no point in letting it run over the rest of the files. :)

pkg/services/profiles/add.go
```diff
@@ -0,0 +1,195 @@
...
+	configRepoUrl, err := gitproviders.NewRepoURL(opts.ConfigRepo)
+	if err != nil {
+		return fmt.Errorf("failed to parse url: %w", err)
+	}
+
+	repoExists, err := gitProvider.RepositoryExists(ctx, configRepoUrl)
+	if err != nil {
+		return fmt.Errorf("failed to check whether repository exists: %w", err)
+	} else if !repoExists {
+		return fmt.Errorf("repository '%v' could not be found", configRepoUrl.String())
```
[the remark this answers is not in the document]
bigkevmcd: It should be `%q` and then `configRepoUrl` (which should be [configRepoURL](https://github.com/golang/go/wiki/CodeReviewComments#initialisms)) wouldn't need to have `.String()` called on it.

You almost never need to use `.String()` on things you pass to the fmt functions as this https://pkg.go.dev/fmt explains:

> 3. If the %v verb is used with the # flag (%#v) and the operand implements the GoStringer interface, that will be invoked.
>
>
>
> 5. If an operand implements method String() string, that method will be invoked to convert the object to a string, which will then be formatted as required by the verb (if any).

pkg/services/automation/generator.go
```diff
@@ -410,6 +411,18 @@ func GenerateResourceName(url gitproviders.RepoURL) string {
...
+	data := "abcdefghijklmnopqrstuvwyz1234567890"
+
+	b := make([]byte, 5)
+	for i := range b {
+		b[i] = data[rand.Intn(len(data))]
+	}
+
+	return s + string(b)
+}
+
```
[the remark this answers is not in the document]
bigkevmcd: It's still not limiting the length of the string that's being returned...which could lead to errors.

Skarlso (approved):
LGTM with that one last `break` addition. :) And Kevin's comment about the possible too long name.

pkg/services/automation/generator.go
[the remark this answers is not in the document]
nikimanoledaki: @bigkevmcd just to make sure I understand correctly: since the `len` of the generated byte slice `b` will always be 5, do you mean that we should limit the input string `s`? What would you suggest as a limit?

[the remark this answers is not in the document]
bigkevmcd: There is a limit on branch names (at least on GitHub)

https://stackoverflow.com/questions/60045157/what-is-the-maximum-length-of-a-github-branch-name

And you might be able to take some ideas from https://github.com/gitops-tools/pkg/blob/main/names/generator.go#L25

cmd/gitops/add/profiles/cmd.go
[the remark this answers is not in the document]
nikimanoledaki: @bigkevmcd I've never used a versioned clientset before but from a quick search, it sounds like a good idea. However, since this could benefit a lot of `gitops`' `cmd`s (especially since most of them use `NewKubeHTTPClient` without making use of the `clientSet` that is returned), I suggest opening an issue to improve all of them rather than just `add profiles`!

pkg/services/gitopswriter/gitopswriter.go
nikimanoledaki: @bigkevmcd?

bigkevmcd: I don't feel too strongly about this, if UUIDs are better for the user-experience (which is what I think @Himangini  is implying) I'm fine with that too.

nikimanoledaki: Let's use `uuid.New().String()` since it's the easiest thing and wait for product/UX feedback. :)

Himangini: Along with the UX improvement, I was thinking this saves us from writing our own string generator and then testing it + less code to maintain 😅

cmd/gitops/add/profiles/cmd.go
```diff
@@ -0,0 +1,122 @@
...
+	cmd := &cobra.Command{
+		Use:           "profile",
+		Short:         "Add a profile to a cluster",
+		SilenceUsage:  true,
+		SilenceErrors: true,
+		Example: `
+		# Add a profile to a cluster
+		gitops add profile --name=podinfo --cluster=prod --version=1.0.0 --config-repo=ssh://git@github.com/owner/config-repo.git
+		`,
+		RunE: addProfileCmdRunE(),
```
Himangini, line 41: Are we skipping the `apply` alias for add command in this PR?

nikimanoledaki: I didn't add it because it isn't really what `add profile` does (it does not reconcile), and neither `add cluster` nor `add app` have the `apply` alias. It seems like a legacy thing that we did with pctl that doesn't fit so well with the new profile implementation, but I'm open to changing this if necessary.

Himangini (approved):
🎉 👍🏻

PR #1366, opened 2022-01-28
Added prototype endpoints to Add and List helmcharts

josecordaz:
Closes: 

**What changed?**

**Why?**

**How did you test it?**

**Release notes**

**Documentation Changes**


core/server/sources.go
```diff
@@ -110,3 +110,54 @@ func (as *appServer) ListHelmRepositories(ctx context.Context, msg *pb.ListHelmR
...
+
+func (as *appServer) ListHelmCharts(ctx context.Context, msg *pb.ListHelmChartReq) (*pb.ListHelmChartRes, error) {
+	k8s, err := as.k8s.Client(ctx)
+	if err != nil {
+		return nil, doClientError(err)
+	}
+
+	list := &sourcev1.HelmChartList{}
+
+	var opts client.MatchingLabels
```
jpellizzari, line 140: I think it might be a good time to turn this logic into a helper func, since we need to repeat it in several places. WDYT?

core/server/sources_test.go
```diff
@@ -144,3 +144,152 @@ func TestListHelmRepositories(t *testing.T) {
...
+		Chart:    "mychart",
+		Version:  "v0.0.0",
+		Interval: &pb.Interval{Minutes: 1},
+	}
+
+	addRes, err := c.AddHelmChart(ctx, r)
+	g.Expect(err).NotTo(HaveOccurred())
+	g.Expect(addRes.Success).To(BeTrue())
+
+	unAssociatedKustomizationReq := &pb.AddHelmChartReq{
```
jpellizzari, line 263: Variable name :). Also, we are repeating ourselves here in a couple places. Can anything be refactored in a helper func?

api/app/source.proto
```diff
@@ -125,4 +125,37 @@ message ListHelmRepositoryReq {
...
+}
+
+message AddHelmChartReq {
+  string namespace = 1;
+  string app_name = 2;
+  string name = 3;
+  SourceRef sourceRef = 4;
+  string chart = 5;
+  string version = 6;
+  Interval interval = 7;
```
palemtnrider, line 147: Can we reuse the HelmChart message here instead of repeating the fields?  i.e.
`HelmChart helm_chart =3;`

core/server/types/helmchart.go
```diff
@@ -0,0 +1,71 @@
...
+			Labels:    labels,
+		},
+		Spec: v1beta1.HelmChartSpec{
+			Chart:   helmChartReq.Chart,
+			Version: helmChartReq.Version,
+			SourceRef: v1beta1.LocalHelmChartSourceReference{
+				Kind: helmChartReq.SourceRef.Kind.String(),
+				Name: helmChartReq.SourceRef.Name,
+			},
+			Interval: metav1.Duration{Duration: time.Minute * 1},
```
palemtnrider, line 31: I may be misreading this but it doesn't appear to be using the actual interval

core/server/types/helmchart.go
```diff
@@ -0,0 +1,71 @@
...
+	hr := &pb.HelmChart{
+		Name:      helmchart.Name,
+		Namespace: helmchart.Namespace,
+		SourceRef: &pb.SourceRef{
+			Kind: kind,
+			Name: helmchart.Name,
+		},
+		Chart:   helmchart.Spec.Chart,
+		Version: helmchart.Spec.Version,
+		Interval: &pb.Interval{
```
palemtnrider, line 58: Same question about the interval.

palemtnrider (commented):
A couple of suggestions/comments.

core/server/types/helmchart.go
josecordaz: This is the default time for interval when we create resources using flux cli. It would be a bit premature to think about using a different value than that. Unless we want to have the default value in the UI and leave the user the option to change it. But as this are prototypes still probably not something we are interested in at this point.

jpellizzari (approved):
LGTM :+1:

Issue #1361, opened 2022-01-28
Add Pagination to Data Table


joshri:
Needs to Have:

- [ ] Row selector dropdown with options passed as prop
- [ ] Text "currently displayed rows out of total rows"
- [ ] Buttons that move forward and back one page as well move to the beginning and end
- [ ] Unit tests 

![image](https://user-images.githubusercontent.com/65822698/151589844-31febfa8-d945-4a51-a935-6f33cf43b8f4.png)

PR #1367, opened 2022-01-28
Adds Pagination to Data Table

joshri:
Closes: #1361

Adds pagination functionality to the data table + unit tests

tested in react storybook `npm run storybook`

image

### DIFF
--- a/ui/components/CommitsTable.tsx
+++ b/ui/components/CommitsTable.tsx
@@ -98,6 +98,7 @@ function CommitsTable({
           { label: "Author", value: "author" },
         ]}
         rows={commits.commits}
+        disablePagination
       />
     </div>
   );
--- a/ui/components/CommitsTable.tsx
+++ b/ui/components/CommitsTable.tsx
@@ -84,6 +84,7 @@ function CommitsTable({
         fields={[
           {
             label: "SHA",
+            displayLabel: "SHA",
             value: (row: Commit) => (
               <Link newTab href={row.url}>
                 {row.hash}
@@ -91,11 +92,12 @@ function CommitsTable({
             ),
           },
           {
-            label: "Date",
+            label: "date",
+            displayLabel: "Date",
             value: (row: Commit) => timestamp(row.date),
           },
-          { label: "Message", value: "message" },
-          { label: "Author", value: "author" },
+          { label: "message", displayLabel: "Message", value: "message" },
+          { label: "Author", displayLabel: "Author", value: "author" },
         ]}
         rows={commits.commits}
         disablePagination
--- a/ui/components/ConditionsTable.tsx
+++ b/ui/components/ConditionsTable.tsx
@@ -12,10 +12,10 @@ function ConditionsTable({ className, conditions }: Props) {
   return (
     <DataTable
       fields={[
-        { label: "Type", value: "type" },
-        { label: "Status", value: "status" },
-        { label: "Reason", value: "reason" },
-        { label: "Message", value: "message" },
+        { label: "type", displayLabel: "Type", value: "type" },
+        { label: "status", displayLabel: "Status", value: "status" },
+        { label: "reason", displayLabel: "Reason", value: "reason" },
+        { label: "message", displayLabel: "Message", value: "message" },
       ]}
       rows={conditions}
       sortFields={[""]}
--- a/ui/components/ConditionsTable.tsx
+++ b/ui/components/ConditionsTable.tsx
@@ -21,6 +21,7 @@ function ConditionsTable({ className, conditions }: Props) {
       sortFields={[""]}
       className={className}
       widths={["10%", "10%", "25%", "55%"]}
+      disablePagination
     />
   );
 }
--- a/ui/components/DataTable.tsx
+++ b/ui/components/DataTable.tsx
@@ -1,4 +1,7 @@
 import {
+  FormControl,
+  MenuItem,
+  Select,
   Table,
   TableBody,
   TableCell,
@@ -30,6 +33,10 @@ export interface Props {
   sortFields: string[];
   /** an optional list of string widths for each field/column. */
   widths?: string[];
+  /** removes bottom pagination bar. */
+  disablePagination?: boolean;
+  /** array of options for rows per page. Defaults to [25, 50, 75, 100]. */
+  paginationOptions?: number[];
 }
 
 const EmptyRow = styled(TableRow)<{ colSpan: number }>`
@@ -64,9 +71,15 @@ function UnstyledDataTable({
   rows,
   sortFields,
   widths,
+  disablePagination,
+  paginationOptions = [25, 50, 75, 100],
 }: Props) {
   const [sort, setSort] = React.useState(sortFields[0]);
   const [reverseSort, setReverseSort] = React.useState(false);
+  const [pagination, setPagination] = React.useState({
+    start: 0,
+    length: paginationOptions[0],
+  });
   const sorted = _.sortBy(rows, sort);
 
   if (reverseSort) {
@@ -101,15 +114,30 @@ function UnstyledDataTable({
     );
   }
 
-  const r = _.map(sorted, (r, i) => (
-    <TableRow key={i}>
-      {_.map(fields, (f, i) => (
-        <TableCell style={widths && { width: widths[i] }} key={f.label}>
-          <Text>{typeof f.value === "function" ? f.value(r) : r[f.value]}</Text>
-        </TableCell>
-      ))}
-    </TableRow>
-  ));
+  const r = [];
+  for (
+    let i = pagination.start;
+    i < pagination.start + pagination.length;
+    i++
+  ) {
+    if (sorted[i]) {
+      r.push(
+        <TableRow key={i}>
+          {_.map(fields, (f, index) => (
+            <TableCell style={widths && { width: widths[index] }} key={f.label}>
+              <Text>
+                {typeof f.value === "function"
+                  ? f.value(sorted[i])
+                  : sorted[i][f.value]}
+              </Text>
+            </TableCell>
+          ))}
+        </TableRow>
+      );
+    } else {
+      break;
+    }
+  }
 
   return (
     <div className={className}>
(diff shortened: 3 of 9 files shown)


ui/components/CommitsTable.tsx
```diff
@@ -84,20 +84,23 @@ function CommitsTable({
...
+            displayLabel: "Date",
             value: (row: Commit) => timestamp(row.date),
           },
-          { label: "Message", value: "message" },
-          { label: "Author", value: "author" },
+          { label: "message", displayLabel: "Message", value: "message" },
+          { label: "author", displayLabel: "Author", value: "author" },
         ]}
         rows={commits.commits}
+        disablePagination
```
jpellizzari, line 103: How about the reverse? That way, we can "opt in" in a few places instead of opting out everywhere.

ui/components/DataTable.tsx
```diff
@@ -141,6 +173,98 @@ function UnstyledDataTable({
           </TableBody>
         </Table>
       </TableContainer>
+      {/* pagination row */}
+      <Spacer padding="xs" />
+      {!disablePagination && (
```
jpellizzari, line 178: This is complex enough to warrant putting all of this block in it's own component. `<PaginationControls />` or something like that

jpellizzari (commented):
One thing we need to keep in mind is what pagination is used for on the back end; we use pagination to show a subset of all rows in the database. 

When the controls are clicked, we need to fetch the next page from the back-end, then show that data in the table.

Instead of paging in the front end, we can simplify the logic a bit to just have click handlers and let the component that wraps the `<DataTable />` figure out how to fetch the page.

Let me know if this makes sense or if you want to discuss on a call.

jpellizzari (commented):
I think I see a way to make this API a little nicer. What about:

```jsx
function MyComponent() {
  const handleForward = () => {
    console.log("forward!");
  };

  const handleBack = () => {
    console.log("back");
  };
  return (
    <DataTable rows={[]} fields={[]}>
      <PaginationControls onForward={handleForward} onBack={handleBack} />
    </DataTable>
  );
}
```
Or, if you don't like having the controls as children, what about as a prop:

```jsx

function MyComponent() {
  const handleForward = () => {
    console.log("forward!");
  };

  const handleBack = () => {
    console.log("back");
  };
  return (
    <DataTable
      rows={[]}
      fields={[]}
      paginationControls={
        <PaginationControls onForward={handleForward} onBack={handleBack} />
      }
    />
  );
}

```

This lifts the controls up and you avoid prop drilling.

ui/components/Pagination.tsx
```diff
@@ -0,0 +1,113 @@
...
+  /** CSS MUI Overrides or other styling. */
+  className?: string;
+  /** func for forward one page button */
+  forward: () => void;
+  /** func for skip to last page button */
+  skipForward: () => void;
+  /** func for back one page button */
+  back: () => void;
+  /** func for skip to start button */
+  skipBack: () => void;
```
jpellizzari, line 20: Typically, you would prefix an event handler with `on` to indicate its a handler: `onSkipBack` etc

bia:
It's beautiful!

Some design-system nits (that @mmoulian and I will prepare better for you in the future)
- All of the text should be `14px` instead of `16px`and `20px`, with the title weight semi-bold (not bold)
- The heading/title which is sorting the table (with the arrow) should change color to `Application/text/dark` or `#1a1a1a`
- could 'No rows' be renamed 'No data'? We're thinking something like this : 
Screenshot 2022-02-01 at 14 24 16
(icon svg [here](https://drive.google.com/file/d/1orX5jeBpj25Tepww_ITmcJodcQbZlcKp/view?usp=sharing))

ui/components/ConditionsTable.tsx
```diff
@@ -12,10 +12,10 @@ function ConditionsTable({ className, conditions }: Props) {
   return (
     <DataTable
       fields={[
-        { label: "Type", value: "type" },
-        { label: "Status", value: "status" },
-        { label: "Reason", value: "reason" },
-        { label: "Message", value: "message" },
+        { label: "type", displayLabel: "Type", value: "type" },
```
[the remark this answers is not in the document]
jpellizzari: I would go with `string.toLowerCase()` myself. As it is in this PR, it is a leaky abstraction, and you are asking the user to think about how the sorting logic works.

ui/components/Link.tsx
```diff
@@ -13,7 +13,11 @@ type Props = {
 };
 
 function Link({ children, href, className, to = "", newTab, ...props }: Props) {
-  const txt = <Text color="primary">{children}</Text>;
+  const txt = (
+    <Text size="small" color="primary">
```
jpellizzari, line 17: We want to change all `<Link />` components to have small text?

ui/components/Pagination.tsx
```diff
@@ -0,0 +1,118 @@
...
+  /** func for forward one page button */
+  onForward: () => void;
+  /** func for skip to last page button */
+  onSkipForward: () => void;
+  /** func for back one page button */
+  onBack: () => void;
+  /** func for skip to start button */
+  onSkipBack: () => void;
+  /** onChange func for perPage select */
+  onSelect: (value) => void;
```
jpellizzari, line 22: Let's give this `value` a type. Is it a string? Number?

ui/components/Pagination.tsx
```diff
@@ -0,0 +1,118 @@
...
+        {index + 1} - {index + length} out of {totalObjects}
+      </Text>
+      <Spacer padding="base" />
+      <Flex>
+        <Button
+          color="inherit"
+          variant="text"
+          aria-label="skip to first page"
+          disabled={index === 0}
+          onClick={() => onSkipBack()}
```
jpellizzari, line 80:
```suggestion
          onClick={onSkipBack}
```

ui/components/__tests__/Pagination.test.tsx
```diff
@@ -0,0 +1,91 @@
+import { fireEvent, render, screen, within } from "@testing-library/react";
+import "jest-styled-components";
+import React from "react";
+import { withTheme } from "../../lib/test-utils";
+import Pagination from "../Pagination";
+
+describe("Pagination", () => {
+  const onForward = jest.fn(() => "");
+  const onSkipForward = jest.fn(() => "");
```
jpellizzari, line 9: This should work:

```suggestion
  const onSkipForward = jest.fn();
```

ui/components/__tests__/Pagination.test.tsx
```diff
@@ -0,0 +1,91 @@
...
+    const forward = screen.getByLabelText("forward one page");
+    const skipForward = screen.getByLabelText("skip to last page");
+    const displayText = screen.getByText(/6 - 6 out of 10/);
+    fireEvent.click(forward);
+    expect(onForward).toHaveBeenCalledTimes(1);
+    fireEvent.click(skipForward);
+    expect(onSkipForward).toHaveBeenCalledTimes(1);
+    fireEvent.click(skipBack);
+    expect(onSkipBack).toHaveBeenCalledTimes(1);
+    fireEvent.click(back);
```
jpellizzari, line 39: Nice work on the event handler tests :+1:

ui/components/__tests__/Pagination.test.tsx
```diff
@@ -0,0 +1,91 @@
...
+    expect(onBack).toHaveBeenCalledTimes(1);
+    expect(displayText).toBeTruthy();
```
jpellizzari, line 41: What should this display text actually be? `.toBeTruthy()` will return ok on literally any string.

jpellizzari (commented):
Looks good, save for the blockers around `displayLabel` and `<Link />` text sizes

ui/components/Pagination.tsx
joshri: it's an event, and it needs to always be an event right? Is there a way to express that in Typescript?

ui/components/ConditionsTable.tsx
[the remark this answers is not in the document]
joshri: The only other thing that's coming to mind is if someone wants to sort columns with a label that won't match when it's shifted to lowercase. Like "Last Commit" with `last_commit` as the key. It makes those columns  not sortable right?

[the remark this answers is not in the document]
jpellizzari: You can do stuff like `_.lowerCase(_.snakeCase(string))` if you want to normalize the values. I would have to see the full filtering implementation to say for sure.

ui/components/Pagination.tsx
jpellizzari: Something like `React.SyntheticEvent` or some such:

```suggestion
  onSelect: (ev: React.SyntheticEvent) => void;
```

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/14d95eb0fe90f5e0579c49df136cccdfe89b2855/types/react/index.d.ts#L1167

You can also just leave it blank if you don't expect the user to do anything with the arg.

jpellizzari (approved):
LGTM, save for the `displayLabel` thing

ui/components/__tests__/Pagination.test.tsx
joshri: Yes but the test will fail if it does not match /6 - 6 out of 10/ because it won't find the string at all

ui/components/DataTable.tsx
```diff
@@ -134,24 +138,44 @@ function UnstyledDataTable({
...
+      {/* optional pagination component */}
+      {children}
     </div>
   );
 }
 
 export const DataTable = styled(UnstyledDataTable)`
   h2 {
+    font-size: 14px;
+    font-weight: 600;
```
bia, line 166: I have played around with this in the inspector and for me this only renders bold (> 600) or regular (<600) – is semi-bold Proxima Nova not getting imported somewhere?

bia (approved):

ui/components/DataTable.tsx
joshri: Hmmmm...I feel like I've seen that before, but I just tested it here's the screenshots:

font-weight 800:
image

font-weight 600:
image

joshri: It DEFINITELY changes but it is subtle for sure

PR #1375, opened 2022-01-31
Added prototype endpoints for adding and listing helm releases

josecordaz:
Closes: 

**What changed?**

**Why?**

**How did you test it?**

**Release notes**

**Documentation Changes**


jpellizzari:
This is all @josecordaz 's work, I was showing how to do a rebase, so my name pops up on the commits as well.

jpellizzari (commented):
@josecordaz Is a helm release valid for use as a source? I think flux treats it like an "Automation"

josecordaz:
> @josecordaz Is a helm release valid for use as a source? I think flux treats it like an "Automation"

I don't think it cab be used as a source but we can reference a helm chart which needs to be associated with one of this 3 sources https://github.com/fluxcd/helm-controller/blob/main/config/crd/bases/helm.toolkit.fluxcd.io_helmreleases.yaml#L86.

jpellizzari (approved):

PR #1377, opened 2022-02-01
Expose test fakes as a package

yiannistri:
**What changed?**
- Exposed test fakes as a separate package
- Added missing interface checks

**Why?**
To allow use of fake objects from other projects that import Weave GitOps

**How did you test it?**
No new tests, mostly moved code around

**Release notes**
N/A

**Documentation Changes**
N/A


pkg/fakes/fakes.go
```diff
@@ -0,0 +1,55 @@
+package fakes
```
foot, line 1: The pattern I have seen around is to nest the fakes under the main implementation, could that work or might introduce cyclic deps?

pkg/server/config_gettter_test.go
```diff
@@ -5,12 +5,13 @@ import (
 	"testing"
```
foot, line 5:
```patch
- pkg/server/config_gettter_test.go
+ pkg/server/config_getter_test.go
```

pkg/fakes/fakes.go
yiannistri: Let me try that 👍

yiannistri: Works fine 👌

foot (approved):
All looks good to me! 👍  Nice syntactic cleanup w/ the `NewDefaultKubeGetter`

PR #1387, opened 2022-02-03
Add instructions on `gitops upgrade` for hosted/on-prem gitlab

foot:
**What changed?**

- Adds some docs around how to configure gitlab on weave-gitops-enterprise 
- Remove `{{ frontMatter.title }}` trick, it doesn't work properly w/ the page metadata and so the string `{{ frontMatter.title }}` appears in the slack link preview when pasting links there.

**Why?**

We only had github until now.


website/docs/enterprise/upgrading.mdx
```diff
@@ -84,15 +158,35 @@ The example templates provided in this guide have been tested with `clusterctl`
 Download a specific version of clusterctl from the [releases page](https://github.com/kubernetes-sigs/cluster-api/releases).
 :::
 
-In order to be able to provision Kubernetes clusters, a CAPI provider needs to be installed.
+In order to be able to provision Kubernetes clusters, a CAPI provider needs to be installed. See [Cluster API Providers](../cluster-management/cluster-api-providers.mdx) page for more details on providers.
+He're we'll continue with our example instructions for CAPD and CAPA.
```
yiannistri, line 162: Should this be `Here` ?

foot: Good catch, the internet concurs that _He're_ is not a valid contraction of anything.

foot:
```suggestion
Here we'll continue with our example instructions for CAPD and CAPA.
```

PR #1391, opened 2022-02-04
Add groups scope

yiannistri:
**What changed?**
Added `groups` scope so that OIDC returns this information if available during login.

**Why?**
To allow platform admins to set RBAC permissions using groups.

**How did you test it?**
Tested manually

**Release notes**
Not needed

**Documentation Changes**
Not needed


foot (approved):
Nice, make sense

And groups! ⭐

PR #1401, opened 2022-02-07
Add deploy key functionality for AddGitRepo page

jpellizzari:
Adds deploy key creation to the UI for private git repos:
https://drive.google.com/file/d/1L9b5DOshibxEOagyM2uMO-YSlIduHFR7/view?usp=sharing

Only works for GitHub as of this PR.

@JamWils PTAL at the UX

@Callisto13 @josecordaz WDYT about the stubbing out of the GitHub API in the `RoundTripper`, good idea or bad idea?


Callisto13:
I am not sure about the context of this PR (have no idea what a deploy key is for, but I can leave that aside for now).

1. Is this anything to do with either thread of work (clear, observability)?
2. Is any part of this likely to be removed by clearance? (thus I should ignore it)
3. Is this part of another stream of product work and is there a ticket I can read?

core/server/app_test.go
```diff
@@ -0,0 +1,133 @@
+package server
```
Callisto13, line 1: this should probably be `server_test`, but I can see that the rest of the files in this suite are written as internal tests although the interfaces are external facing, so this could lead to the contract boundary being blurred. do you know the history of this? it may be irrelevant: the clearance team will probs get to it.

Callisto13:
> WDYT about the stubbing out of the GitHub API in the RoundTripper, good idea or bad idea?

I think it's an alright idea, although there is the old adage that you "should never mock what you don't own". But we need to test the thing and shy of an integration test we want it nice and light at this level.

Having said that I feel we are pulling it out at the wrong layer of abstraction maybe? Unless these are meant to be integration tests? In `core/server/app_test.go` we are directly testing `core/services/deploykey/deploykey.go`, and implicitly testing `pkg/gitproviders/provider.go`. This is a long way down. We could get finer grain tests by testing the deploykey `Manager` interface in `core/server/app_test`, the provider interface at `core/services/deploykey/deploykey.go`, and the "real mcoy" in `pkg/gitproviders/provider.go` which looks to be a thin-ish shim around the bits that actually talk to GitHub (or whatever lib use that ends up talking to github*). This would give us much finer-grained lighter-weight tests at each layer which would be more robust (commenting out bits of the code and running the tests show how only some happy paths and fewer sad paths are covered).

> \* if we only communicate with Git{Hub/Lab} through other libraries, then can we not test only on interfaces? Otherwise we will be testing those libraries for them which would be weird.

This may take some refactoring featuring some more DI. But if this is the sort of thing which will be taken care of by the clearance folks, then we can leave it.

Caveat to all the above: I am of course brand new to this code so if I has missed stuff which makes the above rubbish please lmk 😄 .

Questions (which I may have found the answers to by the time you read this):
1. Does wego ever talk to Git directly itself? Or does it use some flux library or other for everything? (I think it is the latter from what I have seen so far). 
2. coming soon once i remember it...

Callisto13 (commented):
initial comments, more coming soon i just need a break

core/server/app_test.go
jpellizzari: We have been using internal tests as a convention. My understanding is that this is the current flavor for the `go` community. Being able to write tests for complex un-exported funcs seems like an advantage; any strong feelings either way?

Callisto13: I tend to lean towards writing exported tests wherever possible as it encourages me to design along clean lines, keep things in isolated units, maintain clear "contracts" between components, etc. I like internal tests for utils or helper packages, basically things which have no interface or are not designed to be consumed by anything else.
I think in this case since all the funcs being tested are exported, and it is only the `appServer` object which isn't, we could probably test the package externally.

But I am new to this code so I am fine going with and coming to understand convention and don't need to double down on any opinions right now.

jpellizzari:
> Is this anything to do with either thread of work (clear, observability)?

This is a prototype @josecordaz and I have been working on with James that (hopefully) aligns with the new initiatives.

> Is any part of this likely to be removed by clearance? (thus I should ignore it)

Everything in the `/core` directory should stick around, although it may be moved to a more canonical folder in the future. The `/core` dir was an effort not to disturb the other source code in `pkg`.

> Is this part of another stream of product work and is there a ticket I can read?

Nope, this is pulled from a design that can be found here (let me know if you need access): https://www.figma.com/file/IVHnM9iyeFWpd11evtY8ux/Weave-GitOps?node-id=4355%3A3124

> But... if we only communicate with Git{Hub/Lab} through other libraries, then can we not test only on interfaces and not have a roundtripper at all? Otherwise it feels like we will be testing those libraries for them which would be weird.

> Caveat to all the above: I am of course brand new to this code so if I have missed stuff (like if I have completely missed the reason why you added the roundtripper) which makes the above rubbish please lmk smile.

Maybe this is a chance to have a more philosophical discussion about "how much mocking is too much mocking"? If all the interfaces are stubbed out, are we really testing anything?

> But I am struggling to see the reason we need it based on the things we have, and if we do, are we pulling it out at the wrong layer of abstraction? Unless these are meant to be integration tests?

The single test included in the PR aims to be a bit integration-y, since it is testing a handler.

> But... if we only communicate with Git{Hub/Lab} through other libraries, then can we not test only on interfaces and not have a roundtripper at all? Otherwise it feels like we will be testing those libraries for them which would be weird.

:+1: 

> 1. Does wego ever talk to Git directly itself? Or does it use some flux library or other for everything? (I think it is the latter from what I have seen so far).

Yes, we use `go-git` in several places. We also use `go-git-providers` to duplicate all the code paths :smile:.

I'll take another pass at this and re-request a review.

core/services/deploykey/deploykey.go
```diff
@@ -0,0 +1,58 @@
...
+	Create(ctx context.Context, name types.NamespacedName, repoURL gitproviders.RepoURL, token string) (models.GeneratedSecretName, error)
+}
+
+func NewManager(k8s client.Client, rt http.RoundTripper) Manager {
+	return manager{k8s: k8s, http: rt}
+}
+
+type manager struct {
+	k8s  client.Client
+	http http.RoundTripper
```
josecordaz, line 26: It looks ok to me. I think it could be useful for integration tests. Or even cache mechanisms to speed them up. This is what I used a while ago when I cached the Github api interactions to speed unit tests/integration tests up.

josecordaz (approved):
LGTM

jpellizzari:
No longer relevant. Closing.

PR #1407, opened 2022-02-08
Use snyk for js, go & docker, upload results properly

ozamosi:
Run snyk against javascript, go, and use the CLI directly to scan both                                                    
in one invocation.                                                                                                        
                                                                                                                          
Also run snyk against docker (separately, in parallel). It's a bit                                                        
slow right now, but a) I expect to speed this up later b) we'll have                                                      
to build the docker image anyway, and the scan takes about a second                                                       
extra c) this is still faster than the v2 pipeline.       
                                                                                                                          
Upload the snyk test results to github even if they fail - otherwise                                                      
what's the point.                                                                                                         
                                                                                                                          
Upload main branch updates to snyk directly (this will also trigger                                                       
for v2) - this means they can report newly discovered security                                                            
problems with the main branch. Don't upload PRs though - I can't see                                                      
that snyk is particularly branch aware.

Remove CodeQL - it takes over 20 minutes, which isn't anywhere near                                                       
acceptable for interactive PRs. We can add it back as a nightly job,                                                      
if it does anything snyk doesn't.

### DIFF
--- a/.github/workflows/scan.yaml
+++ b/.github/workflows/scan.yaml
@@ -4,9 +4,11 @@ on:
   push:
     branches:
       - main
+      - v2
   pull_request:
     branches:
       - main
+      - v2
   workflow_dispatch:
 
 jobs:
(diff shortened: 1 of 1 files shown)


jpellizzari (approved):
LGMT, but 8m to scan a docker container seems excessive.

ozamosi:
8 minutes to _build_ the docker image - it's 7 seconds to scan it.

Issue #1406, opened 2022-02-08
Tilt up results in service account not found error in wego-app


Skarlso:
Run:

- `tools/kind-with-registry.sh`
- `make cluster-dev`

Observe wego-app logs:

```
[K8s EVENT: ReplicaSet wego-app-68ffdb4f5f (ns: wego-system)] Error creating: pods "wego-app-68ffdb4f5f-" is forbidden: error looking up service account wego-system/wego-app-service-account: serviceaccount "wego-app-service-account" not found
[K8s EVENT: ReplicaSet wego-app-68ffdb4f5f (ns: wego-system)] Error creating: pods "wego-app-68ffdb4f5f-" is forbidden: error looking up service account wego-system/wego-app-service-account: serviceaccount "wego-app-service-account" not found
```

Skarlso:
I remember now... That's by design actually.... You should already have a working cluster before running tilt. That said, I think that's absolutely counterintuitive. I think `make cluster-dev` should be able to make a proper, running, dev cluster without the need to set up ANYTHING previously.

PR #1408, opened 2022-02-08
Modify the local build to install and configure tilt for first runners

Skarlso:
Tilt install assume you already have a working cluster. That's not really needed.

Now that we have acceptance test change how we handle the image in the real deployment yaml file, we can bootstrap a dev environment and then let tilt take over so it monitors all files.

Screenshot 2022-02-08 at 12 53 25

This is now working with a booted up dev cluster from 0 -> 100 without running anything before hand just this:

```
TEST_CONFIG_REPO=ssh://git@github.com/Skarlso/gitops-training.git make cluster-dev
```

There is also a convenient make target to clean everything up: `make clean-dev-cluster`.

# Question

Do people would like to have an option to disable the bootstrap part and still use tilt with an existing cluster?

Closes https://github.com/weaveworks/weave-gitops/issues/1406

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -61,8 +61,15 @@ install: bin ## Install binaries to GOPATH
 api-dev: ## Server and watch gitops-server, will reload automatically on change
 	reflex -r '.go' -R 'node_modules' -s -- sh -c 'go run -ldflags $(LDFLAGS) cmd/gitops-server/main.go'
 
-cluster-dev: ## Start tilt to do development with wego-app running on the cluster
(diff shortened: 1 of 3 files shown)


nikimanoledaki:
awesome, trying it out now! could you also update/add instructions to the contributing guide please? :))

jpellizzari (approved):
👍 Nice work

nikimanoledaki (approved):

jpellizzari:
@Skarlso Don't we also have to set the ports to `5001` in the `./tools/kind-with-registry` script?

Skarlso:
You appear to have an old kind registry. Please delete it and try again?

PR #1410, opened 2022-02-08
No tools in docker image

ozamosi:
The tools thingy is used for two different things - tools for gitops
to use, and tools to use for dealing with gitops. Flux is the only one
of the former - we need flux in the image (for now). The others are
development tools to use for working with flux.

By not importing all the tools into the docker image, we remove a
bunch of flaky bits and speed up the docker build. Zoom!

Turn on buildkit for `make docker` too, because it's better and
faster.


jpellizzari (commented):
Can we clear all these `glibc` warnings?

jpellizzari:
@ozamosi What is the difference between these `Code Scan` jobs and the `security/snyk` jobs?

ozamosi:
> @ozamosi What is the difference between these Code Scan jobs and the security/snyk jobs?

The security/snyk jobs check project dependencies, the snyk source code scans the code for insecure code patterns, and the container one does kinda-both for the OS.

> Can we clear all these glibc warnings?

So. They were supposed to show as not-new and thus not error. I'll see if I've done something wrong tomorrow.

jpellizzari (approved):
LGTM so long as the security vulns get resolved

ozamosi:
The glibc warnings are because I'm faffing enough with the docker image to confuse snyk - but they're not new, they're part of debian, which in turn makes them part of the `distroless/base` image.

Debian doesn't seem to keen on backporting the fixes at all, or at least not with any great haste.

The easiest way around that that I can think of is #1443 which removes the debian bits of the image altogether.

PR #1423, opened 2022-02-09
use nightly-build wego-app image tag for nightly runs

### LINKED ISSUE #1421: Nightly build fails
Nightly build which deploys and runs the acceptance tests against GKE/EKS https://github.com/weaveworks/weave-gitops/actions/workflows/nightly.yaml is failing

I believe that its failing because the `wego-app` pod does not come up, and it has been like this **for a long time**. Prior to https://github.com/weaveworks/weave-gitops/pull/1400 only the `profiles` test were failing- these are the only tests that interact with the pod. Example: https://github.com/weaveworks/weave-gitops/runs/5102080314?check_suite_focus=true. Since then https://github.com/weaveworks/weave-gitops/pull/1400 was merged, and now all the tests are failing, as the `wego-app` pod not coming up causes all the tests to fail.

We need to find a way to make sure that the tests that run on GKE/EKS are using a `wego-app` image containing the latest code changes. The local kind cluster tests achieve this using a local registry https://github.com/weaveworks/weave-gitops/pull/1392. We probably need to push a dev version of the image and have that be deployed when running the tests on GKE/EKS.

aclevername:
wip closes #1421


test/acceptance/test/profiles_test.go
```diff
@@ -64,7 +64,7 @@ var _ = Describe("Weave GitOps Profiles API", func() {
 		deleteNamespace(namespace)
 	})
 
-	It("gets deployed and is accessible via the service", func() {
+	FIt("gets deployed and is accessible via the service", func() {
```
Skarlso, line 67: TODO, remove. :)

Skarlso (approved):

.github/workflows/nightly.yaml
```diff
@@ -4,6 +4,11 @@ on:
     - cron: '0 0 * * Sun'
   workflow_dispatch:
 
+env:
+  REGISTRY: ghcr.io
+  IMAGE_NAME: weaveworks/wego-app
+  USE_NIGHTLY_BUILD_TAG: "true"
```
Skarlso, line 10: One small nit. Can we have this as `1` and then just check that it's set? There is usually no point in doing a string `"true"` in bash. :)

Skarlso: Oh right, you are already doing this:
```go
	} else if os.Getenv("USE_NIGHTLY_BUILD_TAG") != "" {
		version = "nightly-build"
	}
```

so just set it to `1`. :)

PR #1424, opened 2022-02-09
Start deleting some things

ozamosi:
* This deletes all the explicitly deprecated commands, plus the profile commands.
* This also breaks e.g. cluster commands - they make PRs, and I've deleted the git code.
* This replaces install/uninstall with a printed instruction. If you change the branch name to pull from from `v2` to `start-deleting`, then I have installed it successfully. This is not automatically tested, however. Upgrade, similarly, prints a message, but it's made up.
* This leaves the `v1alpha2`/`core` application APIs. That seems to be the starting point to where the new UI will go.
* This leaves the javascript untouched, except for changing the hard-coded javascript - so if you're using the "old" UI, it won't work because the backend doesn't exist, but the "new" UI should work maybe?
* This removes a bunch of test framework-y stuff, because the old UI doesn't exist, so I got unused code errors. It's in the git history.


charts/weave-gitops/Chart.yaml
```diff
@@ -0,0 +1,24 @@
+apiVersion: v2
+name: weave-gitops
+description: Weave Gitops helps you manage your flux installation
```
jpellizzari, line 3: This probably should be it's own PR no? Seems strange to lump this into the "Delete things PR"

charts/weave-gitops/templates/role-binding.yaml
```diff
@@ -2,11 +2,11 @@ apiVersion: rbac.authorization.k8s.io/v1
 kind: RoleBinding
 metadata:
   name: read-resources
-  namespace: {{.Namespace}}
+  namespace: {{.Release.Namespace}}
```
jpellizzari, line 5: More stuff I wouldn't expect to see int he "Delete :allthetheings:" PR

charts/weave-gitops/values.yaml
```diff
@@ -0,0 +1,2 @@
+images:
+  wegoApp: ghcr.io/weaveworks/wego-app:latest
```
jpellizzari, line 2: Probably isn't the right container image. I think we would almost always want this to be a release, but I don't know if we are going to do releases in the short/medium term? Something to thing about.

jpellizzari (commented):
"Clear the way" indeed! Probably a bit too early for this considering we haven't quite nailed down scope on The New Thing :tm:. It might be nice to have some of the old stuff around for reference or re-use initially.

Also some helm chart stuff sneaked in that I would not have expected.

PR #1425, opened 2022-02-09
clean up docs references to `wego` CLI

kingdonb:
It is called `gitops` CLI now, someone pointed this out as an error during our WOUG demo of Weave GitOps today. 👍 

(I don't know if we should update the ADR too, it has some `wego` references in there. I have done it in a separate commit. If ADRs shouldn't be updated in the same PR for any reason at all, let me know and I'll withdraw that commit.)

### DIFF
--- a/cmd/gitops/flux/cmd.go
+++ b/cmd/gitops/flux/cmd.go
@@ -15,7 +15,7 @@ var Cmd = &cobra.Command{
 	Use:                "flux [flux commands or flags]",
 	Short:              "Use flux commands",
 	DisableFlagParsing: true,
-	Example:            "wego flux install -h",
+	Example:            "gitops flux install -h",
 	Run:                runCmd,
 }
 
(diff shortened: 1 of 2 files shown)


dhwthompson (commented):
Good catch, @kingdonb, and thanks for submitting the PR.

I think I’d lean towards keeping the original naming in the ADRs, and leaving them as historical artifacts. Good thinking putting that into a separate commit; if you remove that second commit, I’ll be happy to approve this.

jpellizzari (approved):
LGTM, thanks @kingdonb!

kingdonb:
OK, I removed the commit e4dca32d from the PR.

The ADR will be preserved intact for historical reference, that makes sense to me. 👍

dhwthompson (approved):
Great! Thanks for that, @kingdonb. Merge when ready.

(In case anyone is worried about all the breaking tests, I’m confident it’s not a problem with this PR. It looks like that’s a side-effect of the fact that this PR is coming from a fork, so a bunch of the secret keys we’re using in our test suite aren’t working.)

kingdonb:
(someone with collaborator status/write access will have to merge it)

dhwthompson:
I had to spin this up as #1428 to get the tests to run properly. That’s now merged, but since it was a squash-merge, GitHub hasn’t picked up that this commit is effectively now in the `main` branch. [But it is](https://github.com/weaveworks/weave-gitops/commit/0ca3c1c78b27824f1bcaa445f6d535b9dbf1d5ca).

Thanks again, @kingdonb. 🎉

PR #1428, opened 2022-02-09
clean up docs references to `wego` CLI

dhwthompson:
It is called `gitops` CLI now.

This is a duplicate of #1425, by @kingdonb. Unfortunately, some bits of our test suite that are required for merging don’t work when run against a forked repo.


dhwthompson:
I’m going to go ahead and merge this, since the only reason it exists is to make sure our tests run correctly. The one commit in this PR is the exact same commit that we’ve got over in #1425 ([`68bc89c`](https://github.com/weaveworks/weave-gitops/pull/1428/commits/68bc89cf04f104214d0af7c9e14af922b1ec2715)), which has already been approved.

PR #1430, opened 2022-02-10
Fix the notification controller address

Skarlso:
The default notification-controller endpoint is incorrect.

### DIFF
--- a/cmd/gitops/ui/run/cmd.go
+++ b/cmd/gitops/ui/run/cmd.go
@@ -75,7 +75,7 @@ func NewCommand() *cobra.Command {
 	cmd.Flags().StringVar(&options.ProfileCacheLocation, "profile-cache-location", "/tmp/helm-cache", "the location where the cache Profile data lives")
 	cmd.Flags().StringVar(&options.WatcherHealthzBindAddress, "watcher-healthz-bind-address", ":9981", "bind address for the healthz service of the watcher")
 	cmd.Flags().StringVar(&options.WatcherMetricsBindAddress, "watcher-metrics-bind-address", ":9980", "bind address for the metrics service of the watcher")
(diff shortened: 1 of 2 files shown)


cmd/gitops/ui/run/cmd.go
```diff
@@ -75,7 +75,7 @@ func NewCommand() *cobra.Command {
 	cmd.Flags().StringVar(&options.ProfileCacheLocation, "profile-cache-location", "/tmp/helm-cache", "the location where the cache Profile data lives")
 	cmd.Flags().StringVar(&options.WatcherHealthzBindAddress, "watcher-healthz-bind-address", ":9981", "bind address for the healthz service of the watcher")
 	cmd.Flags().StringVar(&options.WatcherMetricsBindAddress, "watcher-metrics-bind-address", ":9980", "bind address for the metrics service of the watcher")
-	cmd.Flags().StringVar(&options.NotificationControllerAddress, "notification-controller-address", "http://notification-controller./", "the address of the notification-controller running in the cluster")
+	cmd.Flags().StringVar(&options.NotificationControllerAddress, "notification-controller-address", "http://notification-controller.wego-system.svc.cluster.local./", "the address of the notification-controller running in the cluster")
```
Skarlso, line 78: i'm debating whether or not this should be discoverable. ¯\_(ツ)_/¯

cmd/gitops/ui/run/cmd.go
```diff
@@ -152,6 +152,11 @@ func runCmd(cmd *cobra.Command, args []string) error {
 		return fmt.Errorf("failed to create cacher: %w", err)
 	}
 
+	if options.NotificationControllerAddress == "" {
+		namespace, _ := cmd.Parent().Flags().GetString("namespace")
```
Skarlso, line 156: This won't work. This command doesn't have a parent which has this I believe? @foot Can I get this namespace from somewhere?

PR #1431, opened 2022-02-10
fix profiles tests in nightly-build

aclevername:
Two failures:
```
[Fail] Weave GitOps Profiles API [It] gets deployed and is accessible via the service 
/home/runner/work/weave-gitops/weave-gitops/test/acceptance/test/profiles_test.go:118

[Fail] Weave GitOps Profiles API [It] profiles are installed into a different namespace 
/home/runner/work/weave-gitops/weave-gitops/test/acceptance/test/profiles_test.go:140
```

First is because the `cluster-name` is incorrect, the tests wasn't picking it up from the environment variable, this PR fixes that.

Second is because `other` namespace isn't being cleaned up, so creation was failing, this PR fixes that also.

### DIFF
--- a/.github/workflows/nightly.yaml
+++ b/.github/workflows/nightly.yaml
@@ -166,7 +166,6 @@ jobs:
       run: |
         export WEGO_BIN_PATH=$(pwd)/bin/gitops-${{matrix.os}}-nightly
         export CLUSTER_PROVIDER=kubectl
-        export CLUSTER_NAME=wego-nightly-cluster-$OS_NAME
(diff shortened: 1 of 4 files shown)


Callisto13:
this happening or no?

aclevername:
> this happening or no?

I couldn't get this fixed, it was a hot mess. I'll close this PR.

The profiles tests go green on the PR & when merged, its only because of some obscure way the tests run in the nightly build that prevented it from passing, not because of the actual code. Happy to leave it as is (pended for nightly built)

PR #1435, opened 2022-02-11
Cleanup version

J-Thompson12:
**What changed?**
Removes the flux version check when running `gitops version` command. We are removing the flux binary from weave-gitops so as part of decoupling flux and gitops we need to remove the checks on flux.

**Why?**

**How did you test it?**

**Release notes**

**Documentation Changes**

### DIFF
--- a/cmd/gitops/version/cmd.go
+++ b/cmd/gitops/version/cmd.go
@@ -2,11 +2,9 @@ package version
 
 import (
 	"fmt"
+
 	log "github.com/sirupsen/logrus"
 	"github.com/weaveworks/go-checkpoint"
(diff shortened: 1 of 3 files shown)


dhwthompson (commented):
The code changes look good to me. Always nice to see how even a small change allows a bunch of complexity to fall out of the code.

Before we merge this, could you add a bit more context (either into the PR description or the Git commit messages) as to why we’re making this change? I know we understand it now, but our future selves (or other team members) might need a bit more to go on.

dhwthompson (approved):
👍 Nice context-setting. Thanks for doing that.

PR #1436, opened 2022-02-11
Add Filtering Structure for Data Table

joshri:
image

Sets up a filtering component to be used in parent component of `<DataTable />` - functionality will be passed down. I'm fully expecting to have to add some styling polish. so hit me with what needs to change. 

Check it out by pulling down the `filter-bar` branch and running `npm run storybook`

### DIFF
--- a/ui/components/FilterBar.tsx
+++ b/ui/components/FilterBar.tsx
@@ -55,7 +55,7 @@ function UnstyledFilterBar({
   const [anchorEl, setAnchorEl] = React.useState(null);
   const [showFilters, setShowFilters] = React.useState(false);
 
-  const onCheck = (e, option) => {
+  const onCheck = (e: React.ChangeEvent<HTMLInputElement>, option: string) => {
     e.target.checked
       ? setActiveFilters([...activeFilters, option])
       : setActiveFilters(
--- a/ui/components/Icon.tsx
+++ b/ui/components/Icon.tsx
@@ -12,7 +12,6 @@ import NavigateBeforeIcon from "@material-ui/icons/NavigateBefore";
 import NavigateNextIcon from "@material-ui/icons/NavigateNext";
 import RemoveCircleIcon from "@material-ui/icons/RemoveCircle";
 import SaveAltIcon from "@material-ui/icons/SaveAlt";
-import SearchIcon from "@material-ui/icons/Search";
 import SkipNextIcon from "@material-ui/icons/SkipNext";
 import SkipPreviousIcon from "@material-ui/icons/SkipPrevious";
 import * as React from "react";
@@ -38,7 +37,6 @@ export enum IconType {
   SkipPreviousIcon,
   RemoveCircleIcon,
   FilterIcon,
-  SearchIcon,
   ClearIcon,
 }
 
@@ -99,9 +97,6 @@ function getIcon(i: IconType) {
 
     case IconType.FilterIcon:
       return FilterIcon;
-
-    case IconType.SearchIcon:
-      return SearchIcon;
 
     case IconType.ClearIcon:
       return ClearIcon;


ui/components/FilterBar.tsx
```diff
@@ -0,0 +1,157 @@
...
+`;
+
+/** Form Filter Bar */
+function UnstyledFilterBar({
+  className,
+  activeFilters,
+  setActiveFilters,
+  filterList,
+}: Props) {
+  const [anchorEl, setAnchorEl] = React.useState(null);
```
jpellizzari, line 44: This should probably be a `ref`: https://reactjs.org/docs/hooks-reference.html#useref

ui/components/FilterBar.tsx
```diff
@@ -0,0 +1,157 @@
...
+              onChange={(e) => setSearch(e.target.value)}
+            />
+          </Flex>
+        </Spacer>
+        <Divider />
+        <List>
+          {Object.keys(filterList).map((header: string, index: number) => {
+            return (
+              <>
+                <ListItem key={index}>
```
jpellizzari, line 85: Does the `key=` prop need to be on the top level `React.Fragment`? I would need to look this one up myself.

ui/components/FilterBar.tsx
```diff
@@ -0,0 +1,157 @@
...
+            <Input
+              placeholder="SEARCH"
+              defaultValue={search}
+              onChange={(e) => setSearch(e.target.value)}
+            />
+          </Flex>
+        </Spacer>
+        <Divider />
+        <List>
+          {Object.keys(filterList).map((header: string, index: number) => {
```
jpellizzari, line 82: You can call `_.map()` on an object here to get access to both the key and value. That might simplify this a bit.

ui/components/FilterBar.tsx
```diff
@@ -0,0 +1,157 @@
...
+                    <List>
+                      {filterList[header].map(
+                        (option: string, index: number) => {
+                          if (search) {
+                            for (let i = 0; i < search.length; i++) {
+                              if (
+                                search[i].toLowerCase() !==
+                                option[i].toLowerCase()
+                              ) {
+                                return;
```
jpellizzari, line 97: This block is difficult to understand. Can this be converted into a `filter`?

Also, I think React will complain about returning `undefined` here.

ui/components/FilterBar.tsx
```diff
@@ -0,0 +1,157 @@
...
+                              }
+                            }
+                          }
+                          return (
+                            <ListItem key={index}>
```
jpellizzari, line 102: I think encapsulating this into it's own component might be a good idea. It has enough logic to warrant it's own component.

You can declare it at the top of this file if it is only used here.

ui/components/Text.tsx
```diff
@@ -13,7 +14,11 @@ type Props = {
 const Text = styled.span<Props>`
   font-family: ${(props) => props.theme.fontFamilies.regular};
   font-size: ${(props) => props.theme.fontSizes[props.size]};
-  font-weight: ${(props) => (props.bold ? "600" : "400")};
+  font-weight: ${(props) => {
```
jpellizzari, line 17: :+1:

ui/components/FilterBar.tsx
joshri: It doesn't work as a ref I tried it like a billion different ways. In the example in MUI docs they actually do it in state as well: https://mui.com/components/popover/

ui/components/FilterBar.tsx
joshri: I was just looking at that - it definitely needs to be in the fragment but I really hate that for some reason...it kind of ruins the whole point of the fragment right? If I make it a div it's not a big deal but I hate having a trillion divs on the page you know?

ui/components/FilterBar.tsx
joshri: oh la la I love that

ui/components/FilterBar.tsx
joshri: What's going to be undefined? I was doing the search letter by letter but I could also just filter it with a `.includes(search)` if we don't care about the order?

ui/components/FilterBar.tsx
joshri: Is it okay if I just move the onChange outside?  Defining all the props and passing it all down just seems like so many extra lines :(

ui/components/Text.tsx
joshri: 🌞

ui/components/FilterBar.tsx
```diff
@@ -0,0 +1,157 @@
...
+          </Flex>
+        );
+      })}
+      <Chip label="Clear All" onDelete={() => setActiveFilters([])} />
+    </Flex>
+  );
+}
+
+const FilterBar = styled(UnstyledFilterBar).attrs({
+  className: UnstyledFilterBar.name,
```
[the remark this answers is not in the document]
jpellizzari: Ok change the `attrs.className` to be a semantic name, ie `FilterBar`

ui/components/FilterBar.tsx
jpellizzari: From in person code review: `continue` is the magic work here

ui/components/FilterBar.tsx
jpellizzari: Sure that might clean it up a big

ui/components/FilterBar.tsx
```diff
@@ -0,0 +1,130 @@
...
+  className,
+  activeFilters,
+  setActiveFilters,
+  filterList,
+}: Props) {
+  /** why isn't this a ref? It doesn't work. In The MUI Popover docs they do it with setState too... :( https://mui.com/components/popover/ */
+  const [anchorEl, setAnchorEl] = React.useState(null);
+  const [showFilters, setShowFilters] = React.useState(false);
+
+  const onCheck = (e, option) => {
```
jpellizzari, line 58: Can we get some type annotations here? What will `e` and `option` look like when they get passed in here?

ui/components/Icon.tsx
```diff
@@ -91,6 +97,15 @@ function getIcon(i: IconType) {
     case IconType.RemoveCircleIcon:
       return RemoveCircleIcon;
 
+    case IconType.FilterIcon:
+      return FilterIcon;
+
+    case IconType.SearchIcon:
+      return SearchIcon;
+
+    case IconType.ClearIcon:
```
jpellizzari, line 106: Merge conflict incoming from this PR: https://github.com/weaveworks/weave-gitops/pull/1451

ui/components/__tests__/ChipGroup.test.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+  it("should render chips", () => {
+    render(
+      withTheme(
+        <ChipGroup activeChips={chipList} setActiveChips={setActiveChips} />
+      )
+    );
+    expect(screen.queryByText("app")).toBeTruthy();
+    expect(screen.queryByText("app3")).toBeTruthy();
+    expect(screen.queryByText("Clear All")).toBeTruthy();
```
jpellizzari, line 19: :+1:

jpellizzari (approved):
Excellent work on the tests. No blockers :+1:

PR #1440, opened 2022-02-14
cli: Remove application suspend/resume

ozamosi:
This removes the command itself, it removes the service/app functions
the commands call, it removes the flux function the service/app
functions call, it re-generates the flux fakes, and it removes
acceptance tests specifically related to suspension/resumption.

This does not remove creating a suspended source from the proto
API, as it's a different code path.

This does not remove reporting suspension status in the output of
e.g. `gitops get app podinfo` - and admittedly the acceptance tests I
deleted were verifying this functionality as well. In my defence,
`gitops get apps` is about to be deleted too, so I don't want to spend
time re-architecting those tests.


J-Thompson12 (approved):
LGTM. I dont know if we would want to remove the suspended status from the status call. That is something we would probably still want.

PR #1441, opened 2022-02-14
Remove add/remove app

ozamosi:
This removes it from both the protobuf API, the CLI, and the JS.

It also removes various helpers and interfaces that became unused from
removing that, and tests that tested these things.

It's worth noting that I've not deleted all of gitopswriter because
it's still used by installation, but the tests didn't test the
remaining functionality separately from the add/delete bits, so I've
deleted all the tests - installation is being rewritten, too.

This deletes any acceptance/integration tests that were clearly marked
as testing adding or removing applications, but I've not been over
them in any great detail to understand/refactor them - in fact I'm
certain there's still bits using add/remove in there.


J-Thompson12 (approved):
I cant speak for if you deleted everything you should have but I can say you didnt delete anything we still need.

PR #1445, opened 2022-02-14
Implement `update profile` to patch a profile's HelmRelease

nikimanoledaki:
Closes: 

**What changed?**

**Why?**

**How did you test it?**

**Release notes**

**Documentation Changes**

### DIFF
--- a/cmd/gitops/update/cmd.go
+++ b/cmd/gitops/update/cmd.go
@@ -1,0 +1,23 @@
+package update
+
+import (
+	"github.com/weaveworks/weave-gitops/cmd/gitops/update/profiles"
+
+	"github.com/go-resty/resty/v2"
+	"github.com/spf13/cobra"
+)
+
+func UpdateCommand(endpoint *string, client *resty.Client) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "update",
+		Short: "Update a Weave GitOps resource",
+		Example: `
+# Update an installed profile to a certain version
+gitops update profile --name <profile-name> --version <version> --config-repo <config-repo-url> --cluster <cluster-name> --namespace <ns-name>
+`,
+	}
+
+	cmd.AddCommand(profiles.UpdateCommand())
+
+	return cmd
+}
--- a/pkg/helm/releases.go
+++ b/pkg/helm/releases.go
@@ -4,9 +4,9 @@ import (
 	"bytes"
 	"fmt"
 	"io"
+	"strings"
 	"time"
 
-	"github.com/fluxcd/go-git-providers/gitprovider"
 	"github.com/fluxcd/helm-controller/api/v2beta1"
 	helmv2beta1 "github.com/fluxcd/helm-controller/api/v2beta1"
 	sourcev1beta1 "github.com/fluxcd/source-controller/api/v1beta1"
@@ -45,49 +45,44 @@ func MakeHelmRelease(name, version, cluster, namespace string, helmRepository ty
 	}
 }
 
-// AppendHelmReleaseToFile appends a HelmRelease to a gitProvider CommitFile given a condition, file name and file path.
-func AppendHelmReleaseToFile(files []*gitprovider.CommitFile, newRelease *v2beta1.HelmRelease, condition func(r, newRelease v2beta1.HelmRelease) error, fileName, filePath string) (gitprovider.CommitFile, error) {
-	var content string
+// FindHelmReleaseInString finds all HelmRelease(s) in a given string that can be split into YAML.
+func FindHelmReleaseInString(s string, newRelease *v2beta1.HelmRelease) ([]v2beta1.HelmRelease, error) {
+	manifestByteSlice, err := splitYAML([]byte(s))
+	if err != nil {
+		return nil, fmt.Errorf("error splitting into YAML: %w", err)
+	}
 
-	for _, f := range files {
-		if f.Path != nil && *f.Path == filePath {
-			if f.Content == nil || *f.Content == "" {
-				break
-			}
+	found := []v2beta1.HelmRelease{}
 
-			manifestByteSlice, err := splitYAML([]byte(*f.Content))
-			if err != nil {
-				return gitprovider.CommitFile{}, fmt.Errorf("error splitting %s: %w", fileName, err)
-			}
-
-			for _, manifestBytes := range manifestByteSlice {
-				var r v2beta1.HelmRelease
-				if err := kyaml.Unmarshal(manifestBytes, &r); err != nil {
-					return gitprovider.CommitFile{}, fmt.Errorf("error unmarshaling %s: %w", fileName, err)
-				}
-
-				if err := condition(r, *newRelease); err != nil {
-					return gitprovider.CommitFile{}, err
-				}
-			}
-
-			content = *f.Content
-
-			break
+	for _, manifestBytes := range manifestByteSlice {
+		var r v2beta1.HelmRelease
+		if err := kyaml.Unmarshal(manifestBytes, &r); err != nil {
+			return nil, fmt.Errorf("error unmarshaling: %w", err)
 		}
+
+		if profileIsInstalled(r, *newRelease) {
+			found = append(found, r)
+		}
+	}
+
+	return found, nil
+}
+
+// AppendHelmReleaseToString appends a HelmRelease to a string.
+func AppendHelmReleaseToString(content string, newRelease *v2beta1.HelmRelease) (string, error) {
+	var sb strings.Builder
+	if content != "" {
+		sb.WriteString(content + "\n")
 	}
 
 	helmReleaseManifest, err := kyaml.Marshal(newRelease)
 	if err != nil {
-		return gitprovider.CommitFile{}, fmt.Errorf("failed to marshal new HelmRelease: %w", err)
+		return "", fmt.Errorf("failed to marshal HelmRelease: %w", err)
 	}
 
-	content += "\n---\n" + string(helmReleaseManifest)
+	sb.WriteString("---\n" + string(helmReleaseManifest))
 
-	return gitprovider.CommitFile{
-		Path:    &filePath,
-		Content: &content,
-	}, nil
+	return sb.String(), nil
 }
 
 // splitYAML splits a manifest file that may contain multiple YAML resources separated by '---'
@@ -116,4 +111,8 @@ func splitYAML(resources []byte) ([][]byte, error) {
 	}
 
 	return splitResources, nil
+}
+
+func profileIsInstalled(r, newRelease v2beta1.HelmRelease) bool {
+	return r.Name == newRelease.Name && r.Namespace == newRelease.Namespace && r.Spec.Chart.Spec.Version == newRelease.Spec.Chart.Spec.Version
 }
--- a/pkg/helm/releases.go
+++ b/pkg/helm/releases.go
@@ -7,12 +7,11 @@ import (
 	"strings"
 	"time"
 
-	"github.com/fluxcd/helm-controller/api/v2beta1"
 	helmv2beta1 "github.com/fluxcd/helm-controller/api/v2beta1"
 	sourcev1beta1 "github.com/fluxcd/source-controller/api/v1beta1"
-	"gopkg.in/yaml.v2"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
+	apimachinery "k8s.io/apimachinery/pkg/util/yaml"
 	kyaml "sigs.k8s.io/yaml"
 )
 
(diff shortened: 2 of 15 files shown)


test/acceptance/test/profiles_test.go
```diff
@@ -129,6 +129,29 @@ Namespace: %s`, clusterName, namespace)))
 			resp, statusCode, err = kubernetesDoRequest(namespace, clusterName+"-"+profileName, "9898", "/healthz", clientSet)
 			return statusCode
 		}, "120s", "1s").Should(Equal(http.StatusOK))
+
+		By("Updating the version of the installed profile")
+		time.Sleep(4 * time.Minute)
```
aclevername, line 134: :thunk: is this the github API issue? Can we replace it with a robust `Eventually` instead?

test/acceptance/test/profiles_test.go
```diff
@@ -129,6 +129,29 @@ Namespace: %s`, clusterName, namespace)))
...
+		Eventually(func() string {
+			stdOut, stdErr = runCommandAndReturnStringOutput(fmt.Sprintf("%s update profile --name %s --version 6.0.0 --namespace %s --cluster %s --config-repo %s --auto-merge", gitopsBinaryPath, profileName, namespace, clusterName, appRepoRemoteURL))
+			return stdErr
+		}, "120s", "5s").Should(BeEmpty())
```
aclevername, line 138: isn't this going to re-run `update profile` every 5s?

pkg/helm/releases.go
```diff
@@ -37,3 +43,70 @@ func MakeHelmRelease(name, version, cluster, namespace string, helmRepository ty
 		},
 	}
 }
+
+// FindReleaseInNamespace iterates through a slice of HelmReleases to find one with a given name in a given namespace, and returns it with its index.
+func FindReleaseInNamespace(existingReleases []helmv2beta1.HelmRelease, name, ns string) (*helmv2beta1.HelmRelease, int, error) {
```
aclevername, line 48:
```suggestion
func FindReleaseFromList(existingReleases []helmv2beta1.HelmRelease, name, ns string) (*helmv2beta1.HelmRelease, int, error) {
```

or something like that, the `Namespace` part seems a bit off

pkg/helm/releases.go
```diff
@@ -37,3 +43,70 @@ func MakeHelmRelease(name, version, cluster, namespace string, helmRepository ty
...
+	sb.WriteString("---\n" + string(helmReleaseManifest))
+
+	return sb.String(), nil
+}
+
+// SplitHelmReleaseYAML splits a manifest file that contains one or more Helm Releases that may be separated by '---'.
+func SplitHelmReleaseYAML(resources []byte) ([]helmv2beta1.HelmRelease, error) {
+	var helmReleaseList []helmv2beta1.HelmRelease
+
+	decoder := apimachinery.NewYAMLOrJSONDecoder(bytes.NewReader(resources), 100000000)
```
aclevername, line 79: lets figure out a good value for `100000000` :D

pkg/helm/releases.go
```diff
@@ -37,3 +43,70 @@ func MakeHelmRelease(name, version, cluster, namespace string, helmRepository ty
...
+		}
+
+		helmReleaseList = append(helmReleaseList, value)
+	}
+
+	return helmReleaseList, nil
+}
+
+func PatchHelmRelease(existingReleases []helmv2beta1.HelmRelease, patchedHelmRelease helmv2beta1.HelmRelease, index int) (string, error) {
+	existingReleases[index] = patchedHelmRelease
```
aclevername, line 98: We could switch to passing around `[]*helmv2beta1.HelmRelease` instead, and then just edit the list in-place. Then when it comes to patching the list has already been modified, so this func just becomes marshaling the list of helm releases into the string

test/acceptance/test/profiles_test.go
nikimanoledaki: Yep so due to the Github API issue, it'll fail until the `GET` request returns the contents of `profiles.yaml` with:
```
Error: failed to find installed profiles in '.weave-gitops/clusters/kind-kind/system/profiles.yaml' of config repo "ssh://git@github.com/nikimanoledaki/fun-times.git"
```

Pinging until it returns the changes would be okay if it doesn't count towards the rate limit. To do this, we could make conditional requests that check if anything has changed like [this](https://docs.github.com/en/rest/guides/getting-started-with-the-rest-api#conditional-requests). If the response's status code is `304` (meaning that nothing has changed), Github kindly does not count that request towards the rate limit.

That being said, I don't think we need to worry about this here.

pkg/helm/releases.go
nikimanoledaki: done! :)

pkg/helm/releases.go
nikimanoledaki: completely changed this bit! pls review again

pkg/helm/releases_test.go
```diff
@@ -57,6 +60,76 @@ var _ = Describe("MakeHelmRelease", func() {
 				Interval: metav1.Duration{Duration: time.Minute},
 			},
 		}
-		Expect(cmp.Diff(actualHelmRelease, expectedHelmRelease)).To(BeEmpty())
+		Expect(cmp.Diff(&actualHelmRelease, &expectedHelmRelease)).To(BeEmpty())
+	})
+})
+
+var _ = Describe("AppendHelmReleaseToString", func() {
```
aclevername, line 67: missing error case test

test/acceptance/test/profiles_test.go
```diff
@@ -129,6 +129,29 @@ Namespace: %s`, clusterName, namespace)))
 			resp, statusCode, err = kubernetesDoRequest(namespace, clusterName+"-"+profileName, "9898", "/healthz", clientSet)
 			return statusCode
 		}, "120s", "1s").Should(Equal(http.StatusOK))
+
+		By("Updating the version of the installed profile")
+		time.Sleep(3 * time.Minute)
+		Eventually(func() string {
+			stdOut, stdErr = runCommandAndReturnStringOutput(fmt.Sprintf("%s update profile --name %s --version 6.0.0 --namespace %s --cluster %s --config-repo %s --auto-merge", gitopsBinaryPath, profileName, namespace, clusterName, appRepoRemoteURL))
+			return stdErr
+		}, "120s", "5s").Should(BeEmpty())
```
aclevername, line 138: Can we just remove the sleep and have the `Eventually` run over a long period of time instead? Less sleeps the better :smile:

aclevername (approved):
lgtm couple comments

test/acceptance/test/profiles_test.go
nikimanoledaki: I reduced the `sleep` to 3 minutes and having the `Eventually` be `120s` works. I think using `sleep` is justified since we know that this command will definitely not be successful in the first 3-5 minutes after doing `add profile`. Polling in those first few minutes will only count towards the rate limit since we're not doing [conditional requests](https://docs.github.com/en/rest#conditional-requests)

nikimanoledaki: scratch that - I just moved it all to the `Eventually`s!

pkg/helm/releases.go
nikimanoledaki: if anyone has a good suggestion for this, it would be much appreciated :D

aclevername: https://github.com/fluxcd/flux2/blob/f6afe7f0ec7d7722fd9e561d690c22d50ce10f23/internal/utils/utils.go#L307 

lgtm? @nikimanoledaki

cmd/gitops/add/profiles/cmd.go
```diff
@@ -61,8 +59,6 @@ func AddCommand() *cobra.Command {
 
 func addProfileCmdRunE() func(*cobra.Command, []string) error {
 	return func(cmd *cobra.Command, args []string) error {
-		rand.Seed(time.Now().UnixNano())
```
Skarlso, line 64: This is needed for the math randomizer. Why remove it?

cmd/gitops/update/cmd.go
```diff
@@ -0,0 +1,23 @@
...
+	cmd := &cobra.Command{
+		Use:   "update",
+		Short: "Update a Weave GitOps resource",
+		Example: `
+	# Update a profile that is installed on a cluster
+	gitops update profile --name=podinfo --cluster=prod --config-repo=ssh://git@github.com/owner/config-repo.git  --version=1.0.0
+		`,
+	}
+
+	cmd.AddCommand(profiles.UpdateCommand())
```
Skarlso, line 20: `add update`? 🤔

cmd/gitops/add/profiles/cmd.go
nikimanoledaki: This isn't needed anymore because we switched to using UUIDs instead, for now: https://github.com/weaveworks/weave-gitops/pull/1360#discussion_r798409841 :)

pkg/helm/releases.go
```diff
@@ -37,3 +43,58 @@ func MakeHelmRelease(name, version, cluster, namespace string, helmRepository ty
 		},
 	}
 }
+
+// AppendHelmReleaseToString appends a HelmRelease to a string.
```
Skarlso, line 47: :D Can you be a tiny bit more descriptive please? :) Why? How? What's the purpose of the append?

pkg/helm/releases.go
Skarlso: Actually that's a frigging LARGE buffer size. :D Usually it's either 4096 or 2048 which is perfectly enough. :) I opts for using what Jake linked in. :)

pkg/services/profiles/profiles.go
```diff
@@ -34,3 +40,29 @@ func NewService(clientSet kubernetes.Interface, log logger.Logger) *ProfilesSvc
...
+	return types.NamespacedName{
+		Name:      availableProfile.HelmRepository.Name,
+		Namespace: availableProfile.HelmRepository.Namespace,
+	}, version, nil
+}
+
+func getGitCommitFileContent(files []*gitprovider.CommitFile, filePath string) string {
+	for _, f := range files {
+		if f.Path != nil && *f.Path == filePath {
+			if f.Content == nil || *f.Content == "" {
```
Skarlso, line 59:
```suggestion
			if f.Content == nil || f.Content != nil && *f.Content == "" {
```

nikimanoledaki: isnt this implied? if `f.Content == nil` is false, that means `f.Content != nil`

pkg/helm/releases.go
nikimanoledaki: Thanks yep I had changed this to `2048`, was waiting for more feedback. :)

pkg/services/profiles/update.go
```diff
@@ -0,0 +1,146 @@
...
+	Cluster      string
+	ConfigRepo   string
+	Version      string
+	ProfilesPort string
+	Namespace    string
+	Kubeconfig   string
+	AutoMerge    bool
+}
+
+// Update updates an installed profile
```
Skarlso, line 31: It's not perfectly apparent that the fact that there is a HelmRelease IS the way to get that it is installed. Shall we do something like ` checkIfInstalled` at the begin of the function and fail early? So it doesn't have to download and check anything before it even begun that.

pkg/services/profiles/update.go
```diff
@@ -0,0 +1,146 @@
...
+
+	fileContent := getGitCommitFileContent(files, git.GetProfilesPath(opts.Cluster, models.WegoProfilesPath))
+	if fileContent == "" {
+		return fmt.Errorf("failed to find installed profiles in '%s' of config repo %q", git.GetProfilesPath(opts.Cluster, models.WegoProfilesPath), configRepoURL)
+	}
+
+	content, err := updateHelmRelease(helmRepo, fileContent, opts.Name, opts.Version, opts.Cluster, opts.Namespace)
+	if err != nil {
+		return fmt.Errorf("failed to update HelmRelease for profile '%s' in %s: %w", opts.Name, models.WegoProfilesPath, err)
+	}
```
Skarlso, line 76: This is a bit weird here. The `getGitCommitFileContent` is not used anywhere else but `updateHelmRelease`. But also, `updateHelmRelease` doesn't use `helmRepo` at all. It's unused. :D I think something can be shuffled around here.

Skarlso: Which is super weird, because then the result of this:

```go
helmRepo, version, err := s.discoverHelmRepository(ctx, GetOptions{
		Name:      opts.Name,
		Version:   opts.Version,
		Cluster:   opts.Cluster,
		Namespace: opts.Namespace,
		Port:      opts.ProfilesPort,
	})
```

Has no actual result other than maybe version being set on options. 🤔

pkg/services/profiles/profiles.go
Skarlso: Yep, you're right. Nice. :)

cmd/gitops/add/profiles/cmd.go
Skarlso: ah ok!

Skarlso (approved):
Well done! :)

PR #1446, opened 2022-02-14
Remove[*] install/uninstall commands

ozamosi:
Also remove a bunch of services, now that most direct writes to k8s is
    gone. Not all - there's still profiles - but they're very limited in
    what they use.

[*]: Doesn't actually remove install, it replaces it with a
informational note - we agreed that we need to build a helm chart, and
that that is a separate task, so this just suggests a way for you to
continue working while we get to that bit.


pkg/flux/flux.go
```diff
@@ -268,20 +224,6 @@ func (f *FluxClient) runFluxCmd(args ...string) ([]byte, error) {
...
-
-	out, err := f.runner.RunWithOutputStream(fluxPath, args...)
-	if err != nil {
-		return []byte{}, fmt.Errorf("failed to run flux with output: %s and error: %w", string(out), err)
-	}
-
-	return out, nil
-}
-
 func (f *FluxClient) fluxPath() (string, error) {
```
J-Thompson12, line 227: fluxPath can be removed as well. Since we dont have the binary embedded anymore. Whether its part of this or not is fine with me.

pkg/services/gitrepo/service.go
```diff
@@ -6,83 +6,11 @@ import (
...
-func (rw *RepoWriterSvc) CheckoutBranch(newBranch string) error {
-	err := rw.GitClient.Checkout(newBranch)
-	if err != nil {
-		return fmt.Errorf("error checking out branch %s, %w", newBranch, err)
-	}
-
-	return nil
-}
-
 func CommitAndPush(ctx context.Context, client git.Git, commitMsg string, logger logger.Logger, filters ...func(string) bool) error {
```
J-Thompson12, line 14: since you removed the add cmd does anything still use this? I know we will still use the commit part of it but not the push part.

J-Thompson12: Okay I did a quick check and upgrade is the only spot still using this. Upgrade will be getting updated soon anyway so its up to you if we treat it like install with a message or just leave it as is for now.

J-Thompson12 (approved):
Just had a few minor comments but I am sure you have a plan so its up to you if you want to address or ignore.

pkg/flux/flux.go
ozamosi: The embedded binary is still in there - I've avoided doing anything to it since we talked on Friday about the need for _some_ "verified flux", and I don't think we decided what that is. So gonna leave it right now.

PR #1447, opened 2022-02-14
Add Title To Logo

joshri:
Adds Weave GitOps text from Figma as an svg to the `Logo` component, and adds a snapshot test as well.

image


jpellizzari (approved):

bia:
`80px` looked huge to me because it is indeed `96px`! Opened #1796 
Screenshot 2022-03-24 at 11 16 00

PR #1451, opened 2022-02-15
Add releases and flux runtime pages

jpellizzari:
Adds the list pages according to the provided designs https://www.figma.com/file/IVHnM9iyeFWpd11evtY8ux/Weave-GitOps?node-id=5399%3A12874

Releases/Applications:
Screenshot from 2022-02-15 15-15-11

Sources Page:
Screenshot from 2022-02-15 15-08-23

Flux Runtime:
Screenshot from 2022-02-15 15-13-46


api/app/apps.proto
```diff
@@ -119,7 +119,7 @@ service Apps {
     */
   rpc ListKustomizations(ListKustomizationsReq) returns (ListKustomizationsRes) {
     option (google.api.http) = {
-      get : "/v1/namespace/{namespace}/app/{app_name}/kustomization"
+      get : "/v1/kustomizations"
```
JamWils, line 122: Nit: should this be plural?  Usually the REST semantics are /v1/kustomization, I would expect a list.  Then /v1/kustomization/:id, I would expect to get one.

jpellizzari: Sure, but not sure why we would try to min/max this as a REST API. I would hope someone would use the open API output to generate a client library in their language of choice.

We already gen the `go` and `ts` clients. I didn't have to change anything in my "user space" code to accommodate this change.

sympatheticmoose:
Good progress being made 👍 thanks Jordan

Some comments:
**Applications page**
- Types should be kustomization / Helm Release rather than kustomize
- Source in the designs doesn't show the type, but actually I'm ok with it cc @bia 
- Missing filters (assuming follow up PR)
- Name is only one with sort arrow, is this easy to add to other columns?

**Sources page**
- should be sub-menu item under Applications similar to Clusters > templates in WGE.
- Types should be Git Repository / Helm Repository / Bucket (i.e. not just Git)
- Missing Reference (currently marked as Release in Figma)
- Designs are yet to address comments/reflect the doc in notion and as such are missing cluster, namespace and interval/last synced

**Flux Runtime page**
- Image not Images
- Default should sort by Cluster.

jpellizzari:
> * Types should be kustomization / Helm Release rather than kustomize

Fixed

> * Missing filters (assuming follow up PR)

@joshri is working on that

> * Name is only one with sort arrow, is this easy to add to other columns?

Fixed

> **Sources page**
> 
> * should be sub-menu item under Applications similar to Clusters > templates in WGE.

Follow up created: https://github.com/weaveworks/weave-gitops/issues/1459

> * Types should be Git Repository / Helm Repository / Bucket (i.e. not just Git)

Fixed


> * Missing Reference (currently marked as Release in Figma)

Added, but we need to think about what this means for Helm/Bucket

> * Designs are yet to address comments/reflect the doc in notion and as such are missing cluster, namespace and interval/last synced

Interval added

> 
> **Flux Runtime page**
> 
> * Image not Images

Fixed, but a `Deployment` can have multiple images.

> * Default should sort by Cluster.

Added

sympatheticmoose:
Many thanks @jpellizzari 

> Added, but we need to think about what this means for Helm/Bucket

Both still have `status.artifact.revision` or do you feel this isn't valuable/there's something else we should look to show?

> Fixed, but a Deployment can have multiple images.

Fair point, I wouldn't expect the flux controllers to though? Or am I missing something?

core/server/app.go
```diff
@@ -111,6 +113,46 @@ func (as *appServer) ListApps(ctx context.Context, msg *pb.ListAppRequest) (*pb.
...
 
+func (as *appServer) ListFluxRuntimeObjects(ctx context.Context, msg *pb.ListFluxRuntimeObjectsReq) (*pb.ListFluxRuntimeObjectsRes, error) {
+	k8s, err := as.k8s.Client(ctx)
+	if err != nil {
+		return nil, doClientError(err)
+	}
+
+	list := &appsv1.DeploymentList{}
+
+	if err := k8s.List(ctx, list, client.InNamespace(msg.Namespace)); err != nil {
```
luizbafilho, line 124: I think we should avoid making assumptions about flux namespace (https://github.com/weaveworks/weave-gitops/pull/1451/files#diff-35acf2adc3a7b1021fb41d3473b47e030d3708ba2e8ff7a4b76c13ee80b4076cR42) and grab it by using something like: https://github.com/weaveworks/weave-gitops/blob/v2/pkg/kube/kubehttp.go#L496

I'm not sure we want to keep that interface but the actual method is useful.

ui/components/KubeStatusIndicator.tsx
```diff
@@ -0,0 +1,27 @@
...
+  return (
+    <Flex start className={className} align>
+      <Icon color={color} size="base" type={IconType.Circle} text={readyText} />
+    </Flex>
+  );
+}
+
+export default styled(KubeStatusIndicator).attrs({
+  className: KubeStatusIndicator.name,
+})``;
```
joshri, line 27: 💕 ❤️ 💕

ui/pages/v2/ApplicationList/ApplicationList.tsx
```diff
@@ -1,52 +1,97 @@
...
-                </Link>
-              ),
+      <DataTable
+        sortFields={[
+          "name",
+          "namespace",
+          "type",
+          "cluster",
+          "status",
+          "last synced at",
```
joshri, line 32: I'm curious if my sort is handling "last synced at" correctly...

joshri (commented):
Front end looks stunning but I'll leave the approval to the back end experts

@sympatheticmoose are all sortable fields intended to have a sort arrow? Right now only the field that is actively being sorted has one

core/server/app.go
jpellizzari: @luizbafilho Since we allow the client to specify the `namespace`, can I fetch the correct namespace in a different call (in a future PR)?

ui/pages/v2/ApplicationList/ApplicationList.tsx
jpellizzari: Probs not; we will need to add the concept of data types.

core/server/app.go
luizbafilho: Yeah, it would be a good follow up issue

api/app/apps.proto
jpellizzari: https://github.com/weaveworks/weave-gitops/issues/1465

core/server/app.go
jpellizzari: https://github.com/weaveworks/weave-gitops/issues/1465

luizbafilho (approved):

ui/pages/v2/ApplicationList/ApplicationList.tsx
jpellizzari: https://github.com/weaveworks/weave-gitops/issues/1467

ui/components/KubeStatusIndicator.tsx
sympatheticmoose: > @sympatheticmoose are all sortable fields intended to have a sort arrow? Right now only the field that is actively being sorted has one

I believe that is the design... @bia could you confirm? I am assuming the column head is clickable and thats what applies/flips the sort?

bia: Yes, in the current design only the field that is actively being sorted has an arrow (facing up or down based on sort direction). The other column headers that can sort the table should be clickable and the sort would be applied on click.

PR #1461, opened 2022-02-17
Stop describing releases as straightforward

dhwthompson:
If the release process actually is straightforward, we don't really gain much from saying it's straightforward, because people can see how straightforward it is. If, on the other hand, the release process *isn't* straightforward, saying it is will only make people feel bad when they struggle to understand or follow it, and will discourage them from speaking up and calling attention to problems.


jpellizzari (approved):
:+1:

JamWils (approved):

PR #1469, opened 2022-02-17
Separate UI

SamLR:
**What changed?**

Remove the `ui run` command and replace with a separate server binary

**Why?**

* This reduces the size of the main gitops binary (as it no longer needs to embed the UI). 
* This is a more intuitive configuration, it is clear which portions of code do what.
* This increases security by removing capabilities from the binary which will be deployed to clusters.
* This will simplify testing by removing the requirements to work around the CLI when testing the server and visa-versa.

**How did you test it?**

The binaries build and successfully host the UI/operate on the CLI.

**Release notes**

This is a breaking change that removes the `ui run` command (the `ui` command has been replaced with a placeholder that can be extended to help access the ui).

**Main questions/decisions to focus on for review**

I've mostly left the Go code intact as I'm not so confident there and it feels like more of that will be wanted by the observability folks. 

This means the focus of a lot of this has been around dockerfiles and the makefile. The later especially as the next bit of work I have is to tidy it up further. To that end I have some questions about how comfortable folks are with various usages of makefiles: 

* Am I breaking something by merging the `cmd/gitops/ui/run/dist/`, `cmd/gitops/ui/run/dist/index.html` and `cmd/gitops/ui/run/dist/index.js` targets?
* Should I use implicit targets (`bin/%: cmd/%/main.go`) to reduce repetition or are they more overhead than is worth it?
* Are folks happy with the pattern of using variables to change the behaviour of makefile targets (this:
```make
_docker:
  docker build $(ARGS) -f $(FILE)

docker-gitops: FILE:=gitops.dockerfile
docker-gitops: _docker

docker-gitops-server: FILE:=gitops-server.dockerfile
docker-gitops-server: ARGS:=--build-arg "SOME_FLAG=foo"
docker-gitops-server: _docker
```


gitops.dockerfile
```diff
@@ -0,0 +1,31 @@
...
+    apt-get install -y kubectl
+# Add known_hosts entries for GitHub and GitLab
+RUN mkdir ~/.ssh
+RUN ssh-keyscan github.com >> ~/.ssh/known_hosts
+RUN ssh-keyscan gitlab.com >> ~/.ssh/known_hosts
+
+COPY Makefile /app/
+COPY tools /app/tools
+WORKDIR /app
+RUN make dependencies
```
ozamosi, line 19: Do we really need to add make dependencies here? This is the one that starts using sudo inside your container.

SamLR: fixed

ozamosi (approved):

SamLR:
@JamWils @katya-makarevich @dlopro not sure who else I should ask to review this

cmd/gitops-server/cmd/cmd.go
```diff
@@ -0,0 +1,328 @@
...
+
+  "github.com/weaveworks/weave-gitops/cmd/gitops/cmderrors"
+  "github.com/weaveworks/weave-gitops/pkg/helm/watcher"
+  "github.com/weaveworks/weave-gitops/pkg/helm/watcher/cache"
+  "github.com/weaveworks/weave-gitops/pkg/kube"
+  "github.com/weaveworks/weave-gitops/pkg/server"
+  "github.com/weaveworks/weave-gitops/pkg/server/auth"
+)
+
+// Options contains all the options for the `ui run` command.
```
luizbafilho, line 30: nit: no ui run here, the same reference to ui run command repeats in other places

luizbafilho (approved):
just a very minor comment, LGTM

cmd/gitops-server/cmd/cmd.go
SamLR: Ah good catch, I'll go sort these, thanks!

SamLR: Think [0fc3639](https://github.com/weaveworks/weave-gitops/pull/1469/commits/0fc36393a61c93e8e9e63ad4fc9df7c8ecc1b245) should fix these

PR #1470, opened 2022-02-17
Resolve wait times issues for tests

rokshana-b:
**What changed?**
A couple of the acceptance tests were failing due to short waits. I've increased the wait time to a reasonable amount and removed skip tag from those failing tests.

**Why?**
GitOps takes around 60s to reconcile — so I've set the wait time to that.

### DIFF
--- a/test/acceptance/test/add_tests.go
+++ b/test/acceptance/test/add_tests.go
@@ -225,7 +225,7 @@ var _ = Describe("Weave GitOps Add App Tests", func() {
 		})
 
 		By("And app should not get deleted until the PR is merged", func() {
-			err = waitForAppRemoval(appName, EVENTUALLY_DEFAULT_TIMEOUT)
+			err = waitForAppRemoval(appName)
 			Expect(err).To(HaveOccurred())
 		})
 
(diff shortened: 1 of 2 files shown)


J-Thompson12 (approved):

dhwthompson (commented):
So, it looks like this PR does four things:

1. Raise the timeout for `gitAddCommitPush` from 10 seconds to 30
2. Raise the timeout for app deletion to 60 seconds in the cases where it was 30 seconds
3. Remove the parameter to customize that timeout, as it’s always the same now
4. Re-enable the disabled tests

Is that right? I’m just trying to make sure I’m following along correctly.

Also, do you have a guess how much difference this will make to how long the acceptance tests take to run? I see they took about 30–35 minutes on this PR, which seems on the upper end of what we were seeing already, but I appreciate that these times aren’t always super consistent.

rokshana-b:
@dhwthompson you are right about the 4 points, that's exactly what I'm doing. 

> how much difference this will make to how long the acceptance tests take to run? I see they took about 30–35 minutes on this PR

I did see the longer run time but re-enabling the tests means there's 3 more tests running now compared to the last runs. So, I don't think this will make any significant impact overall because the functions are written to quit polling as soon the conditions become true. Therefore, tests will run as per usual time - they'll just stop failing(hopefully) due to quicker timeouts.

dhwthompson (approved):
Cool cool. Thanks for clarifying. 👍

PR #1483, opened 2022-02-18
Serve `ui run` over https

foot:
**What changed?**

- Use https to serve the UI by default (can be switched off w/ `--no-tls`

**Why?**

So we can support OIDC token and user/pass credentials properly (don't leak them over the network on 0.0.0.0)

**How did you test it?**

Manually:

`go run ./cmd/gitops/main.go ui run`

**Release notes**

Yes, we should mentioned it will use https by default now.

**Documentation Changes**

Included doc changes in this PR

### DIFF
--- a/cmd/gitops/ui/run/cmd.go
+++ b/cmd/gitops/ui/run/cmd.go
@@ -44,8 +44,8 @@ type Options struct {
 	OIDC                          OIDCAuthenticationOptions
 	NotificationControllerAddress string
 	Insecure                      bool
-	TlsKey                        string
-	TlsCert                       string
+	TLSCert                       string
+	TLSKey                        string
 	NoTLS                         bool
 }
 
@@ -84,8 +84,8 @@ func NewCommand() *cobra.Command {
 	cmd.Flags().IntVar(&options.WatcherPort, "watcher-port", 9443, "the port on which the watcher is running")
 
 	cmd.Flags().BoolVar(&options.Insecure, "insecure", false, "allow insecure TLS requests")
-	cmd.Flags().StringVar(&options.TlsKey, "tls-key", "", "filename for the TLS key, in-memory generated if omitted")
-	cmd.Flags().StringVar(&options.TlsCert, "tls-cert", "", "filename for the TLS certficate, in-memory generated if omitted")
+	cmd.Flags().StringVar(&options.TLSCert, "tls-cert", "", "filename for the TLS certficate, in-memory generated if omitted")
+	cmd.Flags().StringVar(&options.TLSKey, "tls-key", "", "filename for the TLS key, in-memory generated if omitted")
 	cmd.Flags().BoolVar(&options.NoTLS, "no-tls", false, "do not attempt to read TLS certificates")
 
 	if server.AuthEnabled() {
@@ -310,7 +310,7 @@ func listenAndServe(srv *http.Server, options Options, log logrus.FieldLogger) e
 		return srv.ListenAndServe()
 	}
 
-	if options.TlsCert == "" && options.TlsKey == "" {
+	if options.TLSCert == "" && options.TLSKey == "" {
 		log.Info("TLS cert and key not specified, generating and using in-memory keys")
 
 		tlsConfig, err := wego_tls.TLSConfig([]string{"localhost", "0.0.0.0", "127.0.0.1"})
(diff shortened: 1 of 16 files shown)


pkg/server/tls/tls.go
```diff
@@ -0,0 +1,91 @@
...
+	"math/big"
+	"net"
+	"time"
+)
+
+// TLSConfig is adapted from http.Server.ServeTLS
+func TLSConfig(hosts []string) (*tls.Config, error) {
+	certPEMBlock, keyPEMBlock, err := generateKeyPair(hosts)
+	if err != nil {
+		return nil, fmt.Errorf("Failed to generate TLS keys %s", err)
```
bigkevmcd, line 23: This should be `%w` and not `%s` https://pkg.go.dev/fmt#Errorf

Same issue below.

pkg/server/tls/tls.go
```diff
@@ -0,0 +1,91 @@
...
+
+// Adapted from https://go.dev/src/crypto/tls/generate_cert.go
+func generateKeyPair(hosts []string) ([]byte, []byte, error) {
+	priv, err := ecdsa.GenerateKey(elliptic.P256(), rand.Reader)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	template := x509.Certificate{
+		SerialNumber: big.NewInt(1),
```
bigkevmcd, line 46: We could randomise this with something like:

```go
maxSerialNumber := new(big.Int).Lsh(big.NewInt(1), 128)
serialNumber, err := rand.Int(rand.Reader, maxSerialNumber)
```

And use that value as the `SerialNumber`.

pkg/server/tls/tls.go
```diff
@@ -0,0 +1,91 @@
...
+	}
+
+	return tlsConfig, nil
+}
+
+// Adapted from https://go.dev/src/crypto/tls/generate_cert.go
+func generateKeyPair(hosts []string) ([]byte, []byte, error) {
+	priv, err := ecdsa.GenerateKey(elliptic.P256(), rand.Reader)
+	if err != nil {
+		log.Fatal(err)
```
bigkevmcd, line 42: Why not return the error here?

cmd/gitops/ui/run/cmd.go
```diff
@@ -27,6 +27,7 @@ import (
 	"github.com/weaveworks/weave-gitops/pkg/kube"
 	"github.com/weaveworks/weave-gitops/pkg/server"
 	"github.com/weaveworks/weave-gitops/pkg/server/auth"
+	wego_tls "github.com/weaveworks/weave-gitops/pkg/server/tls"
```
bigkevmcd, line 30: You can import this as just tls (i.e. without the import alias) as it's not conflicting with anything else?

cmd/gitops/ui/run/cmd.go
```diff
@@ -78,6 +83,11 @@ func NewCommand() *cobra.Command {
 	cmd.Flags().StringVar(&options.NotificationControllerAddress, "notification-controller-address", "", "the address of the notification-controller running in the cluster")
 	cmd.Flags().IntVar(&options.WatcherPort, "watcher-port", 9443, "the port on which the watcher is running")
 
+	cmd.Flags().BoolVar(&options.Insecure, "insecure", false, "allow insecure TLS requests")
+	cmd.Flags().StringVar(&options.TLSCert, "tls-cert", "", "filename for the TLS certficate, in-memory generated if omitted")
```
bigkevmcd, line 87: We could reuse the same naming scheme as kube-apiserver here... `--tls-cert-file` and `--tls-private-key-file` ?

cmd/gitops/ui/run/cmd.go
```diff
@@ -290,6 +306,31 @@ func runCmd(cmd *cobra.Command, args []string) error {
...
+func listenAndServe(srv *http.Server, options Options, log logrus.FieldLogger) error {
+	if options.NoTLS {
+		log.Info("TLS connections disabled")
+		return srv.ListenAndServe()
+	}
+
+	if options.TLSCert == "" && options.TLSKey == "" {
+		log.Info("TLS cert and key not specified, generating and using in-memory keys")
+
+		tlsConfig, err := wego_tls.TLSConfig([]string{"localhost", "0.0.0.0", "127.0.0.1"})
```
bigkevmcd, line 318: This should probably fail if only one of them is provided?

bigkevmcd:
No tests at all for this code?

pkg/server/tls/tls.go
foot: Failed to adapt all the code..! Good catch

pkg/server/tls/tls.go
foot: > A CA is supposed to choose unique serial numbers, that is, unique for the CA

Cool 👍

cmd/gitops/ui/run/cmd.go
foot: sgtm!

cmd/gitops/ui/run/cmd.go
foot: Yes!

yiannistri (approved):
LGTM 👍

PR #1489, opened 2022-02-18
Add reconciled and child object API endpoints and KustomizationDetail page

### LINKED ISSUE #1453: Move "object reconciliation" handlers to MVP grpc service
Blocked by #1452 

According to our proposed designs, we will need to be able to render a table/tree view of the objects that a Flux automation (`Kustomization` | `HelmRelease`) outputs. We have existing endpoints from the `applications` service that we have implemented for this:

https://github.com/weaveworks/weave-gitops/blob/14800da7c5dc76fea69df8876cf380697b0bc6f0/api/applications/applications.proto#L62

https://github.com/weaveworks/weave-gitops/blob/14800da7c5dc76fea69df8876cf380697b0bc6f0/api/applications/applications.proto#L73

Acceptance Criteria:
* Add those endpoints/handlers with test coverage to our GRPC service

jpellizzari:
Closes #1453 

Adds the endpoint necessary to render the `KustomizationDetail` page, and the page itself to the UI:
Screenshot from 2022-02-22 14-18-42

These endpoints are mostly copy-paste from the `pkg/server/server.go` endpoints with some basic test coverage added.


core/server/automations_test.go
```diff
@@ -49,6 +51,66 @@ func TestListKustomizations(t *testing.T) {
...
+	g.Expect(k.Create(ctx, kust)).To(Succeed())
+
+	t.Run("gets a kustomization", func(t *testing.T) {
+		_, err = c.GetKustomization(ctx, &pb.GetKustomizationRequest{Name: appName, Namespace: ns.Name})
+		g.Expect(err).NotTo(HaveOccurred())
+
+		// The kustomization status does not get populated when running in a test environment,
+		// probably because there is no kustomization-controller running to populate it.
+		// Even though we populate it in the kustomization.Status object above, it still does not populate.
+		// Need some help from the flux team on this.
```
Callisto13, line 97: do we care about those status bits? as in, is that something we would add/control anyway or is that something the flux controller would add? if it is a flux thing, then I think we are find not checking those things here (we can do that in an e2e) and just check that the result name is the right one because that is behaviour that we control.

ui/pages/v2/KustomizationDetail.tsx
```diff
@@ -0,0 +1,114 @@
...
+const InfoList = styled(
+  ({
+    items,
+    className,
+  }: {
+    className?: string;
+    items: { [key: string]: any };
+  }) => {
+    return (
+      <table className={className}>
```
joshri, line 33: normal HTML table?

ui/pages/v2/KustomizationDetail.tsx
```diff
@@ -0,0 +1,114 @@
...
+      title={kustomization?.name}
+      loading={isLoading}
+      error={error}
+      className={className}
+    >
+      <Info>
+        <h3>{kustomization?.namespace}</h3>
+        <InfoList
+          items={{
+            Source: (
```
joshri, line 74: Do we not have the other fields that are in the Figma yet (namespace, lastUpdated, and interval)?

ui/pages/v2/KustomizationDetail.tsx
```diff
@@ -0,0 +1,114 @@
...
+type Props = {
+  name: string;
+  className?: string;
+};
+
+const Info = styled.div`
+  padding-bottom: 32px;
+`;
+
+const InfoList = styled(
```
joshri, line 24: I'm seeing this pattern in a couple other pages so we might want to separate this out?

core/server/automations_test.go
jpellizzari: In order for the endpoint to be "working" from the perspective of the UI, we need the status of the Kustomization to be populated.

If we can simulate *without* running a Flux reconciler in the test environment, that would be ideal. @hiddeco Any tricks for getting this status populated?

ui/pages/v2/KustomizationDetail.tsx
```diff
@@ -0,0 +1,114 @@
...
+import { formatURL } from "../../lib/nav";
+import { V2Routes, WeGONamespace } from "../../lib/types";
+
+type Props = {
+  name: string;
+  className?: string;
+};
+
+const Info = styled.div`
+  padding-bottom: 32px;
```
joshri, line 21: It looks like the padding is the space for the Filter and Search things - will probs have to change later

joshri (approved):

ui/pages/v2/KustomizationDetail.tsx
jpellizzari: Yes, to get the columns to all line up.

core/server/automations_test.go
hiddeco: The problem I think, is with the usage of `Create` which does not persist the Status. First creating the object, and then patching it with a Status should work.

ui/pages/v2/KustomizationDetail.tsx
joshri: It looks like it's in all the different Release Detail and Source Detail pages

jpellizzari: Can you link to those components? Not seeing them on the `v2` branch.

joshri: Oh no I'm saying it's in the Figma designs for those pages

jpellizzari: I typically wait to move stuff until we need to, then the commits make more sense. Good call out though :+1:

ui/pages/v2/KustomizationDetail.tsx
jpellizzari: Added!

jpellizzari:
I tried to find the right `go.mod` incantation to get this to pass `snyk`, but gave up. The vulns are only medium.

Issue #1477, opened 2022-02-18
Makefile improvements


ozamosi:
The makefile can be leveraged by engineering and the github action commands.  Redundant commands are removed and any commands that require prerequisites to run are clearly documented in the README or in the docs folder of the repository.  For example, the protobuf command requires some dependencies to be installed.

In addition, commands do what they are supposed to.  Does the clear command remove everything including tools and dependencies?  Does the dependencies command actually run and add new dependencies when something is added?  Today, I think it just blindly looks for any dependency and then doesn’t run at all.

PR #1495, opened 2022-02-22
Pre commit hooks

SamLR:
Part of #1477 

**What changed?**

Add pre-commit hooks. This can be used to replace several small makefile entries (e.g. `fmt`, `vet` etc) by running them as part of a git hub action on pull requests and locally on push and commit. The checks are relatively fast with the longest being `go vet` (which only runs on push) and `golangci-lint` (which runs on commit). Some ad-hoc testing suggests running with hooks adds ~7s to committing (for 5 changed go files). 

**Why?**

These checks should be run regularly to maintain good code quality. Adding a system to run them automatically as part of pull requests and locally on commit helps this. The checks are opt-in and can be disabled easily on the command line (via the `-n/--no-verify` flag). 

**How did you test it?**

I have checked the changes by commiting to the branch and running `pre-commit run --all` in order to check the current state of the repo.

**Release notes**

* add pre-commit hooks to aid development


SamLR:
@JamWils @katya-makarevich @dlopro not sure who else I should ask to review this

.github/workflows/scan.yaml
```diff
@@ -62,20 +62,23 @@ jobs:
...
+          DOCKER_REGISTRY: ghcr.io/weaveworks
+        run: make docker-${{ matrix.docker-target }}
       - name: Scan docker image
         uses: snyk/actions/docker@master
         env:
           SNYK_TOKEN: ${{ secrets.SNYK_API_TOKEN }}
         with:
-          image: ghcr.io/weaveworks/wego-app:latest
-          args: --file=Dockerfile
+          image: ghcr.io/weaveworks/${{ matrix.docker-target }}:latest
```
jpellizzari, line 80: :question: Does this mean that we run these each of these matrices? So this job runs twice?

README.md
```diff
@@ -75,6 +75,24 @@ To set up a development environment for the CLI
 6. Start the in-cluster API replacement job (powered by [http://tilt.dev](tilt.dev)) with `make cluster-dev`
 7. make or make unit-tests to ensure everything built correctly.
 
+### Pre-commit hooks
+
+This repo has [pre-commit hooks](https://pre-commit.com/) configured on it, they are run by github action for pull requests but it is recommended install them locally too:
+
+```bash
+$ pip install pre-commit
```
jpellizzari, line 83: I would prefer not to involve Python in our build process. I am not familiar enough with it or it's ecosystem to know how to troubleshoot it.

jpellizzari (commented):
I am a little confused by this. The name "pre-commit" hooks is confusing me, since it appears these run during CI. How can CI do something before a commit unless CI is committing itself?

Also, this appears to be doing much more than adding pre-commit hooks. There appears to be the beginnings of the deprecation of `gitops ui run`? The PR description is not matching up with the changes I am seeing.

I am happy to discuss on a call.

SamLR:
> I am a little confused by this. The name "pre-commit" hooks is confusing me, since it appears these run during CI. How can CI do something before a commit unless CI is committing itself?

The CI is not commiting anything. I added the hooks to PRs as a final-check as you can disable checks locally.

> Also, this appears to be doing much more than adding pre-commit hooks. There appears to be the beginnings of the deprecation of `gitops ui run`? The PR description is not matching up with the changes I am seeing.

Ah sorry, I should update the PR info as I'd built this off another branch I was working on (#1469) so it's showing work from that; I'll rebase so it's directly on v2 and push only those changes.

.github/workflows/scan.yaml
SamLR: Yes (although this is work done in the PR #1469 rather than here, will split & re-update)

README.md
SamLR: Is the problem here using a program written in python or using pip to install it? pre-commit can also be [installed](https://pre-commit.com/#install) via brew on conda (although I'm not familiar with conda myself) and I can suggest one of those, or just link to the relevant documentation.

jpellizzari: It’s really about the pip command itself. I haven’t had good luck with it.

SamLR: I've updated the readme with details on installation via brew and conda. Are those better options?

SamLR:
@jpellizzari forgot to say in my last comment: happy to jump on a call if you want to talk through this? Given you're much more likely to be using it than I am it only makes sense if it looks useful to to .

README.md
jpellizzari: I am on linux, and not familiar with conda.

SamLR: yea I'd not encountered conda before.

I looks like some linux distros have pre-commit available in their package managers (e.g. ubuntu jammy) but I suspect it's only in very recent ones.

Is this a blocker?

jpellizzari: Yes I would prefer we not involve a new runtime in the build if possible.

JamWils: Is there a go version?

SamLR: Not that I am aware of

Issue #1503, opened 2022-02-23
Fix snyk integration


ozamosi:
In order to fully evaluate #1484, the snyk integration needs to be sorted out, so we can see how well it solves our problems, properly integrated - it's currently running in too many places, is alerting in the wrong ways, and can't create PRs.

Fix it.

PR #1504, opened 2022-02-23
Upgrade weaveworks/go-checkpoint to version with go.mod

ozamosi:
Related to: #1503  

**What changed?**
This upgrades the version of go-checkpoint. The new version has absolutely no code changes.

Yes, this clear-the-way PR is against main. It's supposed to be against main. Snyk's github integration only works against the default branch.

**Why?**
This is a bit of a stab in the dark to try to get the snyk github
integration working (part of #1503) - apparently we can't have
dependencies without go.mod files?

**How did you test it?**
There are checkpoint tests already, and this changes literally no code. I did try dumping the response from checkpoint to see that it's still returning "sensible" values.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -46,7 +46,7 @@ require (
 	github.com/spf13/cobra v1.2.1
 	github.com/spf13/viper v1.9.0
 	github.com/stretchr/testify v1.7.0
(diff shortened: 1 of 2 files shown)


mnowster:
It's part of what we're investigating so let's try it. We can revert if it something unexpected happens ⚡

mnowster (approved):

JamWils (approved):

PR #1507, opened 2022-02-23
feat: Create endpoint for fetching the flux namespace

### LINKED ISSUE #1465: Create endpoint for fetching the flux namespace
I think we should avoid making assumptions about flux namespace (https://github.com/weaveworks/weave-gitops/pull/1451/files#diff-35acf2adc3a7b1021fb41d3473b47e030d3708ba2e8ff7a4b76c13ee80b4076cR42) and grab it by using something like: https://github.com/weaveworks/weave-gitops/blob/v2/pkg/kube/kubehttp.go#L496

I'm not sure we want to keep that interface but the actual method is useful.

_Originally posted by @luizbafilho in https://github.com/weaveworks/weave-gitops/pull/1451#discussion_r809097989_

Acceptance Criteria:
* Make an endpoint that works like the one linked above, so that the UI can figure out which namespace(s) to query

yitsushi:
Add new endpoint to query flux namespace. For simplicity, no arguments
on the request and returns only the name of the namespace. That fulfills
the original request so we can remove the constant[^1] from the UI.

== Considered solutions

1. Create a general endpoint

The consumer could query namespaces and filter them with provided
filters.

Dropped this solution, because it could open up some weird surfaces
where the client can query arbitrary filters.

2. Create a general endpoint with strict filtering options

The consumer could query namespaces with a set of filter options and if
they are not specified, throw back an error.

Dropped this solution, because changing the filtering rules would
require code change. The final solution in this change-set has the same
issue, but less complex. If we don't get extra value out of the
complexity, it's not worth it.

== Follow-up

If/when we need a complex endpoint to query all namespaces with filter
options, then we can add one with a plan like:

* Who/what will use it?
* What is the desired input and output?
* How to achieve this?
* What limitations do we want to implement?
* How much freedom do we want to give to the consumer?

== Other changes

* `make proto` updated a lot of auto-generated files.
* In `tools/dependencies.toml`, removed the system-wide installation of
  a tool, because the last thing I want from a project to download a
  binary and place it somewhere with root permission. I'm not giving
  root permissions just to install dependencies.
* Use `tools/bin/tilt` in `Makefile` (see previous entry).
* Replaced the logic behind `getMatchingLabels`. Using the `with option`
  pattern, we can build up more complex labeling queries, and the
  codebase will remain clean.

  ```go
  match := matchLabel(
    withPartOfLabel(appName),
    withInstanceLabel(appInstance),
    withManagedByLabel(appManager),
  )
  ```

[^1]: https://github.com/weaveworks/weave-gitops/pull/1451/files#diff-35acf2adc3a7b1021fb41d3473b47e030d3708ba2e8ff7a4b76c13ee80b4076cR42

Closes #1465


core/server/namespace_test.go
```diff
@@ -0,0 +1,63 @@
...
+
+	ctx := context.Background()
+
+	coreClient, cleanup := makeGRPCServer(k8sEnv.Rest, t)
+	defer cleanup()
+
+	_, client, err := kube.NewKubeHTTPClientWithConfig(k8sEnv.Rest, "")
+	g.Expect(err).NotTo(HaveOccurred())
+
+	// k8sClient := fake.NewClientBuilder().WithScheme(kube.CreateScheme()).Build()
```
jpellizzari, line 26: Left-over comment?

jpellizzari (approved):
Looks good! It appears we have some proto files changing that probably don't need to change; this is probably due to `buf` version mismatches or something of the sort. Not unique to this PR though.

core/server/namespace_test.go
yitsushi: yes

buf.lock
```diff
@@ -4,14 +4,8 @@ deps:
   - remote: buf.build
     owner: googleapis
     repository: googleapis
-    branch: main
-    commit: 0039142542b74d11b0ae5a1e9fd4d523
-    digest: b1-AzPoqjtlLpx5dVojvgRauI1re5i4LPsghzU0KO-16o4=
-    create_time: 2022-01-05T15:03:17.46488Z
+    commit: bdfe2a6dbea44b2aadd202aef3b02139
```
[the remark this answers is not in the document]
yitsushi: reverted

yitsushi:
Closing this in favor of #1515 

Reason: Some PR checkers are using org secrets and does not like forks.

PR #1508, opened 2022-02-23
Added validation to registry script

josecordaz:
Closes: 

**What changed?**
- Added validation with creating registry

**Why?**
- We used to use port 5000 for the local registry. But the current script uses port 5001 and only creates the registry if it is running. But if the register is using different ports from what tilt is expecting then tilt won't be able to push to the registry. 

**How did you test it?**
- Manually

**Release notes**

**Documentation Changes**

### DIFF
--- a/tools/kind-with-registry.sh
+++ b/tools/kind-with-registry.sh
@@ -43,6 +43,14 @@ if [ "${running}" != 'true' ]; then
   docker run \
     -d --restart=always -p "127.0.0.1:${reg_port}:5000" --name "${reg_name}" \
     registry:2
+else
+  # Validate the ports used by the running register are the ones we need.
+  right_mapping_ports="5000/tcp -> 127.0.0.1:${reg_port}"
+  registry_port=$(docker port ${reg_name})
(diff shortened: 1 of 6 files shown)


tools/kind-with-registry.sh
```diff
@@ -41,8 +41,16 @@ esac
...
   docker run \
-    -d --restart=always -p "${reg_port}:5000" --name "${reg_name}" \
+    -d --restart=always -p "127.0.0.1:${reg_port}:5000" --name "${reg_name}" \
     registry:2
+else
+  # Validate the ports used by the running register are the ones we need.
+  right_mapping_ports="5000/tcp -> 127.0.0.1:${reg_port}"
+  registry_port=$(docker port ${reg_name})
+  if [ "${registry_port}" != "${right_mapping_ports}" ]; then
+    echo "It seems the current registry is running on different port configuration than expected:\n\n Current => ($registry_port) \n Expected ($right_mapping_ports). \n\nTry deleting registry manually using 'docker rm -f $reg_name' "
```
jpellizzari, line 51: :tada: Nice

jpellizzari (approved):
LGTM nice work

PR #1514, opened 2022-02-24
Remove explicit snyk code scan

ozamosi:
The Snyk github integration is working now, so we shouldn't run the
    CLI.
    
Part of #1503


mnowster (approved):
:shipit:

PR #1519, opened 2022-02-24
Use ginkgo2's test runner to run all tests

ozamosi:
Note: this has no influence on whether we write our tests using ginkgo or not! This is just about swapping the test runner we use.

It also isn't introducing a new dependency - we already used ginkgo for writing tests, and for running _some_ tests.
    
It has Several Neat-Looking Features™, none of which we're using. However, there's one thing I consider a killer feature: at the end of a test run, if there were errors, it prints what those errors were.
    
That is, when `go test` finishes a run, it prints:
```
FAIL
```
so you know to go look through the logs looking for something that failed. Whereas when `ginko` finishes a run, it prints:
```
Ginkgo ran 42 suites in 36.107255816s
    
There were failures detected in the following suites:
  server ./pkg/server
  server ./test/integration/server [Compilation failure]
    
Test Suite Failed
```
which is actually helpful.


Makefile
```diff
@@ -27,15 +27,20 @@ endif
 ##@ Default target
 all: gitops ## Install dependencies and build Gitops binary
 
+
+TEST_TO_RUN=./...
 ##@ Test
-unit-tests: dependencies  ## Run unit tests
+unit-tests: dependencies ## Run unit tests
+	which ginkgo || go install github.com/onsi/ginkgo/v2/ginkgo
```
dhwthompson, line 34: It looks like `which ginkgo` will succeed for folks with Ginkgo v1 installed. Is that a problem? I’m guessing from the fact that you’ve upgraded the library that the v1 CLI doesn’t have the functionality this PR needs; is that right?

dhwthompson (commented):
I feel like the description on this PR doesn’t quite capture the whole story. This is also upgrading our library from Ginkgo v1 to v2, right?

From a quick skim of the [Ginkgo migration guide](https://onsi.github.io/ginkgo/MIGRATING_TO_V2), it looks like there are quite a few features that have either changed or been removed. How confident are we that none of these changes affect our various test suites and CI workflows?

Makefile
ozamosi: This is not a problem, however the test output won't be as pretty (the table I want won't be there).

Using the ginkgo2 CLI with ginkgo v1 tests is a problem, however, so if I want to be able to use the v2 CLI, I have to update the library to v2.

ozamosi:
>I feel like the description on this PR doesn’t quite capture the whole story. This is also upgrading our library from Ginkgo v1 to v2, right?

It is - the upgrade is a separate commit, I can make it a separate PR if you would prefer. I haven't seen any real difference, though - there's two timeouts I've removed but the tests appear to be way faster than that timeout anyway so I suspect we've deleted the reason for them, two tables that now don't need a table import any more, and a bunch of line noise.

> How confident are we that none of these changes affect our various test suites and CI workflows?

The full unit-test test suite runs here already as of... recently, so they're all being exercised. The integration tests are all gone because all functionality they tested is gone. I'm not gonna worry about the acceptance tests.

Makefile
```diff
@@ -27,15 +27,20 @@ endif
 ##@ Default target
 all: gitops ## Install dependencies and build Gitops binary
 
+
+TEST_TO_RUN=./...
```
dhwthompson, line 31: Is the intent of this variable that people can change it to override which unit tests they’re running? If so, would it make sense to assign it with `?=`, so that it’s possible to override it on the command line? That solution seems to work for me, but I’m no Make expert, so I might have missed something.

If it’s intended for something else, what is that? I’m struggling to figure it out.

If it’s a remnant of trying out a different approach, how about we fold it back into the `unit-tests` command for consistency’s sake?

Makefile
dhwthompson, line 34: When I run this locally, I always get the output:

```
which ginkgo || go install github.com/onsi/ginkgo/v2/ginkgo
/Users/david/go/bin/ginkgo
```

Given this is going to run on every test command, is there something we can do to make it quieter, at least in the case where Ginkgo is already installed?

dhwthompson, line 34: Another question on the Ginkgo check: is that worth pulling out somewhere, so that it only occurs in one place rather than four? That seems to make sense to me as a separate phony target, but there may well be more idiomatic Make-y ways to do it.

dhwthompson (commented):
Cool cool. I’m satisfied that this isn’t going to break our tests in any particularly nasty ways, and I like the extra clarity we’re getting from the new test runner.

Just a couple of extra questions and suggestions on how we could go about setting it up.

Makefile
ozamosi: Right. Will do.

Somewhat pointless context:

I wrote these patches on top of some of my other PRs to enable more tests. I've used it as `make unit-tests`, wait for the test to finish, and then copy a failing suit name and ploink it at the end, e.g. `make unit-tests TEST_TO_RUN=./pkg/server` - this works just fine, as explicit make variables can override defaults with `=`.

However `TEST_TO_RUN=./pkg/server make unit-tests` doesn't work now. Allowing that too would be more helpful. I think `?=` is a GNU Make-ism, but we don't to my knowledge have any passionate OpenBSD people, so that's fine.

Makefile
ozamosi: The integration tests are gone, the acceptance tests are almost gone, and the `test` and `unit-tests` target do basically the same thing so I implemented one in terms of the other.

dhwthompson: Fair enough. That sounds like a good argument for it not being worth pulling out.

Makefile
dhwthompson: That was super helpful context; thank you. Not pointless at all. 😄

Makefile
dhwthompson: I see what you did here: using `@go install github.com/onsi/ginkgo/v2/ginkgo` and letting Go quietly handle the case where it’s already installed. Nice. 👍

dhwthompson (approved):
Looks good to me. Thanks for making some of those changes; more importantly, thanks for providing me with some extra surrounding context so I could understand for myself.

I’m happy with this; merge away.

Issue #1520, opened 2022-02-24
Implement the source detail page(s)


jpellizzari:
Implement the source detail page: https://www.figma.com/file/IVHnM9iyeFWpd11evtY8ux/Weave-GitOps?node-id=6074%3A20012

Figma drawing is `bianca - Source Detail Page - Git Repository`

PR #1522, opened 2022-02-24
Add GitRepositoryDetail page

jpellizzari:
Part of #1520 

A GitRepo that is working:
Screenshot from 2022-02-24 12-33-59

One that is in an error state:
Screenshot from 2022-02-24 12-34-10

### DIFF
--- a/ui/components/InfoList.tsx
+++ b/ui/components/InfoList.tsx
@@ -1,0 +1,40 @@
+import _ from "lodash";
+import * as React from "react";
+import styled from "styled-components";
+import Text from "./Text";
+
+const InfoList = styled(
+  ({
+    items,
+    className,
+  }: {
+    className?: string;
+    items: { [key: string]: any };
+  }) => {
+    return (
+      <table className={className}>
+        <tbody>
+          {_.map(items, (v, k) => (
+            <tr key={k}>
+              <td>
+                <Text capitalize bold>
+                  {k}:
+                </Text>
+              </td>
+              <td>{v || "-"}</td>
+            </tr>
+          ))}
+        </tbody>
+      </table>
+    );
+  }
+)`
+  tbody tr td:first-child {
+    min-width: 200px;
+  }
+  tr {
+    height: 16px;
+  }
+`;
+
+export default InfoList;
--- a/ui/components/SourceDetail.tsx
+++ b/ui/components/SourceDetail.tsx
@@ -1,0 +1,105 @@
+import _ from "lodash";
+import * as React from "react";
+import styled from "styled-components";
+import { useListAutomations } from "../hooks/automations";
+import { useListSources } from "../hooks/sources";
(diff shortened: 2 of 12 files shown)


ui/components/InfoList.tsx
```diff
@@ -0,0 +1,40 @@
...
+                <Text capitalize bold>
+                  {k}:
+                </Text>
+              </td>
+              <td>{v || "-"}</td>
+            </tr>
+          ))}
+        </tbody>
+      </table>
+    );
```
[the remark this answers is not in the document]
joshri: Here's the spacing with short error message - I'd personally rather have everything in one column from the start if we could like in the long error message @bia.

image

ui/components/SourceDetail.tsx
```diff
@@ -0,0 +1,105 @@
...
+              text={`Error: ${msg}`}
+            />
+          )}
+        </div>
+      </Flex>
+      {error && (
+        <Alert severity="error" title="Error" message={error.message} />
+      )}
+      <div>
+        <h3>{s.type}</h3>
```
joshri, line 80: `<Text color='neutral30' fontSize='large' >{s.type}</Text >`

ui/pages/v2/GitRepositoryDetail.tsx
```diff
@@ -0,0 +1,47 @@
...
+    </Page>
+  );
+}
+
+export default styled(GitRepositoryDetail).attrs({
+  className: GitRepositoryDetail.name,
+})`
+  ${TitleBar} {
+    margin-bottom: 0;
+  }
```
joshri, line 42: What was happening without this override?

joshri (approved):

ui/pages/v2/GitRepositoryDetail.tsx
jpellizzari: Extra margin on `TitleBar`. It assumes that there is a title, and in this case I am keeping the title in the component instead of the page.

ui/components/InfoList.tsx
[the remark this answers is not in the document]
jpellizzari: > I'd personally rather have everything in one column from the start

Same.

ui/pages/v2/GitRepositoryDetail.tsx
joshri: Mmmmmm I see 🌞

ui/components/SourceDetail.tsx
jpellizzari: Created a follwup https://github.com/weaveworks/weave-gitops/issues/1542

bia:
Looking good.

Two nits that I don't know if it's worth opening an issue over :
- The distance between `h1` and `h2` should be `12px`, between `h2` and the info boxes should be `24px`
- `h2` as well as the semibold field names ("URL:") should be `Neutral 30`, `#737373`

PR #1524, opened 2022-02-24
Add a RoleBinding for `User: admin`

sarataha:
Closes: weaveworks/weave-gitops-enterprise#507

**What changed?**
Added a default rolebinding for `admin` user.

**Why?**
To give access rights to user `admin` when logging in with the superuser, as user doesn't have any roles currently.

**How did you test it?**
Tested manually.

**Release notes**

**Documentation Changes**


yiannistri (commented):
Does it matter that this role binding has the same name as an existing one ? 🤔 

I tried locally against a kind cluster and I can't see this role binding. Not sure if I'm missing something.

yiannistri (approved):
LGTM 👍

Issue #1459, opened 2022-02-16
Re-work nav menu to support sub-menu options


jpellizzari:
The current navigation implementation uses the MUI `Tabs` component that won't support sub-menu items.

Acceptance Criteria:

Make this:
![Screenshot from 2022-02-16 10-06-12](https://user-images.githubusercontent.com/2802257/154329667-053b3b98-7561-4d54-9d1f-28e8484a41f4.png)

Look like this:
![Screenshot from 2022-02-16 10-16-08](https://user-images.githubusercontent.com/2802257/154329696-ff408b15-d953-4c0c-8532-2b4a07c4a350.png)

PR #1525, opened 2022-02-24
Adjust Nav Styling

joshri:
Closes: #1459

image

All you have to do to make a Nav Item gray is add the key/value `{sub: true}` to the navItems object array.

Also did some fixes for the logo, background, and border radius on main containers.


ui/components/Logo.tsx
```diff
@@ -15,10 +15,10 @@ type Props = {
 function Logo({ className }: Props) {
   return (
     <Spacer padding="medium">
-      <Flex className={className}>
+      <Flex className={className} align>
         <img src={logoSrc} />
         <Spacer padding="xxs" />
-        <img src={titleSrc} />
+        <img src={titleSrc} style={{ height: "90%" }} />
```
jpellizzari, line 21: What is this change trying to accomplish?

ui/pages/v2/Sources.tsx
```diff
@@ -17,7 +17,7 @@ function SourcesList({ className }: Props) {
       loading={isLoading}
       className={className}
     >
-      <SourcesTable sources={sources} />
+      <SourcesTable sources={sources || []} />
```
jpellizzari, line 20: This looks like a leaky abstraction. Can we make `<SourcesTable />` handle `undefined` instead?

jpellizzari (commented):
This doesn't look like the designs. We probably need to change the font of the sub menu item and make sure that menu items are grouped together.

ui/pages/v2/Sources.tsx
joshri: I'm going to remove this and handle it when I do the filtering. It has nothing to do with the Nav I was just checking out the errors

ui/components/Logo.tsx
joshri: Check it out without the set height - it stretches the text out to fit the container.

joshri:
@jpellizzari good call out on the grouping together - I missed that. Sub menu item font I think needs to go from 600 - 400 font weight based on figma

joshri:
image

FIXED

ui/components/Layout.tsx
```diff
@@ -141,12 +144,26 @@ function Layout({ className, children }: Props) {
...
+                  <Flex column>
+                    <StyleLinkTab
+                      key={n.value}
+                      label={n.label}
+                      to={formatURL(n.value)}
+                      value={n.value}
+                    />
+                    {_.map(n.subItems, (sub) => {
+                      return (
+                        <StyleLinkTab
```
jpellizzari, line 156: Looking at Figma, sub items should be `12px` below the main one. This looks a like a little bigger gap.

jpellizzari (approved):
Nice work on the sub items pattern. One nit about the distance, but assuming that matches the Figma drawing, LGTM.

jpellizzari (approved):

PR #1531, opened 2022-02-25
Turn off gitlab acceptance tests

ozamosi:
Our gitlab user has got themselves banned, so our gitlab tests can't
run, so temporarily disable the tests that are currently failing
because of it.

This should be reverted when weaveworks/corp#2502 is resolved.


mnowster:
Thanks. We're working on getting the account unblocked and issue resolved.

mnowster (approved):

PR #1534, opened 2022-02-25
feat: Implement the GetHelmRelease endpoint

### LINKED ISSUE #1498: Implement the `GetHelmRelease` endpoint
Blocks #1497 

We need to be able to render some details about a HelmRelease in the context of a Automation. The endpoint will probably look very similar to this `GetKustomization` endpoint:

https://github.com/weaveworks/weave-gitops/blob/e691bde49162c1f7470165617684e4ee0f713365/core/server/automations.go#L41

See https://github.com/weaveworks/weave-gitops/pull/1489 for more details

Acceptance Criteria:
* Needs to have a list of `ReconciledKinds` like the `GetKustomization` endpoint

yitsushi:
Add endpoint to get a single HelmRelease.

Additional changes:

* Moved `ListHelmReleases` and `GetHelmRelease` endpoint definitions into
  their own file, to get a better separation (`helm.go`).
* Moved the tests for `ListHelmReleases` and `GetHelmRelease` endpoints into
  the corresponding test file (`helm_test.go`).

For Kustomization, the getInventory function is under `types`, but to
get the inventory for a `HelmRelease`, we need a lot of extra queries from
k8s and we already have a lot of helper functions in `server`.

Refactored tests to use `server_test` module instead of `server`.

Benefits:
* If a branch is not covered with test, it might be deleted, or some
  cases are not tested.
* We test the same interface as a consumer will use.
* We can't "reuse" private functions for testing, that way we can see
  during testing if we have to publish a function or not.

Closes #1498

### DIFF
--- a/core/server/helm_release.go
+++ b/core/server/helm_release.go
@@ -129,8 +129,6 @@ func getHelmReleaseInventory(ctx context.Context, helmRelease v2beta1.HelmReleas
 	found := map[string]bool{}
 
 	for _, entry := range objects {
-		entry.GetAPIVersion()
-
 		idstr := strings.Join([]string{entry.GetAPIVersion(), entry.GetKind()}, "_")
 
 		if !found[idstr] {


core/server/automations.go
```diff
@@ -57,25 +56,3 @@ func (cs *coreServer) GetKustomization(ctx context.Context, msg *pb.GetKustomiza
 
 	return &pb.GetKustomizationResponse{Kustomization: res}, nil
 }
-
-func (cs *coreServer) ListHelmReleases(ctx context.Context, msg *pb.ListHelmReleasesRequest) (*pb.ListHelmReleasesResponse, error) {
```
jpellizzari, line 61: Moving this to a new file makes the naming inconsistent. If we are really going for that level of separation, we also need to rename `automations` to `kustomizations`, since that is all it contains now.

yitsushi: I would like to split them up, later on much easier to find and change without worrying about a conflict between PRs. I'll rename the file to `helm_release` and `automations` to `kustomization`

yitsushi:
I would be happy to test the inventory part, but I have no idea how to even approach that part, with creating a secret, and create encoded stuff with that secret in there. If required, I'll spend more time and figure it out.

core/server/helm_release.go
```diff
@@ -0,0 +1,148 @@
...
+	}
+
+	var gvk []*pb.GroupVersionKind
+
+	found := map[string]bool{}
+
+	for _, entry := range objects {
+		entry.GetAPIVersion()
+
+		idstr := strings.Join([]string{entry.GetAPIVersion(), entry.GetKind()}, "_")
```
Callisto13, line 134: Do you know why we call `entry.GetAPIVersion` here and then again on 134? I can see we are doing it somewhere else in the code, so am curious what repeating but not keeping the value gives us.

core/server/helm_release_test.go
```diff
@@ -0,0 +1,109 @@
+package server
```
Callisto13, line 1: can we make this `server_test`? we are testing an exposed interface, so we can test against it from the outside

core/server/helm_release_test.go
```diff
@@ -0,0 +1,109 @@
...
+
+	res, err := c.ListHelmReleases(ctx, &pb.ListHelmReleasesRequest{
+		Namespace: ns.Name,
+	})
+	g.Expect(err).NotTo(HaveOccurred())
+	g.Expect(res.HelmReleases).To(HaveLen(1))
+	g.Expect(res.HelmReleases[0].Name).To(Equal(appName))
+}
+
+func TestGetHelmRelease(t *testing.T) {
```
Callisto13, line 38: are you going to test any of the `getHelmReleaseInventory` bits? it feels important to us that we are populating that bit of the returned value correct

Callisto13: ah just literally saw your comment on that 👍

core/server/helm_release.go
yitsushi: I have no idea, and because I have no idea, I just used the same logic XD

core/server/helm_release_test.go
yitsushi: I'm 100% on the that side (blackbox testing), but the rest of the code follows this logic (whitebox testing), so I did the same.

core/server/helm_release.go
Callisto13: does it hurt anything if you remove the first one?

core/server/helm_release_test.go
yitsushi: Did a few poke on it, all the tests are using the same functions (and global variable), so I moved everything and now it's all `server_test`

core/server/helm_release.go
yitsushi: while testing it does not hurt, but I don't know if it does with real k8s objects coming from a real k8s cluster.

Callisto13: ack, makes sense, would be good to prove and remove at somepoint

yitsushi: Removed... Why?

```
entry.GetAPIVersion()

idstr := strings.Join([]string{entry.GetAPIVersion(), entry.GetKind()}, "_")
```

The function call sequence is:
1. `entry.GetAPIVersion()`
2. `entry.GetKind()`
3. `strings.Join()`

With that extra call the difference is an extra call only:
1. `entry.GetAPIVersion()`
2. `entry.GetAPIVersion()`
3. `entry.GetKind()`
4. `strings.Join()`

I see only one case when it can have any effects, if the underlying code is far from optimal and has an internal counter (how many times it was called) or something, and it does not make sense.

jpellizzari (approved):
LGTM :+1:

core/server/fluxruntime_test.go
```diff
@@ -1,11 +1,12 @@
-package server
+package server_test
```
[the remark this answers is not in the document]
Callisto13: It's probably a per-codebase decision. It's not the end of the world to white box test some layers, but it is not _great_ practice when it comes to external facing contracts (which these are). Given we are tidying things, now feels like a good time to clean this up.

core/server/helm_release_test.go
yitsushi: My issue with testing that function... well I have to create a full secret + store + data + "i don't know what else", because unlike kustomization,  to get helm inventory, we query the api to get a helm secret, get release from the secret storage, decode + unzip (if gzipped), and finally read object from the storage. or something like that, so that would require a huge bootstrapping. I'll try it, because I would like to test it.

Issue #1499, opened 2022-02-23
Ensure we use TLS when serving API


richardcase:
We need to ensure that the API is served with TLS so that the transport is encrypted. Without this tokens can be sniffed etc.

jpellizzari:
We probably need to do an `--insecure` flag to disable this as well.

jpellizzari:
Related https://github.com/weaveworks/weave-gitops-enterprise/issues/405

josecordaz:
Some questions:
1.- Which domain the certificate would need to point to?
2.- Where can I get one for production? corp-it? Do we need one for local development? Or we just skip TLS locally?
3.- Should we store the certificate in the repository?

richardcase:
> 1.- Which domain the certificate would need to point to?
This depends on the user who will be using Core. We can't stipulate anything here.

> 2.- Where can I get one for production? corp-it? Do we need one for local development? Or we just skip TLS locally?
I think we need to provide docs for generating self-signed certs using openssl and to specify these when installing core.  We could also write docs for using cert-manager/lets encrypt. Perhaps even the TLS certs feature directly in k8s as well.

For local dev we can rely on self-signed certs (and we could include a script/make target for generating them). And we should have the `--insecure` option where we don't use TLS. But the `--insecure` option shouldn't be the default....it needs to be an explicit decision by the user to run insecurely.

> 3.- Should we store the certificate in the repository?
I don't think you can unless you use sealed secrets or sops.

@josecordaz - shall we chat?

josecordaz:
@richardcase  Thanks for answering. I'll try to address your last comments. I'll definitely reach out if I get some more questions.

josecordaz:
@richardcase I just addressed your comments in this draft PR https://github.com/weaveworks/weave-gitops/pull/1537. Please take a look once you have a chance. Let me know if that would work.

PR #1537, opened 2022-02-25
Added TLS security

josecordaz:
Closes: #1499 

**What changed?**
- Added parameters to pass in certificate and key. If they are not provided, they will be generated. Very convenient for development mode.

**Why?**
- Add security to all the API interactions.

**How did you test it?**
- Manually

**Release notes**

**Documentation Changes**


Makefile
```diff
@@ -213,6 +213,13 @@ merged.lcov:
 	lcov --add-tracefile coverage/unittest.info --add-tracefile coverage/integrationtest.info -a coverage/lcov.info -o merged.lcov
 
 ##@ Utilities
+tls-files:
+ifeq (, $(shell which mkcert))
+ $(error "mkcert is not installed, consider following this instructions: https://github.com/FiloSottile/mkcert#installation ")
```
jpellizzari, line 218: Does there need to be a `dependencies.toml` or a README entry to have devs install this tool?

jpellizzari (approved):

README.md
```diff
@@ -72,8 +72,9 @@ To set up a development environment for the CLI
 3. Run `make all` to install dependencies and build binaries and assets
 4. Start a `kind` cluster like so: `KIND_CLUSTER_NAME=<some name> ./tools/kind-with-registry.sh`
 5. Run `./bin/gitops install --config-repo=<repo url>`
-6. Start the in-cluster API replacement job (powered by [http://tilt.dev](tilt.dev)) with `make cluster-dev`
-7. make or make unit-tests to ensure everything built correctly.
+6. Run `make tls-files`
```
luizbafilho, line 75: since we are generating tls keys in memory are these new tools and generating localhost keys really necessary?

cmd/gitops-server/cmd/cmd.go
```diff
@@ -74,6 +81,10 @@ func NewCommand() *cobra.Command {
 	cmd.Flags().StringVar(&options.NotificationControllerAddress, "notification-controller-address", "http://notification-controller./", "the address of the notification-controller running in the cluster")
 	cmd.Flags().IntVar(&options.WatcherPort, "watcher-port", 9443, "the port on which the watcher is running")
 
+	cmd.Flags().StringVar(&options.TLSCert, "tls-cert-file", "", "filename for the TLS certificate, in-memory generated if omitted")
+	cmd.Flags().StringVar(&options.TLSKey, "tls-private-key", "", "filename for the TLS key, in-memory generated if omitted")
```
luizbafilho, line 85: to keep consistency with the other flag and with `kube-api`, we could add the `-file` suffix here.

cmd/gitops-server/cmd/cmd.go
```diff
@@ -40,6 +43,9 @@ type Options struct {
 	LoggingEnabled                bool
 	OIDC                          OIDCAuthenticationOptions
 	NotificationControllerAddress string
+	TLSCert                       string
```
luizbafilho, line 46: Just to match the flag and avoid confusion we could change `TLSCert` and `TLSKey` to `TLSCertFile` and `TLSKeyFile`

luizbafilho (changes requested):

cmd/gitops-server/cmd/cmd.go
```diff
@@ -74,6 +81,10 @@ func NewCommand() *cobra.Command {
 	cmd.Flags().StringVar(&options.NotificationControllerAddress, "notification-controller-address", "http://notification-controller./", "the address of the notification-controller running in the cluster")
 	cmd.Flags().IntVar(&options.WatcherPort, "watcher-port", 9443, "the port on which the watcher is running")
 
+	cmd.Flags().StringVar(&options.TLSCert, "tls-cert-file", "", "filename for the TLS certificate, in-memory generated if omitted")
+	cmd.Flags().StringVar(&options.TLSKey, "tls-private-key", "", "filename for the TLS key, in-memory generated if omitted")
+	cmd.Flags().BoolVar(&options.NoTLS, "no-tls", false, "do not attempt to read TLS certificates")
```
richardcase, line 86: I think perhaps `--insecure` would be a better flag name?

cmd/gitops-server/cmd/cmd.go
```diff
@@ -74,6 +81,10 @@ func NewCommand() *cobra.Command {
...
+
```
richardcase, line 87: Should we also add a flag so that we can also enable client certificate validation as well?

README.md
josecordaz: Fair question. The certificates generated in memory are not valid, but the ones generated here are. I guess this is a matter of preference.

pkg/server/tls/tls.go
```diff
@@ -0,0 +1,89 @@
...
+	}
+
+	tlsConfig := &tls.Config{
+		Certificates: []tls.Certificate{cert},
+	}
+
+	return tlsConfig, nil
+}
+
+func generateKeyPair(hosts []string) ([]byte, []byte, error) {
```
yitsushi, line 37: Potentially we can reuse an existing rootCA. On my local environment I have one rootCA,added that cert as trusted cert and i just generate keys per "site".

https://gist.github.com/yitsushi/2cd4ec3158c2f334d07f493da7f829e0

so basically the `priv, err :=` part is not generated, it can come from a file.

cmd/gitops-server/cmd/cmd.go
yitsushi: ^^ this is should be there 100%, in production use, I would never turn it off.

README.md
luizbafilho: Given that we are providing steps and tools to generate valid certs and having invalid auto-generated in-memory certs is a bit unusual, I have personally never seen any other project doing that, shouldn't we just remove the invalid auto-generate ones?

Since also we set up our dev environment to use the valid certs with tilt that auto-generate option will be mostly unused.

Makefile
josecordaz: I tried to add a toml entry but this dependency doesn't have the specific binary for x86_64 architectures. 
```
Downloading https://github.com/FiloSottile/mkcert/releases/download/v1.4.3/mkcert-v1.4.3-darwin-x86_64
curl: (22) The requested URL returned error: 404
```
image

pkg/server/tls/tls.go
josecordaz: Actually, I'm deleting this function as it is not relevant anymore.

tools/validate-mkcert.sh
```diff
@@ -0,0 +1,9 @@
+# shellcheck shell=bash
+if ! command -v mkcert &> /dev/null
+then
+    echo "mkcert is not installed, consider following this instructions: https://github.com/FiloSottile/mkcert#installation "
+    exit
```
yitsushi, line 5: `exit 1` to indicate error, and make it fail if we use in a pipeline (like any other tasks that has `tls-files` as a dependency)

josecordaz: This is not supposed to run as part of a pipeline. But I would like to start using this patter as this could end up in a pipeline at some point.

yitsushi (approved):

cmd/gitops-server/cmd/cmd.go
```diff
@@ -269,6 +285,34 @@ func runCmd(cmd *cobra.Command, args []string) error {
 	return nil
 }
 
+func listenAndServe(srv *http.Server, options Options) error {
+	if options.Insecure {
+		log.Println("TLS connections disabled")
+		return srv.ListenAndServe()
+	}
+
```
richardcase, line 293: Given insecure is false here we should probably check that the `options.TLSCertFile` and `options.TLSKeyFile` are not empty. If they are empty then we should return an error.

richardcase: And check that the files exist?

josecordaz: @richardcase You are right we should check that those flags are not empty. On the other hand, if the files do not exist we will get this type of errors already:
image
image

PR #1540, opened 2022-02-25
Update upgrade cmd to no longer create a pr

J-Thompson12:
**What changed?**
So this is just an idea and I would like some thoughts on it. This pr changes upgrade cmd to no longer create a PR. Instead it just adds the files locally and the user pushes it up. There are some issue I think may occur. I am assuming the top level folder being used is "clusters". I dont think we can assume that in the future but I am not really sure how we can determine what the file structure for all users will look like. The command also now has to be run from the app config dir. The config repo url is determined based off the .git file of the current dir. I am also assuming that the subfolder for clusters is named after the cluster. I know in flux this isnt the case so that can also be an issue.

Things to address before merging:
-  I am fairly certain I have broken profiles with this pr. Honestly still not sure how they work.
-  I will update the tests after I know if this is a working solution.

**Why?**

**How did you test it?**

**Release notes**

**Documentation Changes**

### DIFF
--- a/cmd/gitops/upgrade/cmd.go
+++ b/cmd/gitops/upgrade/cmd.go
@@ -4,6 +4,8 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"os/exec"
+	"strings"
 
 	"github.com/spf13/cobra"
 	wego "github.com/weaveworks/weave-gitops/api/v1alpha1"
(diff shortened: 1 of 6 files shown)


ozamosi:
I don't like having this problem.

On one hand, I don't think we can assume any kind of file structure - https://fluxcd.io/docs/guides/repository-structure/ suggests many ways of structuring your files. And with the new multi-tenant flux features, even asking the cluster what resources it's applying to itself is fraught.

On the other, you can't upgrade something you can't find.

Is this the kind of thing we'd need to ask the user? Like, "Is this repo the one where you store your flux manifests? [y/n]", "The upgrade manifests will be created in `cluster/kind-kind`, is this right? [y/n]" This is straying far into product territory, but I can't see a way to solve this problem, so I feel we need to take a step back and ask ourselves if we're solving the right problem.

J-Thompson12:
I have been brainstorming this and thought that maybe we just have the user provide the path. We would no longer get the cluster name but just ask the user to give us the path to the cluster. 

The other issue I found today is that we also need to remove the old weave-gitops helm chart when enterprise is added. In that case I have yet to come up with a solution because I am unsure if the user will always add enterprise in the same place core was installed and we dont know the file name anymore.

J-Thompson12:
Wrote up a story to discuss this https://github.com/weaveworks/weave-gitops/issues/1560.

dhwthompson:
Given this has been succeeded by #1560 and the `v2` branch is no longer current, I’m assuming we can close this.

PR #1549, opened 2022-02-28
Cherry-pick superuser stuff from `main`

jpellizzari:
Picks some stuff of of `main` for us to work on in V2. Having the super-user stuff on the `v2` branch will unblock some other work.


.proxyrc
```diff
@@ -1,6 +1,6 @@
 {
   "/v1": {
-    "target": "http://localhost:9001/",
+    "target": "http://localhost:9000/",
```
ozamosi, line 3: I must admit I'm quite confused about the 9000 vs 9001 - the manifests/helm chart only exports 9001, so what's listening to 9000?

jpellizzari: [The `](https://github.com/weaveworks/weave-gitops/blob/c11faf222d0c648c16589fb01477d841c5a2d0e8/Tiltfile#L29)

yiannistri (approved):
The changes for `admin` flow look fine 👍. Just a note that you'll need to add the relevant RBAC manifests for the `admin` user to work.

Callisto13:
> Just a note that you'll need to add the relevant RBAC manifests for the admin user to work.

Will we need https://github.com/weaveworks/weave-gitops/pull/1524 as well then?

Callisto13:
I'll take this as-is now, and will pull over the role pr when i need it.

yiannistri:
@Callisto13 apologies just seen your message. Yes you'll likely need that otherwise the `admin` user won't be able to do much. But you may also need additional permissions, depending on what you need to access as that user.

By the way, that role should not be called `resources-reader` as it allows all verbs for some resources.

Callisto13:
> Yes you'll likely need that otherwise the admin user won't be able to do much. But you may also need additional permissions, depending on what you need to access as that user.

Cool i will pull the PR over into this branch and add more things as i need them

> By the way, that role should not be called resources-reader as it allows all verbs for some resources.

Yeh i think I saw that sara changed the name of that role

Issue #1509, opened 2022-02-23
Move breadcrumbs from enterprise to core


jpellizzari:
Implement these breadcrumbs in core:

![Screenshot from 2022-02-23 12-32-25](https://user-images.githubusercontent.com/2802257/155403435-c709e3a4-377d-4a50-a2fd-6b34bcaea231.png)

There is already a component built here: https://github.com/weaveworks/weave-gitops-enterprise/blob/main/ui-cra/src/components/Breadcrumbs.tsx

We won't be able to show the counts in the breadcrumbs, so that element should be optional.

Acceptance Criteria:
* Copy the component into the `weave-gitops` repo and implement it globally in the app
* The "count" element should be optional
* Export the component via our index file
* Create a follow up issue to point the `weave-gitops-enterprise` component at our exported component

PR #1550, opened 2022-02-28
Move Breadcrumbs and reconfigure for V2Routes

joshri:
Closes: #1509 

image

Adds Breadcrumbs component, and should also fix some errors caused by data tables mixing with the new sorting pr.

### DIFF
--- a/ui/components/AutomationsTable.tsx
+++ b/ui/components/AutomationsTable.tsx
@@ -4,6 +4,7 @@ import { Automation } from "../hooks/automations";
 import { formatURL } from "../lib/nav";
 import { AutomationType, V2Routes } from "../lib/types";
 import DataTable, { SortType } from "./DataTable";
+import KubeStatusIndicator, { computeReady } from "./KubeStatusIndicator";
 import Link from "./Link";
 
 type Props = {
@@ -46,6 +47,24 @@ function AutomationsTable({ className, automations }: Props) {
           label: "Namespace",
           value: "namespace",
         },
+        {
+          label: "Cluster",
+          value: "cluster",
+        },
+        {
+          label: "Status",
+          value: (a: Automation) =>
+            a.conditions.length > 0 ? (
+              <KubeStatusIndicator conditions={a.conditions} />
+            ) : null,
+          sortType: SortType.bool,
+          sortValue: ({ conditions }) => computeReady(conditions),
+        },
+        {
+          label: "Revision",
+          value: "lastAttemptedRevision",
+        },
+        { label: "Last Synced At", value: "lastHandledReconciledAt" },
       ]}
       rows={automations}
     />
--- a/ui/components/AutomationsTable.tsx
+++ b/ui/components/AutomationsTable.tsx
@@ -4,7 +4,6 @@ import { Automation } from "../hooks/automations";
 import { formatURL } from "../lib/nav";
 import { AutomationType, V2Routes } from "../lib/types";
 import DataTable, { SortType } from "./DataTable";
-import KubeStatusIndicator from "./KubeStatusIndicator";
 import Link from "./Link";
 
 type Props = {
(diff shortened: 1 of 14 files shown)


ui/components/Breadcrumbs.tsx
```diff
@@ -0,0 +1,53 @@
...
+    color: ${(props) => props.theme.colors.neutral00};
+  }
+`;
+
+export const Breadcrumbs = () => {
+  const { currentPage } = useNavigation();
+  const parentValue = getParentNavValue(currentPage);
+
+  function pageLookup(route) {
+    const parsed = qs.parse(location.search);
```
jpellizzari, line 23: Blocker: Is this the global `window.location`? This will make the component hard to test. Instead, we should pull location out of the context via the `useLocation` hook that `react-router` provides.

ui/components/DataTable.tsx
```diff
@@ -96,7 +96,7 @@ export const sortWithType = (rows: any[], sort: field) => {
 function UnstyledDataTable({
   className,
   fields,
-  rows,
+  rows = [],
```
jpellizzari, line 99: Not necessary to handle this in both places. I would say lets handle it in `sourceWithType` in case someone uses that in the future in another component.

ui/components/ReconciledObjectsTable.tsx
```diff
@@ -35,23 +35,33 @@ function ReconciledObjectsTable({
...
+            value: "name",
+            label: "Name",
+            sortType: SortType.string,
+            sortValue: ({ name }) => name,
+          },
           {
             label: "Type",
             value: (u: UnstructuredObject) => `${u.groupVersionKind.kind}`,
+            sortType: SortType.string,
+            sortValue: (u: UnstructuredObject) => `${u.groupVersionKind.kind}`,
```
jpellizzari, line 49: Not sure why the original was in a string template:

```suggestion
            sortValue: (u: UnstructuredObject) => u.groupVersionKind.kind,
```

ui/components/ReconciledObjectsTable.tsx
```diff
@@ -35,23 +35,33 @@ function ReconciledObjectsTable({
...
           },
           {
             label: "Namespace",
             value: "namespace",
+            sortType: SortType.string,
+            sortValue: ({ namespace }) => namespace,
```
jpellizzari, line 55: No longer necessary, right?

https://github.com/weaveworks/weave-gitops/pull/1522/commits/746b1beefa0b8cf476cdad61868f09e2d0dcd5aa

ui/components/__tests__/Breadcrumbs.test.tsx
```diff
@@ -0,0 +1,19 @@
+import React from "react";
+import renderer from "react-test-renderer";
+import { createMockClient, withContext, withTheme } from "../../lib/test-utils";
+import Breadcrumbs from "../Breadcrumbs";
+
+describe("snapshots", () => {
+  it("renders", () => {
+    const tree = renderer
```
jpellizzari, line 8: :+1: 

Can we get another snapshot case that shows a child route being rendered?

ui/components/__tests__/Breadcrumbs.test.tsx
```diff
@@ -0,0 +1,19 @@
...
+      .create(
+        withTheme(
+          withContext(<Breadcrumbs />, "/automations", {
```
jpellizzari, line 11: @jpellizzari TODO: rework these tests helpers to not be so cumbersome to use

jpellizzari (commented):
One blocker around `window.location`. Other than that, LGTM.

ui/components/ReconciledObjectsTable.tsx
joshri: My branch was out of date :( will fix

joshri: ALSO - noticed a problem with this - right now DataTable is using the field having a SortType as an indication of whether it's sortable or not - we need a different indication of which columns are sortable

ui/lib/nav.ts
```diff
@@ -41,6 +42,24 @@ export const getParentNavValue = (
   }
 };
 
+export const getPageLabel = (currentPage: string): string => {
+  const parsed = qs.parse(useLocation().search);
```
jpellizzari, line 46: :thinking:  Not sure how this is working. I was under the impression that you cannot call hooks from plain old JS like this. I thought it had to be within another hook.

jpellizzari (approved):
Looks good. No blockers for me.

ui/lib/nav.ts
joshri: Me too but it works so I'm not asking questions

ui/components/AutomationsTable.tsx
```diff
@@ -47,22 +46,6 @@ function AutomationsTable({ className, automations }: Props) {
           label: "Namespace",
           value: "namespace",
         },
-        {
```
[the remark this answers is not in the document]
joshri: Me neither put it back - I might have changed it trying to get the table to not break

PR #1558, opened 2022-03-01
Delete GetApplication, ListApplications endpoints, take 2

ozamosi:
This removes the endpoints and the backend implementation, the
same as #1455.

It removes some frontend bits - e.g. hooks - that merely wraps the
removed API calls.

It also removes 2 pages, Applications and ApplicationDetails, because
neither of them a) work b) can be accessed in the UI.

This breaks embedding - these pages are publicly exported to
enterprise - but I don't think that's important right now.


ui/__tests__/library.test.tsx
```diff
@@ -1,51 +0,0 @@
-import { render, screen } from "@testing-library/react";
```
jpellizzari, line 1: TODO: write a test for library integration for Core:reloaded

jpellizzari (approved):
LGTM

PR #1561, opened 2022-03-01
human readable logs added with flag for ui run

sourikghosh:
Signed-off-by: Sourik Ghosh <sourikghosh31@gmail.com>

Closes: 1559

**What changed?**
adds an extra flag to enable human-readable logs for  UI 

**How did you test it?**

**Release notes**

**Documentation Changes**


sourikghosh:
ping reviewer
@yiannistri

sourikghosh:
any suggestion for a good name for `human-readable-logs` for cli --flag

sourikghosh:
> Hi Sourik, thank you for your help with this issue 🙏
> 
> A couple of comments:
> 
> * If you run`gitops ui run` you'll notice that there are additional logs originating from the profile watcher. That controller initialises a logger and uses it without any configuration. The appropriate fix therefore would be to pass 2 additional options to that controller `LoggingEnabled` and `HumanLogs` and depending on their values to configure that logger accordingly. You don't need to worry about reconfiguring the existing logger in `cmd.go`, just make the profile watcher logger respect the CLI flags that have been passed.
> * Please squash your commits into one when your PR is ready for review.

just to confirm, are you talking about this method `StartWatcher()`??
https://github.com/weaveworks/weave-gitops/blob/c11faf222d0c648c16589fb01477d841c5a2d0e8/pkg/helm/watcher/server.go#L71-L74

I need to pass both the logging flags there and remove the reconfiguration changes from `ui/run/cmd.go`

yiannistri:
Yes, that's the one 👍

sourikghosh:
> Yes, that's the one 👍

I don't think we can do the no-op logger here.  So, sending `humanLog` flag only to change the logger encoding.
https://github.com/weaveworks/weave-gitops/blob/c11faf222d0c648c16589fb01477d841c5a2d0e8/pkg/helm/watcher/server.go#L71-L74

`cmd/gitops-server/main.go` also calls this `StartWatcher()`. Should I explicitly pass false over there?
https://github.com/weaveworks/weave-gitops/blob/c11faf222d0c648c16589fb01477d841c5a2d0e8/cmd/gitops-server/main.go#L87

cmd/gitops/ui/run/cmd.go
```diff
@@ -78,6 +79,7 @@ func NewCommand() *cobra.Command {
 	options = Options{}
 
 	cmd.Flags().BoolVarP(&options.LoggingEnabled, "log", "l", false, "enable logging for the ui")
+	cmd.Flags().BoolVar(&options.HumanLogs, "pretty-log", false, "enable human readable logging for the ui")
```
yiannistri, line 82: s/pretty-log/human-logs

yiannistri (changes requested):
Could we use a `logr.Discard()` instance if no logs are needed?

Callisto13:
Thanks for your work here @sourikghosh! Since you opened this PR, the codebase has started to be heavily refactored. This includes the removal of several commands. We are auditing how we do logging, so will likely end up superseding this PR.

For now we are not accepting any external PRs, but feel fee to come back in a month or so if you feel there are still things missing from the logs 👍

PR #1565, opened 2022-03-01
Fix error on calling .localeCompare on undefined in Datatable

jpellizzari:
Fixes an issue where a string can be `undefined` but we still call `.localeCompare` on it.

### DIFF
--- a/ui/components/DataTable.tsx
+++ b/ui/components/DataTable.tsx
@@ -98,7 +98,7 @@ export const sortWithType = (rows: Row[], sort: Field) => {
         else return 1;
 
       default:
-        return sortFn(a).localeCompare(sortFn(b));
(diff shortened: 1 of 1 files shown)


ui/components/DataTable.tsx
```diff
@@ -98,7 +98,7 @@ export const sortWithType = (rows: Row[], sort: Field) => {
         else return 1;
 
       default:
-        return sortFn(a).localeCompare(sortFn(b));
+        return (sortFn(a) || "").localeCompare(sortFn(b));
```
joshri, line 101: I think the `|| ''` also has to go on `sortFn(b)`

jpellizzari: It probably wouldn't have exploded, but added it anyway

joshri (approved):

PR #1566, opened 2022-03-02
chore(test): Add inventory tests on GetHelmRelease

yitsushi:
Extra tests for #1534 

(Merged #1534 to make sure I'm not blocking any UI related works while I'm trying to figure out how to test this properly)

### DIFF
--- a/core/server/helm_release_test.go
+++ b/core/server/helm_release_test.go
@@ -1,13 +1,21 @@
 package server_test
 
 import (
+	"bytes"
+	"compress/gzip"
 	"context"
+	"encoding/base64"
+	"encoding/json"
+	"fmt"
 	"testing"
 
 	helmv2 "github.com/fluxcd/helm-controller/api/v2beta1"
 	. "github.com/onsi/gomega"
+	"github.com/weaveworks/weave-gitops/core/server/types"
 	pb "github.com/weaveworks/weave-gitops/pkg/api/core"
 	"github.com/weaveworks/weave-gitops/pkg/kube"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/rand"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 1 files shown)


core/server/helm_release_test.go
```diff
@@ -82,6 +90,199 @@ func TestGetHelmRelease(t *testing.T) {
...
+
+	g.Expect(k.Status().Patch(ctx, &helmRelease, client.Apply, opt...)).To(Succeed())
+
+	// Create helm storage.
+	storage := types.HelmReleaseStorage{
+		Name:     "",
+		Manifest: helmReleaseInventoryObjects(),
+	}
+
+	storageData, _ := json.Marshal(storage)
```
Callisto13, line 231: missing err check?

core/server/helm_release_test.go
```diff
@@ -82,6 +90,199 @@ func TestGetHelmRelease(t *testing.T) {
...
+
+	var compressed bytes.Buffer
+
+	compressor := gzip.NewWriter(&compressed)
+	_, _ = compressor.Write(storageData)
+	compressor.Close()
```
Callisto13, line 237: any values we need to check here?

core/server/helm_release_test.go
yitsushi: not missing... ignored on purpose.

core/server/helm_release_test.go
yitsushi: Well, if we don't test compressed data, we can't be sure it's working.

Callisto13: no i mean, does `Write` return anything we might care about? like an error

yitsushi: it returns with the number of bytes written and an error if it failed, but it shouldn't fail, it's a write and we are not using filesystem io, buffer writing will not fail, if it does somehow, the test will fail anyway.

yitsushi: The reason I put the `_, _ =` there because linter cried.

jpellizzari (approved):
Nice work!

PR #1606, opened 2022-03-03
Add table filtering interaction

### LINKED ISSUE #1506: Apply filtering and search to pages with a data table
The `<FilterBar />` component requires a couple things from the parent component in order to function properly -

1. Filtering options need to be grabbed from data and passed down in the format of `{<columnName>: [...options], }` - could be the job of a utils function. Do we want every column in every table to have every possible option filterable?

2. Active filters state needs to be stored in parent and passed down, and also applied to the data in the `rows` prop of `<DataTable />`.

3. Component needs to be rendered above `<DataTable />` - which brings up the question of potentially removing the `children` prop of `<DataTable />` and just rendering `<Pagination />` under the table to keep things consistent?

We also need to add the check/uncheck all option to each filter heading - along with the css changes to match.

The `<SearchInput />` is similar - but is only meant to filter the `Name` column of the table it's applied to - `search` and `setSearch` have to be passed down from the parent, and `search` has to be applied to `rows` like in step 2.

jpellizzari:
Closes #1506 

Adds a slide-out animation and interaction. Only supports basic string filtering at the moment:
https://drive.google.com/file/d/1aXsjywUAD85wXSohtW_xIbiwazvQWuWK/view?usp=sharing


jpellizzari:
Follow ups:

- [ ] Add filtering/sorting by boolean and number values

Issue #1594, opened 2022-03-03
Smooth out development setup


luizbafilho:
Currently, development [setup instructions](https://github.com/weaveworks/weave-gitops/tree/v2#cliapi-development) don't work out of the box. 

We need to do a couple of manual interventions applying some manifests to get it to work.

Make sure the process is more automated and minimize the setup steps, including making use of the chart in the repo: https://github.com/weaveworks/weave-gitops/tree/v2/charts/weave-gitops and if doable make tilt to installed it instead of just applying a deployment manifest.

Callisto13:
I am already doing this as part of https://github.com/weaveworks/weave-gitops/issues/1454 so i can take this

PR #1608, opened 2022-03-04
fix: smooth out v2 dev process

Callisto13:
This is not a "finished product", I am simply doing the bare minimum to
unblock people and make life easier while I tidy some things up.

Here I am just adding some more manifests to the `Tiltfile`; I have
stashed these in `tools/dev-manifests` for now, but this may change.
Either way people will not notice as they are just running `make
cluster-dev`.

The manifests under `manifests/wego-dev` are technically out of use as
we move onto a Helm Chart install flow, but I am leaving them there for
now as I tinker, plus I know other people are working on the Helm bit so
I am staying out of their way.

I have also left the SuperUser and the "stripping out the service
account" bit for now, as they do not need to be in place to unblock
people. They are next on my list.

Closes https://github.com/weaveworks/weave-gitops/issues/1594

### DIFF
--- a/tools/dev-manifests/role.yaml
+++ b/tools/dev-manifests/role.yaml
@@ -1,0 +1,21 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: resources-reader
+  namespace: flux-system
+rules:
+  - apiGroups: [""]
+    resources: ["secrets"]
+    verbs: [ "get", "list" ]
+  - apiGroups: ["apps"]
+    resources: [ "deployments" ]
+    verbs: [ "list", "get" ]
+  - apiGroups: ["kustomize.toolkit.fluxcd.io"]
+    resources: [ "kustomizations" ]
+    verbs: [ "*" ]
+  - apiGroups: ["helm.toolkit.fluxcd.io"]
+    resources: [ "helmreleases" ]
+    verbs: [ "*" ]
+  - apiGroups: ["source.toolkit.fluxcd.io"]
+    resources: [ "buckets", "helmcharts", "gitrepositories", "helmrepositories" ]
+    verbs: [ "*" ]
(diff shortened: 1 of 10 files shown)


tools/dev-manifests/role.yaml
```diff
@@ -0,0 +1,21 @@
...
+rules:
+  - apiGroups: [""]
+    resources: ["secrets"]
+    verbs: [ "get", "list" ]
+  - apiGroups: ["apps"]
+    resources: [ "deployments" ]
+    verbs: [ "list", "get" ]
+  - apiGroups: ["kustomize.toolkit.fluxcd.io"]
+    resources: [ "kustomizations" ]
+    verbs: [ "*" ]
```
yitsushi, line 15: isn't `*` too broad? We have only read operations on our short term roadmap and removed everything that would write.

tools/dev-manifests/role.yaml
```diff
@@ -0,0 +1,21 @@
...
+  - apiGroups: ["helm.toolkit.fluxcd.io"]
+    resources: [ "helmreleases" ]
+    verbs: [ "*" ]
```
yitsushi, line 18: Same

tools/dev-manifests/role.yaml
```diff
@@ -0,0 +1,21 @@
...
+  - apiGroups: ["source.toolkit.fluxcd.io"]
+    resources: [ "buckets", "helmcharts", "gitrepositories", "helmrepositories" ]
+    verbs: [ "*" ]
```
yitsushi, line 21: Same

yitsushi (approved):
The `allow everything because why not, #yolo` is seems a bit too much, but I'm happy with that, we will update those over time.

Callisto13:
All the `"*"` bits are from existing manifests, for speed of unblocking people I did not spend time working out what they actually need. And to be clear: these are dev manifests which will never get to prod.

Issue #1580, opened 2022-03-03
Bug: Unable to view Flux Runtime information


sympatheticmoose:
<img width="1424" alt="Screenshot 2022-03-03 at 11 32 16" src="https://user-images.githubusercontent.com/19860021/156556892-f4473278-cb9d-423f-a53a-c7dc4fbaa1a0.png">

Steps to reproduce:

- kind create cluster
- flux bootstrap
- Installed via:
```
kubectl apply -n flux-system -f - << EOF
apiVersion: source.toolkit.fluxcd.io/v1beta1
kind: GitRepository
metadata:
  name: ww-gitops
  namespace: flux-system
spec:
  interval: 1m0s
  ref:
    branch: jlw-helm-patch
  secretRef:
    name: flux-system
  url: ssh://git@github.com/weaveworks/weave-gitops
  ignore: |
    # exclude all
    /*
    # include charts directory
    !/charts/
EOF
```
and

```
kubectl apply -n flux-system -f - << EOF
apiVersion: helm.toolkit.fluxcd.io/v2beta1
kind: HelmRelease
metadata:
  name: ww-gitops
  namespace: flux-system
spec:
  chart:
    spec:
      chart: ./charts/weave-gitops
      sourceRef:
        kind: GitRepository
        name: ww-gitops
  values:
    image: v2-preview
  interval: 1m0s
EOF
```

Callisto13:
For now, if you want to check out what the UI for that tab looks like and give feedback on that, you can add the following to the `rules` in `manifests/wego-app/role.yaml.tpl`:

```
  - apiGroups: ["apps"]
    resources: ["deployments"]
    verbs: ["get","list"]
```

If you are using the Helm Chart (not sure what state that is in or if anyone is using it), then you can find the equivalent Role in `charts/weave-gitops/templates/wego-app.yaml`.

sympatheticmoose:
Many thanks @Callisto13

Callisto13:
Hopefully this will do the trick https://github.com/weaveworks/weave-gitops/pull/1609

PR #1609, opened 2022-03-04
fix: Update sa role to include apps/deployments

Callisto13:
Closes https://github.com/weaveworks/weave-gitops/issues/1580

# I am about to do a lot more with that Role, but this should be enough to unblock people for now.


sympatheticmoose:
@Callisto13 appreciate this isn't strictly related to your change, but why do we need full permissions on the flux objects for a read-only UI? This seems at odds with what is described in the superuser docs https://docs.gitops.weave.works/docs/next/gitops-dashboard#login-via-an-oidc-provider cc @luizbafilho

JamWils (approved):
LGTM until the super user is in place.

Issue #1589, opened 2022-03-03
Bug: incorrect interval reported for GitRepository source objects


sympatheticmoose:
I believe all my sources have an interval of 1m, most are showing as 0h1m60s... what is that, 2min? The one that is correct is a HelmRepository
<img width="122" alt="Screenshot 2022-03-03 at 12 04 17" src="https://user-images.githubusercontent.com/19860021/156561487-5acd3b31-461e-4041-8229-92db9dedb8a8.png">

Additionally, designs show:
`1m 0s` rather than `XhXmXXs`

PR #1610, opened 2022-03-04
fix: incorrect interval reported on the Sources list page

yitsushi:
**UI**

* Utility function to convert `interval` to string, that way we have to
  change only in one place if we want to change the format.

**Back-end**

* Fix interval calculation on `HelmRepository` and `GitRepository`.
  The issues was: `.Minutes()` and `.Seconds()` return with the value of
  the whole interval, not just the minutes/seconds parts.

Closes: #1589

### DIFF
--- a/.proxyrc
+++ b/.proxyrc
@@ -3,6 +3,6 @@
     "target": "http://localhost:9000/",
   }, 
   "/oauth2": {
-    "target": "http://localhost:9000 /"
+    "target": "http://localhost:9000/"
   },
 }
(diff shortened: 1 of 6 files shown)


jpellizzari (commented):
@yitsushi Can you combine this logic with this component?: https://github.com/weaveworks/weave-gitops/blob/v2/ui/components/Interval.tsx

jpellizzari (approved):
Ignore my previous comment, you are already ahead of me on this! LGMT, nice work.

PR #1612, opened 2022-03-04
feat: Wire core server for Impersonation

Callisto13:
2 commits here:

337d703f7:

feat: Refactor chart service account

1. Strip out the service account (mostly) in the Helm Chart. To keep the
   app's scope limited we are making it "impersonate only". The caveat
   to this is that the Profiles functionality is always on, with queries
   not made by the user. Therefore in this case there is nothing to
   impersonate. So *for now*, the wego-app service account has an
   extra role which will let it query those resources.
2. Add an Admin user to the Chart. With the "impersonate only" plan, we
   require users to enable auth by default. I have therefore added a
   template to create an Admin user with the required permissions.
   The actual password is configurable via values and is left empty.
3. Update the Tiltfile to deploy this Chart locally. All current local
   dev instructions remain true. The dev password is `dev`.
4. Create a default const to hold the value of the default claims
   subject (admin).

----------------------

f56c707f49:

feat: Wire the core server to Impersonate queries

1. Move the `ImpersonatingConfigGetter` into the `kube` package. This
   felt like a "natural" place to me, feel free to suggest elsewhere if
   you disagree. It has to be moved _somewhere_ to avoid import cycles
   with t'other server.
2. Update the core server to use the default client getter rather than
   our placeholder. With the import cycle fixed this is now doable.
   This default ClientGetter takes a ConfigGetter, so now core will
   Impersonate.
3. Update the server wiring to receive a `CoreServerConfig`, following
   the same pattern as seen in the `ApplicationsServer` and the
   `ProfilesServer`. I have included a `Logger` on the server, which I
   intend to use in my next ticket, but feel free to complain that it
   can technically be dropped from this PR.


Callisto13:
@JamWils @sympatheticmoose could you 2 look at the Chart changes in commit 337d703f7?

yitsushi (approved):
I love it with all the readability changes :heart:

Callisto13:
No-one get keen and merge anything - looking into something curious in the superuser flow.

README.md
```diff
@@ -74,6 +74,8 @@ To set up a development environment for the CLI
 5. Run `./bin/gitops install --config-repo=<repo url>` (or just `flux install -n flux-system` if you don't care about doing the whole dance.)
 6. Start the in-cluster API replacement job (powered by [http://tilt.dev](tilt.dev)) with `make cluster-dev`
 7. make or make unit-tests to ensure everything built correctly.
+8. The UI will start immediately on port `9001`. Auth is now always on (I do not recommend
+  turning it off). The password is `dev`.
```
jpellizzari, line 78: :+1:

jpellizzari (approved):
Excellent work! I was originally wondering if there is any test coverage to be added here, but I don't see any obvious places to add any.

Callisto13:
> looking into something curious in the superuser flow

false alarm, all good

Issue #1592, opened 2022-03-03
Table column headers are inconsistently large for source detail view


sympatheticmoose:
Designs show 14px across the board. This is the only one which sticks out.

<img width="1401" alt="Screenshot 2022-03-03 at 12 20 23" src="https://user-images.githubusercontent.com/19860021/156563703-03eb8015-d6db-4485-a7a5-4073c42af2f7.png">

Issue #1590, opened 2022-03-03
As a user, I want to sort tables by any column to help me find the information I seek.


sympatheticmoose:
Missing and inconsistent across tables

On Applications > releases

- [x] Name
- [ ] Type
- [ ] Namespace
- [ ] Cluster
- [x] Status
- [ ] Revision
- [ ] Last updated / Last Synced At

On Applications > releases > object
- [ ] Name
- [x] Type
- [x] Namespace
- [ ] Cluster
- [x] Status
- [x] Message - Creates bug https://github.com/weaveworks/weave-gitops/issues/1591

On Sources 
- [x] Name
- [ ] Type
- [ ] Status
- [ ] Cluster
- [ ] URL
- [ ] Reference
- [ ] Interval

On Sources > object
- [x] Name
- [ ] Type
- [ ] Namespace
- [ ] Cluster
- [x] Status
- [ ] Revision
- [ ] Last updated / Last Synced At

Issue #1587, opened 2022-03-03
Last Synced At should be Last Updated


sympatheticmoose:
Based on discussion in notion around using `status.artifact.lastUpdateTime` as the value here. 

Is this the correct field? Might be a question for the flux team.

Additionally, need to ensure the column is single line rather than 3
<img width="85" alt="Screenshot 2022-03-03 at 11 57 29" src="https://user-images.githubusercontent.com/19860021/156560581-2c351c05-6b59-4906-a6ce-3d2133ae1e5c.png">

sympatheticmoose:
This additionally causes an annoying jump in where the table is when going between Applications and Sources
<img width="583" alt="Screenshot 2022-03-03 at 12 17 30" src="https://user-images.githubusercontent.com/19860021/156563206-ee3c4497-10f8-4d1b-83d2-7c70074bc1e8.png">
<img width="709" alt="Screenshot 2022-03-03 at 12 17 24" src="https://user-images.githubusercontent.com/19860021/156563209-cf984a54-8805-4fcd-956e-47c926c36429.png">

Issue #1579, opened 2022-03-03
Sort arrow is off center on details pages


sympatheticmoose:
From applications > sources (correct)
<img width="125" alt="Screenshot 2022-03-03 at 11 28 45" src="https://user-images.githubusercontent.com/19860021/156556062-66c61e1d-b6d3-4177-8e18-2df2d2544502.png">

From applications > sources > source (incorrect)
<img width="301" alt="Screenshot 2022-03-03 at 11 28 43" src="https://user-images.githubusercontent.com/19860021/156556065-8fcb82fa-f9ee-48dd-9668-3caa16b2c1ae.png">

joshri:
Hey @sympatheticmoose - I can't reproduce 

![image.png](https://images.zenhubusercontent.com/617c4a20ede496b2fae35baf/59c00f87-4f9c-48ee-b6df-9c959028699c)

@jpellizzari do you see this on your system?

jpellizzari:
No mine looks good. Chrome on linux. @joshri Check it out in Safari?

sympatheticmoose:
Mine was chrome on Mac

PR #1614, opened 2022-03-04
Add Numerous UI Fixes

joshri:
Closes: #1590 - sort all columns
Closes: #1587 - change 'last synced at' to 'last updated'
Closes: #1579 - off center sort arrow
Closes: #1592 - t-head font size


ui/components/AutomationsTable.tsx
```diff
@@ -69,7 +69,7 @@ function AutomationsTable({ className, automations }: Props) {
           label: "Revision",
           value: "lastAttemptedRevision",
         },
-        { label: "Last Synced At", value: "lastHandledReconciledAt" },
+        { label: "Last Updated", value: "lastHandledReconciledAt" },
```
jpellizzari, line 134: I don't think this is accurate. Can you check the flux documentation to see what this field actually means?

jpellizzari (approved):
LGTM, with the caveat about the `Synced At` column. I don't know if we can say a thing was "updated" from that field.

My understanding is that it is possible for something to be reconciled without being updated. In this case, I think "reconciled" could mean "we checked it and it there was nothing to do".

ui/components/AutomationsTable.tsx
joshri: @sympatheticmoose 👀

Issue #1584, opened 2022-03-03
Switch status icons to have indicative mark as well as colour to help make more accessible.


sympatheticmoose:
Currently just using red and green circles

Designs show:

Ready
<img width="397" alt="Screenshot 2022-03-03 at 11 47 40" src="https://user-images.githubusercontent.com/19860021/156559264-af2aa3f1-2d49-4b54-bd8d-4931a43976be.png">

Suspended
<img width="1030" alt="Screenshot 2022-03-03 at 11 47 46" src="https://user-images.githubusercontent.com/19860021/156559250-f3ced347-7643-4fb8-9136-7623c354cf69.png">

Error
<img width="200" alt="Screenshot 2022-03-03 at 11 48 25" src="https://user-images.githubusercontent.com/19860021/156559224-97613113-9428-456a-84db-7faf8987aec4.png">

Issue #1583, opened 2022-03-03
As a user I want to see if any of my resources are suspended


sympatheticmoose:
Suspended my `flux-system` kustomization, and it still reported as Green/Ready. Should indicate that it is suspended as shown below

<img width="1446" alt="Screenshot 2022-03-03 at 11 45 47" src="https://user-images.githubusercontent.com/19860021/156558619-754d0696-f054-4c05-88e9-5c7e28c5ac07.png">

PR #1622, opened 2022-03-04
Add Status Icons

joshri:
Closes: #1584

Adds correct status icons from Figma - we are currently missing implementation for the suspended icon which is in issue #1583 

image


ui/components/__tests__/__snapshots__/KubeStatusIndicator.test.tsx.snap
```diff
@@ -80,18 +78,11 @@ exports[`KubeStatusIndicator snapshots renders error 1`] = `
...
-      viewBox="0 0 100 100"
-    >
-      <circle
-        cx="50"
-        cy="50"
-        r="45"
-      />
-    </svg>
+    <img
+      src=""
```
jpellizzari, line 82: :thinking: Probably needs to be populated, unless we are using `background-image`?

ui/images/failed.svg
```diff
@@ -0,0 +1,3 @@
+<svg width="24" height="24" viewBox="0 0 24 24" fill="none" xmlns="http://www.w3.org/2000/svg">
```
jpellizzari, line 1: Let's format these. An HTML formatter should work.

jpellizzari (approved):
No blockers

ui/components/__tests__/__snapshots__/KubeStatusIndicator.test.tsx.snap
joshri: I wonder why nothing is there - maybe because it figures out IconType on load?

PR #1626, opened 2022-03-07
Upgrade logr

ozamosi:
The purpose of this PR is to upgrade logr from 0.4 -> 1. Alas, there's a couple of "unrelated" changes in separate commits that makes the final commit simpler.

First, reformat the go.mod (e.g. remove the different blocks and let `go mod tidy` re-sort the dependencies) so it contains one block of direct requirements and one block of indirect requirements, which helps future library upgrade PRs create smaller diffs.

Second, remove a duplicate logr fake, and while doing that also remove a duplicate http fake. Without this, not all tests would have been mocking the right logr after the upgrade.

Third, upgrade logr, and rewrite all the tests that test our logs. This triggered an upgrade of klog, which triggered an upgrade of many k8s things. Those upgrades aren't major versions, though.


dhwthompson (approved):
Looks reasonable to me. Thanks for the thorough commit messages; they were super helpful for figuring out what was going on. 👍

Issue #1600, opened 2022-03-03
Implement ClustersClientsMiddleware


luizbafilho:
Implement the middleware responsible for fetching the cluster and creating clients for each and injecting them in the context.

That work should be based of https://github.com/weaveworks/weave-gitops/pull/1556

Implementation Notes
- If no `clusters` configmap is present, we understand that as running in a non-multi-cluster setup, so just return that single cluster client in the clusters clients' lists

Callisto13:
Recapping what was discussed in our planning meeting:
- Multi-cluster is *not* a feature of core
- So core cannot have any way to add multiple clusters
- So we will be doing the multi bit in Enterprise
  - Including all the config map hack bits
- From core's side we need to:
  - Define the clusters fetcher interface
  - Fulfil that interface by adding a single, in-memory cluster to a pool which is added to the context (basically doing what we do now, but putting the client somewhere else.

More tickets will be coming up for the rest 👍

PR #1627, opened 2022-03-07
ClustersClientsPool Middleware

luizbafilho:
Closes: #1600

**What changed?**
- Implements the ClustersClientsPool Middleware  based on https://github.com/weaveworks/weave-gitops/pull/1556

**Why?**

**How did you test it?**
- Manually and Unit tests

**Release notes**

**Documentation Changes**


jpellizzari (commented):
Code looks good, but one more `cluster` field could really move the whole initiative along.

pkg/server/auth/server.go
```diff
@@ -323,6 +323,10 @@ func (s *AuthServer) UserInfo() http.HandlerFunc {
 			return
 		}
 
+		if s.provider == nil {
+			return
+		}
```
Callisto13, line 328: This fix + a whole load of auth tests is coming in https://github.com/weaveworks/weave-gitops/pull/1699, so you can drop this bit

core/multicluster/middleware.go
```diff
@@ -0,0 +1,50 @@
+package multicluster
```
Callisto13, line 1: is there any way that we can "hide" this a bit better?

given that core is technically not supposed to have a feature called multicluster, it may look sus when people read the code.

perhaps we can rename all this to `clustersmanager` or something?

core/multicluster/multicluster.go
```diff
@@ -0,0 +1,92 @@
...
+	"context"
+	"fmt"
+
+	"github.com/weaveworks/weave-gitops/pkg/kube"
+	"github.com/weaveworks/weave-gitops/pkg/server/auth"
+	"k8s.io/client-go/rest"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+//go:generate go run github.com/maxbrunsfeld/counterfeiter/v6 -generate
```
Callisto13, line 13: 🙌  yay love counterfeiter! so much better that the disaster that is gomock/mockgen

core/multicluster/middleware.go
luizbafilho: we'll not hide that enterprise has that feature, so that's not an issue, although having a `SingleClusterFetcher` inside `multicluster` package looks weird. 😄  I'll change it.

core/server/kustomization.go
```diff
@@ -2,34 +2,40 @@ package server
...
 	}
 
 	var results []*pb.Kustomization
 
-	for _, kustomization := range l.Items {
-		k, err := types.KustomizationToProto(&kustomization)
-		if err != nil {
-			return nil, fmt.Errorf("converting items: %w", err)
+	//TODO: handle failures and parallelize
+	for _, c := range clientsPool.Clients() {
```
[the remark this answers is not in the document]
luizbafilho: Given this work is focused on a single cluster, I'll leave this to later because adding these names is nontrivial since we'll have to change the response format and update the UI to comply with it.

core/clustersmngr/middleware_test.go
```diff
@@ -0,0 +1,69 @@
...
+
+	defaultHandler := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {})
+	middleware := func(next http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			clientsPool := clustersmngr.ClientsPoolFromCtx(r.Context())
+
+			clients := clientsPool.Clients()
+			if _, ok := clients[cluster.Name]; !ok {
+				g.Fail("leaf cluster client not present")
+			}
```
Callisto13, line 31: we can simplify to this if you like
```suggestion
			g.Expect(clientsPool.Clients()).To(HaveKey(cluster.Name))
```

core/clustersmngr/single_fetcher.go
```diff
@@ -0,0 +1,28 @@
...
+	"context"
+
+	"k8s.io/client-go/rest"
+)
+
+type singleClusterFetcher struct {
+	restConfig *rest.Config
+}
+
+func NewSingleClusterFetcher(config *rest.Config, _ string) (ClusterFetcher, error) {
```
Callisto13, line 13: Is that string var meant to be used for something?

pkg/server/handler.go
```diff
@@ -41,7 +44,18 @@ func NewHandlers(ctx context.Context, cfg *Config) (http.Handler, error) {
 	httpHandler := middleware.WithLogging(cfg.AppConfig.Logger, mux)
 	httpHandler = middleware.WithProviderToken(cfg.AppConfig.JwtClient, httpHandler, cfg.AppConfig.Logger)
 
+	restCfg, _, err := kube.RestConfig()
+	if err != nil {
+		return nil, fmt.Errorf("building rest config: %w", err)
+	}
```
Callisto13, line 50: I think this has been created (a couple of times) before this (IIRC it exists in several parts of the Config, and is created by the calling func in cmd/gitops-server/cmd/cmd.go), does it make sense reuse rather than redefine everywhere?

core/clustersmngr/middleware.go
```diff
@@ -0,0 +1,50 @@
...
+// WithClustersClients creates clusters client for provided user in the context
+func WithClustersClients(clustersFetcher ClusterFetcher, next http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		user := auth.Principal(r.Context())
+		if user == nil {
+			next.ServeHTTP(w, r)
+			return
+		}
+
+		clusters, err := clustersFetcher.Fetch(r.Context())
```
Callisto13, line 20: we are punting the caching of the available clusters to another PR right?

tho i suppose the implementor of the interface can cache on that side. actually that is the better idea, nvm!

Callisto13 (commented):
looks good, just a couple of idle thoughts

core/clustersmngr/single_fetcher.go
luizbafilho: specify the namespace, it was used for the configmap, and will for the multicluster to fetch the crds

core/clustersmngr/middleware.go
luizbafilho: yep, not handling any of that now.

pkg/server/handler.go
luizbafilho: good call, changing it. we have it available there we just need to use it

luizbafilho: done

core/server/suite_test.go
```diff
@@ -74,3 +76,23 @@ func makeGRPCServer(cfg *rest.Config, t *testing.T) pb.CoreClient {
 
 	return pb.NewCoreClient(conn)
 }
+
+func withClientsPoolInterceptor(config *rest.Config, user *auth.UserPrincipal) grpc.ServerOption {
+	return grpc.UnaryInterceptor(func(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (interface{}, error) {
```
jpellizzari, line 81: Are we sure this runs? I have run into issues before where the middlewares don't run with our current implementation: https://github.com/grpc-ecosystem/grpc-gateway/issues/1043

jpellizzari (approved):

core/server/suite_test.go
luizbafilho: yes, if we comment that out the test breaks because the clients pool doesnt get injected

core/clustersmngr/single_fetcher.go
luizbafilho: done

core/server/suite_test.go
jpellizzari: What about when the server is actually running?

luizbafilho: I don't know if you pay attention to that's for test only since we use HTTP middleware for it.

Callisto13 (approved):
:shipit:

PR #1628, opened 2022-03-07
ui: Port graph + tests away from using Applications types

ozamosi:
We've moved the endpoints required to render the graph to the core
proto, however two things hadn't been moved:

Graph was still using some type definitions from applications. This is
an easy fix of just changing the import.

The graph tests were using the plain Application mock, because that's
the only client mock we have. This copy-pastes the mock setup code so
that we can mock Core or Application.

Doing this will allow me to delete both the endpoints and their                 
types from the Applications proto.

This commit is intended to establish a pattern, not finish the work -
    subsequent endpoints will be moved in the same commit that deletes the
    endpoint in question.

### DIFF
--- a/ui/lib/__tests__/graph.test.ts
+++ b/ui/lib/__tests__/graph.test.ts
@@ -1,7 +1,6 @@
 import _ from "lodash";
-import { Core } from "../api/core/core.pb";
 import { getChildren } from "../graph";
-import { createMockClient } from "../test-utils";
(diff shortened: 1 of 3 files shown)


jpellizzari (approved):
LGTM. I wish we had the same `counterfeit` library for TS stuff that we do for `go`. Maybe I will look into that.

joshri (approved):

PR #1630, opened 2022-03-07
Add Suspended Status

joshri:
Closes: #1583

Adds `suspend` to `Conditions` type and applies it in `KubeStatusIndicator`


api/core/types.proto
```diff
@@ -28,6 +28,7 @@ message Condition {
     string reason = 3;
     string message = 4;
     string timestamp = 5;
+    bool suspend = 6;
```
jpellizzari, line 31: I wouldn't expect this to get treated like a condition. I would instead add it as a bool directly on the `GitRepository` type.

jpellizzari (changes requested):

jpellizzari (approved):
:fire:  :fire: BE changes + Unit test coverage :fire: :fire:  Nicely done.

PR #1631, opened 2022-03-07
Unset page width and affix nav bar to left

### LINKED ISSUE #1527: Fix double scroll bars
There are two, and there should be exactly 1.

### LINKED ISSUE #1577: Small table size makes poor use of screen real estate
Screenshot 2022-03-03 at 11 23 38

Is there a reason we want to limit the main section to `max-width: 1400px`? 

At default zoom this will inevitably mean users have to horizontally scroll the table which we should be trying to avoid.

cc @bia

jpellizzari:
Closes #1577 
Closes #1527 

Sets the page-width free for larger monitors and fixes the underlying table-cell overflow problem with `KubeStatusIndicator`:
Screenshot from 2022-03-07 15-25-30

### DIFF
--- a/ui/components/DataTable.tsx
+++ b/ui/components/DataTable.tsx
@@ -168,10 +168,7 @@ function UnstyledDataTable({
           <TableHead>
             <TableRow>
               {_.map(fields, (f) => (
-                <TableCell
-                  style={f.width && { maxWidth: f.width }}
-                  key={f.label}
-                >
+                <TableCell style={f.width && { width: f.width }} key={f.label}>
                   <SortableLabel field={f} />
                 </TableCell>
               ))}
@@ -221,6 +218,8 @@ export const DataTable = styled(UnstyledDataTable)`
   }
 
   td {
+    word-break: break-all;
+    word-wrap: break-word;
     white-space: nowrap;
     overflow: hidden;
     text-overflow: ellipsis;
--- a/ui/components/Layout.tsx
+++ b/ui/components/Layout.tsx
@@ -122,7 +122,6 @@ const NavContent = styled.div`
 
 const ContentContainer = styled.div`
   width: 100%;
-  /* min-height: 100vh; */
   padding-top: ${(props) => props.theme.spacing.medium};
   padding-bottom: ${(props) => props.theme.spacing.medium};
   padding-right: ${(props) => props.theme.spacing.medium};
--- a/ui/components/Layout.tsx
+++ b/ui/components/Layout.tsx
@@ -58,22 +58,33 @@ const StyleLinkTab = styled(LinkTab)`
 
 const AppContainer = styled.div`
   width: 100%;
-  overflow-x: hidden;
+  overflow: hidden;
   height: 100%;
   margin: 0 auto;
   padding: 0;
 `;
 
 const NavContainer = styled.div`
+  position: relative;
   width: 240px;
+  height: 0;
   min-height: 100%;
   margin-top: ${(props) => props.theme.spacing.medium};
+  margin-right: ${(props) => props.theme.spacing.medium};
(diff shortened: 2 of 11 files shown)


ui/components/Layout.tsx
```diff
@@ -111,7 +122,7 @@ const NavContent = styled.div`
 
 const ContentContainer = styled.div`
   width: 100%;
-  min-height: 100vh;
+  /* min-height: 100vh; */
```
yitsushi, line 125: Can we just remove this instead?

ui/components/Page.tsx
```diff
@@ -20,15 +20,13 @@ export type PageProps = {
...
-  min-height: 85vh;
-  max-width: 1400px;
   margin: 0 auto;
   width: 100%;
   min-width: 1260px;
   box-sizing: border-box;
   background-color: rgba(255, 255, 255, 0.75);
   padding-left: ${(props) => props.theme.spacing.large};
-  padding-right: ${(props) => props.theme.spacing.large};
+
```
yitsushi, line 29: Empty line here looks strange. Can we remove it, to move before `padding-left` to at least the padding rules are grouped?

ui/components/Layout.tsx
```diff
@@ -151,25 +162,27 @@ function Layout({ className, children }: Props) {
...
-                    value={n.value}
-                    className={n.sub && "sub-item"}
-                    href={n.href}
-                    newTab={n.newTab}
-                  />
-                ))}
-              </Tabs>
-            </NavContent>
+            <FixedNav>
+              <NavContent>
```
yitsushi, line 166: (for learning) why can't we merge these together? `FixedNav` is a simple `div` with added style. Can it be a added to `NavContent` and reduce the number of HTML nodes we generate?

jpellizzari: Because FixedNav is set to “position: fixed”, it will not take up any space. The NavContent div “pushes” the main content section right, giving us the layout we want.

ui/components/DataTable.tsx
```diff
@@ -168,10 +168,7 @@ function UnstyledDataTable({
           <TableHead>
             <TableRow>
               {_.map(fields, (f) => (
-                <TableCell
-                  style={f.width && { maxWidth: f.width }}
-                  key={f.label}
-                >
+                <TableCell style={f.width && { width: f.width }} key={f.label}>
```
joshri, line 171: I remember I had the widths as percentages originally - now that there's no max width on the table is that worth reconsidering?

joshri (approved):
🌞

ui/components/Layout.tsx
yitsushi: I understand that part of the CSS, but can't be NavContent `fixed` instead?

```
<FixedNav div>
  <NavContent div with pad>
    ...
  </div>
</div>
```
and just add this:
```
  background-color: white;
  position: fixed;
  height: 100%;
  top: 80;
  width: 240px;
  left: 0;
```

to `const NavContent = styled.div`. As I see that div has only 2 padding rule and bunch of definitions for child nodes.

It does not take up extra space on the UI, but in the DOM tree it does.

ui/components/DataTable.tsx
jpellizzari: Yeah I think percentages would be easier to think about. I will change, although with `table-layout: fixed`, I have no idea what is going to happen.

ui/components/Layout.tsx
jpellizzari: `position: fixed` means that it won't take up space in the DOM. You will end up with `NavContent` on top of the main body content. It will look like this:
Screenshot from 2022-03-08 07-31-44

yitsushi: Ah, thanks.

PR #1632, opened 2022-03-08
Remove more endpoints

ozamosi:
This removes another 5 deprecated endpoints.

These were all effectively unused - as far as frontend goes, I only needed to remove a couple of unused hooks and change an import path.


jpellizzari:
I think Enterprise will still rely on core to provide GitProvider auth for doing cluster-related git operations. I’m not entirely sure if this PR affects that, but just wanted to add that note.

ozamosi:
You know what, I'll remove the two auth/token related endpoints from this PR, and I'll have a good answer to the enterprise question next time I try to remove those ones.

dhwthompson (approved):
Looks good to me from the back-end side of things. I don’t profess to understand much about how the front-end works, but it looks like a relatively straightforward [blood diff](https://twitter.com/MissAmyTobey/status/1486023849720172547), so as long as @jpellizzari is happy, I’m happy.

jpellizzari (approved):

Issue #1621, opened 2022-03-04
Wrap error messages in a border


jpellizzari:
https://www.figma.com/file/IVHnM9iyeFWpd11evtY8ux/Weave-GitOps?node-id=6031%3A25278

Drawing name: `Source Detail Page - Helm Repository - long error message`

PR #1635, opened 2022-03-08
Add Border to Source Detail Error Message

joshri:
Closes: #1621

Changed structure of the header area of `SourceDetail` to clean things up and get a spicy border around that error message

image

### DIFF
--- a/ui/components/PageStatus.tsx
+++ b/ui/components/PageStatus.tsx
@@ -9,11 +9,13 @@ type StatusProps = {
   ok: boolean;
   msg: string;
   error: boolean;
+  className?: string;
 };
 
-function PageStatus({ ok, msg, error }: StatusProps) {
+function PageStatus({ ok, msg, error, className }: StatusProps) {
+  console.log(error);
   return (
-    <div className={`page-status${error ? " error-border" : ""}`}>
+    <div className={`${className}${!ok || error ? " error-border" : ""}`}>
       <Flex align>
         <Icon
           type={ok ? IconType.CheckCircleIcon : IconType.FailedIcon}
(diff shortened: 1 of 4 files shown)


ui/components/SourceDetail.tsx
```diff
@@ -23,6 +23,24 @@ type Props = {
   info: <T>(s: T) => InfoField[];
 };
 
+type StatusProps = {
+  status?: string;
+  msg: string;
+};
+const PageStatus = ({ status, msg }: StatusProps) => {
```
jpellizzari, line 30: This needs to be repeated on automations as well right?

joshri: Yeah it needs to be done on everything that doesn't have source table - I made a new issue for it but I should just separate this out here.

ui/components/PageStatus.tsx
```diff
@@ -0,0 +1,40 @@
...
+import Text from "./Text";
+
+type StatusProps = {
+  ok: boolean;
+  msg: string;
+  error: boolean;
+  className?: string;
+};
+
+function PageStatus({ ok, msg, error, className }: StatusProps) {
```
jpellizzari, line 15: What if this accepted a `Conditions[]` type?

jpellizzari (approved):
LGTM, only nits

ui/components/PageStatus.tsx
```diff
@@ -0,0 +1,40 @@
...
+  console.log(error);
```
jpellizzari, line 16: Extra `console.log` here.

jpellizzari (approved):

Issue #1591, opened 2022-03-03
Bug: attempting to sort via message column in release details view causes error


sympatheticmoose:
<img width="725" alt="Screenshot 2022-03-03 at 12 11 50" src="https://user-images.githubusercontent.com/19860021/156562832-a9f8282f-b31f-45a1-a29b-b2fa9733ce49.png">

jpellizzari:
From standup: check the `short` flag on `KubeStateIndicator`

PR #1637, opened 2022-03-08
Fix Crash on Sort By Message

joshri:
Closes: #1591 

Adds the `Available` type back into the computeMessage function!

image


sympatheticmoose:
Hopefully this PR resolves, but FYI running latest v2 this morning with `make cluster-dev` navigating to Flux Runtime view results in the same error message, after this it persists across ALL tables and you have to clear your cache before it will work again
Screenshot 2022-03-09 at 12 38 31

joshri:
Duplicate of #1645 - closing

PR #1643, opened 2022-03-09
feat: List namespaces endpoint

### LINKED ISSUE #1615: Create endpoint for listing available namespace

yitsushi:
Tried to return as many information as I we can list labels, annotations,
and status.

On the request, it does not accept any filtering options. If we want to
pass in filters, we can add when we need filtering on server side. Right
now list endpoints are returning everything without filters and the UI
does the filtering, so following this patter, no filtering options on
the request.

Closes #1615


core/server/namespace.go
```diff
@@ -37,3 +38,26 @@ func (as *coreServer) GetFluxNamespace(ctx context.Context, msg *pb.GetFluxNames
 
 	return &pb.GetFluxNamespaceResponse{Name: nsList.Items[0].Name}, nil
 }
+
+func (as *coreServer) ListNamespaces(ctx context.Context, msg *pb.ListNamespacesRequest) (*pb.ListNamespacesResponse, error) {
+	k8s, err := as.k8s.Client(ctx)
+	if err != nil {
+		return nil, doClientError(err)
```
bigkevmcd, line 45: Small tip, `doClientError` 

```go
return status.Errorf(codes.Internal, "unable to make k8s rest client: %s", err.Error())
```
From https://pkg.go.dev/fmt
>If the format (which is implicitly %v for Println etc.) is valid for a string (%s %q %v %x %X), the following two rules apply:
>
>4. If an operand implements the error interface, the Error method will be invoked to convert the object to a string, which will then be formatted as required by the verb (if any).

core/server/namespace.go
```diff
@@ -37,3 +38,26 @@ func (as *coreServer) GetFluxNamespace(ctx context.Context, msg *pb.GetFluxNames
...
+	}
+
+	nsList := corev1.NamespaceList{}
+
+	if err = k8s.List(ctx, &nsList); err != nil {
```
bigkevmcd, line 50: Is this paginating the request?

core/server/namespace_test.go
```diff
@@ -59,3 +59,40 @@ func TestGetFluxNamespace_notFound(t *testing.T) {
 	_, err = coreClient.GetFluxNamespace(ctx, &pb.GetFluxNamespaceRequest{})
 	g.Expect(err).To(HaveOccurred())
 }
+
+func TestListNamespaces(t *testing.T) {
+	g := NewGomegaWithT(t)
+
+	ctx := context.Background()
+
+	coreClient, cleanup := makeGRPCServer(k8sEnv.Rest, t)
```
bigkevmcd, line 68: You can simplify this by adding the cleanup with [t.Cleanup(func(){...})](https://pkg.go.dev/testing#T.Cleanup) in `makeGRPCServer`?

core/server/namespace.go
yitsushi: No it does not.

core/server/namespace_test.go
yitsushi: I don't think it will work as we want. Last time I had very hard time with namespaces in tests. A `Delete` call did not delete the resources, or did something different because the next test case could query the "deleted" namespace. At the end, there I used `Update` to remove a label from a namespace. It can be useful to improve, but not here.

bigkevmcd: Can you point to this code?

This doesn't sound quite right, it can take a while to delete a namespace.

But, I don't see how this is relevant to the issue of simplifying the API?

core/server/namespace.go
yitsushi: I have no idea what to do with this comment. I had no intention to change what we are using, if it's not a refactor PR. I like not to combine random things in a single PR if they are not strictly related, that makes the process to revert changes much easier.

As a totally personal opinion, even if I wrote that part, I would still use the `Error()` function on it, because we don't know how long this code will live there, it will work no matter what happens and explicitly called a function that returns with a string. In that case, we have a nice and unified way to represent errors. No one can guarantee the the order of the checks will not change, or no one will implement (don't ask why, ppl can be weird) Formatter on an error type, in which case fmt will use that instead of the Error function.

core/server/namespace_test.go
yitsushi: https://github.com/weaveworks/weave-gitops/pull/1515/files#diff-f66d7854f580d2c66e5671211c392900713f66b468390f99207eb5395bb0daf3R35-R40

yitsushi: Oh and now I see, I though you commented on a different line.

yitsushi: https://github.com/weaveworks/weave-gitops/issues/1649

jpellizzari (commented):
:question: @yitsushi Does this return all namespaces, or only the namespaces the user has access to? I guess it depends on which objects they want to access in the namespace.

I think we want the latter. @luizbafilho Can you point us in the direction of the research you did around `kubectl can-i-use` and how that might relate here?

luizbafilho:
What it will return depends entirely on how the RBAC was set up. Usually, RBAC is set up to allow whole resource type, so usually, users would have access to all namespaces. 

But allowing individual ones are possible too by using `resourceNames` more [here](https://kubernetes.io/docs/reference/access-authn-authz/rbac/#referring-to-resources)

We won't have control over that, it will be the Operators' responsibility to set up the RBAC the way they think is the best.

jpellizzari (approved):
LGTM. I am thinking a little harder about how we might use this, given that we might want to query across namespaces, but I think it is ok to have for now.

luizbafilho (approved):

Issue #1642, opened 2022-03-09
Bug viewing Flux Runtime tab


Callisto13:
**Describe the bug**
I see this when I click on the `Flux Runtime` tab:

```
Error
Something went wrong.
Cannot read properties of undefined (reading 'message')
TypeError: Cannot read properties of undefined (reading 'message')
    at ex (http://localhost:9001/index.06e12bcf.js:8:467701)
    at tx (http://localhost:9001/index.06e12bcf.js:8:467853)
    at va (http://localhost:9001/index.06e12bcf.js:1:69959)
    at ru (http://localhost:9001/index.06e12bcf.js:1:122030)
    at Du (http://localhost:9001/index.06e12bcf.js:1:109303)
    at zu (http://localhost:9001/index.06e12bcf.js:1:109231)
    at Lu (http://localhost:9001/index.06e12bcf.js:1:109094)
    at Ru (http://localhost:9001/index.06e12bcf.js:1:106080)
    at http://localhost:9001/index.06e12bcf.js:1:55878
    at b (http://localhost:9001/index.06e12bcf.js:1:133082)
```

I didn't see it yesterday, so I think it got merged in the last 12-16 hours. When I am done with my current thing I will do a quick bisect and see what's up.

PR #1645, opened 2022-03-09
fix: Check for Available condition's message

Callisto13:
## I just had some spare time and thought I would play around, please close if it is garbage 😄 

Closes https://github.com/weaveworks/weave-gitops/issues/1642

The `computeMessage` was returning nothing for Flux runtime objects
since those have `Available` conditions, not `Ready` ones. This PR
updates that.

**How did you test it?**
Manually clocking the UI.
I also added a UI test, but I never got a clean run of those here so not sure if it actually works

**Random thought on UX:**

The `KubeStatusIndicator` takes a `short` bool, but would it make sense to have the `short` be set internally based on the value of `ready`? If the value is `False` (ie the pod is failing) then the message computed for that row will be shown in long-form (per [this](https://github.com/weaveworks/weave-gitops/blob/v2/ui/components/KubeStatusIndicator.tsx#L39)), with all the still `Ready` lines keeping it nice and short. Do we care about the message at all if the pod is running fine?

### DIFF
--- a/ui/components/KubeStatusIndicator.tsx
+++ b/ui/components/KubeStatusIndicator.tsx
@@ -18,11 +18,14 @@ export function computeReady(conditions: Condition[]): boolean {
     // Deployment conditions work slightly differently;
     // they show "Available" instead of 'Ready'
     _.find(conditions, { type: "Available" });
+
   return ready?.status == "True";
 }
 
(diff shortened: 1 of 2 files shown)


jpellizzari (approved):
Looks like we have the makings of a FE specialist on our hands. Come on in, the water is...painful.

PR #1654, opened 2022-03-09
Remove the Application CRD

ozamosi:
This removes the Application CRD (the v1alpha1 one, that is) completely. That is to say, I've left 3 variables in the api/v1alpha1 directory as
    they're used in various places (and, frankly, merge conflicts with
    main are gonna be painful enough as it is without renames all over),
    but remove the CRD itself, remove the code that generates the CRD,
    remove the application model, and remove bits & pieces that still
    referenced those bits.

I also snuck in an extra commit with a few extra deletions that should have been included in the SyncApplication endpoint deletion.


go.mod
```diff
@@ -4,7 +4,6 @@ go 1.17
 
 require (
 	github.com/Masterminds/semver/v3 v3.1.1
-	github.com/benbjohnson/clock v1.3.0
```
dhwthompson, line 7: 💥 Nice. Always satisfying to see a dependency relegated.

dhwthompson (approved):

PR #1660, opened 2022-03-10
Adds a global `--insecure-skip-tls-verify` flag to the cli

foot:
**What changed?**

Adds a global `--insecure-skip-tls-verify` flag to the cli to make requests to an api-server using self-signed TLS certs

**Why?**

API server serves on _only_ https by default now. This allows you to `gitops get clusters --insecure-skip-tls-verify` etc.

**How did you test it?**

Manual so far

### DIFF
--- a/cmd/gitops/root/cmd.go
+++ b/cmd/gitops/root/cmd.go
@@ -1,6 +1,7 @@
 package root
 
 import (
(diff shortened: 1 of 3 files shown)


yiannistri (approved):
Nice 🥇

PR #1663, opened 2022-03-10
refactor: Simplify makeGRPCServer API in tests

### LINKED ISSUE #1649: Simplify makeGRPCServer
You can simplify this by adding the cleanup with [t.Cleanup(func(){...})](https://pkg.go.dev/testing#T.Cleanup) in `makeGRPCServer`?

_Originally posted by @bigkevmcd in https://github.com/weaveworks/weave-gitops/pull/1643#discussion_r822505045_

yitsushi:
We use the `cleanup` function always in `defer`, which can be simplified
and let `makeGRPCServer` handle that part as it can already access `testing.T`.

Closes #1649

### DIFF
--- a/core/server/fluxruntime_test.go
+++ b/core/server/fluxruntime_test.go
@@ -20,8 +20,7 @@ func TestGetReconciledObjects(t *testing.T) {
 
 	ctx := context.Background()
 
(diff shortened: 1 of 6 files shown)


Callisto13 (approved):
nice

PR #1668, opened 2022-03-10
Remove bits and pieces

ozamosi:
This is a grab bag of stuff I found that had been left behind.

This removes the old embedded manifests, as we've got a brand new helm chart to use instead.

This removes several unused bits from our kube client.

This removes some generated protobuf files that got left behind.


SamLR (approved):

ozamosi:
@jpellizzari I noticed you've changed the tpl manifests I was trying to delete recently. Are you still using these for development? If so, can I help make the helm chart more appropriate for local development?

jpellizzari:
> @jpellizzari I noticed you've changed the tpl manifests I was trying to delete recently. Are you still using these for development? If so, can I help make the helm chart more appropriate for local development?

No we are not using these anymore. We are using a helm chart for development.

PR #1679, opened 2022-03-10
Added mkcert installation to support both linux on macOS

josecordaz:
Closes: 

**What changed?**

**Why?**

**How did you test it?**

**Release notes**

**Documentation Changes**


jpellizzari (commented):
Nice! I'll let other folks give the :+1: , added @SamLR as a reviewer since they have been working on Makefile things.

tools/download-deps.sh
```diff
@@ -131,3 +131,10 @@ done
 
 echo "Installing golangci-lint"
 curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b "$(go env GOPATH)"/bin v1.44.0
+
+echo "Installing mkcert"
+git clone https://github.com/FiloSottile/mkcert && (cd mkcert
+go build -ldflags "-X main.Version=$(git describe --tags)"
+cp mkcert "$(go env GOPATH)"/bin)
+rm -rf mkcert
+"$(go env GOPATH)"/bin/mkcert -install
```
yitsushi, line 140: We can use simply `go install`, it will do exactly the same:
```
go install filippo.io/mkcert@v1.4.3
```

yitsushi (changes requested):

tools/download-deps.sh
yitsushi: Why `filippo.io/mkcert` and not `github.com/FiloSottile/mkcert`?

The module has that path in [the `go.mod` file](https://github.com/FiloSottile/mkcert/blob/master/go.mod#L1)

```
❯ go install github.com/FiloSottile/mkcert@v1.4.3
go install: github.com/FiloSottile/mkcert@v1.4.3: github.com/FiloSottile/mkcert@v1.4.3: parsing go.mod:
        module declares its path as: filippo.io/mkcert
                but was required as: github.com/FiloSottile/mkcert
```

yitsushi: We can use `@latest` if we want, I just pinned a version for no reason, it can be `go install filippo.io/mkcert@latest`

josecordaz: Nice finding! 👏🏽👏🏽👏🏽

ozamosi: The problem with this solution is that go install updates your `go.mod`/`go.sum` with a dependency that's not part of our source tree, which you then need to not commit or the static check will be cross with you.

If you put `filippo.io/mkcert` in `tools/tools.go` instead and run `go mod tidy`, then the `go.mod` will manage the version we use for us, and you can just use `go install filippo.io/mkcert` to install that specific version.

josecordaz: @ozamosi Thanks for your comment. 

But it seems that is not the case anymore:
```
Since Go 1.16, go install can install a command at a version specified on the command line while ignoring the go.mod file in the current directory (if one exists). go install should now be used to install commands in most cases.
```
[Source](https://go.dev/doc/go-get-install-deprecation)

I can confirm this because I don't see any mkcert entry in the go.mod or go.sum files.

Please let me know If I'm missing something

ozamosi: Oooh - nevermind me, then.

yitsushi (approved):

Issue #1640, opened 2022-03-09
Spike: What is the best way to query events for a given object? 


jpellizzari:
Right now we loop through all the events and find the relevant ones. Is there a better way?

Labels? Annotations?

PR #1694, opened 2022-03-11
refactor: Query only object related events

yitsushi:
For now, it seems it's enough to query events based on the requested
involved object info. Later we might need to add extra magic if we ever
want to return other events like events from "related objects", but I
highly recommend to do not walk on that path and let the client do extra
queries instead.

Additional fixes:
- Rename `events._test.go` to `events_test.go`

Related to #1640

Removed: Already done in #1693
  - Fix test, somehow I could merge #1663 without rebasing from v2


jpellizzari (approved):
LGTM :+1:

PR #1695, opened 2022-03-11
Put explicit snyk scan workflow back

ozamosi:
This replaces the webhook, because the webhook does the
    wrong thing:
    
 * You can't get the code scan to run using the webhook.
 * You also can't get snyk to look up go dependencies properly using the
       webhook (it uses go mod graph, which lists lots of packages that
       aren't part of your build, whereas the CLI uses go list, which only
       lists ones that are)
    
Thus, run all the checks manually.
    
Rather than putting them in a separate scan workflow, just put it into
    an appropriate part of the overall PR workflow - that is:
 * The static check for all checks that don't require a built docker image
 * A bit of extra logic dunked at the end of the docker build to scan
       the built image
 * A new, post-merge job that uploads all the things.
    
Unfortunately, you can't (as far as I can tell) make github actions
    depend on a status, only jobs, so while these jobs have to run, their
    results aren't actually mandatory.

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -97,6 +97,19 @@ jobs:
       - run: make fakes
       - name: Check that make fakes has been run
         run: git diff --no-ext-diff --exit-code
+      - uses: snyk/actions/setup@master
+      - name: Look for insecure dependencies or bad licenses
+        uses: snyk/actions/golang@master
+        env:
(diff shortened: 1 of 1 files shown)


mnowster (approved):
Thank you for the helpful description on names cos the prodcut names are a bit confusing.

mnowster:
reviewed and discussed 👍🏻

PR #1696, opened 2022-03-11
Improve ci

SamLR:
Build the docker images in parallel with running the go unit tests and then upload the images to Google Artifact Registry for staging.

This is builds upon & replaces PR #1633 

```mermaid
flowchart TD;
  build-gitops-image[build gitops image]-->gitops-image[/gitops image/]:::asset
  gitops-image-->upload-images;
  build-gitops-server-image[build gitops-server image]-->gitops-server-image[/gitops-server image/]:::asset
  gitops-server-image-->upload-images;
  go-tests[run go unit tests]-.->upload-images;
  ui-tests[run ui tests]-.->upload-images;
  lint[run linters etc.]-.->upload-images;
  upload-images-->finish-ci;

  classDef asset fill:#9f9,stroke:#363;
```

### DIFF
--- a/.github/workflows/v2.yml
+++ b/.github/workflows/v2.yml
@@ -9,6 +9,10 @@ on:
       - v2
   workflow_dispatch:
 
+env:
+  CI_CONTAINER_REGISTRY: europe-west1-docker.pkg.dev
+  CI_CONTAINER_REPOSITORY: europe-west1-docker.pkg.dev/weave-gitops-clusters/weave-gitops
+
 name: V2 CI Workflow
 jobs:
   ci-js:
@@ -70,7 +74,6 @@ jobs:
       - run: make unit-tests
       # - run: make lib-test
 
-
   ci-static:
     name: V2 CI Check Static Checks
     runs-on: ubuntu-latest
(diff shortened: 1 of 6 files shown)


SamLR:
Successful run here: https://github.com/weaveworks/weave-gitops/actions/runs/1968917828

.github/workflows/v2.yml
```diff
@@ -96,26 +95,75 @@ jobs:
       - name: Check that fakes are updated
         run: git diff --no-ext-diff --exit-code
 
-  full-ci:
-    name: V2 CI Check (16.13.2, 1.17.5)
```
ozamosi, line 98: This name is actually magic - it needs to match https://github.com/weaveworks/corp/blob/master/github-repo-weave-gitops.tf#L50 or you can't merge.

SamLR: good catch

.github/workflows/v2.yml
```diff
@@ -93,10 +95,77 @@ jobs:
...
+            FLUX_VERSION=${{ env.FLUX_VERSION }}
+            LDFLAGS='${{ env.LDFLAGS }}'
+            GIT_COMMIT=${{ github.sha }}
+      - name: Cache docker image
+        uses: actions/upload-artifact@v2
+        with:
+          name: ${{ matrix.docker-image }}
+          path: /tmp/${{ matrix.docker-image }}.tar
+
+  ci-upload-images:
```
jpellizzari, line 131: Can we have this run *only* on push to the `v2` branch. I don't think we need/want to upload containers on every pull request?

jpellizzari (commented):
The jobs themselves LGTM; my only concern is *when* these images get pushed.

.github/workflows/v2.yml
SamLR: ah good point, I'll add a condition for the push step

jpellizzari (approved):
LGTM. Nice work.

SamLR:
@jpellizzari I think I've got it so that it'll only push on merge but I've had to create a secondary 'all done' check so not sure if the checks will fully work or not so might have to revert that change

jpellizzari (commented):
@SamLR One more thought: do we need to upload a Docker image for the CLI?

SamLR:
@jpellizzari I don't think we need to although tools in docker images seem to becoming more common place and it may make a nice way to get hold of it for testing.

Issue #1652, opened 2022-03-09
Add Helm Chart Detail Page


joshri:
This screenshot is from the Helm RELEASE view design - there may be changes required. 

![image.png](https://images.zenhubusercontent.com/617c4a20ede496b2fae35baf/69709c83-d2b0-4d22-870b-21d838521492)

Issue #1617, opened 2022-03-04
Show HelmCharts as sources


jpellizzari:
`HelmReleases` create `HelmChart` objects that flux represents as sources. We would like to show those as first-class objects.

The API endpoint already exists for listing charts, just needs wiring up in the UI.

I can't find any designs for this currently, but re-use of the `SourceDetail` component should be fine.

PR #1700, opened 2022-03-11
Add Helm Chart Detail Page

joshri:
Closes: #1617 
Closes: #1652 

Adds Helm Charts to Sources and adds Helm Chart Detail page

image


ui/pages/v2/HelmReleaseDetail.tsx
```diff
@@ -57,7 +51,10 @@ function HelmReleaseDetail({ className, name }: Props) {
             ]}
           />
         </Info>
-        <PageStatus ok={ok} msg={msg} error={error ? true : false} />
+        <PageStatus
+          conditions={helmRelease?.conditions}
+          error={error ? true : false}
```
jpellizzari, line 56: This will convert the value to a bool:

```suggestion
          error={!!error}
```

I would question the need for this, since `PageStatus` can compute it's own error state from `conditions` right?

ui/pages/v2/HelmChartDetail.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+export default styled(HelmChartDetail).attrs({
+  className: HelmChartDetail.name,
+})`
+  ${TitleBar} {
+    margin-bottom: 0;
+  }
+
+  ${Content} {
+    padding-top: 0;
```
jpellizzari, line 41: :thinking: We don't do this on any other detail pages. What makes `HelmCharts` special?

joshri: I initially copied the page from GitRepositoryDetail - will check it out and remove if unnecessary

ui/pages/v2/HelmReleaseDetail.tsx
joshri: The error prop is actually from whatever the request is on the page (useListSources, useGetHelmRelease) and not about the source itself - should those be separated?

jpellizzari: This comment confuses me. I was only pointing out a syntactic improvement around ternary operators.

joshri: Oh just the second part of your comment about the need for the error prop! No biggie

ui/components/SourceDetail.tsx
```diff
@@ -31,7 +31,7 @@ function SourceDetail({ className, name, info, type }: Props) {
     return <LoadingPage />;
   }
 
-  const s = _.find(sources, { name });
+  const s = _.find(sources, { name, type: type });
```
jpellizzari, line 34:
```suggestion
  const s = _.find(sources, { name, type });
```

ui/pages/v2/HelmChartDetail.tsx
```diff
@@ -0,0 +1,43 @@
...
+  return (
+    <Page error={null} className={className}>
+      <SourceDetail
+        name={name}
+        namespace={namespace}
+        type={SourceRefSourceKind.HelmChart}
+        info={(ch: HelmChart) => [
+          ["Chart", ch?.chart],
+          ["Ref", ch?.sourceRef?.name],
+          ["Last Updated", ""],
```
jpellizzari, line 24: We should probably have a `<Timestamp />` value here.

jpellizzari (approved):
Nice work, no blockers.

ui/pages/v2/HelmChartDetail.tsx
joshri: Just added Timestamps wherever last updated was blank

Issue #1478, opened 2022-02-18
A smoother, faster release process


ozamosi:
The [current state of the release process](https://github.com/weaveworks/weave-gitops/blob/main@{2022-02-14}/doc/release-process.md) (or at least something close to it) is documented in the Weave GitOps repo.

**What is Consuming our Time?**

The [release process](https://github.com/weaveworks/weave-gitops/blob/main/doc/release-process.md) is better documented than it used to be, but there are still a fair few manual steps. The simpler and more repeatable we can make the release process, the easier it is to release more often if and when we need to, and the less time we spend fixing up mistakes in a release. 

**What does good look like?**

I want a release process that takes a person minutes of their time to run instead of several steps that require 45+ minutes of asynchronous time to step through.

- A simplified release process that any team could trigger without the need for assistance.  Most steps done manually by a person in the current directions are now automated.
- CI/CD pipeline time is significantly reduced.  Optimizations are made such as caching and running non-dependent workflows in parallel.
- Updating our homebrew install is currently a manual process, I want this to be automated.
- Weave GitOps runs on a cluster.
- Snyk report is built and delivered properly.

**Where do we want flexibility?**

- The Enterprise FE team can easily get an NPM package built for absorbing the UI into Enterprise.
- Documentation is released with each new release but can also be released separately.

PR #1727, opened 2022-03-16
Write a more approachable release-tagging script

dhwthompson:
The current version of the script has a few issues:

- Because it uses command-line flags to bump version numbers, it's
  possible to bump multiple parts of the version number at once.
- The version-number logic was, at one point, capable of generating and
  pushing invalid tags (for instance, `v0.5.1-rc0.5.0`). This has
  [probably been fixed][1], but it's hard to be sure because...
- Bash being bash, it's not always easy to tell what the code is doing.
  Given the script has the potential to kick off and publish a new
  release of our product, it's worth building up our collective
  confidence in what the code is doing.

[1]: https://github.com/weaveworks/weave-gitops/pull/1266

I've taken a first stab at a version that addresses these issues, trying
to build up a release tagging script as I might like to see it. I've
documented everything I can think of documenting, and used a combination
of named and positional arguments to mean it's only possible to call for
one type of release (as it's a single named argument).

I decided to test the Weaveworks waters with a spot of Python, partly to
see what a Python-based deploy script could look like, and partly
because I got fed up bashing my head against bash. I've made sure to
keep to the dependencies in Python's standard library, because Python's
package management is a whole other nest of complexity I don't want to
get into.

At least partially related to #1478.


ozamosi:
This might absolutely be me not understanding how the old script worked, but… If the current version is 0.6.1, and I want to start the release process for 0.7.0, then I want to "bump, and make that an RC", yet I think the way you modelled the "state machine of release transitions" means I can only go from 0.6.1 to 0.7.0 directly, or to 0.6.2-rc1? My intuition about release numbers suggest that "rc" on top of stable shouldn't exist, instead "minor" and "patch" should both return rc1.

I also suspect throwing when you run minor on a rc isn't the best behaviour either - I've found myself going "hang on, I already have 0.6.2-rc1 but it turns out we've done non-patch changes, let's re-version this 0.7.0-rc1" more than once when working on other projects in the past.

I do think that not using bash is an excellent proposal, and python should be universal enough to not cause a problem.

tools/tag-release
```diff
@@ -0,0 +1,212 @@
...
+
+    Release tags are of the form "v0.6.3", and release candidates are of the form "v0.6.3-rc2".
+
+    Raise a ValueError if the tag isn't in a recognised format.
+    """
+    pattern = re.compile(
+        r"""
+            v
+            (?P<major>\d+)
+            .
```
SamLR, line 22: minor point but I think these `.` should be `\.` so they only match full stops rather than any character

SamLR (commented):
Looks reasonable but I'd echo @ozamosi's comment about what the allowed transitions are, feels like any-non-rc -> rc+(minor or patch increment) -> rc, minor or patch increment

tools/tag-release
SamLR: gotta love regexps

dhwthompson:
Ok, here’s what I’m thinking for a potential flow (and the flags to go with it).

Some principles I’ve tried to follow here:

* *Fail safe*. As this is kicking off a potential release, the script should abort with an error rather than do anything unexpected.
* *RCs by default*. Release candidates are good, we want to use them pretty much all the time, but let’s provide a get-out for situations we might not have thought about.

I’ve drawn up a map of input versions to output versions, as much for my own clarity as anything.

| Command | `v1.1.1` | `v1.1.1-rc1` |
| -- | -- | -- |
| `tag-release minor` | `v1.2.0-rc1` | ❌ |
| `tag-release patch` | `v1.1.2-rc1` | ❌ |
| `tag-release minor --no-rc` | `v1.2.0` | ❌ |
| `tag-release patch --no-rc` | `v1.1.2` | ❌ |
| `tag-release minor --abandon` | ❌ | `v1.2.0-rc1` |
| `tag-release patch --abandon` | ❌ | `v1.1.2-rc1` |
| `tag-release minor --abandon --no-rc` | ❌ | `v1.2.0`|
| `tag-release patch --abandon --no-rc` | ❌ | `v1.1.2` |
| `tag-release bump` | ❌ | `v1.1.1-rc2` |
| `tag-release promote` | ❌ | `v1.1.1` |

How does that look to people? Does that make sense from your perspective as potential users?

Side-note: at this point, it’s possible that it makes more sense to [build a new branch-based release process](https://www.notion.so/weaveworks/Releasing-GitOps-Core-32ad67a70f714d23a474815b1ffb4d52) rather than spend more time replacing this script, although either way we go about it we’re probably going to:

1. Need at least some of the same logic in place
2. End up with one less Bash script

SamLR:
I'm not 100% sure what the `--abandon` flag is doing here? 

Other than that though I think the flows look reasonable.

ozamosi:
So I went outside. We should ignore all the things I wrote before, and do exactly what you propose for a branch based release process.

I believe that as more conservative organisations starts to rely on gitops core we'll want longer-lived stable branches, and as more developers start to work on it we'll want an easier-to-separate release process. But my workings out above has convinced me that we can add that when we get there, from what you've proposed for a branch based release process.

jrryjcksn:
Thumbs-up for a branch-based release process :+1:   One of the pain points of the current approach is that it's difficult to test changes to the release process. You either have to "test it live" with the real environment or dummy up a separate fork or standalone repo that can stand in for the real one. Neither option is great.

Also, I have high hopes that we can now make the process much simpler and less error-prone because much of the complexity in the original process derived from managing the documents as a separate repo. :crossed_fingers:

SamLR (approved):

PR #1731, opened 2022-03-16
ci: Remove deploy workflow

ozamosi:
I think this used to do a lot more, but now it only did 2 things:
 * Publish binaries to S3. Move this step to the PR workflow, and change it so it works again.
 * Re-run all unit-tests with test coverage enabled. We've said we don't care about code coverage at this point.


ozamosi:
@dhwthompson it's possible something here is or was related to releases, so gonna ask you for review.

dhwthompson:
> Publish binaries to S3. Which, we host official binaries in github, so I don't know why? Maybe related to some testing?

Aha—I happen to know this one.

It's a Unix system! I know this!

The reason we upload these binaries to S3 is so that the Enterprise test flows can download them and test against them. I’m not 100% clear on the way the various workflow incarnations work, and it looks like _some_ of those workflows download and build their own `gitops` binary instead of using these ones, but [these are the bits I can find](https://github.com/weaveworks/weave-gitops-enterprise/search?q=gitops-binary).

ozamosi:
OK, thank you, I've put back the binary uploading.

SamLR (approved):
Looks reasonable to me.

If this starts impacting PR merge time we could look at splitting `ci-upload-binary` into two parts: one that does the compilation regardless of the status of other steps and the other which pushes if linting etc succeeeded.

ozamosi:
We could - however, this was ~3 minutes, and I don't think post-merge time is quite as important as pre-merge time.

SamLR:
> We could - however, this was ~3 minutes, and I don't think post-merge time is quite as important as pre-merge time.

reasonable, it should be an easy enough change to make if it does become needed

PR #1745, opened 2022-03-16
Only return deployments with part-of=flux

### LINKED ISSUE #1713: Flux Runtime page shows too many things
The Flux Runtime will show any pods running within the flux-system namespace.

1. As a user, it isn't always guaranteed that the controllers will be in `flux-system`.
2. As a user, I could  want to install more than one set of controllers installed on a cluster.  E.g. I don't want two apps teams sharing the same controller set for reliability and blast radius purposes.  Or certain apps rely on a specific version of a controller to be running.
3. As a user, I want to see the runtime health and not other resources that happen to fall under the `flux-system` namespace.

The deployments have labels for the controllers.  Is this good enough?  I feel like this is something that could bite us in the future.
```
app.kubernetes.io/part-of=flux
control-plane=controller
```

The screenshot below shows `podinfo` and `wego-app`.

jpellizzari:
Closes #1713

Only show deployments that match the `part-of` labels. Also fixes some nits around the proxy config for local development.


yitsushi (approved):

jpellizzari:
I had an old version of flux that was not populating the label. This looks to be working now
Screenshot from 2022-04-01 08-52-22
:

Issue #1706, opened 2022-03-14
UX: Detail page doesn't update when issue resolved


JamWils:
As a user, I would like to see that an error has occurred on the detail page and then update to "green" when the issue is resolved without reloading the page.

Below I was able to see the UI update but I had to refresh the page.  I waited about 30 seconds after it resolved before I hit the refresh button.

![Screen Shot 2022-03-14 at 10.22.22 AM.png](https://images.zenhubusercontent.com/60f5a90306cf76325d5accdc/eb66f47e-c21d-4063-b4f8-4eec053de58f)

When I hit refresh I got:

![Screen Shot 2022-03-14 at 10.31.34 AM.png](https://images.zenhubusercontent.com/60f5a90306cf76325d5accdc/d060de22-d9f7-4402-8c24-562d6f4111a3)

PR #1750, opened 2022-03-17
Add polling through react-query

joshri:
Closes: #1706

Adds the polling indicator component to all pages where requests will refresh every 5 seconds 

https://user-images.githubusercontent.com/65822698/158902347-f55f0cc2-1199-42bd-a0bf-fd9f97f3e705.mov


jpellizzari (approved):
LGFN (looks good for now 😄 ).

Could you create a follow up issue that says something like "Move page/fetching state to shared context"?. 

I would like us to wire up all of the repeated `useQuery` calls to a central state that can be used to display loading/error states on the page.

Issue #1560, opened 2022-03-01
Upgrade cmd for v2


J-Thompson12:
There are some issues that need to be figured out for upgrade to work in v2. Currently upgrade does a few things:
- Checks for a secret the user has added to say they can use enterprise.
- Adds a weave gitops enterprise helm chart
- Deletes the existing weave gitops core crd

Doing these things in v2 create some problems around the file structure. In v2 we have no knowledge of what the file structure the user will have. We also will not know what the file name is for the weave gitops yaml. 

The current Issues I see:
- Without knowledge of the file structure we do not know where to add the enterprise yaml.
- Without knowledge of the file structure we do not know where the core yaml file is to delete it.

There are a few possible solutions here but each comes with its own set of drawbacks. 

The upgrade cmd can have a path flag for the user to tell us where to put the enterprise yaml. This still leaves the issue of deleting the core yaml. For that we can either have the user provide a path to that as well through a flag or we can just inform the user with a warning to delete the core yaml file for us.

There is the option to remove the upgrade cmd completely and just have the user upgrade the same way they install. Give them the yaml and they can place it where it is needed. The drawback for this is we can no longer check for the enterprise secret during the upgrade cmd.

PR #1751, opened 2022-03-17
1560 upgrade command

jrryjcksn:
Closes: #1560 

**What changed?**
The main changes are around removing dependencies on the old core model, including the directory structure. Since we removed the directory structure and the wego config map:
* We don't know where to find the configuration repo
* We don't know where to find the core configuration manifests (GitRepository and HelmRelease) within the config repo

So, we:
* add a `config-repo` path argument back into the upgrade command (defaulting to "")
* add a `cluster-path` argument pointing to the subtree within the repo for the cluster of interest
* move the enterprise manifests into `cluster-path` rather than nested in the directory structure
* remove the `apps/capi/templates/.keep` file from the old directory structure
* search for the existing core config manifests under the cluster-path
  * if single file found containing both manifests and nothing else, remove them
  * otherwise (only one manifest in the file, cruft manifests in the file, no file found) print a message to the user that we would remove them and move on

**Why?**
To make upgrade available in the new model

**How did you test it?**
* Unit tests for config search
* Manual upgrade testing
* 

**Release notes**
No

**Documentation Changes**
Only CLI docs which should be auto-updated


pkg/upgrade/upgrade.go
```diff
@@ -262,25 +263,11 @@ func makeHelmResources(namespace, version, clusterName, repoURL string, values [
...
-func upgradeGitManifests(gitClient git.Git, cname, wegoEnterpriseManifests string, logger logger.Logger) error {
-	capiKeepPath := filepath.Join(git.WegoRoot, git.WegoAppDir, "capi", "templates", ".keep")
-	capiKeepContents := string(strconv.AppendQuote(nil, "# keep"))
-	kustomizationPath := git.GetSystemQualifiedPath(cname, models.SystemKustomizationPath)
-	wegoAppPath := git.GetSystemQualifiedPath(cname, models.WegoAppPath)
-	wegoEnterprisePath := git.GetSystemQualifiedPath(cname, WegoEnterpriseName)
+func upgradeGitManifests(gitClient git.Git, repoDir, clusterPath, cname, wegoEnterpriseManifests string, logger logger.Logger) error {
+	wegoEnterprisePath := filepath.Join(clusterPath, WegoEnterpriseName)
 
 	manifests := map[string]string{
```
SamLR, line 269: Am I right that this `manifests`//`for path, content` construction can be removed & replaced with a simple call to `gitClient.Write`?

pkg/utils/common.go
```diff
@@ -130,3 +135,101 @@ func MigrateToNewDirStructure(orig string) string {
...
+type WalkResult struct {
+	Status ConfigStatus
+	Path   string
+}
+
+func (wr WalkResult) Error() string {
+	return fmt.Sprintf("found %s: with status: %s", wr.Path, wr.Status)
+}
+
+func FindCoreConfig(dir string) WalkResult {
```
SamLR, line 172: This function feels fragile. Could we just ask the user for the location instead?

pkg/utils/common.go
```diff
@@ -130,3 +135,101 @@ func MigrateToNewDirStructure(orig string) string {
...
+					break
+				}
+
+				docs = append(docs, entry)
+			}
+
+			rootNode := dasel.New(docs)
+			foundPartial := false
+
+			_, err = rootNode.QueryMultiple(".(kind=HelmRelease)(.metadata.name=ww-gitops)")
```
SamLR, line 204: The metadata name here should be a constant, particularly as we don't yet have an official `install` method so the value of this name is likely to change.

pkg/utils/common.go
```diff
@@ -130,3 +135,101 @@ func MigrateToNewDirStructure(orig string) string {
...
+			_, err = rootNode.QueryMultiple(".(kind=GitRepository)(.metadata.name=ww-gitops)")
+			if err != nil {
+				if foundPartial {
+					return WalkResult{Status: Partial, Path: path}
+				}
+
+				return nil
+			}
+
+			val, err := rootNode.Query(".[#]")
```
SamLR, line 218: Could you add a comment what this query is doing? (I think it gets the length?)

pkg/utils/common.go
```diff
@@ -130,3 +135,101 @@ func MigrateToNewDirStructure(orig string) string {
...
+			if err != nil {
+				return nil
+			}
+
+			if val.InterfaceValue() != 2 { // we have two manifests
```
SamLR, line 223: This also feels pretty fragile, I wouldn't be surprised if we end up with more manifests.

SamLR (commented):
I've got a few questions for this. I also think it would be a good idea if, e.g. @ozamosi, gave this a look as he's spent more time working on this than I have and has a better feel for the state of the repo than I.

ozamosi:
I might have missed some context - have we made an active decision to upgrade a remote git repository, instead of a local git clone? The PR description seems to suggest the command upgrades a local clone (because it mentions `$PWD` - or do I misunderstand?), but the code still does the previous clone/modify/create PR flow.

pkg/upgrade/upgrade.go
jrryjcksn: Yikes! Yes, there used to be multiple files written here and when I found I didn't need one of them, I removed it to test and forgot to to go back and clean this up. Thanks!

pkg/utils/common.go
jrryjcksn: James and I discussed this a bit. The main reason for doing this is that I felt like the "single file containing both manifests and nothing else" case should be by far the most common. This provides a very simple upgrade experience for most users (who won't have to concern themselves with the core config at all. In particular, if someone upgrades six months after installing core (potentially not the same person who installed core) no knowledge would be required in the typical scenario.

As for  the fragility, except for the case in which a user put the core manifests in a single file _along with other manifests_ and _still_ wants to the delete the whole file, this function should be correct.

pkg/utils/common.go
jrryjcksn: Will do. (and you are correct)

pkg/utils/common.go
jrryjcksn: That seems quite plausible (and "2" should be a constant) but I don't think it's reasonable to assume that when we add manifests to core we won't need to do any updates to the upgrade process unless we're prepared to lock down the directory structure a bit.

jrryjcksn:
> I might have missed some context - have we made an active decision to upgrade a remote git repository, instead of a local git clone? The PR description seems to suggest the command upgrades a local clone (because it mentions `$PWD` - or do I misunderstand?), but the code still does the previous clone/modify/create PR flow.

The mention of $PWD was a thinko

pkg/utils/common.go
SamLR: That makes sense. I guess this is something we'll just have to make sure is clear in the documentation (both public & developer).

yiannistri (approved):
LGTM 👍

Issue #1757, opened 2022-03-18
Move react-query isFetching state to shared context


joshri:
Using the global `useIsFetching` hook from react-query in shared context (https://react-query.tanstack.com/reference/useIsFetching) would be more efficient than our current polling implementation with `isFetching` on each query. 

To quote resident genius @jpellizzari from PR #1750, "I would like us to wire up all of the repeated useQuery calls to a central state that can be used to display loading/error states on the page."

PR #1758, opened 2022-03-18
Switch isFetching to useIsFetching

joshri:
Related to and might close: #1757

Source Detail pages now use the `Page` component to render their title, and `Page` holds the global is fetching state with the `useIsFetching` hook. I don't like that it's pushed the `PageStatus` component down, but there's no way to render that one outside of `SourceDetail`

image

`SourceDetail` errors appear under title - could set error state of page in parent component of source detail, or have all queries alter app context error state?

image


jpellizzari (approved):
Much nicer!

PR #1770, opened 2022-03-21
Remove actively misleading help texts

ozamosi:
* Add, delete and get all assumed you want to add or delete an app. We
      no longer have apps. Get also mentioned getting commits, which also
      isn't available.
* The main command had many examples of how to install gitops and
  manage applications. None of that is applicable any more, so remove
  any mention of it - we'll probably want to do a _something_ with
  installation instructions, but this clears the way to do so.


sympatheticmoose:
@ozamosi still references to apps under `gitops get`:

```
# Get all applications under gitops control
gitops get apps

# Get last 10 commits for an application
gitops get commits <app-name>
```

sympatheticmoose (approved):
LGTM

Issue #1778, opened 2022-03-22
Production version of a helm chart


JamWils:
Main changes:

* Use base helm template (e.g. allows annotations, correct full-name, etc etc)
* Make args & envvars modifiable 
* Provide mechanisms to limit `impersonation` options to specific groups and/or users

PR #1779, opened 2022-03-22
Helm upgrades

SamLR:
Closes: #1778 

**What changed?**

* Replaced previous chart with one built out from the recommended helm template (i.e. the one created by `helm create`)
* Added options to restrict service account `impersonation` scope. This is important for production usage as without a limited scope on this permission the app can impersonate `system:masters` which is unacceptable
* Made the `wego-admin` user (now renamed `wego-test-user`) an optional creation
* Added quality of life fixed like `additionalArgs` and `envVars` for greater flexibility

**Why?**

* I needed to be able to pass `--insecure` for deployment (and handle TLS termination in the LB)... things got out of hand

**How did you test it?**

Running `helm template` with various settings, e.g.:
```bash
$ helm template -n TEST --dry-run --debug \
      --set resources.limits.cpu=320,resources.limits.memory=666 \
      --set resources.requests.cpu=125,resources.requests.memory=333 \
      --set additionalArgs='{a,b,c}' \
      --set envVars\[0\].name=TEST,envVars\[0\].value=foo,envVars\[1\].name=bar \
      --set rbac.impersonationResources='{user,group,serviceaccout}' \
      --set rbac.impersonationResourceNames='{foo,bar}' test ./charts/gitops-server
```

**Documentation Changes**

The notes from the README.md should be reflected in the main line website docs (probably in `gitops-dashboard`) but those look like they need a top-to-bottom overhaul so I wasn't sure where to start.

**Note**

Ideally the permissions granted to the service account should be further reduced. As I understand it the non-`impersonate` stanzas exist to support profiles which is not a default use of the UI so should probably be removed but doing so (currently) stops the UI working. I think this should be fixed.


ozamosi:
* The chart is called `weave-gitops` in a directory called `gitops-server` - I think the chart should move to `weave-gitops` as the old chart is deleted.
* I believe this needs some changes to `tools/helm-values-dev.yaml` so that it doesn't break the dev env.

SamLR:
> * The chart is called `weave-gitops` in a directory called `gitops-server` - I think the chart should move to `weave-gitops` as the old chart is deleted.

I think the question here is should this be "weave-gitops" or "gitops-server" (there're also places we refer to it as "wego-app" still). I don't mind which we go with and, yes, once one is chosen happy to update as many places as I can find.

>     * I believe this needs some changes to `tools/helm-values-dev.yaml` so that it doesn't break the dev env.

Good catch, will go and check

josecordaz:
This work looks awesome. Assuming we are using `Helm 3`... should we also plan on defining a json schema for the file values.yaml ? (`values.schema.json`). A couple advantages I see are that we can enforce specific value types and required fields when installing, updating or using template subcommand. Or maybe we want to wait a bit more to let weave-gitops to get a bit more mature? 🤔

charts/gitops-server/README.md
```diff
@@ -0,0 +1,52 @@
...
+It installs the weave-gitops-server component as a 1-replica deployment.
+
+Optionally it will also install:
+* Service Account
+* ClusterRoleBinding (to the service account) and ClusterRole with the
+  permissions required to run Gitops.
+* Service, this is optional as you may want to limit access to the UI to via
+  port-forwarding
+* Ingress
+* Test User -- A test user with hard-coded username & password with minimal
```
Callisto13, line 14: The test user is actually a backup admin user in the case that OIDC is failing and cluster ops need to do things.
It is currently undergoing some feature realignment, see the thread here: https://weaveworks.slack.com/archives/C03244W0C8H/p1647967858091769

charts/gitops-server/templates/test-user.yaml
```diff
@@ -41,9 +41,9 @@ rules:
...
 metadata:
-  name: cluster-user-auth
-  namespace: {{.Values.namespace}}
+  name: wego-test-user-auth
 type: Opaque
 data:
-  username: {{.Values.adminUsername}}
-  password: {{.Values.adminPassword}}
+  username: {{ .Values.testUser.username | b64enc | quote }}
+  password: {{ .Values.testUser.password | b64enc | quote }}
```
Callisto13, line 48: do these need to be `b64enc` here? i think i remember that k8s will do that automagically when the secret is created

charts/gitops-server/values.yaml
```diff
@@ -0,0 +1,106 @@
+# Default values for chart.
+# This is a YAML-formatted file.
+# Declare variables to be passed into your templates.
+
+replicaCount: 1
+
+image:
+  # FIXME check the app name
```
Callisto13, line 8: is this a thing you wanted to fix now?

charts/gitops-server/values.yaml
```diff
@@ -0,0 +1,106 @@
...
+fullnameOverride: ""
+
+logLevel: info
+# Additional arguments to pass in
+# additionalArgs: []
+
+# To enable enterprise OIDC:
+# envVars:
+# - name: WEAVE_GITOPS_AUTH_ENABLED
+#   value: true
```
Callisto13, line 25: this flag is also needed for the admin user login flow, just in case you wanted to note that.
i think it gates all login functionality, in core as well as ent

charts/gitops-server/values.yaml
```diff
@@ -0,0 +1,106 @@
...
+  create: true
+  # List of resources that the service account can impersonate
+  impersonationResourceNames: []
+  # Limit the type of principal that can be impersonated
+  # impersonationResources: [ "users", "groups" ]
+  # Specify whether additional permissions should be granted to the service account
+  # in order to work with weave-gitops enterprise
+  viewSecrets: []
+
+# This should not be used in production
```
Callisto13, line 47: this will be used in production

charts/gitops-server/values.yaml
```diff
@@ -0,0 +1,106 @@
...
+podAnnotations: {}
+
+podSecurityContext: {}
+  # fsGroup: 2000
+
+securityContext: {}
+  # capabilities:
+  #   drop:
+  #   - ALL
+  # readOnlyRootFilesystem: true
```
Callisto13, line 65: just for context, if users make the pod fs rootfs read only, i think the profiles mechanism will fail as it writes a disk-based cache

charts/gitops-server/values.yaml
```diff
@@ -0,0 +1,106 @@
...
+  viewSecrets: []
+
+# This should not be used in production
+testUser:
+  # Whether the wego-admin user should be created.
+  # If you use this make sure you add it to rbac.impersonationResourceNames.
+  create: false
+  # helm will base64 encode these
+  username: not-production-worthy
+  password: seriously-change-me
```
Callisto13, line 54: we should note that these need to be bcrypted not plain-text

charts/gitops-server/values.yaml
SamLR: negotiations are on going https://weaveworks.slack.com/archives/C03244W0C8H/p1647972249151359 :p 

Ideally yes but it's not a blocker on this and it may need a wider conversation

charts/gitops-server/templates/test-user.yaml
SamLR: you do (I just checked against a kind cluster with a simple helm chart). I think the auto-b64encode is only when you use `kubectl`

charts/gitops-server/README.md
SamLR: Why would you have a backup admin with only read permissions?

charts/gitops-server/values.yaml
SamLR: Would it make more sense to turn this into a variable then, something like:
```yaml
# Sets the WEAVE_GITOPS_AUTH_ENABLED environment variable to enable login
enable_login: true
```

charts/gitops-server/templates/test-user.yaml
Callisto13: i swear i have seen that as well when just doing like a:

```go
secret := &v1.Secret{
        ObjectMeta: metav1.ObjectMeta{
                Name:      "foo",
                Namespace: "bar",
        },
        Type: "Opaque",
        Data: map[string][]byte{
                "key": []byte("value"),
        },
}

return yaml.Marshal(secret)
```
even then it is encoded

charts/gitops-server/README.md
Callisto13: because those are all i added so far 🙃

charts/gitops-server/values.yaml
SamLR: I'm unclear as to why this would be used in production, is there a ticket for clarifying/resolving it's use there?

charts/gitops-server/values.yaml
SamLR: Is the profiles mechanism supposed to always be available? I was unclear as to whether it was an enterprise only feature. (this is default copy but once its use is clarified I can update & fix this).

charts/gitops-server/values.yaml
SamLR: is that in addition to base64 encoded to make them k8s secrets?

charts/gitops-server/values.yaml
Callisto13: see that thread I linked. the AdminUser feature is a deliberate one for prod.
I have to write up the tickets now

charts/gitops-server/values.yaml
Callisto13: yeh that sounds good

charts/gitops-server/values.yaml
Callisto13: yes. the code will do this: https://github.com/weaveworks/weave-gitops/blob/main/pkg/server/auth/server.go#L308-L313

charts/gitops-server/values.yaml
Callisto13: See this thread here: https://weaveworks.slack.com/archives/C03244W0C8H/p1646387603804719

Callisto13: (It has no resolution and I am still confused, but just for context)

SamLR:
> This work looks awesome. Assuming we are using `Helm 3`... should we also plan on defining a json schema for the file values.yaml ? (`values.schema.json`). A couple advantages I see are that we can enforce specific value types and required fields when installing, updating or using template subcommand. Or maybe we want to wait a bit more to let weave-gitops to get a bit more mature? 🤔

I think waiting until the interface is a bit more stable would be best. There are a lot of environment variables and parameters that aren't currently exposed which probably should be (I've not because I'm unsure what they all do and how they interact).

SamLR:
I've created a ticket, #1784, to cover the work in exposing the rest of the various flags & envars that are available to configure the server.  @Callisto13  I think this may cover some of the stuff WRT use of the admin user as well?

charts/gitops-server/README.md
SamLR: from a point of least privilige I'd like to only make it a requirement/recommendation when it's actually in a position to fulfill those extended duties ;)

charts/gitops-server/templates/test-user.yaml
SamLR: bold of you to assume there're rules ;)

charts/gitops-server/values.yaml
Callisto13: https://github.com/weaveworks/weave-gitops/issues/1785

Callisto13:
Yep that's fine. We can leave getting the admin user stuff "correct" until we have smoothed out exactly what it is for and what it does.

charts/gitops-server/values.yaml
SamLR: I'm unsure what it does but it feels like it's separate to what the rest of the ui is (currently) doing

Callisto13: yep 💯  Core v2 has no real concept of what profiles are yet. so let's shuffle it away until they come into scope

charts/gitops-server/values.yaml
SamLR: cf. above thread, is there a reason not to change this when those changes are implemented?

Callisto13: you mean when https://github.com/weaveworks/weave-gitops/issues/1787 is done? yep makes sense

charts/gitops-server/values.yaml
SamLR: (turns out helm has a handy [htpasswd function](https://helm.sh/docs/chart_template_guide/function_list/#htpasswd)

SamLR:
@Callisto13 I've reverted a load of the 'admin user' bits to closer how they were previously because they're hard coded elsewhere. I've tested it all and I don't think I've broken anything

charts/gitops-server/README.md
```diff
@@ -0,0 +1,52 @@
...
+* `get`, `list`, `watch` on `helmrepositories` and `secrets`. These permissions
+  are required by the profiles system.
+
+### Impersonate
+
+When deploying gitops-server it is recommended to limit the types of resource
+and specific resources that the service account can impersonate. e.g.
+```yaml
+rbac:
+  create: true
```
luizbafilho, line 36: It is not like users have a choice here, they must have impersonation setup otherwise nothing works. So ideally we could have some default values as `impersonaionResouces` and `impersonationResourceNames` but allow configuration of those here.

luizbafilho (approved):
LGTM

charts/gitops-server/README.md
SamLR: I've improved how the defaults for `impersonationResources` is set in ca342853 

Much as I'd love to provide defaults for `impersonationResourceNames` I can't think of a good way to do so. The best solution I've come up with would be to create some default groups (e.g. `viewer` and `editor`) that are then the only things to be impersonated. The problem with this is that it requires the user's k8s authentication system to be configured to support them. I suspect this is best solved via documenting common processes for it but that feels very out of the scope of this PR (I'm also not 100% sure this is the best way to do it, I'm not sure who might be the best person to cover this sort of security work).

luizbafilho: definitely, we can't set anything there since there is no way for us to know what groups the companies use. I'm good with that, the important bit that's allowing configuration is there, it should be enough.

Issue #1675, opened 2022-03-10
Update core namespace endpoint to use client pool 


Callisto13:
**Short description**
Following on from https://github.com/weaveworks/weave-gitops/issues/1600 and based on https://github.com/weaveworks/weave-gitops/pull/1556, this is to update the endpoint in `core/server/namespace` to:
- Pull a list of clients from the `ctx`
- Loop over those and query each

**Acceptance criteria**
Everything which currently works should still work

Callisto13:
I don't think it is _exactly_. Looks like it is using a single client. But is worth noting that this one should probably wait for #1780 to be merged, just to avoid complications

yitsushi:
The scope of this ticket changes with my PR on namespace cache (https://github.com/weaveworks/weave-gitops/pull/1780) because I replaced the namespace grpc endpoints to use the cache instead. So it's more like:

> update cache/namespace to use client pool, pull and store data respectively.

PR #1780, opened 2022-03-22
feat: Create a namespace client and watcher

### LINKED ISSUE #1724: Create a namespace client and watcher
We want to be able to list the namespaces on a cluster to support cross-cluster querying. We don't want to hit the kube API server on every request, so we need to get the list of namespaces, store them in memory, and provide a client interface for other parts of the code base to get the list of namespaces for a cluster.

Acceptance Criteria:
* Provides a `ListNamespaces` function or equivalent interface
* Watches clusters for new or changed namespaces
* Stores the namespaces for a cluster in memory

Implementation Notes:
* This is probably a long-running `go` routine that watches the cluster namespace for changes

yitsushi:
Tried to use the Watch API, everything worked as expected, but I
couldn't get `testutils` in a good shape because Manager in controller
runtime does not expose WithWatch clients.

Controller-runtime has a separate Client and WithWatch (which is a
client with an extra function). That wouldn't be a problem as they are
basically the same, I tried to update `client.Client` everywhere to be
`client.WithWatch`, and the last piece was in `pkg/testutils/testutils.go`
where we are using Manager that has a `GetClient()` function, but does
not have a single function that would return a `WithWatch`.

Because of that, we are using polling instead. Maybe later we can
revisit this gap, for example if we have a wrapper client around
controller-runtime client and that can handle both `WithWatch` and `Client`.

**Update:**

- GetFluxNamespace endpoint uses the Namespaces cache.
- ListNamespaces endpoint uses the Namespaces cache.

I don't know how to simplify the API for that yet.

- ForceRefresh should be used only if required.
- Originally cache.Conatiner was a capsule and could be instantiated,
  but that way we couldn't access it from tests and we may want to reuse
  the same cache in other parts of the application and not only on
  endpoints (for example some other goroutines).
- Only Namespaces and ForceRefresh functions are exposed to the
  consumer. We can extend it later, even change the underlying structure
  to be a `map[string]v1.Namespace` with multi-cluster.

Closes #1724


charts/weave-gitops/templates/wego-app.yaml
```diff
@@ -63,6 +63,9 @@ rules:
   - apiGroups: [""]
     resources: [ "secrets" ]
     verbs: [ "get", "list" ]
+  - apiGroups: [ "" ]
+    resources: [ "namespaces" ]
+    verbs: [ "get", "list", "watch" ]
```
yitsushi, line 68: We can remove `watch`

core/cache/namespace.go
```diff
@@ -0,0 +1,46 @@
...
+		namespaces: []v1.Namespace{},
+	}
+}
+
+func (n *namespaceStore) Namespaces() []v1.Namespace {
+	return n.namespaces
+}
+
+func (n *namespaceStore) Start(ctx context.Context) {
+	go func() {
```
yitsushi, line 31: Add `cancelFn` so we can stop this loop (for example graceful shutdown)

core/server/fluxruntime.go
```diff
@@ -32,6 +34,10 @@ func (as *coreServer) ListFluxRuntimeObjects(ctx context.Context, msg *pb.ListFl
 
 	l := &appsv1.DeploymentList{}
 
+	for _, ns := range as.cacheContainer.Namespaces() {
+		logrus.WithField("check", "efertone").Infof("NamespaceList: %s", ns.Name)
+	}
```
yitsushi, line 39: Oops, debug ;)

core/server/server.go
```diff
@@ -49,10 +50,16 @@ func NewCoreConfig(cfg *rest.Config, clusterName string) CoreServerConfig {
 
 func NewCoreServer(cfg CoreServerConfig) pb.CoreServer {
 	cfgGetter := kube.NewImpersonatingConfigGetter(cfg.RestCfg, false)
+	k8s := kube.NewDefaultClientGetter(cfgGetter, cfg.clusterName)
+	ctx := context.Background()
```
yitsushi, line 54: That's a context mess right now.

core/cache/namespace.go
```diff
@@ -0,0 +1,79 @@
...
+import (
+	"context"
+	"time"
+
+	"github.com/go-logr/logr"
+	v1 "k8s.io/api/core/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const pollInteralSeconds = 120
```
Callisto13, line 12:
```suggestion
const pollIntervalSeconds = 120
```

core/cache/namespace.go
```diff
@@ -0,0 +1,79 @@
...
+	}
+}
+
+func (n *namespaceStore) Start(ctx context.Context) {
+	var newCtx context.Context
+
+	newCtx, n.cancel = context.WithCancel(ctx)
+
+	go func() {
+		ticker := time.NewTicker(pollInteralSeconds * time.Second)
```
Callisto13, line 52:
```suggestion
		ticker := time.NewTicker(pollIntervalSeconds * time.Second)
```

core/cache/suite_test.go
```diff
@@ -0,0 +1,12 @@
+package cache_test
+
+import (
+	"os"
+	"testing"
+)
+
+func TestMain(m *testing.M) {
```
Callisto13, line 8: what's this for?

core/server/server.go
```diff
@@ -49,11 +51,19 @@ func NewCoreConfig(log logr.Logger, cfg *rest.Config, clusterName string) CoreSe
 }
 
 func NewCoreServer(cfg CoreServerConfig) pb.CoreServer {
+	ctx := context.Background()
+
 	cfgGetter := kube.NewImpersonatingConfigGetter(cfg.RestCfg, false)
+	k8s := kube.NewDefaultClientGetter(cfgGetter, cfg.clusterName)
+	c, _ := k8s.Client(ctx)
+	cacheContainer := cache.NewContainer(c, cfg.log)
```
Callisto13, line 59: I am guessing this will need to go soon with #1675?

Callisto13 (commented):
I think you committed `core/logger/logger.test` by mistake.

How will the cacheContainer work when we add in the implementation to update `GetFluxNamespace` to query across a pool of clients taken from the `ctx`? From what I understand, here a single cache is created with a single client at server config time, which is fine for now as doing multi-cluster is not in scope for this ticket. Just wondering if you have a plan for how it will look when doing 1675.

core/cache/suite_test.go
yitsushi: Left there, initially i did some bootstrapping here.

yitsushi:
> I think you committed core/logger/logger.test by mistake.
Not it's not by mistake, that's the fix for tests with different timezone than GMT.

core/server/server.go
yitsushi: Possibly yes.

Callisto13:
The fix is an empty binary file?

core/server/namespace.go
```diff
@@ -17,45 +16,24 @@ const (
 var ErrNamespaceNotFound = errors.New("namespace not found")
 
 func (as *coreServer) GetFluxNamespace(ctx context.Context, msg *pb.GetFluxNamespaceRequest) (*pb.GetFluxNamespaceResponse, error) {
```
Callisto13, line 18: while you are there, can you do me a massive favour and make `as` `cs` in this file 😁  pleeeeasseee

yitsushi:
> Design Q: can I ask why you did it so that there is a Container bit which wraps the NamespaceStore bit rather than just having the NamespaceStore? Is it so that we can have more types of Containers in the future?

Yes, the container is there so we can add later other caching layers, and because they are isolated themselves, they can act as they want, and exposed through the container only. At least that was my idea.

core/server/namespace.go
yitsushi: done

core/cache/facekubeclient_test.go
```diff
@@ -0,0 +1,66 @@
...
+import (
+	"context"
+
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/meta"
+	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type fakeKubeClient struct {
```
jpellizzari, line 12: Can we use [counterfiter](https://github.com/maxbrunsfeld/counterfeiter) for this?

core/cache/namespace.go
```diff
@@ -0,0 +1,79 @@
...
+
+	go func() {
+		ticker := time.NewTicker(pollIntervalSeconds * time.Second)
+
+		defer ticker.Stop()
+
+		for {
+			list := &v1.NamespaceList{}
+
+			err := n.client.List(newCtx, list)
```
jpellizzari, line 59: Question: so once we switch to the client pool, this will be a loop over all the clients to populate the namespace cache, correct?

jpellizzari (approved):
Nice work. Only questions.

Callisto13:
> container to make sure if we add other resources, they will not be added one by one to the coreServer struct

yesss +1

core/cache/namespace.go
Callisto13: There may need to be a little rethink of how we interleave these 2 things together.

Right now the ClientPool is build ad-hoc per request (see top func of core/server/kustomization.go), whereas this cache is started (with a client) on server boot.

2 things may need to happen:
- It would be helpful if the ClientPool was cached in advance, not built on the fly (I believe we have plans for this)
- This Start may either need to watch the context for the client pool key to appear, or we move _something_ up to where the query is happening


My hot take, off for a covid test

core/cache/facekubeclient_test.go
yitsushi: I never used counterfiter and reading the readme it's not clear how can I implement given logic into that generated mock, and sadly I don't know how can I do that right now as it would take days with try-and-error (unless someone else does it and commits it into the PR).

core/cache/namespace.go
yitsushi: > Question: so once we switch to the client pool, this will be a loop over all the clients to populate the namespace cache, correct?

It can loop over them, but it is possible to make it parallel and each client in the pool has it's own cycle, only `Start()` and `Stop()` would be exposed to do that. I don't really know how client pool is working 100% but I tried to make this logic isolated as possible so we can change anything inside and all callers would still use the `Namespaces()` function, that might be better to change to return `map[string]v1.Namespace` instead.

core/cache/facekubeclient_test.go
jpellizzari: You can also use `DoThingsReturns()`

> it's easier to not use. right now it's a simple mock

OK not a deal breaker.

luizbafilho: To handle this we rely on the official fake client as we do in other places: https://github.com/weaveworks/weave-gitops/blob/a52d6faa9b53af338c665b471d2406d1f70a9174/pkg/services/auth/auth_suite_test.go#L21

That will be the path forward to help us reduce the slowness our tests already have.

core/server/server.go
```diff
@@ -49,11 +51,19 @@ func NewCoreConfig(log logr.Logger, cfg *rest.Config, clusterName string) CoreSe
 }
 
 func NewCoreServer(cfg CoreServerConfig) pb.CoreServer {
+	ctx := context.Background()
+
 	cfgGetter := kube.NewImpersonatingConfigGetter(cfg.RestCfg, false)
+	k8s := kube.NewDefaultClientGetter(cfgGetter, cfg.clusterName)
```
luizbafilho, line 57: we should avoid using this Initializer since it only adds unnecessary complexity.

```go

import (
)

cl, err := client.New(cfg.RestCfg, client.Options{})
```

The snippet above should do the trick

core/cache/namespace.go
luizbafilho: Good point, right now `CientsPool` expects a user to impersonate, but in this context, there is no user, since it's the server calling with its own credentials. We could tweak a little bit to ignore the impersonation if the user is `nil` to make that happen.

The cache container will need to have its own Clients Pool built once on cache container instantiation.

luizbafilho (changes requested):

core/server/server.go
Callisto13: Was there a reason we cannot pass it in from the caller?

luizbafilho (approved):

PR #1783, opened 2022-03-23
docs: Support staging multiple branches, remove docs front page

ozamosi:
This makes the staging do actual staging - that is, for this branch, the docs are in https://staging.docs.gitops.weave.works/docs-per-branch

This also adds a little status update, which you can see by scrolling through the checks in the PR and clicking on Details for "Doc site preview" that takes you to that specific URL.

This finally removes the front page of the docs, and replaces it with a redirect to the intro. I needed to make a docs change to trigger the docs pipeline to see that the previous changes worked, and this was something @sympatheticmoose specifically requested.


sympatheticmoose:
thanks @ozamosi !

SamLR (approved):
I'm not keen on the `sed` use to rewrite the config file but I'm not sure there's a better option.

I'm also a little confused by the pre-existing 'if yarn then yarn else npm` build step but I'm inclined to leave it.

Issue #1761, opened 2022-03-21
Unable to scroll tables resulting in missing information


sympatheticmoose:
We appear to have lost the ability to scroll vertically or horizontally when viewing i.e. an application detail view.

Testing using main on MacOS with chrome

PR #1790, opened 2022-03-23
Allow overflowed data to scroll

joshri:
Closes: #1761 

Fixes various spacing issues, and removes overflow: hidden from the body to allow overflowed pages to scroll. Adjusts nav and content to match.

### DIFF
--- a/ui/components/AutomationsTable.tsx
+++ b/ui/components/AutomationsTable.tsx
@@ -46,23 +46,23 @@ function AutomationsTable({ className, automations, hideSource }: Props) {
         );
       },
       sortValue: ({ name }) => name,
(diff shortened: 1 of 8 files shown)


joshri:
There is a design question here @bia: Do we want the page content to actually fill the page? Or be as tall as it needs to be? It helped me get the nav to match to have everything fill the page but it's different from what it was before. Here's a screenshot: 
image

bia:
What is the behaviour here? Columns scroll horizontally?
It would really make sense for rows to get taller to display the full text without scrolling if there are only two rows.

jpellizzari (approved):

PR #1793, opened 2022-03-23
Added static tls files to helm template

### LINKED ISSUE #1703: Users should be able to set their own TLS certs
**Short description**
Currently TLS is only really usable in the dev env: the certs are built into the image.

For the actual release, we do not want to build certs into the image.

Users should be able to set their own certs.

Users should also have a nice way to enable/disable TLS without editing the Chart (which they can't really anyway unless they clone it).

Using a [configMap as a volume](https://kubernetes.io/docs/concepts/storage/volumes/#configmap) (or secret) may be a nice and easy way to get cert values into the pod, but you don't have to do this if you find something better.

**Acceptance criteria**
As a User I can:
- [ ] Generate my own certs
- [ ] Add them to a config map/pass them to my wego Chart values in some way/ add them to my cluster by some other method
- [ ] And see that wego is running over HTTPS

If I don't want to use TLS:
- [ ] I can set some value in my Chart values
- [ ] And see that my wego app is served over HTTP

josecordaz:
Closes #1703 

**What changed?**
Added a single way to enable TLS on the gitops server. This will be done by specifying the name of a TLS secret in the helm values:
```
...
serverTLS:
  enable: true
  secretName: "my-secret-tls"
...
```

**Why?**
- To enable TLS on the gitops server.

**How did you test it?**
- Manually using a customer TLS secret
- Manually using mkcert files

**Release notes**

**Documentation Changes**


josecordaz:
I wonder if there could be any better way than these to define the tls values:
```yaml
...
serverTLS:
  # This block is meant to use a secret as a source for the tls info
  useSecret: true # Use secret to pass in tls info to the gitops server
  secretName: "mysecret" # Specify the secret name that contains certificate and private key info
  certContentField: "cert" # Specify the key within the secret that contains the certificate
  keyContentField: "key" # Specify the key within the secret that contains the private key
  # This block is meant to pass in local files directly to the gitops server
  useLocalFiles: false # Use local files to pass in tls info to the gitops server
  localCertFile: "" # Specify the path for the local certificate file (it needs to be relative  to the charts folder)
  localKeyFile: "" # Specify the path for the local private key file (it needs to be relative  to the charts folder)
...
```

Callisto13:
Just an FYI, this:

>Closes:
> - https://github.com/weaveworks/weave-gitops/issues/1703

Won't link the issue or trigger Github's auto-closer thingy. This is mainly the fault of our PR template which has a `:` after `Closes`. `Closes #1703` (no punctuation, no bullets, all one line) will do the trick.

Callisto13:
> 2. Using local files. On this option we need to specify the name of the local certificate file(localCertFile) and the local private key file (localKeyFile). Both paths should be relative to the chart folder.

Question: Is option 2 only there for dev purposes? IIUC, users will not be cloning our repo so there will be no "relative to the chart folder".

charts/gitops-server/templates/deployment.yaml
```diff
@@ -36,6 +40,21 @@ spec:
...
+            {{- end }}
+            {{- if .Values.serverTLS.useLocalFiles  }}
+            - "--tls-cert-file"
+            - "/etc/tls-volume/.cert-file"
+            - "--tls-private-key-file"
+            - "/etc/tls-volume/.key-file"
+            {{- end }}
+            {{- if (not (and .Values.serverTLS.useLocalFiles .Values.serverTLS.useSecret) )}}
+            - "--insecure"
+            {{- end }}
```
Callisto13, line 57: I feel this goes against most cert/sec UX patterns. With this logic, if they forget to set TLS files, it will quietly default to insecure. I believe that the majority of tools will default to secure (and complain if you don't have certs) and require explicitly mentioning that you want it to be insecure. (`curl`'s `-k` flag comes to mind). WDYT @SamLR?

charts/gitops-server/templates/deployment.yaml
```diff
@@ -65,6 +84,30 @@ spec:
...
+          volumeMounts:
+            - name: tls-volume
+              readOnly: true
+              mountPath: "/etc/tls-volume"
+          {{- end }}
+      {{- if .Values.serverTLS.useSecret }}
+      volumes:
+        - name: tls-volume
+          secret:
+            secretName: {{ .Values.serverTLS.secretName }}
```
Callisto13, line 97: can we add some docs for how that secret should look? and also include an example template?

charts/gitops-server/values.yaml
```diff
@@ -108,3 +108,14 @@ nodeSelector: {}
 tolerations: []
 
 affinity: {}
+
+serverTLS:
+  # This block is meant to use a secret as a source for the tls info
```
SamLR, line 113: I think we should only offer one method for providing certs and it should probably be by creating a secret.

charts/gitops-server/templates/deployment.yaml
```diff
@@ -36,6 +40,21 @@ spec:
             - "{{ .Release.Namespace }}"
             - "--log-level"
             - "{{ .Values.logLevel }}"
+            {{- if .Values.serverTLS.useSecret }}
+            - "--tls-cert-file"
+            - "/etc/tls-volume/{{ .Values.serverTLS.certContentField }}"
+            - "--tls-private-key-file"
+            - "/etc/tls-volume/{{ .Values.serverTLS.keyContentField }}"
+            {{- end }}
+            {{- if .Values.serverTLS.useLocalFiles  }}
```
SamLR, line 49: This should be `{{- else if....` otherwise it will be possible to set multiple `--tls-private-key-file` flags and `--insecure`

charts/gitops-server/templates/deployment.yaml
SamLR: k8s has an `tls` secret type which we should use https://kubernetes.io/docs/concepts/configuration/secret/#tls-secrets

SamLR (changes requested):

SamLR:
I think there may be a final piece needed here to update the dev env to generate a self signed cert and inject it as a secret

josecordaz:
> > 2. Using local files. On this option we need to specify the name of the local certificate file(localCertFile) and the local private key file (localKeyFile). Both paths should be relative to the chart folder.
> 
> Question: Is option 2 only there for dev purposes? IIUC, users will not be cloning our repo so there will be no "relative to the chart folder". And if we possibly move the Chart to its own repo, which is a possibility, it becomes less useful in a dev context too. Have I missed something?

Option 2 wasn't originally meant for development purposes. I was mostly thinking that a user might want to use local files to pass them in at installation time. But after deeper look and investigation, I've found that this is not possible when using `helm install ...`:

> There is no way to pass files external to the chart during helm install. So if you are asking users to supply data, it must be loaded using helm install -f or helm install --set

[Reference](https://helm.sh/docs/chart_template_guide/accessing_files/)

At this point I see two options:
1) Change the way I'm passing in the file's content to something similar to this: `helm install my-gitops-server repo/gitops-server --set serverTLS.useLocalFiles=true  --set serverTLS.certificateContent="$(cat localhost.pem)" --set serverTLS.privateKeyContent="$(cat localhost-key.pem)"` (this seems an easy task)
2) Get rid of this option completely (if it doesn't seem useful)

@Callisto13  What would be your thoughts on this?

charts/gitops-server/templates/deployment.yaml
josecordaz: Fair point. I'll not default to `insecure` given the reasons you give.

charts/gitops-server/templates/deployment.yaml
josecordaz: working on this now...

josecordaz: @SamLR @Callisto13  I wonder if we should replace the NOT TLS secret type approach (option 1) with the TLS secret type instead? What do you guys think will be the most likely scenario? Will it be worth supporting both approaches?

charts/gitops-server/templates/deployment.yaml
```diff
@@ -11,10 +11,14 @@ spec:
       {{- include "chart.selectorLabels" . | nindent 6 }}
   template:
     metadata:
-      {{- with .Values.podAnnotations }}
       annotations:
-        {{- toYaml . | nindent 8 }}
-      {{- end }}
+        {{- if .Values.serverTLS.useLocalFiles }}
+        tls-cert: {{.Files.Get .Values.serverTLS.localCertFile | quote }}
```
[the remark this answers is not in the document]
josecordaz: The reason behind this is that the downward API can save annotations as files. So we can easily pass them into the gitops server using the tls file flags (`--tls-cert-file`, `--tls-private-key-file`).  Like follows. Reference to the downwardAPI [here](https://kubernetes.io/docs/tasks/inject-data-application/downward-api-volume-expose-pod-information/#the-downward-api).
```
args:
...
            {{- if .Values.serverTLS.useSecret }}
            - "--tls-cert-file"
            - "/etc/tls-volume/.cert-file"
            - "--tls-private-key-file"
            - "/etc/tls-volume/.key-file"
```

charts/gitops-server/templates/deployment.yaml
Callisto13: I think having one approach is sufficient, the TLS secret type feels like the winner to me

Callisto13:
I would go for opt 2: just have the one way to set TLS certs (the TLS secret way)

charts/gitops-server/templates/deployment.yaml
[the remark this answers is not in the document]
SamLR: [Feel free to ignore this if you don't have time to answer, this is for my own curiosity & understanding]

Why do you need to use `metadata.annotations['tls-cert']` and the downwardAPI to template the value `{{.Files.Get .Values.serverTLS.localCertFile | quote }}` into the volume's `fieldpath`? Can't you remove this section and change 97-107 to:
```yaml
      {{- if .Values.serverTLS.useLocalFiles -}}
      volumes:
        - name: tls-volume
          downwardAPI:
            items:
              - path: ".cert-file"
                fieldRef:
                  fieldPath: {{ .Files.Get .Values.serverTLS.localCertFile | quote }}
              - path: ".key-file"
                fieldRef:
                  fieldPath: {{ .Files.Get .Values.serverTLS.localKeyFile | quote }}
      {{ end }}
```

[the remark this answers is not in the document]
josecordaz: Ha! I would think that this should also work. But after testing it I got this error:

```
...ICATE-----": error converting fieldPath: field label not supported: -----B...
```

charts/gitops-server/templates/deployment.yaml
```diff
@@ -11,10 +11,14 @@ spec:
       {{- include "chart.selectorLabels" . | nindent 6 }}
   template:
     metadata:
-      {{- with .Values.podAnnotations }}
```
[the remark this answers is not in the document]
josecordaz: This will no longer be a problem. We are not going to support passing in files contents. Only TLS secrets. I'll leave this block as it originally was.

josecordaz:
> I would go for opt 2: just have the one way to set TLS certs (the TLS secret way)

This is an even better approach because we can use files to create the TLS secret 🙂 , which is great because then we don't need to rely on helm to pass in the info from those files 🎉. 

```
kubectl create secret tls my-tls-secret \
  --cert=path/to/cert/file \
  --key=path/to/key/file
```

Callisto13:
Great work! Love that the Helm course is already paying off 🎉 

I just have a couple of final general questions:
- The `enable` Helm setting does not seem to correspond to the server `--insecure` flag, are we planning to drop the insecure or linking them up somehow?
- Could you also update the `tools/helm-values-dev.yaml` or the `Tiltfile` so we can disable this in dev?

SamLR (approved):
looks great : D

Callisto13 (approved):

PR #1797, opened 2022-03-24
Add a date prefix to the docker tag for PR images

SamLR:
We want to use Flux's image policies to auto deploy the latest version
of main to a staging cluster. To do this the images need to be tagged
in a sortable manner. The easiest way to do this is using the unix time
so we'll add that as a prefix.

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -164,7 +164,7 @@ jobs:
       - name: Push images to gcloud
         run: |
           docker load --input /tmp/${{ matrix.docker-image }}.tar
-          docker push "${{ env.CI_CONTAINER_REPOSITORY }}/${{ matrix.docker-image }}:${{ github.sha }}"
+          docker push "${{ env.CI_CONTAINER_REPOSITORY }}/${{ matrix.docker-image }}:${date +%s}-${{ github.sha }}"
 
   ci-upload-binary:
     name: Upload Binary
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -164,7 +164,7 @@ jobs:
       - name: Push images to gcloud
         run: |
           docker load --input /tmp/${{ matrix.docker-image }}.tar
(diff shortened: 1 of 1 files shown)


.github/workflows/pr.yaml
```diff
@@ -164,7 +164,7 @@ jobs:
       - name: Push images to gcloud
         run: |
           docker load --input /tmp/${{ matrix.docker-image }}.tar
-          docker push "${{ env.CI_CONTAINER_REPOSITORY }}/${{ matrix.docker-image }}:${{ github.sha }}"
+          docker push "${{ env.CI_CONTAINER_REPOSITORY }}/${{ matrix.docker-image }}:${date +%s}-${{ github.sha }}"
```
dhwthompson, line 167:
```suggestion
          docker push "${{ env.CI_CONTAINER_REPOSITORY }}/${{ matrix.docker-image }}:${date -u +%s}-${{ github.sha }}"
```

Is it worth being explicit about using the UTC timestamp here?

dhwthompson (approved):
Looks good to me. I briefly wondered whether we were going to run into problems when the number of digits in the unix timestamp increased, but then did some calculations and figured out that won’t be problem for another 264 years, so I think we’re good.

I do wonder whether it would be more informative to use the ISO 8601 timestamp in there, but I realise that would make things more complicated, and we can always reconstruct the date if we need to. Your call.

yitsushi (approved):

SamLR:
> Looks good to me. I briefly wondered whether we were going to run into problems when the number of digits in the unix timestamp increased, but then did some calculations and figured out that won’t be problem for another 264 years, so I think we’re good.

phew :)

> I do wonder whether it would be more informative to use the ISO 8601 timestamp in there, but I realise that would make things more complicated, and we can always reconstruct the date if we need to. Your call.

I thought about it but I think if you're after metadata on the deployed image the git sha is likely to be more useful in most circumstances and these are mostly just to be deployed to staging so not really intended for general use.

Can always refine it if need be (also this was the call recommend in the flux docs ;) )

PR #1804, opened 2022-03-25
feat: Add over-engineered reboot script

Callisto13:
_If nobody wants this, I don't care about merging, I was just having fun._

Well this one got away from me.

I have paired with a couple of people recently and given them my "turn
it off and on again script" and they were suitably impressed that I
thought I would refactor that script and bequeath it to the masses.

As what often happens when I write bash, it went a bit overboard. So now
I present to you: The Complete CLI experience:

```
Usage: ./tools/reboot.sh <COMMAND> <OPTIONS>

Script to tear down and reprovision a dev environment for weave-gitops.

WARN: will destroy and recreate the kind cluster.

Note: this script is deliberately bare-bones. It will not be extended.

ENV VARS:
  GITHUB_TOKEN              With repo scope. Required for `bootstrap` command.
  KIND_CLUSTER_NAME         Set to configure created kind cluster name. default: wego-dev

COMMANDS:
  install                  Install flux to the kind cluster. This is the default action. If you run `reboot` without any commands, install will just run.
    OPTIONS:
      --namespace, -n      Namespace to install flux to. default: flux-system

  bootstrap                Bootstrap flux in the kind cluster and connect to a given gitops repo
    OPTIONS:
      --git-provider, -p   Git provider. default: github
      --owner, -o          Owner of gitops repo
      --repo, -r           Name of gitops repo
```

If left untouched, the default command `./tools/reboot.sh` will:
- Delete the `wego-dev` kind cluster
- Create the `wego-dev` kind cluster
- Install flux to the `flux-system` namespace

To bootstrap the cluster with flux and a gitops repo, run
`./tools/reboot.sh` with the correct options.

_Lol yes it is more CLI experience that actual stuff happening._

### DIFF
--- a/tools/reboot.sh
+++ b/tools/reboot.sh
@@ -1,0 +1,202 @@
+#!/bin/bash
+
+# This is a deliberately opinionated script for developing Weave Gitops.
+# To get started run ./tools/reboot.sh --help
+#
+# WARN: This script is designed to be "turn it off and on again". It will delete
+# the given kind cluster (if it exists) and recreate, installing everything from
+# scratch.
+
+export KIND_CLUSTER_NAME=${KIND_CLUSTER_NAME:-wego-dev}
+NAMESPACE=flux-system
+
+do_kind() {
+        kind delete cluster --name "$KIND_CLUSTER_NAME"
+
+        "$(dirname "$0")/kind-with-registry.sh" || exit 1
+}
+
+do_bootstrap(){
+        local provider="$1"
+        local owner="$2"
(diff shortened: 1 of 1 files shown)


sbernheim:
Interesting @Callisto13 - I've done a couple different versions of this script to install flux1, flux2, wkp and wego versions for various purposes.  Mine are probably ridiculously over-engineered as well.

I'll check your installer out and see how close we landed to each other.  We should compare notes sometime!

luizbafilho (approved):

Issue #1796, opened 2022-03-24
UI : Reduce padding of menu bar


bia:
Padding of the menu bar should be reduced so that in total the blue part is `80px` high.

The Weave GitOps logo should be `56px` high so that it has `12px` of vertical space on either side within the menubar.

<img width="831" alt="Screenshot 2022-03-24 at 11 16 00" src="https://user-images.githubusercontent.com/92439/159897753-f22cd43d-f6cc-4efc-bdd6-eed27cbd108b.png">

PR #1805, opened 2022-03-25
Remove title bar padding and shrink logo

joshri:
Closes: #1796

Top title bar is now the correct height of 80px. Also aligns breadcrumbs and content title 

image


bia:
LGTM.
While we're here, spacing between the sidebar categories should  be as follows : 

Applications
`12px`
Sources
`32px` (not `24px` as in the screenshot)
Flux Runtime
`32px` 
Docs

Screenshot 2022-03-28 at 12 45 53

bia (approved):

Issue #1777, opened 2022-03-22
Handle Failures and Parallelization of calls to the Clusters


luizbafilho:
Currently, we loop through all clients to all clusters when listing resources https://github.com/weaveworks/weave-gitops/blob/main/core/server/kustomization.go#L25

Given this is done serially it's not very efficient and should be done in parallel to optimize execution.

While doing the parallelization work we should also make sure a failure while calling a single cluster does not exit out the call and keep going until all clusters have been called.

PR #1814, opened 2022-03-28
Clusters Client

luizbafilho:
Closes: https://github.com/weaveworks/weave-gitops/issues/1777

**What changed?**
Introduces the Clusters Client, this is an abstraction to help make calls to all the clusters. It is a thin wrapper around the `controller-runtime/client` that handles the connections to the clusters and when contacting them does it in parallel.

**Why?**
- Simplify resources updates and listing

**How did you test it?**
- automated and manual tests

**Release notes**

**Documentation Changes**


jpellizzari (approved):
LGTM

core/clustersmngr/client.go
```diff
@@ -0,0 +1,168 @@
...
+	"context"
+	"errors"
+	"fmt"
+	"sync"
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+var (
+	ErrClusterNotFound error = errors.New("cluster not found")
```
Callisto13, line 13: can we turn this into an error type that lets us pass in the cluster name? might be useful for debugging

core/logger/logger_test.go
```diff
@@ -157,7 +157,7 @@ func TestBuildConfigWithOptions(t *testing.T) {
 }
 
 var (
-	rfc3339Regexp = `^((\d{4}-\d{2}-\d{2})T(\d{2}:\d{2}:\d{2}(?:\.\d+)?))(Z|[\+-]\d{2}:\d{2})$`
+	rfc3339Regexp = `^((\d{4}-\d{2}-\d{2})T(\d{2}:\d{2}:\d{2}(?:\.\d+)?))(Z|[\+-]\d{2}:?\d{2})$`
```
Callisto13, line 160: i think balazs fixed this in his last pr, check on main?

Callisto13 (commented):
like this a lot! great work! just a couple of questions/requests

core/clustersmngr/client.go
```diff
@@ -0,0 +1,168 @@
...
+				errs = append(errs, fmt.Errorf("cluster=\"%s\" err=\"%s\"", clusterName, err))
+			}
+		}(clusterName, c)
+	}
+
+	wg.Wait()
+
+	if len(errs) > 0 {
+		return fmt.Errorf("failed to list resources: %s", errs)
+	}
```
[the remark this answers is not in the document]
luizbafilho: I made myself the same question, and I don't think the end goal would be this, but we need to coordinate with the UI to settle down the best way to handle this "partial success" response.

core/logger/logger_test.go
luizbafilho: just pulled from master

core/clustersmngr/client.go
```diff
@@ -0,0 +1,171 @@
...
+		pool: clientsPool,
+	}
+}
+
+func (c *clustersClient) ClientsPool() ClientsPool {
+	return c.pool
+}
+
+func (c *clustersClient) Get(ctx context.Context, cluster string, key client.ObjectKey, obj client.Object) error {
+	client := c.pool.Clients()[cluster]
```
yitsushi, line 57: that `c.pool.Clients()[cluster]` hurts my eyes a bit, totally personal taste. I know it does not fail, but my brain constantly says "it should fail if the key is not defined".

```go
func (cp *clientsPool) Client(name string) (client.Client, error) {
  if c, found := cp.clients[name]; found && c != nil {
    return c, nil
  }

  return ClusterNotFoundError{Cluster: name}
}

client, err := c.pool.Client(cluster)
if err != nil {
  return err
}
```

Not a blocker at all. That's only an extra sugar.

luizbafilho: agree, much better, pushing the change

Callisto13 (approved):

Issue #1812, opened 2022-03-25
Fix Docker builds in the release workflow


dhwthompson:
Running the [`release` workflow against `v0.7.0-rc2`](https://github.com/weaveworks/weave-gitops/runs/5695928485?check_suite_focus=true), the following error crops up in the “Build and push Docker image” stage:

```
buildx failed with: error: failed to solve: failed to solve with frontend dockerfile.v0: failed to read dockerfile: open /var/lib/docker/tmp/buildkit-mount530813440/Dockerfile: no such file or directory
```

jpellizzari:
Closed by #1820

PR #1820, opened 2022-03-28
Specify docker file in release job

jpellizzari:
The gitops-server dockerfile name has been changed, and the release job didn't get changed to reflect that


dhwthompson (approved):

dhwthompson:
Looks like this will fix up #1812 nicely.

Issue #1708, opened 2022-03-14
UI: filter options appear within table


bia:
Currently the filter is implemented as an overlay floating above the table, but it would fit better with material design styling if our filters were a [side sheet](https://material.io/components/sheets-side#anatomy)
![sidesheet](https://user-images.githubusercontent.com/92439/158211686-137d3d26-bf16-4d0c-90e1-3ab877010ab1.png)

Compressing the table as in my original design might be problematic, it might be better to keep the filter menu content hovering above and horizontally scroll the table below : 

https://user-images.githubusercontent.com/92439/158212198-2f4a876e-0905-42f2-b189-911c31d042bf.mov

jpellizzari:
This appears to require an upgrade to `v5` of Material UI. I cannot find the `SideSheet` component in `v4`.

PR #1821, opened 2022-03-28
Restyle Filter Modal

joshri:
Closes: #1708

BEFORE:
image
AFTER:
image

Gets the filter modal pretty close to what we want. Here's what's missing:
- ability to scroll data table horizontally when modal is open / shrink table (from #1708) 
- should filter modal always take up the entire content container? It probably needs SOME sort of max-height so that we can scroll when there are too many filter options.

### DIFF
--- a/ui/components/AutomationsTable.tsx
+++ b/ui/components/AutomationsTable.tsx
@@ -115,13 +115,11 @@ function AutomationsTable({ className, automations, hideSource }: Props) {
   if (hideSource) fields = _.filter(fields, (f) => f.label !== "Source");
 
   return (
-    <div className={className}>
-      <FilterableTable
-        fields={fields}
-        filters={initialFilterState}
-        rows={automations}
-      />
-    </div>
+    <FilterableTable
+      fields={fields}
+      filters={initialFilterState}
+      rows={automations}
+    />
   );
 }
 
--- a/ui/components/AutomationsTable.tsx
+++ b/ui/components/AutomationsTable.tsx
@@ -119,6 +119,7 @@ function AutomationsTable({ className, automations, hideSource }: Props) {
       fields={fields}
       filters={initialFilterState}
       rows={automations}
+      className={className}
     />
   );
 }
--- a/ui/components/DataTable.tsx
+++ b/ui/components/DataTable.tsx
@@ -208,11 +208,9 @@ function UnstyledDataTable({
 
 export const DataTable = styled(UnstyledDataTable)`
   width: 100%;
-<<<<<<< HEAD
-=======
   padding-right: ${(props) => props.theme.spacing.medium};
->>>>>>> 0d674b88 (massive height changes everywhere and almost made it on the filter table in automations table component)
   h2 {
+    padding: 0;
     font-size: 18px;
     font-weight: 600;
     color: ${(props) => props.theme.colors.neutral30};
--- a/ui/components/DataTable.tsx
+++ b/ui/components/DataTable.tsx
@@ -208,6 +208,10 @@ function UnstyledDataTable({
 
 export const DataTable = styled(UnstyledDataTable)`
   width: 100%;
+<<<<<<< HEAD
+=======
+  padding-right: ${(props) => props.theme.spacing.medium};
+>>>>>>> 0d674b88 (massive height changes everywhere and almost made it on the filter table in automations table component)
   h2 {
     font-size: 18px;
     font-weight: 600;
--- a/ui/components/DataTable.tsx
+++ b/ui/components/DataTable.tsx
@@ -215,6 +215,8 @@ export const DataTable = styled(UnstyledDataTable)`
     font-weight: 600;
     color: ${(props) => props.theme.colors.neutral30};
     margin: 0px;
+    text-overflow: ellipsis;
+    white-space: nowrap;
   }
   .MuiTableRow-root {
     transition: background 0.5s ease-in-out;
--- a/ui/components/Flex.tsx
+++ b/ui/components/Flex.tsx
@@ -16,7 +16,7 @@ const Styled = (component) => styled(component)`
   display: flex;
   flex-direction: ${(props) => (props.column ? "column" : "row")};
   align-items: ${(props) => (props.align ? "center" : "start")};
-  ${(props) => props.height && `height: ${props.height};`}
+  ${({ tall }) => tall && `height: 100%`};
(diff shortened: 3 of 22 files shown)


jpellizzari (commented):
I'm not seeing a difference between the current styling and this one. Also, this doesn't look like the screenshot in the original issue?

joshri:
> I'm not seeing a difference between the current styling and this one. Also, this doesn't look like the screenshot in the original issue?

image

Here's the original. What else should change?

jpellizzari:
@joshri Sorry what am I looking at here? The second screenshot doesn't look like the first one you submitted in the PR. Can you edit your PR comment to make it more clear please?

joshri:
> Looks nicer to me! I love that it animates in.
> 
> The font sizes on this page should be either `14px` or `20px`, so table titles should be changed from `18px` to `14px` and the word 'Filter' should be `20px`, not `32px`.
> 
> The right side of the table is cut off for me at 100% zoom level - any idea why?
> 
> The current hover states have strange shapes and alignments : chrome-capture
> 
> It should look like [this](https://material.io/design/interaction/states.html#hover): hover



> Looks nicer to me! I love that it animates in.
> 
> The font sizes on this page should be either `14px` or `20px`, so table titles should be changed from `18px` to `14px` and the word 'Filter' should be `20px`, not `32px`.
> 
> The right side of the table is cut off for me at 100% zoom level - any idea why?
> 
> The current hover states have strange shapes and alignments : chrome-capture
> 
> It should look like [this](https://material.io/design/interaction/states.html#hover): hover

This is just the font sizes in the filter modal thing right? I didn't add in horizontal scrolling on the table so when the modal is open the last column or two of the table will not be visible right now.

bia:
Hmmm - but I mean the space in between the white rectangle and the side of the page. On my laptop it's cut off : 
Screenshot 2022-03-29 at 16 59 26

and on my monitor it's fine : 
Screenshot 2022-03-29 at 16 59 51

joshri:
> Hmmm - but I mean the space in between the white rectangle and the side of the page. On my laptop it's cut off : Screenshot 2022-03-29 at 16 59 26
> 
> and on my monitor it's fine : Screenshot 2022-03-29 at 16 59 51

GOT IT thank you

joshri:
@bia I'm going to create a separate issue for the hover states in order to keep this to just the filter modal if that's alright

JamWils:
Per the design, shouldn't this go all the way to the bottom of the screen @joshri?  Oh you asked that question above, yes it should go down the entire right side of the page.  It looks odd as a half modal / half right side menu.

bia:
@jpellizzari here's how [material design filters](https://material.io/components/sheets-side#anatomy) look - it's more similar to what @joshri has here : 

side-sheet

@JamWils has a point, it should go all the way down to the bottom of the screen.

jpellizzari:
Thar component doesn’t exist in material. It is marked as “research” at the
moment.

ui/components/AutomationsTable.tsx
```diff
@@ -115,13 +115,11 @@ function AutomationsTable({ className, automations, hideSource }: Props) {
   if (hideSource) fields = _.filter(fields, (f) => f.label !== "Source");
 
   return (
-    <div className={className}>
```
jpellizzari, line 118: We lost this `className`, so our component is no longer a `styled-component`. You can pass `className` straight to `FilterableTable` if you want.

ui/components/DataTable.tsx
```diff
@@ -208,11 +208,15 @@ function UnstyledDataTable({
...
 export const DataTable = styled(UnstyledDataTable)`
   width: 100%;
+  padding-right: ${(props) => props.theme.spacing.medium};
   h2 {
+    padding: 0;
     font-size: 18px;
     font-weight: 600;
     color: ${(props) => props.theme.colors.neutral30};
     margin: 0px;
+    text-overflow: ellipsis;
```
jpellizzari, line 218: Which columns are overflowing? There may be a bigger problem happening if we have overflowing column headings.

ui/components/Flex.tsx
```diff
@@ -16,7 +16,7 @@ const Styled = (component) => styled(component)`
   display: flex;
   flex-direction: ${(props) => (props.column ? "column" : "row")};
   align-items: ${(props) => (props.align ? "center" : "start")};
-  ${(props) => props.height && `height: ${props.height};`}
+  ${({ tall }) => tall && `height: 100%`};
```
jpellizzari, line 19: :+1:

ui/components/Layout.tsx
```diff
@@ -61,6 +61,7 @@ const AppContainer = styled.div`
   min-height: 100vh;
   margin: 0 auto;
   padding: 0;
+  overflow-x: hidden;
```
jpellizzari, line 64: :thinking: Not sure about this one. Never scroll horizontally?

ui/lib/theme.ts
```diff
@@ -14,7 +14,7 @@ export const theme: DefaultTheme = {
   fontSizes: {
     huge: "48px",
     extraLarge: "32px",
-    large: "22px",
+    large: "20px",
```
jpellizzari, line 17: Really changing this globally?

ui/components/DataTable.tsx
joshri: I added `nowrap` for the header row so that Last Updated wouldn't wrap - It wasn't ever going to the ellipsis for me but I put it there just in case.

ui/components/Layout.tsx
joshri: It's how I keep the filter bar off the screen - when do we want to scroll horizontally?

ui/lib/theme.ts
joshri: Bianca said higher up font sizes on the page should be either 14px or 20px - seemed like it'd be better to stay consistent than to put in a separate size for 22

bia: Yes, this matches our type scale https://www.figma.com/file/gnodK9xtoOahXBqa6ONsA5/WW-Design-System-%2F-Draft?node-id=4%3A5

ui/components/Layout.tsx
jpellizzari: > when do we want to scroll horizontally?

Whenever we need to? As long as there isn't information that we can't get to :shrug:

jpellizzari (approved):

PR #1828, opened 2022-03-29
website: Replace static file redirect with JS redirect

ozamosi:
It seems the static redirect didn't always work - it worked poorly
locally, and it didn't work at all when you click on the logo, however
it worked when you just opened the URL. I take it there's a react
router sometimes rendering a 404 instead of serving the file.

This introduces a new page, that's known by the routing system, which
seems to fix the problem.


sympatheticmoose:
Tested locally and works well for me, thanks @ozamosi. Will defer to you whether a second/technical review is required 👍

PR #1834, opened 2022-03-29
Add a fallback option to output the latest tag

dhwthompson:
This is available either by running `tag-release current` or just `tag-release` with no arguments. I realized I wanted some way to ask "hey, where are we at right now?", and figured that this was a reasonable way to do it. It also seems more approachable than spitting out an error when no arguments are provided, which feels like a useful property to have in a potentially dangerous script.

### DIFF
--- a/tools/tag-release
+++ b/tools/tag-release
@@ -230,8 +230,10 @@ def _build_parser():
 
     parser.add_argument(
         "action",
-        choices=["promote", "bump", "minor", "patch"],
-        help="the type of release to tag",
+        nargs="?",
+        choices=["promote", "bump", "minor", "patch", "current"],
+        default="current",
+        help="the type of release to tag (or output the current tag)",
     )
 
     return parser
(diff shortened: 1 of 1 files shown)


tools/tag-release
```diff
@@ -230,8 +230,10 @@ def _build_parser():
 
     parser.add_argument(
         "action",
-        choices=["promote", "bump", "minor", "patch"],
-        help="the type of release to tag",
+        nargs="?",
```
dhwthompson, line 233: [Documentation for the cryptically-named `nargs` argument](https://docs.python.org/3/library/argparse.html#nargs).

SamLR (approved):
Looks reasonable. 

I did spot, though, that this requires python3.8 (for the walrus operator `:=`) while the default python version on mac is `3.7` I'm not sure if this needs to downgrade? Apaprently the version of python3 that ubuntu ships seems to be 3.8.1

dhwthompson:
> I did spot, though, that this requires python3.8 (for the walrus operator `:=`) while the default python version on mac is `3.7`

That’s a good thought, although I don’t seem to be seeing quite the same picture of macOS Python versioning as you. I couldn’t find any official documentation, but from a bit of Stack Overflow spelunking and checking on my own machine, it looks like the last few versions have been:

| OS version | Name | Python version |
|--|--|--|
| `10.15` | Catalina | [`3.7.3`](https://apple.stackexchange.com/questions/361840/when-updating-to-macos-catalina-will-the-default-python-installation-get-removed#comment509502_364188) |
| `11` | Big Sur | [`3.8.2`](https://apple.stackexchange.com/a/406703) |
| `12` | Monterey | `3.8.9` |

I don’t know what distribution of macOS versions our Mac-based developers have, but it seems worth the small amount of effort it would take to de-walrus-ify the code, just in case. I’ll make a follow-up PR.

PR #1835, opened 2022-03-29
Convert DirectedGraph d3 logic into class

jpellizzari:
Our `DirectedGraph` was a bit clunky after polling was added. This splits the imperative `d3` logic out from the declarative `react` logic to ensure zoom/scale/pan settings are preserved between polls.


joshri (approved):
it's beautiful

PR #1836, opened 2022-03-29
Generate documentation PRs against `main`

dhwthompson:
When the code and the docs were stored in separate repos, this would happen by default, as the release workflow would [checkout the docs repo using its `main` branch][1], which the `create-pull-request` action would pull in as a default base. Now that we're running this workflow from a release tag (which has no concept of which branch it's on), we have to be explicit about where the documentation updates go.

[1]: https://github.com/weaveworks/weave-gitops/blob/e83c56bb2969a087679a9a3d7e163d2098696e57/.github/workflows/release.yaml#L122-L127

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -145,5 +145,6 @@ jobs:
           author: weave-test-user <weave-test-user@example.com>
           signoff: true
           branch: ${{ env.BRANCH }}
+          base: main
           title: "Documentation updates for ${{ env.GITOPS_VERSION }}"
           body: "Update version references to ${{ env.GITOPS_VERSION }} and create new versioned documentation set."


dhwthompson:
From a previous [failed release](https://github.com/weaveworks/weave-gitops/runs/5727441144?check_suite_focus=true):

    /usr/bin/git symbolic-ref HEAD --short
    fatal: ref HEAD is not a symbolic ref
    /usr/bin/git rev-parse HEAD
    a809033e549154425170e935ab1c1d5e6d307490
    Working base is commit 'a809033e549154425170e935ab1c1d5e6d307490'
    Error: When the repository is checked out on a commit instead of a branch, the 'base' input must be supplied.

ozamosi (approved):

Issue #1781, opened 2022-03-22
Make gitops enterprise compatible with gitops core reloaded


ozamosi:
Enterprise imports several paths from gitops core, and some of those have changed. In order to make them work, we need to:
 * Make an RC release (related to #1685)
 * Upgrade the imported go code in enterprise, and fix all issues.
   * In particular, the [request router in enterprise](https://github.com/weaveworks/weave-gitops-enterprise/blob/7b3e86cc8f710fdbc5a8fd96e39b6e1e7db8bf12/cmd/clusters-service/app/server.go#L289) needs to mux the new endpoints. 
 * Upgrade the imported js code in enterprise, and fix all issues.
   * There's several pieces from core imported straight up - [example](https://github.com/weaveworks/weave-gitops-enterprise/blob/7b3e86cc8f710fdbc5a8fd96e39b6e1e7db8bf12/ui-cra/src/components/Applications/index.tsx#L6)
   * This might involve adapting the core js exports in [index.js](https://github.com/weaveworks/weave-gitops/blob/main/ui/index.ts)

There's an [adr](https://github.com/weaveworks/weave-gitops/blob/main/doc/adr/0007-embedding-weave-gitops-core.md) on how the integration works.

This [thread](https://weaveworks.slack.com/archives/C032Y3LFQBB/p1647970803322789) has more suggestions.

PR #1842, opened 2022-03-30
refactor: Rename Applications -> GitProviderAuth

### LINKED ISSUE #1830: Remove all mentions of the old App Server
This is now basically just a git provider auth server, so we can re name some things, have a general tidy up and get some cohesion between core and the old stuff

Callisto13:
Applications are not a "thing" anymore, and all the old Applications
server does these days is interact with Git Providers.

I have chosen to call this server/handlers GitProviderAuth because:
a) Plain GitProvider is taken (as a package and an enum in this grpc
proto
b) GitAuth may be confused with the other Auth server we have going on
c) I could not think of a better name, suggestions welcome

The bulk of the change is just a rename, but I have also tidied up how
we build config and add handlers.

This is a **breaking change**, so think about whether we want it.

Closes #1830


Callisto13:
Right now this PR is an illustration, we do not need to merge. (We also probably want to wait until after #1781 anyway)

One thing to consider is whether we want this server to live in core at all, given we are not using it. IDK what the plan is for core going forward after observability)

PR #1843, opened 2022-03-30
feat: Use client from pool to Get kustomization

Callisto13:
I have updated the List and Get Kustomization protos.
The List will return each Kustomization with its corresponding
clusterName. This clusterName will then be passed by the UI client into
a subsequent Get request.
The Get will take a clusterName, and the method will use that to find
the correct client from the pool.

Where previously the UI had hardcoded the clusterName into the Dash, now
it pulls it from the List result, and then uses that information for
Gets.

Closes [#1674](https://github.com/weaveworks/weave-gitops/issues/1674)

So far I have only done Kustomizations on the applications table, and the info section of when you click on a thing from that table.
The rest will come later, so don't worry if you see HelmReleases or whatever without a cluster name, that will come next

Screenshot from 2022-03-31 14-57-16
Screenshot from 2022-03-31 14-59-16


core/server/kustomization.go
```diff
@@ -46,18 +46,19 @@ func (cs *coreServer) ListKustomizations(ctx context.Context, msg *pb.ListKustom
...
+	clustersClient := clustersmngr.ClientFromCtx(ctx)
 
 	k := &kustomizev1.Kustomization{}
+	key := client.ObjectKey{
+		Name:      msg.Name,
+		Namespace: msg.Namespace,
+	}
 
-	if err := get(ctx, k8s, msg.Name, msg.Namespace, k); err != nil {
+	if err := clustersClient.Get(ctx, msg.ClusterName, key, k); err != nil {
```
Callisto13, line 57: damn this is just so easy to use ❤️  @luizbafilho

Callisto13:
@joshri could you look at the UI stuff? And let me know if there are any tests I should be adding to? I was basically just typing words.

core/clustersmngr/clustersmngr.go
```diff
@@ -58,7 +58,7 @@ type ClusterFetcher interface {
 type ClientsPool interface {
 	Add(user *auth.UserPrincipal, cluster Cluster) error
 	Clients() map[string]client.Client
-	Client(clsuter string) (client.Client, error)
+	Client(string) (client.Client, error)
```
yitsushi, line 61: Why the removal? I know it had a type, but adding the "variable name" is useful, for example documentation when we just check the interface, or `:GoImpl` can populate the name properly.

(and not consistent as we have variable names on the `Add` function)

ui/App.tsx
```diff
@@ -34,11 +34,14 @@ import Sources from "./pages/v2/Sources";
 
 const queryClient = new QueryClient();
 
-function withName(Cmp) {
+function withParams(Cmp) {
```
yitsushi, line 37: That name feels a bit too broad, technically this function handles only two `params` values.
I'm ok with that and I have no better suggestion.

yitsushi: Actually, now I understand what this function does, it can be `withSeachPrams` to be more specific.

yitsushi (approved):
lgtm, only 2 small suggestions

ui/App.tsx
Callisto13: sound good, will change 👍 
at some point it will also pass in the namespace, which is currently defaulted/hardcoded to "flux-system"

core/clustersmngr/clustersmngr.go
Callisto13: lol i did not look at the `Add`. i first corrected the typo then was like "eh, i can just remove it" 😂

joshri (approved):
🌞  BIG APPROVE 🌞

PR #1844, opened 2022-03-30
Proper error format

### LINKED ISSUE #1717: Random error message instead of being logged out
As a user, when my session is no longer valid I see a random error message (see screenshot below) instead of

a. Automatically being redirected to the login screen. (Preferred situation)
b. Simply seeing an error that my session has expired. (Less ideal)

If you check the console logs the error message first appears after the session expires.  Upon hitting refresh I am then taking to the login screen.

luizbafilho:
Closes #1717 

**What changed?**
- Authentication middleware was sending back a plain text error instead of json, this PR is a quick fix to have it a expected json format on the client side

**Why was this change made?**

**Release notes**

**Documentation Changes**

### DIFF
--- a/pkg/server/auth/server.go
+++ b/pkg/server/auth/server.go
@@ -161,7 +161,7 @@ func (s *AuthServer) oauth2Config(scopes []string) *oauth2.Config {
 func (s *AuthServer) OAuth2Flow() http.HandlerFunc {
 	return func(rw http.ResponseWriter, r *http.Request) {
 		if !s.oidcEnabled() {
-			http.Error(rw, "oidc provider not configured", http.StatusBadRequest)
+			Error(rw, "oidc provider not configured", http.StatusBadRequest)
 			return
 		}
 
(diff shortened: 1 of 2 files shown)


Callisto13 (approved):
👍 :shipit:

pkg/server/auth/server.go
```diff
@@ -490,3 +490,7 @@ func contains(ss []string, s string) bool {
 
 	return false
 }
+
+func Error(rw http.ResponseWriter, err string, code int) {
```
bigkevmcd, line 494: Rather than making the string like this, you could marshal an anonymous struct...

```Go
    response := struct {
        Message string `json:"message"`
    }{Message: "testing"}
    b, err := json.Marshal(response)
```

Also, you should probably set the content-type

bigkevmcd: You aren't controlling the message https://github.com/weaveworks/weave-gitops/pull/1844/files#diff-f96b9f66364af6512f2689b1500deebf882f76fc92556b66bc74f3c40005940aR250 which means that you could get something that didn't simply marshal like this.

pkg/server/auth/server.go
```diff
@@ -491,6 +491,16 @@ func contains(ss []string, s string) bool {
...
+func JSONError(w http.ResponseWriter, errStr string, code int) {
+	w.Header().Set("Content-Type", "application/json; charset=utf-8")
+	w.Header().Set("X-Content-Type-Options", "nosniff")
+	w.WriteHeader(code)
+
+	response := struct {
+		Message string `json:"message"`
+	}{Message: errStr}
+
+	if err := json.NewEncoder(w).Encode(response); err != nil {
```
bigkevmcd, line 503: I think you want to just log the error if you fail to write it to the response.

Usual reason for failing to write it will be client-closed.

bigkevmcd (approved):
Looks better!

PR #1845, opened 2022-03-30
Make the release script compatible with Python 3.7

dhwthompson:
Python's "walrus operator" was [introduced in Python 3.8][1], and while it makes the code a little more elegant in certain cases (specifically, in this case, regex matching), it's better in this case to make sure we preserve compatibility, as some older OS versions bundle Python 3.7.x.

[1]: https://docs.python.org/3/whatsnew/3.8.html#assignment-expressions

### DIFF
--- a/tools/tag-release
+++ b/tools/tag-release
@@ -30,8 +30,10 @@ def parse_tag(tag_name):
         """,
         re.VERBOSE,
     )
-    if tag_match := pattern.match(tag_name):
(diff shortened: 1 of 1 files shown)


SamLR (approved):
hopefully soon the walrus will rise again

dhwthompson:
Goodnight, walrus. Rest well.

A walrus, settling into repose

PR #1846, opened 2022-03-30
V2 restructure

sympatheticmoose:
Closes

**What changed?**

**Why was this change made?**

**Release notes**

**Documentation Changes**


pfcurtis:
My only comment on the restructure and WGE. Instead of installing a CAPI provider, we should install the CAPI management first to allow for the checking of the WGE installation, including the profile and bootstrap controller steps. These two steps should be a part of the enterprise upgrade rather than a configuration step.
http://localhost:3000/docs/next/installation#weave-gitops-enterprise
Specifically, add those two steps between step 5 and 6. They can just be references to the right place rather than re-copying the text.

Once all that is finished, a CAPI provider should be installed. This keeps WGE separate from the CAPI provider. Why? Because WW doesn't own the provider, and we want to ensure all our bits are installed and running properly **first.**

website/docs/configuration/tls.md
```diff
@@ -0,0 +1,48 @@
...
+
+By default the dashboard will listen on 0.0.0.0:9001 with TLS enabled. A self-signed certificate and key pair are generated when the server starts.
+As the certificate is _self-signed_, Chrome and other browser will show a warning you will have to click through to view the dashboard.
+
+| Parameter           | Type   | Description                                                     | Default |
+| ------------------- | ------ | --------------------------------------------------------------- | ------- |
+| `--no-tls`          | bool   | Disable TLS, access the dashboard on default port via http      | false   |
+| `--tls-private-key` | string | Filename for the TLS certficate, in-memory generated if omitted |         |
+| `--tls-cert-file`   | string | Filename for the TLS key, in-memory generated if omitted        |         |
+| `--host`            | string | Host to listen on                                               | 0.0.0.0 |
```
ozamosi, line 16: I'm trying to suggest how to change this, and I'm struggling to write something helpful. This is my current best attempt, but it's both too technical and too vague:

By default the dashboard will listen on port 9001 with TLS disabled and without exposing itself to connections from the outside. If you wish to expose it, you first need to set up TLS. You'll need to open the weave-gitops HelmRelease and add some values to the bottom of it. There's two general options.

Either you're already running an ingress controller that terminates TLS for you, in which case you can configure the weave-gitops helm chart to enable an ingress. The exact parameters you need depends on what ingress you use.
```
  values:
    ingress:
      enable: true
      annotations:
        # kubernetes.io/tls-acme: "true"
      hosts:
        # - host: "weave-gitops.mycompany.com",
        #   paths:
        #     - path: "/" 
```

Or you want the dashboard to do TLS termination itself, in which case you need to first create a secret in your cluster:
```
kubectl create secret tls my-tls-secret \
  --cert=path/to/cert/file \
  --key=path/to/key/file
```
and then configure the weave-gitops helm chart to enable TLS termination:
```
  values:
    serverTLS:
      enable: true
      secretName: "my-tls-secret"
```

Either way, when you're done your helm release should look similar to:
```
---
kind: HelmRelease
metadata:
  name: ww-gitops
  namespace: flux-system
spec:
  chart:
    spec:
      chart: weave-gitops
      sourceRef:
        kind: HelmRepository
        name: ww-gitops
  interval: 1m0s
  values:
    # Either remove serverTLS or ingress:
    serverTLS:
      enabled: true
      # etc
    ingress:
      enabled: true
      # etc
```

website/docs/installation.mdx
```diff
@@ -0,0 +1,520 @@
...
+  chart:
+    spec:
+      chart: weave-gitops
+      sourceRef:
+        kind: HelmRepository
+        name: ww-gitops
+  interval: 1m0s
+  values:
+    additionalArgs:
+    - --insecure
```
ozamosi, line 45: This also isn't needed

website/docs/installation.mdx
```diff
@@ -0,0 +1,520 @@
...
+## Installing the gitops CLI
+
+The `gitops` command line interface provides a set of commands to make it easier to interact with Weave GitOps, including both the free open source project, and commercial Weave GitOps Enterprise product. It is currently supported on Mac (x86 and Arm), and Linux including WSL.
+
+Windows support is a [planned enhancement](https://github.com/weaveworks/weave-gitops/issues/663).
+
+To install the `gitops` CLI, please follow the following steps:
+
+```console
+curl --silent --location "https://github.com/weaveworks/weave-gitops/releases/download/v0.6.2/gitops-$(uname)-$(uname -m).tar.gz" | tar xz -C /tmp
```
ozamosi, line 73: Is this version supposed to be hard-coded? :thinking:

website/docs/installation.mdx
```diff
@@ -0,0 +1,520 @@
...
+```
+
+You should see:
+
+```console
+Current Version: v0.6.2
+GitCommit: cf7b11b8
+BuildTime: 2022-01-20_20:15:15
+Branch: HEAD
+Flux Version: v0.24.1
```
ozamosi, line 85: We don't print Flux Version any more, so this is old.

ozamosi (commented):
I've flipped through the manual and noticed out of date chart things.

I've not run through the manual myself yet, I'll do that next.

ozamosi:
I meant to say, my chart comments aren't wrong in this branch right now. They become outdated when the branch is rebased against latest main.

website/docs/installation.mdx
jrryjcksn: Okay

website/docs/installation.mdx
jrryjcksn: The release process says to update this manually after a successful release

website/docs/installation.mdx
jrryjcksn: This gets updated automatically during the release process:
```
ex - installation.md << EOS
/download\/
%s,download/\(.*\)/,download/${GITOPS_VERSION}/,
/Current Version
.,+4! ${WEAVE_GITOPS_BINARY} version
+5d
wq!
EOS
```
although I notice the installation file name has changed... I'll fix that

website/docs/configuration/tls.md
jrryjcksn: Thanks. Let me see what I can massage it into...

website/docs/getting-started.mdx
```diff
@@ -0,0 +1,455 @@
...
+```
+kubectl create secret generic cluster-user-auth \
+  --namespace flux-system \
+  --from-literal=username=admin \
+  --from-literal=password='$2b$12$FFokfwM/p8RPr0YMTdHWk.MXe3Kg1/UVyj63pT9FU8/mthUjKjm7W'
+```
+
+6. Configure RBAC for the admin user
+
+Still in your local clone of `fleet-infra`, create a new YAML file and add the following:
```
ozamosi, line 127: It's only in the next section we're told to clone `fleet-infra`.

Unlike the next section, this doesn't tell you to put the file in `clusters/my-cluster`

website/docs/getting-started.mdx
```diff
@@ -0,0 +1,455 @@
...
+
+```
+apiVersion: source.toolkit.fluxcd.io/v1beta1
+kind: HelmRepository
+metadata:
+  name: ww-gitops
+  namespace: flux-system
+spec:
+  interval: 1m0s
+  url:
```
ozamosi, line 206: This ought to be `url: https://helm.gitops.weave.works`

website/docs/getting-started.mdx
```diff
@@ -0,0 +1,455 @@
...
+![Applications details view for podinfo showing 2 pods](/img/dashboard-podinfo-details.png)
+
+### Customize podinfo
+
+To customize a deployment from a repository you don’t control, you can use Flux in-line patches. The following example shows how to use in-line patches to change the podinfo deployment.
+
+1. Add the following to the field spec of your podinfo-kustomization.yaml file:
+
+```
+patches:
```
ozamosi, line 406: This line should have 2 leading spaces, so this block can just be copy-pasted into the file. The rest of the block is already appropriately indented.

website/docs/configuration/securing-access-to-the-dashboard.mdx
```diff
@@ -54,16 +11,20 @@ There are 2 supported methods for logging in to the dashboard:
 
 The recommended method is to integrate with an OIDC provider, as this will let you control permissions for existing users and groups that have already been configured to use OIDC. However, it is also possible to use a cluster user account to login, if an OIDC provider is not available to use. Both methods work with standard Kubernetes RBAC.
 
-:::note FEATURE TOGGLE 
+:::note FEATURE TOGGLE
```
ozamosi, line 14: This isn't behind a feature flag.

In fact, this page is a bit funny now that half of this has been moved to the quickstart. I wonder if this should be re-written to be an "Enabling SSO" page :thinking:

website/docs/configuration/securing-access-to-the-dashboard.mdx
```diff
@@ -128,15 +122,15 @@ b'$2b$12$nLfl7lKBiYzgAN2aI3ii6exZSZ9KRsj18C7CEWY8kscj9.c6bRXim'
...
+kubectl create secret generic cluster-user-auth \
+  --namespace flux-system \
+  --from-literal=username=admin \
   --from-literal=password='$2b$12$nLfl7lKBiYzgAN2aI3ii6exZSZ9KRsj18C7CEWY8kscj9.c6bRXim'
 ```
 
 You should now be able to login via the cluster user account using your chosen username and password. Follow the instructions in the next section in order to configure RBAC correctly.
 
-### RBAC authorization
+## RBAC authorization
```
ozamosi, line 133: This whole section needs updating.

I'd replace the whole thing by only the role we created in the quickstart in point 5 of https://staging.docs.gitops.weave.works/v2-restructure/docs/next/getting-started#configure-access-to-the-dashboard

ozamosi (approved):
I got the core quickstart working all the way, with only small bits I found surprising.

I think the bits I've found can be addressed in a follow-up, so hitting approve.

PR #1847, opened 2022-03-30
update-upgrade-command-to-deal-with-weave-gitops-helm-repo

jrryjcksn:
**What changed?**
With this change we can upgrade from a core installation pulled from a helm repository.

**Why was this change made?**
We want to install core from a helm repository.

The change was implemented conservatively by allowing either a GitRepository or a HelmRepository to be specified.

* Unit tests
* Manual tests

**Release notes**
No

**Documentation Changes**
No


sympatheticmoose:
Apologies @jrryjcksn given its my last day I'm unlikely able to give sufficient review, from a quick glance - do you think we need to support both git/helm repositories? I would expect we're only advertising the helm repository mechanism. @yiannistri - FYI for enterprise

jrryjcksn:
I wasn't sure and it was easy to do this. If we're sure we don't need the git repo, I'm happy to prune it back.

ozamosi (approved):

PR #1850, opened 2022-03-31
feat: Use clientpool to Get and List HelmReleases

### LINKED ISSUE #1673: Update core helmrelease endpoints to use client pool
**Short description**
Following on from https://github.com/weaveworks/weave-gitops/issues/1600 and based on https://github.com/weaveworks/weave-gitops/pull/1556, this is to update the endpoints in `core/server/helmrelease` to:
- Pull a list of clients from the `ctx`
- Loop over those and query each

**Acceptance criteria**
Everything which currently works should still work

_I am off to make dinner now, feel free to flesh out if you want more._

Callisto13:
Same as #1843, I have
- Updated the List and Get HelmRelease protos.
- The List will return each HelmRelease with its corresponding
clusterName.
- This clusterName will then be passed by the UI client into
a subsequent Get request.
- The Get will take a clusterName, and the method will use that to find
the correct client from the pool.

Where previously the UI had hardcoded the clusterName into the Dash, now
it pulls it from the List result, and then uses that information for
Gets. (Note that there might currently be a separate issue with this where we are getting
the Chart not the Release? idk)

Closes #1673

Again this is just the tables that the List and Get HelmReleases methods are called.
So the only place this impacts is the Applications table and the info section of a Helm
Release if you click into one from there


yitsushi (approved):

joshri (approved):
🌞  BIG APPROVE 🌞

PR #1852, opened 2022-03-31
fix: Ordering within filter menu changes periodically

### LINKED ISSUE #1759: Ordering within filter menu changes periodically
https://user-images.githubusercontent.com/19860021/159256313-6a7360b4-d67b-49a6-a03e-8a49e5e1e1a6.mov

This can be pretty jarring and strange

yitsushi:
Approach: Sort filter options alphabetical

I don't know if react will reload things or not if the `options` list is
changed or not, so went with a immutable option using
`[...options].sort()`

Closes #1759

### DIFF
--- a/ui/components/FilterDialog.tsx
+++ b/ui/components/FilterDialog.tsx
@@ -120,7 +120,7 @@ function UnstyledFilterDialog({
                             </Text>
                             <List>
                               {_.map(
-                                [...options].sort(),
+                                _.sortBy(options),
                                 (option: string, index: number) => {
                                   return (
                                     <ListItem key={index}>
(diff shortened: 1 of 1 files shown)


jpellizzari (approved):
:+1: Nice work on the immutable pattern. I had to look up if [`.sort()` mtuates](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort):

I tend to use `lodash` for stuff like this, as it will handle `undefined` and usually returns a new array: https://lodash.com/docs/4.17.15#sortBy.

One issue with this would be if `options` was `null` or `undefined`; the `...` operator would cause an explosion. Very low likelihood for that in this case, however.

yitsushi:
Updated to use `_.sortBy()`

PR #1854, opened 2022-03-31
feat: Use clientpool to Get flux runtime objects

### LINKED ISSUE #1672: Update core flux runtime endpoints to use client pool
**Short description**
Following on from https://github.com/weaveworks/weave-gitops/issues/1600 and based on https://github.com/weaveworks/weave-gitops/pull/1556, this is to update the endpoints in `core/server/fluxruntime` to:
- Pull a list of clients from the `ctx`
- Loop over those and query each

**Acceptance criteria**
Everything which currently works should still work

Callisto13:
Following on from #1850 and #1843, I have:
- Updated the various Flux Get/List protos
- Updated the UI to thread through the clusterName from the relevant
  object to the requests
- Used that clusterName to fetch the correct client

Again this is only concerned with the elements of the UI which call
those endpoints, any gaps or missing things you notice will come along
next.

Closes #1672


Callisto13:
i will get those CI failures tomorrow

luizbafilho (approved):

PR #1855, opened 2022-03-31
Implementing dev mode

luizbafilho:
Closes

**What changed?**

**Why was this change made?**

**Release notes**

**Documentation Changes**

### DIFF
--- a/cmd/gitops-server/cmd/cmd.go
+++ b/cmd/gitops-server/cmd/cmd.go
@@ -55,6 +55,8 @@ type Options struct {
 	TLSKeyFile                    string
 	Insecure                      bool
 	MTLS                          bool
+	DevMode                       bool
+	DevUser                       string
 }
 
 var options Options
@@ -89,6 +91,11 @@ func NewCommand() *cobra.Command {
 	cmd.Flags().StringVar(&options.OIDC.ClientSecret, "oidc-client-secret", "", "The client secret to use with OpenID Connect issuer")
 	cmd.Flags().StringVar(&options.OIDC.RedirectURL, "oidc-redirect-url", "", "The OAuth2 redirect URL")
 	cmd.Flags().DurationVar(&options.OIDC.TokenDuration, "oidc-token-duration", time.Hour, "The duration of the ID token. It should be set in the format: number + time unit (s,m,h) e.g., 20m")
+
+	cmd.Flags().BoolVar(&options.DevMode, "dev-mode", true, "Enables development mode")
+	cmd.Flags().StringVar(&options.DevUser, "dev-user", v1alpha1.DefaultClaimsSubject, "Sets development User")
+	cmd.Flags().MarkHidden("dev-mode")
+	cmd.Flags().MarkHidden("dev-user")
 
 	return cmd
 }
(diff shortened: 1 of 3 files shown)


cmd/gitops-server/cmd/cmd.go
```diff
@@ -90,6 +92,9 @@ func NewCommand() *cobra.Command {
 	cmd.Flags().StringVar(&options.OIDC.RedirectURL, "oidc-redirect-url", "", "The OAuth2 redirect URL")
 	cmd.Flags().DurationVar(&options.OIDC.TokenDuration, "oidc-token-duration", time.Hour, "The duration of the ID token. It should be set in the format: number + time unit (s,m,h) e.g., 20m")
 
+	cmd.Flags().BoolVar(&options.DevMode, "dev-mode", true, "Enables development mode")
+	cmd.Flags().StringVar(&options.DevUser, "dev-user", v1alpha1.DefaultClaimsSubject, "Sets development User")
```
Callisto13, line 96: i think we can probably do without letting people set this and just have the dev user be the default

SamLR:
Minor think but could we have this log a warning at start up to make it clear that you shouldn't be running this in dev mode? Something like `WARNING: dev mode enabled. This should be used for local work only`.

cmd/gitops-server/cmd/cmd.go
luizbafilho: The intent here is to allow us to have multiple users we can use to check behaviors. Now we have the admin one, but shortly we'll get the limited one, we might also want to have a middle term semi-limited one for instance.

luizbafilho: there are actually no real users at all, they are all just role binding(https://github.com/weaveworks/weave-gitops/blob/main/charts/gitops-server/templates/rolebinding.yaml#L18), so it's just easier to have another set of role and role binding in place deployed, via a `dev` chart, than override the current one.

Callisto13: that role binding you linked is for the wego-server, did you mean https://github.com/weaveworks/weave-gitops/blob/main/charts/gitops-server/templates/admin-user.yaml#L16 ?

luizbafilho: yep that one.

Callisto13: either way, the k8s user that the claims subject is bound to does not change, that username doesn't turn into anything concrete, so we can't really have more than one user at the same time

Callisto13: users can set [this](https://github.com/weaveworks/weave-gitops/blob/main/charts/gitops-server/templates/admin-user.yaml#L56) but not [this](https://github.com/weaveworks/weave-gitops/blob/main/charts/gitops-server/templates/admin-user.yaml#L16), so there can only ever be one user at a time

luizbafilho: It can change, that's why I added the dev user, it's being set here: https://github.com/weaveworks/weave-gitops/pull/1855/files#diff-f77070871fc35c299f82f2108c7fe3c52e8a113e0fa100fdc03f353d07307770R73

So when making the calls, it will have that as impersonated user. So we can have as many roles/rolesbinding to test different permitions.

Callisto13: ah so we are sort of implementing static users for this mode?

luizbafilho: no really, because we don't have any concept of password and user fetching, we just hardcode the user there and no matter the username and password sent, it will return the `dev-user` and the authenticated one.

As the name implies is for development only.

jpellizzari (approved):

PR #1856, opened 2022-03-31
Make core work to embed in into enterprise

ozamosi:
This does a few things:
* Put some of the new pages into components. That seems more embed friendly.
* Expose _all the things_ in the js library. It's a lot. Maybe I should put a better abstraction into place somewhere, happy to get pointers.
* Put a bunch of question marks in types in places. Given that we send the result of an HTTP request straight to a component, the contents of the response should be optional, right?
* Handle a couple of places where, for some reason, on my system, I had inputs be undefined when the code expected them not to be.


.parcelrc
```diff
@@ -1,6 +1,7 @@
 {
   "extends": "@parcel/config-default",
   "transformers": {
-    "*.{ts,tsx}": ["@parcel/transformer-typescript-tsc"]
+    "*.{ts,tsx}": ["@parcel/transformer-typescript-tsc"],
+    "*.svg": ["...", "@parcel/transformer-svg-react"]
```
jpellizzari, line 5: Curious about this. Were the `svg` causing build issues?

ui/index.ts
```diff
@@ -22,6 +34,12 @@ import OAuthCallback from "./pages/OAuthCallback";
 import SignIn from "./pages/SignIn";
 
 export {
+  SourceRefSourceKind,
```
jpellizzari, line 37: Maybe alphabetize this to group things a bit better?

jpellizzari (approved):
LGTM, only nits. Nice work.

.parcelrc
ozamosi: There were no build issues, but the links to the SVG file that the core code generated when run inside enterprise just pointed at e.g. `file:///success.3e193089.svg` instead of being resolved on the server, so they wouldn't load.

The actual code generated by `make ui-lib` contained
```
var $72fdb1cb9e084dbe$exports = {};
$72fdb1cb9e084dbe$exports = new URL("success.3e193089.svg", "file:" + __filename
).toString();
```
which looks to me like parcel does nothing extra to indicate that the URL needs to be resolved, and react-scripts doesn't think to try to read it from the local directory - and I suspect there's some way to make it work this way cleanly through various faffing around with bundlers, and also I suspect that'd take longer than I expect.

So I just grabbed one of the other approaches in https://parceljs.org/languages/svg/#importing-as-a-react-component

ozamosi:
Because I got an approval this when it was still in draft state: I intend to merge this _now-ish_, because the frontend changes are proving surprisingly susceptible to merge conflicts. I assume there'll still be follow-up PRs before this is completely done.

PR #1857, opened 2022-03-31
Add nsaccess package to determine if a user "has access" to a namespace

### LINKED ISSUE #1725: Return a list of namespaces that a user can access
Part of #1725

We want to show GitOps resources across namespaces. Given how RBAC works in kubernetes, we need to know which namespaces to query.

We determine "access" as a user who has these permissions in a given namespace:

```yaml
apiVersion: rbac.authorization.k8s.io/v1
kind: Role
metadata:
  name: wego-admin-user-role
  namespace: {{.Values.namespace}}
rules:
  - apiGroups: [""]
    resources: ["secrets", "pods","events", "namespaces"]
    verbs: [ "get", "list" ]
  - apiGroups: ["apps"]
    resources: [ "deployments", "replicasets", "pods" ]
    verbs: [ "get", "list" ]
  - apiGroups: ["kustomize.toolkit.fluxcd.io"]
    resources: [ "kustomizations" ]
    verbs: [ "get", "list" ]
    resources: [ "helmreleases" ]
    verbs: [ "get", "list" ]
  - apiGroups: ["source.toolkit.fluxcd.io"]
    resources: [ "buckets", "helmcharts", "gitrepositories", "helmrepositories" ]
    verbs: [ "get", "list" ]
```

Acceptance Criteria:
* Some function or `interface` needs to return a list of namespaces that represent "access" for a given user

Reference PoC PR: https://github.com/weaveworks/weave-gitops/pull/1715

jpellizzari:
Closes #1725

The goal of this work was to be able to provide a list of Kubernetes namespaces that a user can access. "Access" in this context, is determined by a minimum set of actions a user can take. In the default case, those actions map directly to the things that the `wego-app` can do.

Some thoughts:
* This feels wrong: is there not already some `k8s` machinery that can do this? I looked at lots of different source code, but could not find anything that does exactly this.
* The `hasAllRules` function is very un-`go` like to me, but given how flexible the rules can be, I chose to model this as a string-matching problem. Any input on that section would be appreciated.

CC @JamWils This is the brute force approach we had discussed, with an opening for a different (better?) implementation in the future.


core/server/kustomization_test.go
```diff
@@ -132,11 +132,11 @@ func TestGetKustomization(t *testing.T) {
...
 
-func newNamespace(ctx context.Context, k client.Client, g *GomegaWithT) *corev1.Namespace {
+func newNamespace(ctx context.Context, k client.Client, g *GomegaWithT) corev1.Namespace {
 	ns := &corev1.Namespace{}
 	ns.Name = "kube-test-" + rand.String(5)
 
 	g.Expect(k.Create(ctx, ns)).To(Succeed())
 
-	return ns
+	return *ns
```
yitsushi, line 141: This can be
```go
func newNamespace(ctx context.Context, k client.Client, g *GomegaWithT) corev1.Namespace {
	ns := corev1.Namespace{}
	ns.Name = "kube-test-" + rand.String(5)

	g.Expect(k.Create(ctx, &ns)).To(Succeed())

	return ns
}
```
It's kind of the same, but looks less weird with `return *ns`.

yitsushi (commented):
a small nit, otherwise lgtm

(`hasAllRules` is a bit chaotic)

core/nsaccess/nsaccess_test.go
```diff
@@ -0,0 +1,365 @@
...
+	adminClient, err := client.New(testCfg, client.Options{
+		Scheme: kube.CreateScheme(),
+	})
+
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	createRole(t, adminClient, roleName, rules)
+	secretName := createKubeConfig(t, adminClient, testCfg)
```
luizbafilho, line 349: we can simplify the user config creation by just:
```go
userCfg := *testCfg

userCfg.Impersonate = rest.ImpersonationConfig{
	UserName: userName,
}
```

luizbafilho (approved):
Minor comment to help out clean up the tests

jpellizzari:
TODO:

```
	t.Run("works when a user has * permissions on a resource", func(t *testing.T) {

	})
```

Issue #1859, opened 2022-04-01
Ensure UI gets clusterName from objects everywhere


Callisto13:
ie: no hardcoded "Default" string anywhere

The final piece of the core work for #1677, tying up lose ends basically

Issue #1861, opened 2022-04-01
No "Cluster" field on HelmChart info page?


Callisto13:
All other source's info pages have a "Cluster" field but not HelmCharts. Is that deliberate or did we miss it?

Issue #1677, opened 2022-03-10
Epic: Multi-Cluster


Callisto13:
This is the breakdown of work. Some tickets can be done in the same PR, some you may want to tackle separately:

POC can be found here: https://github.com/weaveworks/weave-gitops/pull/1556.

Decision write up here: https://docs.google.com/document/d/1kxcyznya57gAW28_jUU4ZQVffAyrSze7Py_y0T_z6jE/edit#

### Context

- Multi-cluster is *not* a feature of core
- So core cannot have any way to add multiple clusters
- So we will be doing the multi bit in Enterprise
- From core's side we need to:
  - Define the clusters fetcher interface
  - Fulfil that interface by adding a single, in-memory cluster to a pool which is added to the context (basically doing what we do now, but putting the client somewhere else).
- From enterprise's side we will:
  - First implement a temporary ConfigMap based fulfilment of the fetcher interface (to unblock anyone waiting on this)
  - Then (when the official CRDs become available) replace the temporary solution and fulfil the fetcher interface using those CRDs.

### Work to be done in Core
- [x] https://github.com/weaveworks/weave-gitops/issues/1600
- [x] https://github.com/weaveworks/weave-gitops/issues/1676
- [x] https://github.com/weaveworks/weave-gitops/issues/1675
- [x] https://github.com/weaveworks/weave-gitops/issues/1674
- [x] https://github.com/weaveworks/weave-gitops/issues/1673
- [x] https://github.com/weaveworks/weave-gitops/issues/1672
- [x] https://github.com/weaveworks/weave-gitops/issues/1859

### Work to be done in Enterprise

- [x] ~https://github.com/weaveworks/weave-gitops/issues/1690~
- [x] https://github.com/weaveworks/weave-gitops/issues/1691
- [ ] https://github.com/weaveworks/weave-gitops/issues/2000

PR #1862, opened 2022-04-01
feat: Derive Cluster from Source object response

Callisto13:
The final UI piece for Mutli-Cluster #1677 (I think).

Previously in the UI we hardcoded "Default" everywhere in the Source
table and source Info pages. Now the UI will derive the object's Cluster
from its server response and fill the table/fields with that.

I have also added a Cluster field in the HelmChart source info page,
since we did not have that before and I wondered if it had been missed?

Closes https://github.com/weaveworks/weave-gitops/issues/1859  and https://github.com/weaveworks/weave-gitops/issues/1861

To test you can set [this](https://github.com/weaveworks/weave-gitops/blob/main/core/clustersmngr/single_fetcher.go#L9) to `"foo"` or whatever, and you should not see a cluster called `"Default"` anywhere.


yitsushi (approved):

joshri (approved):
🌞  PHAT APPROVE 🌞

PR #1873, opened 2022-04-04
Remove CLI auth code

ozamosi:
This started in `pkg/services/factory.go`, where I realised the git
setup code in there that I've removed is only invoked with either (in
upgrade) no URL, or (in profiles) a helm chart. Both of those cases
are times when we skip trying to install a deploy key.

If we don't create a deploy key in the CLI, then we don't do git auth
in the CLI. If we don't do git auth in the CLI, then half of the git
auth code is just not relevant any more.


JamWils:
@jpellizzari, no.  Basically we should mimic the functionality from the Flux CLI and nothing more.  I don't think they enable OAuth via any of their commands.

jpellizzari (approved):

PR #1826, opened 2022-03-29
initial rewrite of readme and contributing

sympatheticmoose:
Closes: https://github.com/weaveworks/weave-gitops/issues/1816

**What changed?**
Updates readme to be more user-focussed
Moves development guide under Contributing.md. NOTE: this will require follow up work to tidy-up which I am not intending to address in this PR as feel this can be tackled soon after. My focus is on the readme itself.

**Why?**

**How did you test it?**

**Release notes**

**Documentation Changes**

PR #1877, opened 2022-04-04
Update README to reflect current development steps

jpellizzari:
Re-arranges and updates the README to reflect how to effectively do development.

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,9 +62,14 @@ Use "gitops [command] --help" for more information about a command.
 
 For more information please see the [docs](https://docs.gitops.weave.works/docs/cli-reference)
 
-## CLI/API development
+## Development
 
-To set up a development environment for the CLI
+### To use a bootstrapped, ready made environment
+
+1. Install go v1.17
+2. Install [buf](https://github.com/bufbuild/buf)
+3. Run `make all` to install dependencies and build binaries and assets
+4. Run `make cluster-dev` which should install and bring up everything and then start `tilt` to take over monitoring
 
 ### To use an existing environment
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -62,9 +62,14 @@ Use "gitops [command] --help" for more information about a command.
 
 For more information please see the [docs](https://docs.gitops.weave.works/docs/cli-reference)
 
-## CLI/API development
+## Development
 
-To set up a development environment for the CLI
+### To use a bootstrapped, ready made environment
```
josecordaz, line 67: 🤔  I wonder if we can clarify somehow that the `ready made environment` means to have a kind cluster created by the command `KIND_CLUSTER_NAME=<some name> ./tools/kind-with-registry.sh`. Otherwise `make cluster-dev` will fail. Perhaps we need to switch the order of the blocks `To use an existing environment`(which actually seem like it needs to change to something like `creating a dev env from scratch` or something similar) and `To use a bootstrapped, ready made environment`. 🙂

josecordaz:
I see a couple of issues:
image
image

josecordaz:
It seem like the second screenshot error is because `make all` generates a binary using whichever platform the developer is on. I would probably suggest add a comment to make the differentiation between macOS/linux developers. It would help for a good UX experience. The first error went away somehow, I can't reproduce it anymore.

README.md
Callisto13: I think let's remove this thing. People can use the other section and figure out that they can leave their cluster for reuse if they want to. This section has always really confused me to the point that I pretend it isn't there

Callisto13:
Note also: https://github.com/weaveworks/weave-gitops/pull/1826

jpellizzari:
> It seem like the second screenshot error is because `make all` generates a binary using whichever platform the developer is on. I would probably suggest add a comment to make the differentiation between macOS/linux developers. It would help for a good UX experience. The first error went away somehow, I can't reproduce it anymore.

Click the "refresh" buttons to fix

jpellizzari:
Looks like other folks are working on this

PR #1885, opened 2022-04-05
Revert "Include SVG images as react components"

yitsushi:
This reverts commit 6944695bfd78de1420cd986c95a756770b1dcbb4.


ozamosi (approved):

yitsushi:
It seems we may have to go forward instead of revert because a lot other things are built on top of this (non linear structure), so I think the "easier" solution is to figure out why the svg stuff breaks, fix it or remove it in a new PR instead of  trying to revert it :(

PR #1889, opened 2022-04-06
Change graph node shape + add status + add cluster name

### LINKED ISSUE #1866: Freshen up reconciliation graph nodes
CURRENT CIRCLE NODES:
image

HOT NEW RECTANGLE NODES:
image

### LINKED ISSUE #1874: [UI] Graph outline should change color according to status
@LutzLange brought it to my attention that failed pods still display with green outline, when the outline should be red.
Screenshot from 2022-04-04 16-03-36

He's running MCCP chart version 0.0.17-56-gffbf05f, which is 0.0.18.

### LINKED ISSUE #1880: [UI] Move status on graph nodes
We've received some feedback that the current graph implementation looks like boobs (ha! not sure there is a professional way to explain this) : 
Screenshot 2022-04-05 at 11 36 39

So graph node status icons should be updated to the new design (moved to the bottom left) : 
20220405app-graph

joshri:
Closes #1866
Closes #1874 
Closes #1880

Here's the screenshot:
image

Graph nodes did not correctly display status changes and the circle shape did not use space efficiently.


ui/components/ReconciliationGraph.tsx
```diff
@@ -37,18 +38,32 @@ type NodeHtmlProps = {
 };
 
 const NodeHtml = ({ object }: NodeHtmlProps) => {
+  console.log(object);
```
jpellizzari, line 41:
```suggestion
```

jpellizzari (approved):

bia:
This looks too much like Argo.
Screenshot 2022-04-06 at 17 16 41
Screenshot 2022-04-06 at 17 15 39

@mmoulian and I can propose round nodes in the style of #1880 that fit three lines of text. The rest of the info will be displayed in a tooltip.

bia:
Why is the Git Repository a child of flux-system instead of the parent?

ui/components/HashRouterTabs.tsx
```diff
@@ -65,45 +66,42 @@ function HashRouterTabs({ className, children, defaultPath, history }: Props) {
...
+        </Switch>
+      </HashRouter>
+    </Flex>
   );
 }
 
 export default styled(HashRouterTabs).attrs({
   className: HashRouterTabs.name,
 })`
+  height: 100%;
```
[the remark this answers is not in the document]
joshri: I made that 100% to get the graph to extend to the bottom of the page

ui/components/ReconciliationGraph.tsx
joshri: oopsie 🗑️

JamWils:
The round nodes do not show the status and information efficiently.  The round circles truncate too much text and don't offer enough insights into how the application is laid out.  In addition, hovering over each node to get info such as which namespace is time consuming.  

Just because Argo uses an appropriate shape doesn't mean we should have to choose some other form factor.  Android phones aren't circular because the iPhone was a rectangle.  While I like the design of the circles more there is less utility by going that route.

Per the image below, the `KustomizationAutomation` is truncated.  It's hard to read each node.  I don't know which namespace they are in.  Ideally, we would also show the images running on the pods or some other useful info to look into at a glance.

Screen Shot 2022-04-06 at 12 40 44 PM
route.

JamWils:
Also @joshri it has a `Default/` before every object.  Can that be removed?  Can you also provide a screenshot with some "flux" object suspended as well as something failing.

joshri:
> Also @joshri it has a `Default/` before every object. Can that be removed? Can you also provide a screenshot with some "flux" object suspended as well as something failing.

Default is the cluster name - like in the figma: 
image

Easy to remove if you'd like!

jpellizzari:
We should probably remove the `default/` for now, as it will be confusing as `default` is also used to describe the default namespace in Kubernetes.

JamWils:
Also, @joshri it looks like the top node is not showing a status.

JamWils:
Agree, I didn't even notice I did that with the labels when I did the designs.

joshri:
@JamWils screenshot of suspended and failed apps on graph

image

PR #1890, opened 2022-04-06
Move (most) core/server tests to not use envtest

ozamosi:
So, I've been shaving some yaks.

I wanted to fix enterprise's unit tests. However, those tests don't run with an envtest environment, and the recent cache changes in #1780 mean we can't call `StartCoreServer` without something that responds to requests a bit like kube API - basically, you can't turn `rest.Config` into `client.Client` without making a connection, and without an envtest I don't have anything to connect to. That's unfortunate.

I pulled that kube Client out of the server creation itself, so that the cache can be fed by the official fake client. This was enough to make the enterprise tests work (they only really want to see that they added the routes properly). I was pleased.

Then I realised, this will undoubtedly happen again as long as there's no test that tests that we _can_ start the core server without running an envtest. So I wrote a test for that.

Alas, once I'd done that, I realised that I'd effectively just duplicated one of the tests for the core server. So I ported the original test instead, and then I ported some more. When all the tests are ported, they take about 1 second instead of 10 to run `make unit-tests TEST_TO_RUN=./core/server`

I left one remaining server test that uses the kube client (for ListFluxEvents) - it wasn't trivial to inject the fake client in there, and @Callisto13 appears to be on a mission to remove all uses of that (#1854, etc), so I decided that can wait.

As I started to remove sleeps in the tests, I got into a fight with the namespace cache, because ForceRefresh isn't synchronous or instant. I'm not quite sure what to do with it because there's no uses that needs a global cache yet, so I've left it leaking a few goroutines inside the tests, so that I can at least ask for a new cache and get a new cache.


core/server/fluxruntime_test.go
```diff
@@ -62,7 +55,12 @@ func TestGetReconciledObjects(t *testing.T) {
 		},
 	}
 
-	g.Expect(k.Create(ctx, &reconciledObj)).Should(Succeed())
+	k := fake.NewClientBuilder().
```
jpellizzari, line 58: :thinking: Not understanding why we need to move to this lib? Do we still get some of the same validation and RBAC stuff?

jpellizzari (commented):
Based on the PR description, I'm not sure why we would move away from `envtest`.

core/server/fluxruntime_test.go
ozamosi: I've opened #1895 without test improvements.

PR #1891, opened 2022-04-06
Remove animation from sign_on screen

### LINKED ISSUE #1788: Login page causes CPU saturation on firefox
The login page for core (http://localhost:9001/sign_in) causes CPU saturation in Firefox (60-90%) I've uploaded a profile here https://share.firefox.dev/3tyFTxX

SamLR:
There is apparently a bug in how firefox renders animated SVGs on Mac OS X that causes it to saturate the CPU.

This removes the animation tags and fixes the problem

Fixes #1788

**What changed?**

Removed the animation from the `ui/images/SignInBackground.svg`

**Why was this change made?**

Because the signon page was un-useable on firefox.

Local testing that the page renders.

### DIFF
--- a/ui/images/SignInBackground.svg
+++ b/ui/images/SignInBackground.svg
@@ -1007,22 +1007,6 @@
 			M3.7,705.6c0,0,342.8-486.5,665.5-359.6c322.7,126.9,281,360.8,449,348c167.9-12.8,327.7-138.4,327.7-138.4"/>
 	</g>
 	<g>
-			<animateTransform
-				attributeType="xml"
-				attributeName="transform"
-				type="translate"
-				dur="10s"
-				values="0,0; -100,-100; 0,0; -100,-150;0,0"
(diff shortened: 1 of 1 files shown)


ozamosi:
I'm not gonna hit approve, but I can see that it solves the problem.

Before this, when I open the login page in firefox on linux, it hogs more than 5 CPU cores, I get about 1FPS of background (well, slightly less, I measured with a stopwatch), and it takes just over 3s for any character I type to appear on the screen.

With this change, I can type my username properly.

AlinaGoaga:
Let's go ahead and remove the animation bits for now. I'll look into how we can improve the performance separately. Thanks!

AlinaGoaga (approved):

SamLR:
> Let's go ahead and remove the animation bits for now. I'll look into how we can improve the performance separately. Thanks!

Thanks @AlinaGoaga! LMK if there're things that need testing to help with this.

PR #1893, opened 2022-04-06
final changes to generate a release

### LINKED ISSUE #1685: Make sure our release process works in new `main`

### LINKED ISSUE #1858: Different npm versions break the release workflow

jrryjcksn:
Closes #1685 
Closes #1858

**What changed?**
* Pinned npm version in workflow to avoid reformats
* Removed broken flag from gitops-server build
* Fixed issue with doc generation when CLI reference directory needs to be deleted

**Why was this change made?**
To make the release process work in the new model

Made a fork of weave-gitops and built releases

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -35,9 +35,7 @@ jobs:
           echo "GORELEASER_PREVIOUS_TAG=$(git describe --abbrev=0 --tags $(git rev-list --tags --skip=1 --max-count=1))" >> $GITHUB_ENV
           echo "GORELEASER_CURRENT_TAG=$(git describe --abbrev=0 --tags)" >> $GITHUB_ENV
       - name: "Make All"
-        run: |
-          npm i -g npm@8.1.2
-          make all
+        run: make all
       - name: Check Git State
         run: git diff --no-ext-diff --exit-code
       - name: Build Changelog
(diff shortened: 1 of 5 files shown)


gitops-server.dockerfile
```diff
@@ -28,7 +28,6 @@ COPY . /app
 # These are ARGS are defined here to minimise cache misses
 # (cf. https://docs.docker.com/engine/reference/builder/#impact-on-build-caching)
 # Pass these flags so we don't have to copy .git/ for those commands to work
-ARG LDFLAGS="-X localbuild=true"
```
SamLR, line 31: Why is this being removed? It's needed in order to correctly inject the `$LDFLAGS` as set in the makefile without copying the entire `.git` directory into the docker file.

.github/workflows/release.yaml
```diff
@@ -35,7 +35,9 @@ jobs:
           echo "GORELEASER_PREVIOUS_TAG=$(git describe --abbrev=0 --tags $(git rev-list --tags --skip=1 --max-count=1))" >> $GITHUB_ENV
           echo "GORELEASER_CURRENT_TAG=$(git describe --abbrev=0 --tags)" >> $GITHUB_ENV
       - name: "Make All"
-        run: make all
+        run: |
+          npm i -g npm@8.1.2
```
ozamosi, line 39: I don't think this is necessary - an npm is built into the release of node, so it will be the same version each time. In this case:
```bash
$ curl -s https://nodejs.org/dist/index.json | jq -r '.[] | select(.version == "v16.13.2") | .npm '
8.1.2
```

ozamosi (approved):

PR #1895, opened 2022-04-07
Make namespace cache mockable when starting server

ozamosi:
This means we can start a server and send it a request, without spinning up a k8s api server.
    
It doesn't mean this server is very useful, but we can test things like "are the endpoints present"


mnowster (approved):

Callisto13:
goddamit ya'll are killing me 😂 

merge conflicts galore today

PR #1897, opened 2022-04-07
Fix width and height of Kustomization Detail

joshri:
OLD:
image

NEW:
image

### DIFF
--- a/ui/components/KustomizationDetail.tsx
+++ b/ui/components/KustomizationDetail.tsx
@@ -26,12 +26,14 @@ const Info = styled.div`
 
 const TabContent = styled.div`
   margin-top: 52px;
+  width: 100%;
+  height: 100%;
 `;
 
 export default function KustomizationDetail({ kustomization, name }: Props) {
   const hashHistory = createHashHistory();
   return (
-    <>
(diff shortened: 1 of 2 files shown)


ui/components/KustomizationDetail.tsx
```diff
@@ -26,12 +26,14 @@ const Info = styled.div`
...
   margin-top: 52px;
+  width: 100%;
+  height: 100%;
 `;
 
 export default function KustomizationDetail({ kustomization, name }: Props) {
   const hashHistory = createHashHistory();
   return (
-    <>
+    <Flex wide column height="100%">
```
jpellizzari, line 36: Weird sort-of-inline styling here. I would expect `style={{ height: '100%' }}`.

jpellizzari (approved):

ui/components/KustomizationDetail.tsx
joshri: It's a prop - I didn't know that we had that but it's there. It was messing with my attempt at inline styling so if it's weird we should get rid of it

jpellizzari: Agreed

Issue #1746, opened 2022-03-16
Flux Reconcile, Resume, and Suspend


jpellizzari:

jpellizzari:
Example endpoint of how to force-reconcile (aka sync) a resource: https://github.com/fluxcd/webui/blob/3ee44015f4105bb13857cdd173a0a351f89100d1/pkg/clustersserver/clustersserver.go#L523

PR #1902, opened 2022-04-08
Add automation sync endpoints and UI

jpellizzari:
Part of #1746 

Adds a sync interaction to the UI:
Screenshot from 2022-04-08 10-11-11

What it looks like when it is successful:

Screenshot from 2022-04-06 10-10-30

This is weird and ugly, but refactoring the Toast hopup thingy was out of scope for the PR ^^


yitsushi (approved):
That toggle solution looks awesome. :shipit:

PR #1903, opened 2022-04-11
feat: Add cross-namespace querying to all listing endpoints

### LINKED ISSUE #1901: Add cross-namespace querying to helmreleases and sources
Now that #1888 is merged, it should be pretty straightforward to add cross-namespace querying to other objects.

Acceptance Criteria
* If no `namespace` is specified to the API request, I see all objects for a given `list` route that exist across namespaces for a cluster.

yitsushi:
Closes #1901

### DIFF
--- a/core/server/list.go
+++ b/core/server/list.go
@@ -1,0 +1,50 @@
+package server
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/weaveworks/weave-gitops/core/clustersmngr"
+)
+
+// listFn is a function that contains the logic how to list objects
+// from a single namespace.
+type listFn func(context.Context, clustersmngr.Client, string) ([]interface{}, error)
+
+// listObejcts lists objects across all namespaces.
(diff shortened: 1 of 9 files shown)


core/server/list.go
```diff
@@ -0,0 +1,50 @@
...
+
+	if namespace != "" {
+		res, err := list(ctx, clustersClient, namespace)
+
+		return res, err
+	}
+
+	var results []interface{}
+
+	nsList, found := cs.cacheContainer.Namespaces()[clustersmngr.DefaultCluster]
```
luizbafilho, line 33: I think we might want to handle this differently. After we sort out the namespace thing, we need to make sure this implementation calls all clusters in that list because even tho this code live in the core, the way enterprise integrate it is by hooking the `coreServer` in tho their server but with a different clusters fetcher. If we "hardcode" the cluster name here that will be broken once they start using it.

PR #1908, opened 2022-04-12
Fix Spacing in Page Component

joshri:
CURRENT - filter bar grey line visible
image

FIXED - note visible background
image

### DIFF
--- a/ui/components/Page.tsx
+++ b/ui/components/Page.tsx
@@ -32,8 +32,6 @@ export const Content = styled(Flex)`
   padding-bottom: ${(props) => props.theme.spacing.medium};
   padding-left: ${(props) => props.theme.spacing.large};
   padding-top: ${(props) => props.theme.spacing.medium};
-  width: 100%;
-  height: 100%;
 `;
 
 const Children = styled(Flex)``;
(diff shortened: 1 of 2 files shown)


ui/components/Page.tsx
```diff
@@ -32,7 +32,6 @@ export const Content = styled(Flex)`
   padding-bottom: ${(props) => props.theme.spacing.medium};
   padding-left: ${(props) => props.theme.spacing.large};
   padding-top: ${(props) => props.theme.spacing.medium};
-  padding-right: ${(props) => props.theme.spacing.large};
```
jpellizzari, line 35: I just added this back in here so the page wouldn't extend all the way off the screen: https://github.com/weaveworks/weave-gitops/commit/3d63640ab4b132a03696a31388bff030f1f2c061#diff-5172b3d6127736059f59dbba793deb31dfbed2b93c0211f9c5cb4c397607ed4aR35

Was that incorrect?

jpellizzari (approved):

ui/components/Page.tsx
joshri: When I deleted it the filter bar moved off the screen again - so I guess so? That padding pushes the filter bar to the left.

PR #1913, opened 2022-04-13
ref: Update cache container to sync several caches

Callisto13:
## Tests still to come

There is a fair amount here which will only make sense with the context
of the Multi-Cluster fetcher and cache which will not live in this repo.

(I will link that PR or branch when I have it.)

**There is a lot I have left not-perfect as I was keen to get what I have
out so that others can start iterating on other things. Plus it has taken
me days of getting through all the merge conflicts as this is apparently
a popular package right now. So I wash my hands of this thing!**

In short the highlights are:
- The CacheContainer holds a map of storage types to fullfillers of the
  Cache interface.
- The interface allows the Container to iterate over caches, rather than
  call them independently
- Note the `List() interface{}` call which is not ideal and requires
  Callers to typecheck. This can be changed in future (or now if anyone
  has a quick suggestion).
- There is still capacity for the caches to be called by name for some
  actions
- There are builder functions supplied by each cache to:
  - Add them to the container
  - Set whether then will be synced with another cache in the container
- The Synced/NotSynced builder funcs are deliberately simple for now
- The NamespaceStorage has a clusterChan which will be used for
  Multi-Cluster. If left nil the NamespaceStorage will not need to use
  it, or in future we can change it so the single cluster implementation
  adds the same cluster (self) each time.

I have also bumped some flux packages for the Helm/Profile watcher
because it was yelling at me for "reasons", and updated the Chart SA
with more resource perms as it was also yelling at me. Basically feel
free to ignore all Helm and Chart stuff.


core/cache/namespace.go
```diff
@@ -8,31 +8,45 @@ import (
...
 }
 
-func newNamespaceStore(c clustersmngr.Client, logger logr.Logger) namespaceStore {
-	return namespaceStore{
-		client:     c,
-		namespaces: map[string][]v1.Namespace{},
-		logger:     logger,
-		lock:       sync.Mutex{},
-		cancel:     nil,
+func newNamespaceStore(logger logr.Logger, restCfg *rest.Config, clusterChan <-chan mngr.Cluster) *namespaceStore {
```
yitsushi, line 37: Can we use a `NamespaceStoreOption{}` as an argument instead? That way we can add extra fields later easier, for example add option to set `interval`.

Usually after 3+ arguments I move everything into an Option.

100% not a blocker.

yitsushi (approved):

core/cache/namespace.go
Callisto13: no probs 👍  can add that now while I am doing the container interface tests

luizbafilho (approved):

PR #1917, opened 2022-04-13
Use url: tag instead of data-url on images

### LINKED ISSUE #1916: UI `data-url` in-lining makes builds take forever
The `data-url` inlining that was added (here)[https://github.com/weaveworks/weave-gitops/pull/1896] has made our builds take a long time and appear to have broken hot-reloading.

Reproduce:
1. Make a trivial text change
2. Note that the build takes ~20s
3. Comment out all of the import statements in `ui/lib/images.ts`
4. Note that the rebuild takes ~3s

jpellizzari:
Closes #1916 

Fixes slow builds, but probably breaks enterprise. @foot I assume `data-url` thing was a way to inline the SVGs, any other details you can share on that?

### DIFF
--- a/ui/lib/images.ts
+++ b/ui/lib/images.ts
@@ -1,18 +1,18 @@
 /* eslint-disable */
 // @ts-nocheck
 
-import bg from "data-url:../images/General-backgroundok.png";
-import failedSrc from "data-url:../images/failed.svg";
-import fancyCardBackground from "data-url:../images/fancy-card-background.svg";
-import fancyCardBackgroundBlue from "data-url:../images/fancy-card-background-blue.svg";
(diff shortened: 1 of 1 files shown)


foot:
Sorry, I've been thinking about this and tinkering a little bit.

The closest I've come is exporting esmodules from parcel, it does the `import.meta.url` trick and webpack "forwards" the images/assets through the build (https://github.com/webpack/webpack/issues/7353). BUT I ran into https://github.com/facebook/create-react-app/issues/11770, where the svgs are turned into react components before weave-gitops gets them. 🤷 

I'll do a quick test w/ parcel to see if it can consume esmodules w/ import.meta.url, maybe we conform the build systems..

foot:
Okay that actually seems to work okay.. so switching to parcel might be a thing.

makkes:
closing as stale.

PR #1920, opened 2022-04-14
fix: Namespace cach can be called before the first results come back

### LINKED ISSUE #1919: Test flake: `TestListFluxRuntimeObjects_inMultipleNamespaces`
Link to failed build: https://github.com/weaveworks/weave-gitops/runs/6010034043?check_suite_focus=true

```
=== RUN   TestListFluxRuntimeObjects_inMultipleNamespaces
    fluxruntime_test.go:205: 
        Unexpected error:
            <*status.Error | 0xc00044a130>: {
                s: {
                    s: {
                        state: {
                            NoUnkeyedLiterals: {},
                            DoNotCompare: [],
                            DoNotCopy: [],
                            atomicMessageInfo: nil,
                        },
                        sizeCache: 0,
                        unknownFields: nil,
                        Code: 2,
                        Message: "default cluster not found",
                        Details: nil,
                    },
                },
            }
            rpc error: code = Unknown desc = default cluster not found
        occurred
--- FAIL: TestListFluxRuntimeObjects_inMultipleNamespaces (0.03s)
```

This has been seen once in CI and anecdotally locally. I would guess that it is a race and maybe the namespace cache was not ready after a ForceRefresh.

yitsushi:
It was possible to call `List()` before the first `update` finished and
populated the cache `map[string][]v1.Namespace{}`. Requests before the
first results arrived would return "default cluster not found" error.

Solution: Force sync update operation on `Start()` and move the `update`
call after the `select` for `Tick` to avoid multiple calls on `Start()`.

Closes #1919

---

feat: Force cacheContainer to maintain the order or cacheStores

Later on, some of our cache stores have dependencies between them.
Thinking of cluster clients and cluster list, these can be used in other
cache stores, so they should exists before the rest is initialised.

To achieve this, we can use a `[]cacheInfo` slice instead of a
`map[StorageType]Cache`, so the order we defined with `WithXXX`
functions will be the same as we initialise them.

### DIFF
--- a/core/cache/container.go
+++ b/core/cache/container.go
@@ -108,6 +108,7 @@ func With(l logr.Logger, cachesFn CachesFunc) CacheCollection {
 
 	for _, opt := range options {
 		name, store := opt(l, clusterChan)
+
 		caches = append(caches, cacheInfo{StorageType: name, Cache: store})
 	}
 
--- a/core/cache/container.go
+++ b/core/cache/container.go
@@ -15,7 +15,7 @@ const DefaultPollIntervalSeconds = 120
 type (
 	// StorageType is the name of the cache, eg NamespaceStorage
 	StorageType string
-	// CacheCollection is a map of cache type to cache interface held
+	// CacheCollection is a list of cache info held
 	// in the container.
 	CacheCollection []cacheInfo
 )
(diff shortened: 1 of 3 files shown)


Callisto13 (approved):
aaaah sneaky

yitsushi:
Somewhere there is a deadlock :( working on it

Callisto13:
awww :(

that's going to be an interesting thing to debug whenever someone adds caches in the wrong order :D

Callisto13 (approved):

PR #1922, opened 2022-04-14
Introducing limited user

### LINKED ISSUE #1671: dev: Create a less powerful user for dev purposes
# Claudia needs to update this, ignore until then

**Short description**
The Chart deployed by Tilt in out dev cycle creates an Admin user.
This role will get gradually more powerful, so we want to make sure that we are still doing restrictions properly.

Create a new role which is deployed *only in a dev environment* as part of our Tilt set up, which has only the most basic powers. Basically it should be able to `get` and `list` a selection of objects in a specific namespace.

**Will require pulling https://github.com/weaveworks/weave-gitops/pull/1624 into v2 so that we are able to set usernames.**

**Acceptance criteria**
- [ ] I run `make cluster-dev`
- [ ] I navigate to UI
- [ ] I log on as the `anti-hero` user
- [ ] I see that I can only see some resources in a specific namespace
- [ ] I cannot see or do anything in other namespaces.

luizbafilho:
Closes #1671 

**What changed?**
When doing developing we might need to test with users with limited power, simulating a more real-world environment, this PR introduces the `limited` user. It basically can do all the necessary operations but only in the `team-a` namespace.

Important note, there is no concept of login with this user here, we are basically forcing the `limited` user to be returned as the authenticated user, no matter the credentials you type on the login page.

To use this you should just set the `DEV_USE` as an environment variable so tilt can execute the server with the proper value.

```sh
$ DEV_USER=limited tilt up
```

> Given that currently, we are passing `flux-system` as the namespace to query the objects if you run with a limited user you won't be able to load anything.

**Why was this change made?**

**Release notes**

**Documentation Changes**

### DIFF
--- a/Tiltfile
+++ b/Tiltfile
@@ -18,7 +18,9 @@ docker_build_with_restart(
         './bin',
     ],
     dockerfile="dev.dockerfile",
-    entrypoint='/app/build/gitops-server --log-level=debug --insecure',
+    entrypoint="/app/build/gitops-server --log-level=debug --insecure --dev-mode --dev-user {dev_user}".format(
+        dev_user=os.getenv("DEV_USER", "wego-admin")
+    ),
     live_update=[
         sync('./bin', '/app/build'),
     ],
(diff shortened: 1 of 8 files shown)


jpellizzari (approved):
Very nice solution

PR #1930, opened 2022-04-14
Replace the release process with a workflow

ozamosi:
This pretty much replaces the release process with entering the release number in a web form and approving a PR - no terminal required, just write permissions to this repo.

You can't see it work in this repo until it's merge, but in my own fork of this repo I can go to Actions, click on prepare-release (there's a link in the release-process document), click on Run workflow, and I'm greeted with this:
Screenshot from 2022-04-14 16-40-22

So when I've entered the version number and triggered the workflow, then 1-5 minutes later I have a PR like this: https://github.com/ozamosi/weave-gitops/pull/23 - this example is for an RC so it only has javascript changes in it. If the release is stable it also includes doc updates, readme updates, and helm chart updates, like so: https://github.com/ozamosi/weave-gitops/pull/24

Once the PR is approved, the bot will trigger actions to tag, build, upload, and merge (and the merge, in turn, triggers chart and docs site updates). And then the release is done!

Note: this removes the ability to just push a tag to get a new javascript package. I hope that this flow is low-effort enough that having to open a browser won't impede anybody.


josecordaz (approved):
This looks awesome, great work.

.github/workflows/prepare-release.yaml
```diff
@@ -0,0 +1,84 @@
...
+  REGISTRY: ghcr.io
+  IMAGE_NAME: weaveworks/wego-app
+
+jobs:
+  release-tasks:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: Unshallow
```
SamLR, line 20: I think this can be replaced  by adding this to the `checkout` step
```yaml
with:
  fetch-depth: 0
```

ozamosi:
So it's almost my end of day and easter is happening, so I'm not merging this until when I'm back on Tuesday, because the probability of teething problems seems high.

SamLR (approved):
minor tweak if you want to otherwise looks fine

.github/workflows/prepare-release.yaml
ozamosi: So. According to docs, yes. From past experience, maybe.

Last time I tried to do that I found that it still didn't fetch all the tags I expected, and the actions/checkout repo is full of issues saying the same, many closed without comment.

So, I'm going to leave this, because this seems to work.

PR #1931, opened 2022-04-14
Remove flux-system as the default namespace from requests

jpellizzari:
The new default should be `NoNamepsace`, which is an empty string. Objects for all namespaces should be returned.

### DIFF
--- a/ui/components/AddGitRepoForm.tsx
+++ b/ui/components/AddGitRepoForm.tsx
@@ -3,7 +3,7 @@ import _ from "lodash";
 import * as React from "react";
 import styled from "styled-components";
 import { GitProvider } from "../lib/api/applications/applications.pb";
-import { WeGONamespace } from "../lib/types";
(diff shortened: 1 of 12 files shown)


joshri (approved):
Looks good - with NoNamspace being an empty string, would it mean anything to have it be like a dash or something instead? The empty strings inside the data table just end up looking like errors to me. I guess it's....not likely that a user wouldn't have an object with no namespace? Doesn't matter that much, just a thought.

jpellizzari:
> Looks good - with NoNamspace being an empty string, would it mean anything to have it be like a dash or something instead? The empty strings inside the data table just end up looking like errors to me. I guess it's....not likely that a user wouldn't have an object with no namespace? Doesn't matter that much, just a thought.

@joshri The `NoNamespace` variable is only used in request logic. It never shows up in the UI. You are correct, objects will always have a namespace.

PR #1932, opened 2022-04-14
Redirect to sign in on AJAX requests

### LINKED ISSUE #1894: The user should be taken to the login screen when there token is no longer valid
As a user when I am in the app and I see that there is an error: authentication required.  I should in fact not see that error and should be taken back to the login screen to login again.  This is an app wide behavior and not a specific screen.

jpellizzari:
Closes #1894

Wraps the `Core` client to ensure that 401s always redirect the user to the `sign_in` page.


joshri (approved):
🧠  GENIUS 🧠

PR #1937, opened 2022-04-15
Improve Filter Animation

joshri:
Adds `left` property to transition, padding and active state to hamburger button, and removes X button. I think it looks nice. 

https://user-images.githubusercontent.com/65822698/163601578-2b5f858c-4be7-402f-8536-7164dd9500b7.mov


jpellizzari (approved):
Looks great!

PR #1941, opened 2022-04-18
Stop the testEnv

### LINKED ISSUE #1915: `envtest/kube-apiserver` processes continue running after tests, consume machines
A couple of times I noticed that my computer was whirring off into orbit, and that there were dozens of `envtest/kube-apiserver` processes eating all the memory.

These hung around after `make unit-tests` (or `make tests`, we should probably start joining those up?).
I am not sure whether they always stay, or if it is only on test failure that cleanup does not happen.

jpellizzari:
Closes #1915 

Show the `testEnv` who is boss.

Tested via:
```sh
$ make unit-tests
$ ps -a | grep api
```


josecordaz (approved):
LGTM

Callisto13 (approved):
🙏

yitsushi (approved):

PR #1944, opened 2022-04-19
Docs fixes

ozamosi:
This combines 3 different fixes to the manual:
* Make the bcrypt generation more straightforward, and applies a couple of small bug fixes to the getting started instructions regarding this step. This is done for both main and stable.
* Fix download link in stable. This takes the fixed link (as of fce5eab4) and applies it to stable.
* Adds slashes to all URLs. There's a lot of detail in the commit message for that fix - in essence this fixes relative links if you're linked to the manual, so to reproduce see [the installation page](https://docs.gitops.weave.works/docs/installation) and click on "Securing access to the dashboard" in the Pre-requisite section. This should fail now, but [the staging docs](https://staging.docs.gitops.weave.works/docs-fixes/docs/installation/) should pass the same test.

### DIFF
--- a/website/docs/cluster-management/getting-started.mdx
+++ b/website/docs/cluster-management/getting-started.mdx
@@ -12,7 +12,7 @@ import BrowserOnly from "@docusaurus/BrowserOnly";
 
 ## Creating your first CAPD Cluster
 
-If you've followed the [Upgrade steps](../installation#weave-gitops-enterprise) in the [Installation guide](../installation) you should have:
+If you've followed the [Upgrade steps](installation.mdx#weave-gitops-enterprise) in the [Installation guide](installation.mdx) you should have:
 
 1. Weave GitOps Enterprise installed and an agent running on the management cluster.
 2. A CAPI provider installed (With support for `ClusterResourceSet`s enabled).
--- a/website/docs/configuration/securing-access-to-the-dashboard.mdx
+++ b/website/docs/configuration/securing-access-to-the-dashboard.mdx
@@ -97,26 +97,14 @@ If you need to change the OIDC configuration after running `gitops upgrade`, upd
 
 ## Login via a cluster user account
 
-Before you login via the cluster user account, you need to generate a bcrypt hash for your chosen password and store it as a secret in Kubernetes. There are several different ways to generate a bcrypt hash, this guide uses an Alpine Docker image to generate one:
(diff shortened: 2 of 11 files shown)


website/docs/configuration/securing-access-to-the-dashboard.mdx
```diff
@@ -97,26 +97,14 @@ If you need to change the OIDC configuration after running `gitops upgrade`, upd
 
 ## Login via a cluster user account
 
-Before you login via the cluster user account, you need to generate a bcrypt hash for your chosen password and store it as a secret in Kubernetes. There are several different ways to generate a bcrypt hash, this guide uses an Alpine Docker image to generate one:
+Before you login via the cluster user account, you need to generate a bcrypt hash for your chosen password and store it as a secret in Kubernetes. There are several different ways to generate a bcrypt hash, this guide uses a go Docker image to generate one:
```
sympatheticmoose, line 100: incredibly minor nit, but I _think_ it should be Go rather than go

ozamosi: Fixed, in all 4 places.

SamLR (approved):
The bcrypt thing's annoyed me enough that I've got some fixes coming for it but this looks good

website/docs/cluster-management/getting-started.mdx
```diff
@@ -12,7 +12,7 @@ import BrowserOnly from "@docusaurus/BrowserOnly";
 
 ## Creating your first CAPD Cluster
 
-If you've followed the [Upgrade steps](../installation#weave-gitops-enterprise) in the [Installation guide](../installation) you should have:
+If you've followed the [Upgrade steps](installation.mdx#weave-gitops-enterprise) in the [Installation guide](installation.mdx) you should have:
```
sympatheticmoose, line 15: @ozamosi interesting, in chrome, if I click the upgrade link it navigates to the correct section. If I open in a new tab however, it just goes to the top of the page. This isn't major, but I wondered if theres likely a quick fix ~or whether you think we should force opening in a new tab~? EDIT: related, I wonder if we should force opening in a new tab, but with it screwing up navigation, probably best to avoid unless resolvable

sympatheticmoose, line 15: @ozamosi interesting, in chrome, if I click the upgrade link it navigates to the correct section. If I open in a new tab however, it just goes to the top of the page. This isn't major, but I wondered if theres likely a quick fix or whether you think we should force opening in a new tab?

sympatheticmoose:
I tried the new steps and worked perfectly, thanks @ozamosi - also checked all the links opened as expected, raised a couple of comments but nothing major/urgent

PR #1945, opened 2022-04-19
Make sourceRef namespace optional

jpellizzari:
This was causing a crash if the helm release didn't have a source ref defined


joshri (approved):
🌞

PR #1948, opened 2022-04-19
Better bcrypt use

### LINKED ISSUE #1925: Update Getting Started docs to be a bit more brief
The Getting Started section of the docs includes a bunch of stuff which is not necessary any more.

- `Configure access to the dashboard` can be slimmed down a bit to have `docker run -it golang:1.17 go install github.com/bitnami/bcrypt-cli@v1.0.2 && echo -n "super secret password" | bcrypt-cli`
- `Install Weave GitOps` can be shortened to only push the commits once.

SamLR:
Closes #1925

**What changed?**

* `adminUser.password`->`adminUser.passwordHash` in chart which no longer calculates the password hash for the user
* added `adminUser.createSecret` toggle (so the password can be supplied via other methods without requiring manual `RoleBinding` creation

*NB* this builds upon PR #1944

**Why was this change made?**

To discourage users from (potentially) storing passwords in plain-text in git. This way, at worst, they're storing a bcrypt-hashed copy of it which is much harder to attack. This also allows us to simplify the getting started guide.

Helm changes were tested using `helm template` and various flags to check the k8s was appropriate. The changes to the guide were also checked through.

**Release notes**

This is a breaking change to the helm chart and minor update to the docs but I don't believe it requires any major release notes (beyond that)

**Documentation Changes**

### DIFF
--- a/charts/gitops-server/Chart.yaml
+++ b/charts/gitops-server/Chart.yaml
@@ -15,7 +15,7 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 1.1.2
+version: 2.0.0
 
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application. Versions are not expected to
--- a/charts/gitops-server/templates/admin-user.yaml
+++ b/charts/gitops-server/templates/admin-user.yaml
@@ -44,6 +44,7 @@ rules:
   - apiGroups: [""]
     resources: ["events"]
     verbs: ["get", "watch", "list"]
+{{- if .Values.adminUser.createSecret }}
 ---
 apiVersion: v1
 kind: Secret
@@ -56,4 +57,5 @@ data:
   username: {{ .username | b64enc | quote }}
   password: {{ .passwordHash | required "passwordHash must be set!" | b64enc | quote }}
   {{- end }}
+{{- end }}
 {{- end }}
--- a/charts/gitops-server/templates/admin-user.yaml
+++ b/charts/gitops-server/templates/admin-user.yaml
@@ -54,12 +54,6 @@ type: Opaque
 data:
   {{- with .Values.adminUser }}
   username: {{ .username | b64enc | quote }}
-  {{/*
-    htpasswd returns user-file format, e.g. 'username:passwordhash' so as well
-    as using htpasswd to generate the bcrypt hash we also need to extract the
-    final element before base64 enconding it for k8s (and quoting it b/c yaml)
-  */}}
-   {{- $bcryptPasswordHash := .password | required "You must set a password for this user!" | htpasswd "" }}
-  password: {{ regexSplit ":" $bcryptPasswordHash -1 | last | b64enc | quote }}
+  password: {{ .passwordHash | required "passwordHash must be set!" | b64enc | quote }}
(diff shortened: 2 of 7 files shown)


SamLR:
Further info on securing bcrypt here: https://github.com/OWASP/CheatSheetSeries/blob/master/cheatsheets/Password_Storage_Cheat_Sheet.md

SamLR:
When this is merged I'll need to update weave-gitops-clusters

ozamosi:
This only updates the HEAD docs, but the backwards incompatible change will be pushed to the stable repo, so I think you need to change the 0.7.0 docs, too.

There's a second set of installation instructions in installation.mdx - I believe this needs to be updated, too.

SamLR:
> This only updates the HEAD docs, but the backwards incompatible change will be pushed to the stable repo, so I think you need to change the 0.7.0 docs, too.

That makes sense, is there a command I should use that you know of to generate the 0.7.0 docs or is it best just to copy the files over?

> There's a second set of installation instructions in installation.mdx - I believe this needs to be updated, too.

I'm not sure how to make this change make sense as the installation docs ultimately point at the 'securing access to the dashboard' pages which.. I'm not sure the intention of? Given the title it feels like a "here's the best way to do it" in which case we should probably recommend using SOPS or something similar to encrypt & deploy the secrets into the cluster where as if that page is intended more as a "here's what these bits mean" I'm guessing it should be a bit more in depth on using bcrypt securely and pass that to helm/flux.

@sympatheticmoose any thoughts on what I should be putting in these pages?

sympatheticmoose:
@SamLR the quick answer - yes the installation and securing pages need updates. Secret management recommendations could potentially be a guide in itself? The pages should lean more towards, "here's the recommended way to do it".

ozamosi:
> is there a command I should use that you know of to generate the 0.7.0 docs or is it best just to copy the files over?

Copy over. They're "forked" by automatic copying each time we make a stable release, and after that they're "supposed to" diverge.

> I'm not sure how to make this change make sense

I only meant, installation.mdx tells the reader to set `password:` and you renamed that field, so you also need to rename the example - or am I confused?

Rewriting the securing access page is definitely out of scope for this change, which I know @sympatheticmoose is already looking at.

sympatheticmoose:
>which I know @sympatheticmoose is already looking at.

I would appreciate any and all help with this 😉

SamLR:
> Copy over. They're "forked" by automatic copying each time we make a stable release, and after that they're "supposed to" diverge.

Ack

> I only meant, installation.mdx tells the reader to set password: and you renamed that field, so you also need to rename the example - or am I confused?

No, you're right. I'll fix that

>> Rewriting the securing access page is definitely out of scope for this change, which I know @sympatheticmoose is already looking at.

> I would appreciate any and all help with this 😉

right now? ;)
2b5

sympatheticmoose:
amazing gif choice @SamLR :joy: i'll write up the ticket/tickets for whomever has bandwidth to take it on (including myself), but no please do not feel like this is on you.

SamLR:
> amazing gif choice @SamLR 😂 i'll write up the ticket/tickets for whomever has bandwidth to take it on (including myself), but no please do not feel like this is on you.

TBF it's more that it's one of my favourite gifs ;)

Happy to help with bits of this but I think I'd struggle with a lot of the wider bits as I don't know a lot of how core interfaces with EE and what's needed to support both.

ozamosi (approved):

PR #1951, opened 2022-04-19
Add dedicated components for core details functionality, export flux hook

ozamosi:
This creates components for all of the gitops `SourceDetails`, which is what I took away from the suggestion [here](https://github.com/weaveworks/weave-gitops-enterprise/pull/601#discussion_r840672862)

It also exports `useListFluxRuntimeObjects` so we can show the flux runtime page as suggested [here](https://weaveworks.slack.com/archives/C03244W0C8H/p1649158559625549?thread_ts=1648749524.313119&cid=C03244W0C8H)

And finally, it exports the `CoreClientContextProvider` which seems to be purely a thing that's changed in core recently that I'm catching up with.

### DIFF
--- a/ui/index.ts
+++ b/ui/index.ts
@@ -14,6 +14,7 @@ import HelmChartDetail from "./components/HelmChartDetail";
 import HelmReleaseDetail from "./components/HelmReleaseDetail";
 import HelmRepositoryDetail from "./components/HelmRepositoryDetail";
 import KustomizationDetail from "./components/KustomizationDetail";
+import Page from "./components/Page";
 import SourcesTable from "./components/SourcesTable";
 import Interval from "./components/Interval";
 import Timestamp from "./components/Timestamp";
@@ -78,6 +79,7 @@ export {
   LoadingPage,
   muiTheme,
   OAuthCallback,
+  Page,
   RepoInputWithAuth,
   SignIn,
   SourceRefSourceKind,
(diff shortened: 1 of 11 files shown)


ui/index.ts
```diff
@@ -68,7 +80,6 @@ export {
   OAuthCallback,
   RepoInputWithAuth,
   SignIn,
```
foot, line 82: Shall we bring over the `Page` wrapper and use that in EE too?

jpellizzari (commented):
One thing that will bite us here is that we need to export `styled-components` so that we can override the styling where needed. Also, each top-level of the component should have a `className` prop.

Here is an example of exporting the styled component by default:
https://github.com/weaveworks/weave-gitops/blob/44f95682f7781da8c0219b731930f5dad416d11b/ui/components/ChipGroup.tsx#L31

Here is the snippet I use (vscode) to generate the component boilerplate: https://github.com/weaveworks/weave-gitops#recommended-snippets

ui/index.ts
ozamosi: I can do that - do you want to use it everywhere in enterprise? Only asking as `Page` includes stuff (title bar, loading screen) I'd guess you'd want to keep consistent in enterprise between core bits and other bits?

jpellizzari (approved):
LGTM :+1:

ui/index.ts
foot: Yeah, maybe to play with? I'm still not sure of what the ideal long term API of this should be. It sounds convenient and consistent to export a layout and page system that we can embrace in EE too. But sometimes you wanna bust out..

foot (approved):
Indeed, very tidy 🧹 ⭐

ui/index.ts
ozamosi: I've exported the `Page` wrapper in a sneaky post-approval-pre-merge commit.

PR #1952, opened 2022-04-19
Cross cluster querying

luizbafilho:
Closes #1952 

**What changed?**
This PR aims to answer the question: "_How can we efficiently query resources across all clusters with users' credentials?_"

That was done using couple strategies:
- We periodically fetch all clusters and save the values in memory
- We periodically fetch all namespaces from all clusters and save the values in memory
- When a user requests a list of resources, we loop through all clusters and ask their `kube-api` which namespaces that particular user has access and save those values in memory with an arbitrary TTL. Next time a request of that given user comes through we use the cached namespaces for the queries.

The entity responsible for all that coordination is the [ClientsFactory](https://github.com/weaveworks/weave-gitops/pull/1952/files#diff-73a66520ff9e82583a0c33f086cb8790ceb36faf9b345cd18b35787b99a8ccb2R29). It is a long running process that periodically queries the clusters, namespaces and when returning the list of clients to be used, checks the namespaces the user has permissions before returning.

In possession of the individual impersonated kube clients and the namespaces the user have access to, the [Clusters Client](https://github.com/weaveworks/weave-gitops/pull/1952/files#diff-adf09190525c7aebd977cb99153ff5ab0852d1e026d2d059e1d62f6de13a3156R20)  is responsible to actually do the querying and make sure all calls run in parallels to maximize efficiency.

Pagination support was added but with caveats given the nature of Kube API and the parallelization of calls. The `Limit` param usually specified by the Client caller to limit the number of total items returned in query will have a different meaning here. 

Since we have to dispatch the call to a `N` number of clusters and namespaces, its hard to coordinate all those calls to return at max that number, given it would require kube-api to provide functionally it doesn't support yet if ever will. As a trade off, I decide to use that limit as the params for all calls, which means will endup with a total number o itens of `Limit * Num. Clusters * Num. Namespaces`.  

Given the goal of pagination is to reduce the pressure on Kube API of calling a huge number of resources at once and the desire to minimize the load time the page will have, I think that choice accomplishes that.

- Manual and Automated tests

**Release notes**

**Documentation Changes**


cmd/gitops-server/cmd/cmd.go
```diff
@@ -211,7 +214,15 @@ func runCmd(cmd *cobra.Command, args []string) error {
...
-	coreConfig := core.NewCoreConfig(log, rest, cacheContainer, clusterName)
+	fetcher, err := fetcher.NewSingleClusterFetcher(rest)
+	if err != nil {
+		return err
+	}
+
+	clusterClientsFactory := clustersmngr.NewClientFactory(rawClient, fetcher, nsaccess.NewChecker(nsaccess.DefautltWegoAppRules), log)
+	clusterClientsFactory.Start(context.TODO())
+
+	coreConfig := core.NewCoreConfig(log, rest, cacheContainer, clusterName, clusterClientsFactory)
```
yitsushi, line 225: Do we still need the `cacheContainer`?

core/clustersmngr/factory.go
```diff
@@ -0,0 +1,221 @@
...
+
+func restConfigFromCluster(cluster Cluster) *rest.Config {
+	return &rest.Config{
+		Host:            cluster.Server,
+		BearerToken:     cluster.BearerToken,
+		TLSClientConfig: cluster.TLSConfig,
+	}
+}
+
+func (cf *ClientsFactory) userNsList(ctx context.Context, user *auth.UserPrincipal) map[string][]v1.Namespace {
```
yitsushi, line 133: How can an endpoint fetch available namespaces? With `cacheContainer` is was exposed not only to access them when making queries, but the `ListNamespaces` and the `GetFluxNamespace` uses the raw list of namespaces.

yitsushi (commented):
Overall looks good to me. With the one giant factory it, over time it may grow bigger when we want to add more items to watch/cache, but that's a potential-future-problem, because maybe it will never grow that big.

cmd/gitops-server/cmd/cmd.go
luizbafilho: we don't need it anymore, I was planning to do the clean-up of unused code in another PR to minimize the size of this one which is already big enough.

core/clustersmngr/factory.go
luizbafilho: Regarding `ListNamespaces` and `GetFluxNamespace` those are unused and probably introduced early on when we needed something in place to keep going, but they should not be necessary and can be removed.

jpellizzari: I think we need `GetFluxNamespace` to get the `Flux Runtime` page to work. If we aren't using that now, we should be.

core/server/kustomization_test.go
```diff
@@ -104,7 +106,7 @@ func TestListKustomizationPagination(t *testing.T) {
 		},
 	})
 	g.Expect(err).ToNot(HaveOccurred())
-	g.Expect(res.Kustomizations).To(HaveLen(2))
+	g.Expect(res.Kustomizations).To(HaveLen(existingKust + 2))
```
josecordaz, line 109: @luizbafilho The refactor you did on pagination looks really cool. It's a lot cleaner and easier to read. 

Just a small question. Should this assertion be `g.Expect(res.Kustomizations).To(HaveLen(existingKust + 1))`instead? I mean, in the previous call to `ListKustomization` you have PageSize = 1.

luizbafilho: that's the caveat with this whole multicluster, multinamespace querying, we need to multiplicate things, so it's accounting for 1 kustomization for each of the namespaces, summing 2.

josecordaz: Hmm. I see, So the `PageSize` parameter means, in this case, "bring me 1 kustomization from each namespace" ?.  To be honest that is different from the original plan. My understanding is that the pageSize is the length of the page  I'm requesting (as long as there are enough resources of course).

luizbafilho: That's the natural paging system, but that's not doable when having multiple parallel requests going for each cluster and each namespace, it would require them to be sequential which is not viable.

We can think of our paging just a way to avoid user waiting too much before getting anything and our goal just to get a bunch more data when the user presses the `next` button.

josecordaz: Hmm, that's interesting. 

Ok, just recapitulating a bit to see if I understood this, properly. 

The problem we are trying to solve is `user waiting too much before waiting too much`, I agree 100% ✅. Now, the approach you are following to solve that, is by having `multiple parallel requests`, where each request would be paginated(limited to a specific size), correct?. 

If so, then I must say that I really like this approach because it is aligned to what we are trying to solve. 

However, I see a few implications to consider.

1) Sorting. Let's pretend limit is 2 (2 kustomizations from each namespace)

If the first would look like this:
```
Namespace KustomizationName
      ns1             a1
      ns1             a2
      ns2            b1
      ns2            b2
```

and second page like this:
```
Namespace KustomizationName
      ns1             z1
      ns1             z2
      ns2            z1
      ns2            z2
```

I wonder if it would be a good user experience to still receive kustomizations from namespace `ns1` in the second page. I personally, after seeing page 1 elements, would think that there should be no more kustomizations for namespace `ns1` in the next page. In the previous approach you will get the pages this way:

Page  1
```
Namespace KustomizationName
      ns1             a1
      ns1             a2
      ns1             z1
      ns1             z2
```

Page 2
```
Namespace KustomizationName
      ns2            z1
      ns2            z2
      ns2            b1
      ns2            b2
```

Here, everything is sorted first by namespace and then by the name of the kustomization, ensuring that every page is alphabetically ordered with each other. Which can solve the issue exposed in the previous statement.

2) Page size. Somehow, in the UI. We would need to tell the user, how many kustomizations are going to be available in every page. In the future this could be even configurable. Following the approach on this PR, we can't really ensure that amount of items per page, even if the actual total amount of kustomizations(across namespaces) can be divided into many pages. Unless we explicitly say that this limit is per namespace, which could also lead to have different number of kustomizations in each page, as namespaces may not have the exact amount requested in the PageSize parameter.

3) Complexity. In terms of pagination, I think having simultaneous requests add a bit of unnecessary complexity. If we want the sort to use namespace name and kustomization name, then we can simply just make 1 request and that would be it (assuming enough kustomization to fullfil pageSize). And continue from there in each consecutive request.

Proposal

Now my approach has its disadvantages as well. If I want to query 100 kustomizations across 10 namespaces that have 10 kustomizations each. That would result, in a very slow request as I will be making 10 request to the k8s api sequentially.  

So, considering all this problems from both approaches. I strongly believe we can combine both approaches and have a more complete solution. The key is going to rely on knowing the total amount of kustomizations in each namespace. If we have this information available then, considering the parameters in the page request we can calculate how many requests I will need and to which namepaces pointing it to. 

For instance: 

Data
```
Namespace   Kustomization Name
       .                          
       .
       .
       ns1                      a1
       ns1                      b1
       ns1                      c1
       ns2                      a2
       ns2                      b2
       ns2                      c2
       ns3                      a3
       ns3                      b3
       ns3                      c3
       ns4                      a4
       ns4                      b4
       ns5                      a5
       ns5                      b5
       ns5                      c5
```

Request parameters
```
pageSize: 6
pageToken: XYZ
```

Now, let's say the pageToken received say that we need to continue from namespace namespace `ns3` and just after kustomization `b3`. As the parameter pageSize is `6` it means we need to query 6 kustomizations starting from there. So, by knowing the total amount of kustomizations in each namespace we can calculate that we will need to make 3 parallel request:

1.- Request 1 kustomizations from namespace `ns2` --> Will return `c3`
2.- Request 2 kustomizations from namespace `ns3` --> Will return `a4` and `b4`
3.- Request 3 kustomization from namespace `ns5` --> Will return `a5`, `b5` and `c5`

Now, the 3 previous implications are personal.  I would like to ask @sympatheticmoose his point of view regarding this matter, as I could be wrong on these assumptions. I'm also open to discuss this further in a meeting with the  rest of the team and product. 

I must say, I also have the plan to talk with product next week, about some other issues I saw on my spike regarding pagination. I'll leverage that talk and I'll bring up your approach as well. I just would like to involve everyone on this and conclude something before anyone can start writing more code about it. This way we can safe people's effort and time.

josecordaz (commented):
On a final note regarding pagination, I'm ok if this gets merged like it is, just to avoid having this PR open for too long. We can implement whatever we come up with afterwards.

charts/gitops-server/templates/role.yaml
```diff
@@ -35,4 +35,7 @@ rules:
   - apiGroups: [ "" ]
     resources: [ "namespaces" ]
     verbs: [ "get", "list" ]
+  - apiGroups: [ "apps" ]
```
SamLR, line 38: Why are you granting more permissions to the service account role? Shouldn't the call to query deployments be using the user's permissions via impersonation?

As far as I can tell this is being added to fix tests and we shouldn't be adding permissions to the chart to fix those.

luizbafilho: While I was implementing the `Flux Runtime` page, I realize that most users won't have access to listing deployments in the flux namespace, causing them to always see a red box with an error message, so I added that permission and used the service account credentials to query for deployments in that specific page only, everything else uses impersonated clients.

SamLR: That feels like how you'd want it to operate though? It might be that we want to surpress that specific error message for users we think just can't see that namespace but if I've not granted a user permissions to see deployments in a specific namespace I don't want them to be able to see those deployments

core/clustersmngr/factory.go
```diff
@@ -0,0 +1,306 @@
...
+	clusters *Clusters
+	// the lists of all namespaces of each cluster
+	clustersNamespaces *ClustersNamespaces
+	// lists of namespaces accessible by the user on every cluster
+	usersNamespaces *UsersNamespaces
+}
+
+func NewClientFactory(hubClient client.Client, fetcher ClusterFetcher, nsChecker nsaccess.Checker, logger logr.Logger) ClientsFactory {
+	return &clientsFactory{
+		hubClient:          hubClient,
```
Callisto13, line 60: is this used at all?

core/clustersmngr/factory.go
```diff
@@ -0,0 +1,306 @@
...
+		go func(clusterName string, c client.Client) {
+			defer wg.Done()
+
+			nsList := &v1.NamespaceList{}
+
+			if err := c.List(ctx, nsList); err != nil {
+				cf.log.Error(err, "failed listing namespaces", "cluster", clusterName)
+			}
+
+			cf.clustersNamespaces.Set(clusterName, nsList.Items)
```
Callisto13, line 131: do we want to leave in the case of error? what would happen in the case that `nsList.Items` is `0` (can it be `nil`?) and it is added to the map?

core/clustersmngr/factory.go
```diff
@@ -0,0 +1,306 @@
...
+			return nil, fmt.Errorf("failed creating client for cluster %s: %w", cluster.Name, err)
+		}
+
+		clients[cluster.Name] = c
+	}
+
+	return clients, nil
+}
+
+type Clusters struct {
```
Callisto13, line 234: nit: can we move everything from here down into like `x_cache` files? not bothered if you'd rather not, just a preference for keeping things tidy and organised away from the main "drama" of the file

core/clustersmngr/factory_test.go
```diff
@@ -0,0 +1,118 @@
...
+	g.Expect(err).To(BeNil())
+
+	err = clientsFactory.UpdateNamespaces(ctx)
+	g.Expect(err).To(BeNil())
+
+	_, err = clientsFactory.GetImpersonatedClient(ctx, &auth.UserPrincipal{ID: "user-id"})
+	g.Expect(err).To(BeNil())
+
+	t.Run("checks all namespaces in the cluster when through the filtering", func(t *testing.T) {
+		g.Expect(nsChecker.FilterAccessibleNamespacesCallCount()).To(gomega.Equal(1))
```
Callisto13, line 44:
```suggestion
		g.Expect(nsChecker.FilterAccessibleNamespacesCallCount()).To(Equal(1))
```

think you have some other `gomega`s around too, not the end of the world

core/server/pagination_test.go
```diff
@@ -24,7 +24,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 )
 
-func TestPagination(t *testing.T) {
+func XTestPagination(t *testing.T) {
```
Callisto13, line 27: :sus:

;)

Callisto13 (commented):
looks awesome!

this is all have time for today, if it is still open tomorrow i will review again 👍

core/clustersmngr/factory.go
```diff
@@ -0,0 +1,306 @@
...
+	for _, cluster := range cf.clusters.Get() {
+		if err := pool.Add(restConfigFromCluster, cluster); err != nil {
+			return nil, fmt.Errorf("failed adding cluster client to pool: %w", err)
+		}
+	}
+
+	return NewClient(pool, cf.clustersNamespaces.namespaces), nil
+}
+
+func (cf *clientsFactory) GetClustersNamespaces() map[string][]v1.Namespace {
```
josecordaz, line 164: It seems this function is now being used anywhere. Is this for a specific purpose?

josecordaz (approved):
Impresive work 💯

josecordaz:
The requests made by the UI seem to return faster. Currently in the `main` branch, the responses from sources and flux runtime tabs are consistent every time. They take around ~11seconds for the Sources tab and ~10seconds in Flux Runtime tab. But with this PR the first response in the Sources tab takes around ~6seconds and the subsequent requests take around ~2.5 seconds. On the Flux Runtime tab the first request took 4.5seconds and the subsequent request took ~1.5 seconds. So pretty awesome results. Great work.

cmd/gitops-server/cmd/cmd.go
```diff
@@ -211,7 +214,15 @@ func runCmd(cmd *cobra.Command, args []string) error {
 		),
 	)
 
-	coreConfig := core.NewCoreConfig(log, rest, cacheContainer, clusterName)
+	fetcher, err := fetcher.NewSingleClusterFetcher(rest)
```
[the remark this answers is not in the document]
luizbafilho: That's an interface used on the enterprise side, and that implementation might return errors.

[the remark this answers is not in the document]
Callisto13: This isn't an interface. The `Fetch` used later will be. The `NewSingleClusterFetcher` and the `NewMultiClusterFetcher` are unique funcs, so you can do whatever you like with this, so long as it returns a `ClusterFetcher`.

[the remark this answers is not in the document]
luizbafilho: ohh, indeed, I got myself confused with the `ClusterFetcher` interface. Sounds good, I'll remove the err.

core/server/kustomization_test.go
josecordaz: Right, those are the kind of issues I was referring to. Kube API actually supports some counting, sorting and filter functionality, but they come with some important limitations. That's why I have a meeting setup for next Wednesday with Product, to express these restrictions. 

Now, just to make sure I'm understanding properly. When you say it would be ideal to rely on the UI for the sorting, counting, and filtering, what do you actually mean? Are you saying that the UI will have all the data, maybe in memory(e.g. all kustomizations) and then will do the sorting/counting/filtering in there?.

> The only way to make that happen would have all data from the cluster synced to a central queriable data source, but I have the impression this is out of the table due to security constraints.

Interesting. But I'm confused. What are the security constraints you are referring to?. Are you, perhaps, referring to the location of that data source? Would the cluster itself be insecure ? Would it be safe to use RBAC to secure that data? (k8s student here 😋).

> But if you find I way to do it, we can discuss it later and implement it in follow-up PR.
Actually, my plan is to first ask Product what we really want regarding this matter and then probably have a brainstorm meeting with the team, so everyone is involved and we try to come up with some ideas on how to implement it.

luizbafilho: Do you mind dropping documents on how you sort a query or only get the number of items of a particular list without making the list call?

I didn't say it would be _ideal_ to really be on the ui for sorting, counting, and filtering.  I've said that since the Kube API
> does not provide any functionality regarding counting, sorting, or filtering, so that must be done in the UI only

To accomplish that kind of filtering or searching, we would need to centralize all the data from all the clusters in one single place (ie. elasticsearch) and use that as a data source instead of querying the clusters on the fly. Having all that data in one single place is potentially dangerous and hard to manage properly so users probably want to avoid that, we would also need to build our RBAC to guarantee the users only access the data they can.

core/clustersmngr/factory.go
luizbafilho: nope, removing it.

luizbafilho: Ops sorry it is. in the Flux Runtime page

core/clustersmngr/factory.go
luizbafilho: fair enough

core/clustersmngr/factory.go
josecordaz: Ok I think I'm going to blame my IDE for not finding that reference lol

core/server/kustomization_test.go
luizbafilho: On the counting is that there is no specific call to count things as we would in SQL (`SELECT COUNT(*) FROM table_name;`) and we would that number even before making the first call to distribute the `limit` across the clusters/namespaces calls.

And as the paragraph states, if we use any label or field selector the `remainingItemCount` won't be there, so we can't rely on it.

SamLR (changes requested):
marking this as request changes pending outcome of chat re. granting get deployments to the service account

charts/gitops-server/templates/role.yaml
sympatheticmoose: I take your point @luizbafilho on the tab being useless if the user lacks the permissions to view deployments in the flux-system namespace, however I think we're better off leaning more towards consistent use of impersonation for now. This is something we could speak to users/customers about and change later based on their feedback. As an alternative, could this be configurable? Namely to document changes that an administrative user (one who deploys Weave GitOps) could make to enable the use of the service account with increased permissions?

core/clustersmngr/factory.go
```diff
@@ -0,0 +1,230 @@
...
+
+func (cf *clientsFactory) watchClusters(ctx context.Context) {
+	if err := wait.PollImmediateInfinite(watchClustersFrequency, func() (bool, error) {
+		if err := cf.UpdateClusters(ctx); err != nil {
+			return false, err
+		}
+
+		return true, nil
+	}); err != nil {
+		cf.log.Error(err, "failed pooling clusters")
```
Callisto13, line 81:
```suggestion
		cf.log.Error(err, "failed polling clusters")
```

core/clustersmngr/factory.go
```diff
@@ -0,0 +1,230 @@
...
+
+	// list of clusters returned by the clusters fetcher
+	clusters *Clusters
+	// the lists of all namespaces of each cluster
+	clustersNamespaces *ClustersNamespaces
+	// lists of namespaces accessible by the user on every cluster
+	usersNamespaces *UsersNamespaces
+}
+
+func NewClientFactory(hubClient client.Client, fetcher ClusterFetcher, nsChecker nsaccess.Checker, logger logr.Logger) ClientsFactory {
```
Callisto13, line 57:
```suggestion
func NewClientFactory(fetcher ClusterFetcher, nsChecker nsaccess.Checker, logger logr.Logger) ClientsFactory {
```
this can be removed as an arg now

charts/gitops-server/templates/role.yaml
luizbafilho: Sounds good, that's probably doable, we can investigate it in a follow up issue.

core/clustersmngr/factory.go
luizbafilho: no problem it will just be set as an empty list and web move on since we need to get the other namespaces

charts/gitops-server/templates/role.yaml
luizbafilho: @SamLR access removed, using user's credentials instead.

core/clustersmngr/factory.go
Callisto13: sweet 👍

Callisto13 (approved):
Great work!

PR #1963, opened 2022-04-20
ref: Remove kube.Kube interface and implementation

Callisto13:
Nothing in either repo was using it.

I have also removed the KubeGetter interface and implementation for the
same reason.

I then refactored kube.KubeHTTP further, as by this point the only thing
it added was the ClusterName. Therefore kube.KubeHTTP is now simply a
wrapper around client.Client which adds a little extra. It was trivial
to update uses throughout the code.

We can add this back if we need it, but it may end up making more sense
to create something new to fit our needs when they arise.


josecordaz (approved):
LGTM

PR #1964, opened 2022-04-20
Sizing guidelines

fire-ant:
**What changed?**

Adds a devcontainer so I dont need a local environment to hack on docs.
Adds some sizing guidelines to the docs so we can provide a rule of thumb around WGE installations

**Why was this change made?**

Cluster API does not provide management cluster sizing and does not account for any additional overhead such as running the Flux controllers or other components. We should maintain guidelines specific to the needs of users which we can update as necessary.

learned me some Docusaurus and just tried to do the right thing

As its a docs change so no tests necessary (going on the sage wisdom of other engineers)

**Release notes**

N/A

**Documentation Changes**

N/A


sympatheticmoose:
Thanks @fire-ant - it would be great if you could add a few lines to reference the new devcontainer and its usage for contributing to the docs here https://github.com/weaveworks/weave-gitops/tree/main/website

fire-ant:
updated - appreciate the nudge

makkes:
@fire-ant This PR hasn't seen any movement since April. Is this still something you'd like to see merged?

LappleApple:
@fire-ant pinging with same question as @makkes asked -- what would you like us to do?

PR #1972, opened 2022-04-21
ui: Make CoreClientContext monkeypatch work

ozamosi:
I cannot figure out any way that this ever worked. Whenever
    CoreClientContext was initialised, it tried to loop over the Core
    object's iterable methods. However, it has none, so it just ignored
    the API and returned an empty object. This then means that anything
    trying to use the API will fail.

The fix here is to use `getOwnPropertyNames` instead - static methods
aren't iterable, so we have to use something more dramatic. However,
that also returns builtins like `prototype`, `name`, `length`, and we
don't want to monkeypatch those, so just monkeypatch methods.

I've written a test that fails with the old code and passes with the
new code.

### DIFF
--- a/ui/contexts/CoreClientContext.tsx
+++ b/ui/contexts/CoreClientContext.tsx
@@ -22,16 +22,19 @@ export default function CoreClientContextProvider({ api, children }: Props) {
   const wrapped = {} as typeof Core;
 
   //   Wrap each API method in a check that redirects to the signin page if a 401 is returned.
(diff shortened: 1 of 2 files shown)


jpellizzari (approved):
Nice work.

mnowster (approved):

ozamosi:
Yeah, I can't get it to reproduce when it's compiled to commonjs for enterprise, either. So, yes, very strange.

Issue #1939, opened 2022-04-17
Helm-related sub-pages from Source page show errors


rokshana-b:
**Describe the bug**

When a user navigates from the Source page to Helm-related sub pages, the pages display 404 errors. 

**Screenshots**

https://user-images.githubusercontent.com/76788247/163730136-155c7917-1818-4ac8-8c7e-8552e8737272.mp4

**Is this a UI bug or a server bug?**
- [x] UI
- [ ] Server

**What is the severity of the bug**

- [ ] `severity/Critical`: Weave GitOps is crashing or experiencing data loss, the UI is inaccessible or a key feature is unusable. There is no known workaround
- [x] `severity/Major`: Weave Gitops functionality is broken, there is a workaround, but the workaround requires significant effort
- [ ] `severity/Minor`: Weave Gitops functionality is broken, but there is a fairly straightforward workaround
- [ ] `severity/Low`: Doesn’t affect primary flow/functionality but would be good to fix

**Environment**

 - gitops: v0.7.0
 - How you deployed the Weave GitOps server: Helm Chart
 - kubernetes:  [e.g. 1.20.4]
    - [ ] KinD - _version_]
    - [x] cloud : GKE

**To Reproduce**

1. Go to Source page
2. Click on Helm chart & Helm Release sources
3. See error 404 errors on the pages

**Expected behavior**

UI should display helm chart/release source details just like it does for git repositories.

rokshana-b:
This has not been fixed. I am on main, my gitops binary version: 

```
Current Version: v0.7.0-rc12-68-gac08cb52
GitCommit: ac08cb52
BuildTime: 2022-04-20_12:49:11
Branch: main
```


https://images.zenhubusercontent.com/342592099/4bde2207-2d33-48ca-bb23-61748efd8532/screenshare___2022_04_20_3_57_27_pm.mp4

rokshana-b:
@jpellizzari Helm pages show up now:

https://images.zenhubusercontent.com/342592099/02a84a22-617a-48a9-a078-0be5912d96e6/screenshare___2022_04_21_4_10_51_pm.mp4

jpellizzari:
Closed by #1973

PR #1973, opened 2022-04-21
Fix sub page navigation

jpellizzari:
Closes:
* #1939 

The routing logic needed to be changed to not be `exact` now that we are using path-based tabs.

Also fixes a double scrolly:
Screenshot from 2022-04-21 07-54-03

### DIFF
--- a/ui/App.tsx
+++ b/ui/App.tsx
@@ -59,12 +59,10 @@ const App = () => (
           component={withSearchParams(GitRepositoryDetail)}
         />
         <Route
-          exact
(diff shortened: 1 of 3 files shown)


joshri (approved):
Setting overflow to scroll is the cause of the double scrolly? Noted.

jpellizzari:
> Setting overflow to scroll is the cause of the double scrolly? Noted.

There is some other element that is scrolling in this case, but yes, setting `overflow: scroll` means the bar is always rendered. `auto` means it will only show if there is overflow.

PR #1975, opened 2022-04-21
release: Fix github action for tagging docker images

ozamosi:
Because this pipeline is triggered by a review, not by a tag, it can't
use the plain auto-discovery of tags. Set explicit values.

This tries to preserve the old behaviour of tagging an image both
e.g. `v0.7.1` and `0.7.1`.


mnowster (approved):
pair reviewed with @ozamosi

PR #1978, opened 2022-04-21
Final* small release process changes

ozamosi:
First, it fixes the prepare-release job so it actually triggers the release as an external user, so that CI actually runs. It then sets the exclude from release notes label, because this PR isn't doing anything interesting. And as a result, the whole PR actually turns green: #1977

Second, it rewrites the release process documentation to front-load more of my implicit knowledge.

* probably not actually final.


mnowster (approved):
Thanks for updating the release process docs 👍🏻  I'll give it another run through on the next release.

PR #1983, opened 2022-04-22
Simplify flags - WIP

AlinaGoaga:
Closes

**What changed?**

**Why was this change made?**

**Release notes**

**Documentation Changes**


pkg/server/handler.go
```diff
@@ -43,12 +34,10 @@ func NewHandlers(ctx context.Context, log logr.Logger, cfg *Config) (http.Handle
 	mux := runtime.NewServeMux(middleware.WithGrpcErrorLogging(log))
 	httpHandler := middleware.WithLogging(log, mux)
 
-	if AuthEnabled() {
-		clustersFetcher := fetcher.NewSingleClusterFetcher(cfg.CoreServerConfig.RestCfg)
+	clustersFetcher := fetcher.NewSingleClusterFetcher(cfg.CoreServerConfig.RestCfg)
```
Callisto13, line 37: you can probably remove the fetcher bits as they should not live there anymore

Callisto13: unless @luizbafilho is this something you are taking out now?

website/versioned_docs/version-0.7.0/configuration/securing-access-to-the-dashboard.mdx
```diff
@@ -11,13 +11,6 @@ There are 2 supported methods for logging in to the dashboard:
 
 The recommended method is to integrate with an OIDC provider, as this will let you control permissions for existing users and groups that have already been configured to use OIDC. However, it is also possible to use a cluster user account to login, if an OIDC provider is not available to use. Both methods work with standard Kubernetes RBAC.
 
-:::note FEATURE TOGGLE
-The following instructions describe a feature that is behind a feature toggle. To enable this feature, set the following OS environment variable:
-```sh
-export WEAVE_GITOPS_AUTH_ENABLED=true
-```
-:::
-
```
foot, line 20: Let's leave this in there for the 0.7.0 docs, 0.7.0 release probably requires it?

foot:
It think we need to bump this https://github.com/weaveworks/weave-gitops/blob/7cf5fdc1048f85276461da397116da667bb50a49/charts/gitops-server/Chart.yaml#L17-L18 to 2.0.3 or something, then LGTM :)

ui/pages/Error.tsx
```diff
@@ -2,32 +2,24 @@ import * as React from "react";
...
 import Flex from "../components/Flex";
-import { FeatureFlags } from "../contexts/FeatureFlags";
-import LoadingPage from "../components/LoadingPage";
 
 type Props = {
   className?: string;
   code?: string;
 };
 
 function ErrorPage({ className }: Props) {
```
foot, line 11: Should pass `className` into `Content` I guess

foot (approved):
🧹 🧹 🧹 🧹 !

sympatheticmoose:
FYI @AlinaGoaga @foot this appears to be causing issues where the user is not directed to login in certain circumstances. We have issues that by default our instructions do not pin to a specific chart or image version

foot:
🤔 , I think this change will hopefully help with https://github.com/weaveworks/weave-gitops/issues/2171 actually.

I don't think helm will install `-rc.x` releases unless specifically asked so I'd guess that was still 0.8.0 proper.

sympatheticmoose:
It does, but not whilst folks get the 2.0.7 chart and not 0.8.1 image

sympatheticmoose:
Yeah so with standard install, i.e. `flux create source helm ww-gitops --url=https://helm.gitops.weave.works` and then helm release which looks like:

```
kind: HelmRelease
metadata:
  name: ww-gitops
  namespace: flux-system
spec:
  chart:
    spec:
      chart: weave-gitops
      sourceRef:
        kind: HelmRepository
        name: ww-gitops
  interval: 1m0s
  values:
    adminUser:
      create: true
      passwordHash: $2a$10$Bex8ghY2tdfG5VaNHiBhhO6Mtpx1QJk2Xh5O6bi1LPpqyx8qXlfQK
      username: admin
    rbac:
      create: true
```

what you'll get is:
```
flux get all
...
helmchart/flux-system-ww-gitops	2.0.7   	False    	True 	pulled 'weave-gitops' chart with version '2.0.7'
```

&& 

```
k describe deployment ww-gitops-weave-gitops -n flux-system
...
  weave-gitops:
    Image:      ghcr.io/weaveworks/wego-app:v0.8.0
```

foot:
Right! I see, so helm-chart is published on merge to main, but image ref is not updated in the chart until a release?

sympatheticmoose:
bingo - raised https://github.com/weaveworks/weave-gitops/issues/2180 to potentially address this

sympatheticmoose:
given all this, it doesn't feel like this change was only a patch... feels breaking and more like v3.0.0?

PR #1984, opened 2022-04-22
Move javascript out of the server binary

ozamosi:
The server binary now just loads files from disk, instead of
files embedded into the binary.

This has the advantage that the build process becomes slightly more
straightforward - we can compile the go binary without building
javascript, and we can refresh the javascript in docker images without
re-building the go binary.

I've hard-coded the server binary to look at the sub-directory
"dist" below the binary's directory, and dumped the generated files in
`bin/dist` - it's a bit of an odd place to put the dist files, but
`bin` is a directory of generated files already, and this makes it
very easy to accidentally get right.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -68,9 +68,6 @@ fakes: ## Generate testing fakes
 install: bin ## Install binaries to GOPATH
 	cp bin/$(BINARY_NAME) ${GOPATH}/bin/
 
-api-dev: ## Server and watch gitops-server, will reload automatically on change
-	reflex -r '.go' -R 'node_modules' -s -- sh -c 'go run -ldflags "$(LDFLAGS)" cmd/gitops-server/main.go'
-
 cluster-dev: ## Start tilt to do development with wego-app running on the cluster
 	./tools/bin/tilt up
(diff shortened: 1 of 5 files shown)


SamLR (approved):
Looks fine. One question: is it worth having the server cache the assets rather than always re-read them from disk?

ozamosi:
I'm gonna go with "probably not". I don't think it'll _actually_ re-read from disk every time - the files will land in the linux page cache, right? By the time we've got so many users and features that the performance of this becomes a problem, I'll want to pull down an off-the-shelf load balancer to deal with all the other requirements we'll have by then.

SamLR:
> I'm gonna go with "probably not". I don't think it'll _actually_ re-read from disk every time - the files will land in the linux page cache, right? By the time we've got so many users and features that the performance of this becomes a problem, I'll want to pull down an off-the-shelf load balancer to deal with all the other requirements we'll have by then.

That was mostly my thinking. I guess we should keep it in mind but this shouldn't ever be a high traffic app so we can probably err on the side of simplicity

PR #1987, opened 2022-04-22
refactor: clean up `pkg/models`

Callisto13:
And rename the package, since there are no models or widely used consts
there anyway. One const is used in Enterprise, but since that const was
a dupe of another, they will be able to switch to it on the next bump.


josecordaz (approved):
LGTM

Issue #1900, opened 2022-04-07
Flickering on navigation


jpellizzari:
Navigating causes flickering of the page.

Refactor to use `useQuery` to make routing work if things are cached?

sympatheticmoose:
believe this has been resovled

PR #1989, opened 2022-04-22
Update auth check

AlinaGoaga:
Switch to checking user info instead of loading to stop flickering when navigating

### DIFF
--- a/ui/contexts/AuthContext.tsx
+++ b/ui/contexts/AuthContext.tsx
@@ -26,7 +26,7 @@ export const AuthCheck = ({ children, Loader }: AuthCheckProps) => {
   const { loading, userInfo } = React.useContext(Auth);
 
   // Wait until userInfo is loaded before showing signin or app content
-  if (loading) {
+  if (!userInfo) {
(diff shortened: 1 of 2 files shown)


jpellizzari:
Maybe Closes https://github.com/weaveworks/weave-gitops/issues/1900

AlinaGoaga:
Indeed, thanks @jpellizzari! :)

foot:
> Maybe Closes https://github.com/weaveworks/weave-gitops/issues/1900

I think so. We could refactor the whole AuthCheck thing to useQuery too where `isLoading` behaves a bit more in-align w/ what we're after. This is a bit of a quick-fix so we can get a release out the door. 😬

foot (approved):
💯 

There is still a small jank on the first load, I think as the feature flags come in and force a re-render. That will hopefully be fixed in #1983.

Still, feels much better!

AlinaGoaga:
Yup, I've got that on the list - make more use of `useQuery` (was going to work on it as part of the refactoring on [https://github.com/weaveworks/weave-gitops/pull/1983](url)) ;)

PR #1993, opened 2022-04-22
Align titles, fix fonts, and link logo to applications page

### LINKED ISSUE #1971: [UI] Titles and menu items should be semi-bold, not bold
Like this
Screenshot 2022-04-20 at 16 24 19

Instead of
Screenshot 2022-04-20 at 16 24 31

joshri:
Closes #1971 

The page content title is now semi bold, as it should be. I also threw in some other small changes, such as aligning the title with the nav tabs and breadcrumbs, which was a huge pet peeve of mine, and adding a link to the logo that sends the user to the applications page. The extra padding on the user icon is so that the hover state doesn't brush up against the edge of the page.

image

### DIFF
--- a/ui/components/InfoList.tsx
+++ b/ui/components/InfoList.tsx
@@ -25,9 +25,6 @@ const InfoList = styled(
     );
   }
 )`
-  table {
-    border-spacing: 0;
-  }
   tbody tr td:first-child {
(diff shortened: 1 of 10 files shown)


josecordaz (approved):
LGTM

PR #1994, opened 2022-04-22
Fix HelmRlease typo in docs

kingdonb:
**What changed?**
Typo fix in docs - I am not certain if this will merge or if there are other generated entries that must be updates together as well!


SamLR (approved):

SamLR:
good catch!

PR #2007, opened 2022-04-26
Remove unused code in pkg/utils

ozamosi:
This removes a few generic functions that aren't used, as well as one
small dependency.


Callisto13 (approved):
dammit you beat me to it!

Issue #1959, opened 2022-04-20
Warn Users when running with `--insecure`


bigkevmcd:
**Describe the bug**
When TLS is disabled, authentication tokens are transmitted without encryption.

**Is this a UI bug or a server bug?**
- [X] UI
- [X] Server

**What is the severity of the bug**
Unencrypted Authentication Tokens could be captured.

- [ ] `severity/Critical`: Weave GitOps is crashing or experiencing data loss, the UI is inaccessible or a key feature is unusable. There is no known workaround
- [X] `severity/Major`: Weave Gitops functionality is broken, there is a workaround, but the workaround requires significant effort
- [ ] `severity/Minor`: Weave Gitops functionality is broken, but there is a fairly straightforward workaround
- [ ] `severity/Low`: Doesn’t affect primary flow/functionality but would be good to fix

**Environment**
 - gitops: [e.g. v0.1.0]
 - How you deployed the Weave GitOps server: [e.g. Tilt, Helm Chart, etc]
 - kubernetes:  [e.g. 1.20.4]
    - [ ] KinD - _version_]
    - [ ] k3s - _version_
    - [ ] cloud [e.g., EKS, AKS]  _version_
    - [ ] other - _name_ _version_
 - Browser + version: [e.g. chrome 74, safari 12, firefox 87]

**To Reproduce**

**Expected behavior**
Users should be warned that their credentials are not secure when `--insecure` is configured.

**Config and Logs**
If applicable, add logs to help explain your problem. _please compress the output before attaching_
- [ ] Logs from the `wego-app` pod
- [ ] Events from `flux-system` namespace (Or the namespace you deployed flux and/or Weave GitOps)
- [ ] `kubectl cluster-info dump`
- [ ] Prometheus alerts
- [ ] Flux logs

**Screenshots**

**Additional context**

LappleApple:
Still an issue?

PR #2008, opened 2022-04-26
feat: Set vars when TLS disabled or Dev enabled

Callisto13:
This is so that the UI can query what has been set on the server and
display appropriate warning messages.

It will also be logged on each request whether TLS is enabled or not.

Part of https://github.com/weaveworks/weave-gitops/issues/1959

### DIFF
--- a/cmd/gitops-server/cmd/cmd.go
+++ b/cmd/gitops-server/cmd/cmd.go
@@ -100,7 +100,7 @@ func NewCommand() *cobra.Command {
 }
 
 func runCmd(cmd *cobra.Command, args []string) error {
-	log, err := logger.New(options.LogLevel, options.Insecure)
+	log, err := logger.New(options.LogLevel, options.DevMode)
 	if err != nil {
 		return err
 	}
@@ -181,7 +181,8 @@ func runCmd(cmd *cobra.Command, args []string) error {
 		}
 
 		if options.DevMode {
-			log.Info("WARNING: dev mode enabled. This should be used for local work only")
+			log.Info("WARNING: dev mode enabled. Authentication will be bypassed in some instances. This should be used for LOCAL WORK ONLY.")
+			os.Setenv(server.DevModeFeatureFlag, "true")
 			tsv.SetDevMode(options.DevUser)
 		}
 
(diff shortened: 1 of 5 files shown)


cmd/gitops-server/cmd/cmd.go
```diff
@@ -286,7 +287,10 @@ func runCmd(cmd *cobra.Command, args []string) error {
 
 func listenAndServe(log logr.Logger, srv *http.Server, options Options) error {
 	if options.Insecure {
-		log.Info("TLS connections disabled")
+		log.Info(
+			"WARNING: TLS connections disabled by the `--insecure` flag. All data INCLUDING AUTH TOKENS will be transmitted without encryption.")
```
ozamosi, line 291: I'm not sure this makes sense - we expect `--insecure` to pretty much always be set for security reasons: you'll have a separate ingress to encrypt the data, so you don't need to worry about how to configure allowed ciphers, special-case certificate renewal, and so on. That doesn't mean anything will be transmitted without encryption.

Callisto13: Oh i didn't realised we did that now. In that case, why do we offer TLS at all?

ozamosi: :shrug:

Callisto13: looool excellent

i might be closing this one then

Callisto13:
closing for now, may open again later

cmd/gitops-server/cmd/cmd.go
sympatheticmoose: That other ingress is on the user to configure though right? And I assume we'd have no way of knowing whether they have appropriately configured? In which case a warning still feels valid but perhaps the wording changes to more of a "ensure you have an appropriately configured ingress..."

sympatheticmoose: additionally, if we do indeed expect `--insecure` to be the default in a secure configuration... maybe we need to think about renaming :)

Callisto13: Opened https://github.com/weaveworks/weave-gitops/issues/2010 to discuss what we really want to do around this

Issue #2009, opened 2022-04-26
source links with ssh format attempt to open in terminal


sympatheticmoose:
Trying to click the URL link to go to the appropriate repo 
<img width="700" alt="Screenshot 2022-04-26 at 15 44 42" src="https://user-images.githubusercontent.com/19860021/165327087-8bea1c08-8314-4993-a947-9d6b14faaa13.png">
results in
<img width="666" alt="Screenshot 2022-04-26 at 15 44 37" src="https://user-images.githubusercontent.com/19860021/165327080-5b6245ca-c46b-4c2d-b3d9-ad1445b5a13c.png">

Need to convert to http address for hyperlink

PR #2012, opened 2022-04-26
Convert ssh urls to http

joshri:
Closes: #2009

We had a way to reformat ssh urls, it just never made its way into `GitRepositoryDetail` and `HelmRepositoryDetail`. Do we need to check the helm repo urls? If it's already http it will just return the url as is, so I'm not anticipating these changes breaking anything....is there any other way that a url for a helm repo might be broken/wrong?


sympatheticmoose:
@joshri should just be gitrepositories I believe. To be clear - does your change impact the displayed text or just the hyperlink?

joshri:
> I believe we would want to keep the display text indicating that its configured to be using SSH

Makes sense. Changed!

josecordaz (approved):

PR #2013, opened 2022-04-26
Change doc screenshot to reflect current changes

rokshana-b:
**What changed?**

Dashboard Application Reconciliation picture was updated 

**Why was this change made?**

To reflect current changes

**Before Screenshot**

dashboard-application-reconciliation-012155a9d4afe9f447b0c5dc349ec8ee

**After Screenshot**

dashboard-application-reconciliation


yitsushi (commented):
Two questions:
1. Isn't `3584x2080px`, it will be scaled down to the view of the page and that makes it a bit less readable and the only way to see in full size is to `right click -> open image in new tab`.

2. Do we want to wait until the new rounded design is implemented, or we are good and we will update that image again when it's done?

yitsushi:
Follow-up on image size: Replacing the image with the larger one, that's the difference.

image
image

On the new one it's impossible to read the text on the nodes, but if it's not relevant and it only shows the overall "how it looks like", than it's fine.

yitsushi (approved):

PR #2015, opened 2022-04-26
UI Responsiveness Pass

joshri:
Here's the big list of changes:

A `maxWidth` prop has been added to the `Field` type, allowing longer data table cells (message, name) to wrap. Additionally, the `SourceLink` component has a `short` prop in order to cut the url down inside the data table. One downside of these changes is that it becomes unclear that the table is scrollable/there are additional columns. 
image

The title bar in layout has been removed. Adjustments to the details pages make those look great, but the top level pages look a little empty at the top to me. This is all in service of decreasing the min-width of the app to 768px.

main page: 
image

detail page:
image

The top bar is now set to `position: fixed`. I also set the nav content to fixed but it looked clunky - I want to figure out how to add some kind of delay there or make the entire white bar scroll.
image

Status message in `ReconciledObjectsTable` repeated what's in the message column. Made it consistent with other status indicators and it now displays the `readyText` instead (Ready, Not Ready, Suspended).

Changed the name of the "Relevant Automations" tab to just "Automations".

### DIFF
--- a/ui/components/SourceLink.tsx
+++ b/ui/components/SourceLink.tsx
@@ -7,9 +7,10 @@ import Link from "./Link";
 type Props = {
   className?: string;
   sourceRef?: SourceRef;
+  short?: boolean;
 };
 
-function SourceLink({ className, sourceRef }: Props) {
+function SourceLink({ className, sourceRef, short }: Props) {
   if (!sourceRef) {
     return <div />;
   }
@@ -18,7 +19,7 @@ function SourceLink({ className, sourceRef }: Props) {
       className={className}
       to={formatSourceURL(sourceRef.kind, sourceRef.name, sourceRef.namespace)}
     >
-      {sourceRef.kind}/{sourceRef.name}
+      {!short && sourceRef.kind}/{sourceRef.name}
     </Link>
   );
 }
--- a/ui/components/SourceLink.tsx
+++ b/ui/components/SourceLink.tsx
@@ -19,7 +19,8 @@ function SourceLink({ className, sourceRef, short }: Props) {
       className={className}
       to={formatSourceURL(sourceRef.kind, sourceRef.name, sourceRef.namespace)}
     >
-      {!short && sourceRef.kind}/{sourceRef.name}
+      {!short && sourceRef.kind + "/"}
+      {sourceRef.name}
     </Link>
   );
 }
--- a/ui/components/AutomationsTable.tsx
+++ b/ui/components/AutomationsTable.tsx
@@ -57,7 +57,7 @@ function AutomationsTable({ className, automations, hideSource }: Props) {
       },
       sortValue: ({ name }) => name,
       textSearchable: true,
-      maxWidth: 200,
+      maxWidth: 600,
     },
     {
       label: "Type",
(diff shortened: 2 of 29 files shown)


rokshana-b:
### 1. maxWidth prop:

**before:**

image

**after:**

image

### 2. short prop:

**before:**

image

**after:**

image

### 3. title bar in layout:

**before:**

image


**after:**

image

### 4. new status position:

**before:**

image


**after:**

image

### 5. status message duplication:

**before:**

image

**after:**

image

sympatheticmoose:
minor nit - but the `/` prefix in the shortened source name looks weird to me.

What happened to the bounding box around the status message? And what happens when theres a larger message? Does it push everything down?

sympatheticmoose:
hang on - 4. new status position - you've completely lost the name and namespace of the object you're looking at?

joshri:
> minor nit - but the `/` prefix in the shortened source name looks weird to me (makes it feel like a path) and seems unnecessary.
> 
> What happened to the bounding box around the status message? And what happens when theres a larger message? Does it push everything down?

- Bounding box used to be only on error, but I can change it to always on if we'd like.
- Larger error message will wrap (pushing things down/making the page longer) where it used to ellipsis. 
- I think there's a point to be made about order of the info list/how we're highlighting the name and namespace. If it was unclear at first glance it probably needs to be changed

joshri:
Message table cell update: 
image

We can't have horizontal scrolling and conditional wrapping (wrapping only when max-width is reached) at the same time, as far as I can tell. My compromise here is to add an ellipsis at a specified width as a fail safe for cells that might get obscenely long. @JamWils

jpellizzari (commented):
@joshri Why were the page titles removed?

joshri:
> @joshri Why were the page titles removed?

@JamWils I THINK the idea was that there are three versions of the title very close to each other (nav, content, and the top bar), so we should get rid of one. Getting rid of the content title allows the status to be the most highlighted thing on the page

jpellizzari:
> > @joshri Why were the page titles removed?
> 
> @JamWils I THINK the idea was that there are three versions of the title very close to each other (nav, content, and the top bar), so we should get rid of one. Getting rid of the content title allows the status to be the most highlighted thing on the page

Assuming you are responding to me, was there a discussion about this? Or an issue I can look at?

joshri:
> > > @joshri Why were the page titles removed?
> > 
> > 
> > @JamWils I THINK the idea was that there are three versions of the title very close to each other (nav, content, and the top bar), so we should get rid of one. Getting rid of the content title allows the status to be the most highlighted thing on the page
> 
> Assuming you are responding to me, was there a discussion about this? Or an issue I can look at?

Yep - discussion was with James which is why I tagged him

ui/components/SourceLink.tsx
```diff
@@ -18,7 +19,8 @@ function SourceLink({ className, sourceRef }: Props) {
       className={className}
       to={formatSourceURL(sourceRef.kind, sourceRef.name, sourceRef.namespace)}
     >
-      {sourceRef.kind}/{sourceRef.name}
+      {!short && sourceRef.kind + "/"}
```
jpellizzari, line 22: Same as what @sympatheticmoose said: let's remove the leading `/`

jpellizzari (approved):
Code LGTM. The lack of a ellipsis on the `Message` column in the `DataTable` will eventually bite us and cause a mess when a long value comes through, but :shrug:. 

The lack of a page title looks weird (bad), but if that is what product wants then :+1: .

In the future, let's make sure there is an issue created with the deliverables and discussion points.

PR #2021, opened 2022-04-27
Collection of minor auth changes

### LINKED ISSUE #1949: Auth security improvements
During a security review of the auth code we identified the following improvements:
- [x] `JWTAuthorizationHeaderPrincipalGetter` is not currently used and may be removed
- [x] `offline` scope is not needed as we don't refresh the tokens at the moment
- [ ] /oauth2/userinfo should return 401 when no auth
- [x] Reduce the validation secret cookie TTL as it doesn't need to be set to 1h
- [x] Ensure all cookies are set to HttpOnly/Secure

Callisto13:
- feat: Set Secure flag on cookies

- ref: Remove JWTAuthorizationHeaderPrincipalGetter

    Currently we don't make any requests from the UI using the auth header,
    just the cookie is used for now.
    We can add it back when we need it.

-  feat: Reduce State cookie TTL to 5 mins

    Also remove setting the RefreshToken cookie. We don't do anything with
    that token right now, and need to do some work around how we handle a
    user who is no longer authenticated.

- refactor: Remove offline auth scope

Closes #1949


JamWils:
5 minutes seems aggressive?  Can this be a configurable value and it defaults to say 15 minutes?

JamWils:
I totally misunderstood, thanks for clarifying Claudia.  I'm mixing up tokens.

makkes:
closing as stale

PR #2022, opened 2022-04-27
chore: Alert Slack when CI fails on main

Callisto13:
I am not sure if the `if` syntax is correct, running some tests now but if anyone sees an obvious mistake, plz lmk.

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -291,6 +291,26 @@ jobs:
           args: --all-projects --org=product-engineering-ly9
         continue-on-error: true
 
+  notify-failure:
+    name: Send Slack alert
+    needs:
+      - ci-go
+      - ci-static
+      - ci-js
+      - ci-build-gitops-image
+      - ci-upload-images
+      - ci-upload-binary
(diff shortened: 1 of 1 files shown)


.github/workflows/pr.yaml
```diff
@@ -291,6 +291,26 @@ jobs:
...
 
+  notify-failure:
+    name: Send Slack alert
+    needs:
+      - ci-go
+      - ci-static
+      - ci-js
+      - ci-build-gitops-image
+      - ci-upload-images
+      - ci-upload-binary
```
ozamosi, line 302: I think you can replace this list with just `finish-ci-merge`

ozamosi (approved):
If syntax _looks_ correct

.github/workflows/pr.yaml
Callisto13: i did start with that, changed for... reasons i forget

Callisto13:
tested it works, and made changes. @ozamosi you want to re-review or all good?

ozamosi:
All good!

PR #2025, opened 2022-04-27
Update WGE installation instructions after reloaded integration

foot:
- [x] Mention user-roles supplied by the helm chart and how to disable them if they're no good.
- [x] Mention directory structure options
- [x] Remove references to cli and `gitops upgrade` for now.
- [x] Update examples to use ./clusters as the base path (instead of .weave-gitops)
- [x] Remove `known_hosts` instructions. This was for old core.
- [x] Mention TLS workaround when connecting a cluster w/ self-signed certs
- [x] Update capd-template and CBCC examples that work w/ latest capd


website/docs/installation.mdx
```diff
@@ -302,43 +299,24 @@ kubectl create secret generic git-provider-credentials --namespace=wego-system \
...
-```bash
-gitops upgrade \
-  --git-host-types="git.example.com=gitlab" \
-  --version 0.0.18 \
-  --set "config.extraVolumes[0].name=ssh-config" \
-  --set "config.extraVolumes[0].configMap.name=ssh-config" \
-  --set "config.extraVolumeMounts[0].name=ssh-config" \
-  --set "config.extraVolumeMounts[0].mountPath=/root/.ssh" \
-```
+By default a user `WegoAdmin` is used when you login.
```
sympatheticmoose, line 310: in core this is wego-admin, does WGE change this?

foot: No, bad typo, good catch

foot:
```suggestion
By default a user `wego-admin` is used when you login.
```

yiannistri (approved):

LutzLange:
I found the following issues : 
Screenshot from 2022-05-03 14-08-04
This should be 0.7.0.rc-2 or newer.

Screenshot from 2022-05-03 14-09-02
Deploying CAPA is part of the instructions further up.

Screenshot from 2022-05-03 14-09-41
This should be gitops 0.8.0 right?

The structure of the gitrepository is missing / not mentioned in https://staging.docs.gitops.weave.works/update-enterprise-docs/docs/installation/.

There is no flux configuration for pointing flux @ clusters/management for example. Like : 
GITHUB_TOKEN=<github token> flux bootstrap github --owner=wkp-example-org --repository=demo-02 --branch=main --path=clusters/demo-02

foot:
Hiya!

Yes, I think you might be looking at 0.8.0 instead of `main`? It is very
awkward that.. =/

PR #1967, opened 2022-04-20
Update all Flux (related) dependencies

hiddeco:
**What changed?**

- Update all Flux dependencies to `v0.29.x`
- Move to v1beta2 Flux Source API
- Update source.toolkit.fluxcd.io versions in docs
- Update Flux Source API test CRDs
- (Properly) update Helm to v3.8.2
- Update golang.org/x/{crypto,oauth2}

**Why was this change made?**

To stay up-to-date with latest Flux, and mitigate https://github.com/advisories/GHSA-8c26-wmh5-6g9v.

**How was this change implemented?**
 
Go Modules update, and relevant changes to Go imports (including a change of imports from `github.com/fluxcd/source-controller/api/v1beta1` to `/v1beta2`) and other Source API related files.

:rotating_light: **NB:** the simplicity of this change means the application is no longer able to see Source resources without the `v1beta2` Source API CRD being present on the cluster.

**How did you validate the change?**

:warning: Tests seem to pass, but still requires (human/manual) validation.

**Release notes**

This updates to the Flux Source v1beta2 API, and requires Flux `>=v0.28.0` (or at least source-controller `>=0.22.0` including Source API `v1beta2` CRDs) to be installed on the cluster. The [upgrade instructions for Flux v0.28.0 apply post-upgrade](https://github.com/fluxcd/flux2/discussions/2567).

**Documentation Changes**

Updated `source.toolkit.fluxcd.io/v1beta1` examples to `source.toolkit.fluxcd.io/v1beta2`.

PR #2030, opened 2022-04-28
Don't start the helm watcher as part of core

ozamosi:
This was making gitops crash completely when you a install flux that
has the wrong API version of source.toolkit.fluxcd.io (spoiler: the
version just changed, see #1967).

This isn't a core feature anyway, so we can just disable it.

This doesn't make gitops _work_, it'll just start and find nothing.


ozamosi:
(this doesn't remove the helm chart RBAC permissions, because those can only be removed _after_ this has been released)

mnowster (approved):
pair reviewed with @ozamosi to help unblock another PR 👍🏻

PR #2033, opened 2022-04-28
fix: wego-admin unable to view events for weave gitops app

### LINKED ISSUE #2017: wego-admin unable to view events for weave gitops app
Screenshot 2022-04-27 at 10 10 24

yitsushi:
Removed the `namespace` field from the ListEvents request as
`involvedObject` has this information and does not make sense to manage
two instances of the same value in the request.

Closes #2017


yitsushi:
image

Callisto13 (approved):
Nice!

Callisto13:
> timestamp is empty, i don't know if it's a known issue, or not an issue at all, but that's a different question.

yeh i think that has been noted somewhere? new ticket if not

PR #2037, opened 2022-04-28
release: Thread release notes into the PR, too

ozamosi:
This puts the release notes into the PR body, so it can be edited
before the release happens. This lets us work on general advice on
e.g. breaking changes, without having to edit the release after it's
already been published. While not massively awful to have to make a
    manual change, we ought to want to auto-publish the release to
    e.g. social networks or even email, which means it's better to publish
    the right thing in the first place.


mnowster (approved):
Let's try it.

PR #2040, opened 2022-04-28
Waiting the initial load of clusters

luizbafilho:
Closes

**What changed?**
Waits for the initial load of clusters. Since we fire both, namespaces and clusters go routines to watch for changes, the namespace one can execute before the initial load of clusters, resulting in a empty namespace list for about 30 seconds until the next time it runs.

**Why was this change made?**

**Release notes**

**Documentation Changes**


joshri (approved):
Tested the UI hot reloading, and it works without issue. NICE!

PR #2042, opened 2022-04-28
Avoid ignoring helmreleases on source detail page

### LINKED ISSUE #2016: No deployment details are shown for Helm Releases
For example the Weave GitOps app itself:
Screenshot 2022-04-27 at 10 08 41

This is using wego-admin

josecordaz:
Closes #2016 

**What changed?**
- Added new condition to avoid ignoring helm releases on source detail page when looking at helm chart or helm repository  sources

**Why was this change made?**
- Previous conditions were only considering specific data structure that are characterized for kustomizations.

- Manual testing

**Release notes**

**Documentation Changes**


ui/components/SourceDetail.tsx
```diff
@@ -57,6 +57,10 @@ function SourceDetail({ className, name, info, type }: Props) {
       return true;
     }
 
+    if (a?.helmChart?.sourceRef?.kind == s.type && a?.helmChart?.name == name){
+      return true;
+    }
+
```
yitsushi, line 63: No objection on this, just a note, we can simplify this a bit with a function definition outside of the filter loop:
```typescript
const isRelevant(expectedType, expectedName) {
  return (expectedType == s.type && expectedName == name)
}
```

and later just use

```typescript
  if (isRelevant(a?.sourceRef?.kind, a?.sourceRef?.name)) { return true }
  if (isRelevant(a?.helmChart?.sourceRef?.kind, a?.helmChart?.name)) { return true }
  
  return false;
```

or even chaining them and

```typescript
return isRelevant(a?.sourceRef?.kind, a?.sourceRef?.name) ||
       isRelevant(a?.helmChart?.sourceRef?.kind, a?.helmChart?.name)
```

yitsushi (approved):

ui/components/SourceDetail.tsx
```diff
@@ -48,16 +48,25 @@ function SourceDetail({ className, name, info, type }: Props) {
...
+    return (expectedType == s.type && isNameRelevant(expectedName))
+  }
+
   const relevantAutomations = _.filter(automations, (a) => {
     if (!s) {
       return false;
     }
 
-    if (a?.sourceRef?.kind == s.type && a?.sourceRef?.name == name) {
+    if (type == "HelmChart" && isNameRelevant(a?.helmChart?.name)) {
```
yitsushi, line 64: What's the difference between this check and the following `isNameRelevant` chain? Or why do we need that extra check?

core/server/types/helmrelease.go
```diff
@@ -25,12 +27,14 @@ func HelmReleaseToProto(helmrelease *v2beta1.HelmRelease, clusterName string, in
 		Interval:    durationToInterval(helmrelease.Spec.Interval),
 		HelmChart: &pb.HelmChart{
 			Chart:     helmrelease.Spec.Chart.Spec.Chart,
-			Namespace: helmrelease.Spec.Chart.Spec.SourceRef.Namespace,
-			Name:      helmrelease.Spec.Chart.Spec.SourceRef.Name,
 			Version:   helmrelease.Spec.Chart.Spec.Version,
+			Name:      fmt.Sprintf("%s-%s", helmrelease.Namespace, helmrelease.Name),
```
yitsushi, line 31: Why is it `namespace-name` and not just `name`?

joshri (approved):
Tested locally, details are showing up. NICE!!!

ui/components/SourceDetail.tsx
josecordaz: This page is used to show details from Sources(helm, git) but also Charts, which in this context is like another kind of source. For helm repositories and git repositories, we need to check for the sourceRef info as a way to filter elements. But for Charts there is no sourceRef to compare against, that is why this different check. I mean, there is but the sourceRef is for the chart field not for the release itself.

core/server/types/helmrelease.go
josecordaz: The way I created the testing helm objects was through `flux create source helm ...` and `flux create helmrelease podinfo....`, after that, I saw that flux automatically created a resource called `flux-system-podinfo`.

PR #2044, opened 2022-04-29
chart: Remove --helm-repo-namespace from flags

ozamosi:
We don't have the flag any more.


mnowster:
hang on, do we need to change the chart version?

PR #2046, opened 2022-04-29
fix: HelmChart links from Helm Releases point to an invalid object

### LINKED ISSUE #2018: HelmChart links from Helm Releases point to an invalid object
We will likely need to revisit this to consider how to appropriately link both the Helm Chart and [Helm | Git] Repository. But for now we should address this bug.

You can see in the screenshot it tries for HelmChart/weave-gitops.
Screenshot 2022-04-21 at 11 09 54

If you look at the HelmRelease object, the Chart, aka the actual Helm Chart is called weave-gitops, but the HelmChart Flux object is flux-system-ww-gitops

```


Spec:
  Chart:
    Spec:
      Chart:               weave-gitops
      Reconcile Strategy:  ChartVersion
      Source Ref:
        Kind:   HelmRepository
        Name:   ww-gitops
      Version:  *
  Interval:     1m0s
...
Status:
  Conditions:
    Last Transition Time:          2022-04-20T09:10:41Z
    Message:                       Release reconciliation succeeded
    Reason:                        ReconciliationSucceeded
    Status:                        True
    Type:                          Ready
    Last Transition Time:          2022-04-20T09:10:41Z
    Message:                       Helm install succeeded
    Reason:                        InstallSucceeded
    Status:                        True
    Type:                          Released
  Helm Chart:                      flux-system/flux-system-ww-gitops
```

yitsushi:
Added a new field, it's hard to come up with new names, so it's
HelmChartName.

The main issue was: The flux chart name is only available in Status.

Closes #2018


yitsushi:
image
image

joshri (approved):
NICE. So our source links were only working for helm charts without custom names? Or they were always wrong? Just curious how it worked

yitsushi:
> NICE. So our source links were only working for helm charts without custom names? Or they were always wrong? Just curious how it worked

I think it's always works like that, kept the original logic in there, in case it's not always the case and we get empty value there.

PR #2056, opened 2022-05-02
feat: ClusterRole for default user

### LINKED ISSUE #2053: ClusterRole for default user.
@jpellizzari put it best:

> Now that we have some of the cross namespace stuff in there, I am a little concerned about the Day 1 experience for new users: they won't see much...Since we are using the wego-admin service account, they really only see the flux-system namespace. We should consider some type of ClusterRole for the day 1 user. Otherwise, all of our fancy graphs and "single pane of glass" story goes out the window

We should change the default user to be a cluster role so the day one experience for Weave GitOps is a pleasant one.

yitsushi:
It's created by default if `adminUser.create` is true, but can be turned
off with `adminUser.createClusterRole`.

The clusterRole gets the same permissions as the role, but without namespace restrictions.

Closes #2053

### DIFF
--- a/charts/gitops-server/values.yaml
+++ b/charts/gitops-server/values.yaml
@@ -44,6 +44,9 @@ adminUser:
   # Whether the wego-admin user should be created.
   # If you use this make sure you add it to rbac.impersonationResourceNames.
   create: false
+  # Specifies whether the clusterRole & binding to the admin user should be created.
+  # Will be created only if adminUser.crete is enabled.
+  createClusterRole: true
   createSecret: true
   # Set the username and password, these will be stored in a secret in k8s.
   username: gitops-test-user
--- a/charts/gitops-server/values.yaml
+++ b/charts/gitops-server/values.yaml
@@ -45,7 +45,7 @@ adminUser:
   # If you use this make sure you add it to rbac.impersonationResourceNames.
   create: false
   # Specifies whether the clusterRole & binding to the admin user should be created.
-  # Will be created only if adminUser.crete is enabled.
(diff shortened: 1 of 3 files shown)


yitsushi:
image

charts/gitops-server/values.yaml
```diff
@@ -44,6 +44,9 @@ adminUser:
   # Whether the wego-admin user should be created.
   # If you use this make sure you add it to rbac.impersonationResourceNames.
   create: false
+  # Specifies whether the clusterRole & binding to the admin user should be created.
+  # Will be created only if adminUser.crete is enabled.
```
Callisto13, line 48:
```suggestion
  # Will be created only if adminUser.create is enabled.
```

Callisto13 (approved):

ozamosi:
Why are we keeping the old `wego-admin-role` instead of replacing it?

yitsushi:
> Why are we keeping the old `wego-admin-role` instead of replacing it?

@ozamosi : That was my intention because

1. ClusterRole is created only if that flag is true, otherwise only the Role will be created.
2. It could be an `else if`,, but in my eyes it's more consistent to keep both. That way they have a Role bootstrapped they can use for additional users.

We can remove the Role when we create a ClusterRole instead, I just thought I'll keep both.

ozamosi:
Right, so the usecase is, either the operator sets up per-namespace permissions already in which case they don't need a clusternamebinding but this application gives them a reason to add flux-system bindings, or the operator doesn't in which case they just need it to work across all of them. Gotcha.

PR #2070, opened 2022-05-03
ci: Add step that fails if chart is changed but version isn't

ozamosi:
If 2 different versions of the helm chart are released with the same
version number, then bad things happen: users won't upgrade, users
running the same version will run different revisions, and we won't be
able to debug things properly.

This step won't be part of the mandatory steps, but it'll light up
read, at least.

### DIFF
--- a/.github/workflows/chart.yaml
+++ b/.github/workflows/chart.yaml
@@ -6,11 +6,41 @@ on:
       - 'main'
     paths:
       - 'charts/**'
+  pull_request:
+    branches:
+      - 'main'
+    paths:
+      - 'charts/**'
 
 jobs:
+  helm-verify:
+    runs-on: ubuntu-latest
+    if: github.event_name == 'pull_request'
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+      - name: Find new version
+        id: new_version
+        run: |
+          NEW_VERSION=$(yq e '.version' charts/gitops-server/Chart.yaml)
+          echo "::set-output name=version::$NEW_VERSION"
+      - name: Find old version
+        id: old_version
+        run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/chart.yaml
```diff
@@ -6,11 +6,39 @@ on:
...
+  pull_request:
+    branches:
+      - 'main'
+    paths:
+      - 'charts/**'
 
 jobs:
+  helm-verify:
+    runs-on: ubuntu-latest
+    if: github.event_name != 'push'
```
SamLR, line 18: I think this should only run on `pull_request` as you're explicitly referencing that in the `old_version` step to get the head sha

.github/workflows/chart.yaml
```diff
@@ -6,11 +6,39 @@ on:
...
+          NEW_VERSION=$(yq e '.version' charts/gitops-server/Chart.yaml)
+          echo "::set-output name=version::$NEW_VERSION"
+      - name: Find old version
+        id: old_version
+        run: |
+          git checkout ${{ github.event.pull_request.head.sha }}
+          OLD_VERSION=$(yq e '.version' charts/gitops-server/Chart.yaml)
+          echo "::set-output name=version::$OLD_VERSION"
+      - name: Alert about the need to change chart version
+        run: echo "This PR changed the helm chart. You need to change chart version when you change the chart"
```
SamLR, line 35: I think this needs to return a non-zero status otherwise no one's going to see this check fire (I think)

SamLR (changes requested):

.github/workflows/chart.yaml
ozamosi: :facepalm:

SamLR (approved):
have fun testing this!

PR #2073, opened 2022-05-03
fix: socially distanced if blocks are required, no touching please

jmickey:
**What changed?**

Social distancing required

**Why was this change made?**

The return statement was not at least 2m from the `if` statement


Callisto13 (approved):
yaaaaaaay

yitsushi (approved):

PR #2078, opened 2022-05-03
docs: "fix" (?) update-docs, restore _category_.json

ozamosi:
The script was deleting all CLI files, trying to restore one file, and
apparently failed, and then apparently decided to ignore the mistake
and just continue.

First, don't do that. Just delete the markdown files.

Second, always fail on errors. Why wouldn't you?

Third, because this is run in CI, turn on tracing of commands so we
can see what it failed doing.

Fourth, restore the deleted file.


sympatheticmoose (approved):
Seems good to me 👍

PR #2085, opened 2022-05-04
fix: Clear the clustersNamespaces cache on Update

### LINKED ISSUE #2083: flux-runtime endpoint tries to list deployments in a cluster after the cluster is removed
**Describe the bug**

`/v1/flux_runtime_objects` fails if a cluster is removed.

e.g.

https://localhost:8000/v1/flux_runtime_objects?clusterName=Default

responds with 
```json
{
  "code": 2,
  "message": "could not list deployments in namespace flux-system: cluster=ewq2 not found",
  "details": []
}
```

image

- [ ] UI
- [X] Server

- [ ] `severity/Major`: Weave Gitops functionality or a key feature is broken. There is a workaround, but the workaround requires significant effort.
- [ ] `severity/Minor`: A non-key feature or functionality is broken. There is a fairly straightforward workaround.
- [x] `severity/Low`: Doesn’t affect primary flow/functionality but would be good to fix.

**Environment**
 - Weave-Gitops Version: 0.8.0
 - How you deployed the Weave GitOps server: lib
 - Kubernetes version:  [e.g. 1.21.1]
 - Where are you running your cluster?
    - [x] KinD - _version_

**To Reproduce**
- Add a cluster
- See flux-runtimes sourced from that cluster
- Remove cluster
- See above error message 

**Expected behavior**
- Still see flux-runtime objects from remaining clusters

Callisto13:
This is a simple, unfancy way to ensure that when a cluster is deleted
it is no longer present in this cache. Before the cache was cumulative,
which meant that clusters and their associated namespaces were not removed
despite no longer existing in the separate clusters cache.

Closes #2083


luizbafilho:
This looks good, but it's not enough, since we need to clean up the `cf.usersNamespaces` too, but different from the `clustersNamespaces` we cannot do every 30 seconds, other wise will have to ask user's permissions to the cluster too frequently, slowing things down.

So I propose that we create a hash for the list of clusters, and only reset everything up if that hash changes, so basically you'd do:
- List the cluster names ordered and save a concatenated string in the factory, we could add a `GetHash` to the `Clusters` cache for instance.
- When updating the namespaces, check if current hash is different from the saved one, if so, clear `clustersNamespace` and the `usersNamespaces`, if not leave as is.

With that, we should expect things to be eventually consistent, after 30 seconds or so.

foot:
I get some very strange behaviour..

1. `https://localhost:8000/v1/helmreleases?` returns a `cannot find cluster=` (empty) kind of thing (sorry I'll try and get the corrent error message again)
2. `https://localhost:8000/v1/helmreleases?clusterName=Default` complains it cannot find a secret that only exists in the leaf (ewq2)
3. `https://localhost:8000/v1/helmreleases?clusterName=ewq2` complains it cannot find a secret in mgmt (Default)

Callisto13:
> we need to clean up the cf.usersNamespaces too

I had deliberately left that to handle separately, but you are right that they have to be considered together 👍

Callisto13:
Updated commit and description

core/clustersmngr/factory_caches.go
```diff
@@ -28,6 +30,18 @@ func (c *Clusters) Get() []Cluster {
 	return c.clusters
 }
 
+func (c *Clusters) Hash(clusters []Cluster) string {
```
luizbafilho, line 33: I wonder why accept a list of clusters instead of just accessing them directly and returning the hash?

core/clustersmngr/factory.go
```diff
@@ -211,10 +214,23 @@ func (cf *clientsFactory) userNsList(ctx context.Context, user *auth.UserPrincip
...
 
+func (cf *clientsFactory) userNsList(ctx context.Context, user *auth.UserPrincipal) map[string][]v1.Namespace {
+	userNamespaces := cf.GetUsersNamespaces(user)
+	if len(userNamespaces) > 0 {
+		return userNamespaces
+	}
+
+	cf.UpdateUsers(ctx, user)
+
+	return cf.GetUsersNamespaces(user)
```
luizbafilho, line 231: And update to `GetUserNamespaces` since it's a single user.

luizbafilho (approved):
this looks great beside couple small nits.

@foot would be possible for you to test it from this branch? or does it need to be merged?

foot:
I can test from this branch 👌, will take it for a spin tomo thanks team!

foot:
Still not quite working unfortunately.

/v1/flux_runtime_objects returns:
```
{
  "code": 2,
  "message": "could not list deployments in namespace flux-system: cluster=ewq2 not found",
  "details": []
}
```

We could add some debug logging? 🤔 , I can put together a cluster to play with if it helps too.

core/clustersmngr/factory.go
```diff
@@ -33,10 +33,14 @@ type ClientsFactory interface {
 	UpdateClusters(ctx context.Context) error
 	// UpdateNamespaces updates the namespaces all namespaces for all clusters
 	UpdateNamespaces(ctx context.Context) error
+	// UpdateUsers updates the namespaces all namespaces for all clusters
```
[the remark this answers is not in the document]
Callisto13: 🤦‍♀️ fml

Callisto13:
@foot we could jump on a call together, might be faster?

I only have a couple of hours today before I go on holiday, so someone may need to take this off me if it drags on.

PR #2086, opened 2022-05-04
fix: Weave GitOps username is hard-coded

### LINKED ISSUE #2001: Weave GitOps username is hard-coded
**Describe the bug**

GitOps username is hard-coded to `wego-admin`. Since we're configuring users with user-specified names, the names should show up for the UI user icon. 

- [ ] UI
- [x] Server

- [ ] `severity/Critical`: Weave GitOps is crashing or experiencing data loss, the UI is inaccessible or a key feature is unusable. There is no known workaround
- [ ] `severity/Major`: Weave Gitops functionality is broken, there is a workaround, but the workaround requires significant effort
- [x] `severity/Minor`: Weave Gitops functionality is broken, but there is a fairly straightforward workaround
- [ ] `severity/Low`: Doesn’t affect primary flow/functionality but would be good to fix

**Environment**
 - gitops: main
 - How you deployed the Weave GitOps server: Helm Chart
 - kubernetes:  [e.g. 1.20.4]
    - [x] KinD - _version_]
    - [ ] k3s - _version_
    - [x] cloud [e.g., EKS, AKS]  _version_
 - Browser + version: [e.g. chrome 74, safari 12, firefox 87]

**To Reproduce**

1. Configure username and password using GitOps HelmRelease yaml
2. Run `flux bootstrap`
3. Navigate to UI dashboard
4. Login using configured credentials
5. Click on user icon
6. See `wego-admin` username

**Expected behavior**

User specified username should appear

**Config and Logs**

**Screenshots**

https://images.zenhubusercontent.com/342592099/50078b92-0815-42fa-bb93-efebc41e9caa/screenshare___2022_04_25_11_38_54_am.mp4

**Additional context**

What does this mean for `Dev` environment?

yitsushi:
Create addition RoleBinding and ClusterRoleBinding based on the
username, and keep the original `wego-admin` there to maintain backward
compatibility.

`DevMode` with Tilt still uses `wego-admin` as a hard-coded user,
potentially we can even remove the `DevUser` maybe.

`DevMode` is used only to set `DevUser` on `HMACTokenSignerVerifier` and
the user is used only return with a static claim on `Verify()`.

Additional changes:
* Split up the admin-user.yaml into logical parts like
  * Credentials as they will be created only once.
  * Role and ClustreRole as they will be created only once
  * Two files for RoleBinding and ClusterRoleBinding (as described above)

Fixes #2001

### DIFF
--- a/charts/gitops-server/Chart.yaml
+++ b/charts/gitops-server/Chart.yaml
@@ -14,7 +14,7 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 2.0.2
+version: 2.0.3
 
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application. Versions are not expected to
--- a/charts/gitops-server/templates/admin-user-creds.yaml
+++ b/charts/gitops-server/templates/admin-user-creds.yaml
@@ -1,0 +1,16 @@
+{{- if .Values.adminUser.create }}
+{{- if .Values.adminUser.createSecret }}
+---
+apiVersion: v1
+kind: Secret
+metadata:
+  name: cluster-user-auth
+  namespace: flux-system
+type: Opaque
+data:
+  {{- with .Values.adminUser }}
+  username: {{ .username | b64enc | quote }}
+  password: {{ .passwordHash | required "passwordHash must be set!" | b64enc | quote }}
+  {{- end }}
+{{- end }}
+{{- end }}
--- a/charts/gitops-server/templates/admin-user-roles.yaml
+++ b/charts/gitops-server/templates/admin-user-roles.yaml
@@ -1,0 +1,53 @@
+{{- if .Values.adminUser.create }}
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: wego-admin-role
+  namespace: flux-system
+rules:
+  - apiGroups: [""]
+    resources: ["secrets", "pods" ]
+    verbs: [ "get", "list" ]
+  - apiGroups: ["apps"]
+    resources: [ "deployments", "replicasets"]
+    verbs: [ "get", "list" ]
+  - apiGroups: ["kustomize.toolkit.fluxcd.io"]
+    resources: [ "kustomizations" ]
+    verbs: [ "get", "list", "patch" ]
+  - apiGroups: ["helm.toolkit.fluxcd.io"]
+    resources: [ "helmreleases" ]
+    verbs: [ "get", "list", "patch" ]
+  - apiGroups: ["source.toolkit.fluxcd.io"]
+    resources: [ "buckets", "helmcharts", "gitrepositories", "helmrepositories" ]
+    verbs: [ "get", "list", "patch" ]
+  - apiGroups: [""]
+    resources: ["events"]
+    verbs: ["get", "watch", "list"]
+{{- if .Values.adminUser.createClusterRole }}
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: wego-admin-cluster-role
+rules:
+  - apiGroups: [""]
+    resources: ["secrets", "pods" ]
+    verbs: [ "get", "list" ]
+  - apiGroups: ["apps"]
+    resources: [ "deployments", "replicasets"]
+    verbs: [ "get", "list" ]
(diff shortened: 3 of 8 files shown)


yitsushi:
image

ozamosi (changes requested):
Changes to the chart will (just like documentation) be deployed straight to our end users. I believe this change would break all existing users, hence my change request.

Additionally, when you change the chart, you need to change the chart version, or the deploy becomes weird and we create 2 different "editions" of the same chart version.

yitsushi:
> Additionally, when you change the chart, you need to change the chart version, or the deploy becomes weird and we create 2 different "editions" of the same chart version.

I know that part, first I wanted some review if the approach is acceptable.

> Changes to the chart will (just like documentation) be deployed straight to our end users. I believe this change would break all existing users, hence my change request.

When the Helm Chart is applied, the RoleBinding will be changed. I don't see why would it break. If they log in with an `adminUser`, the username/password pair will not change, the underlying RoleBinding will change, but the change will just use the same value as defined in adminUser (the one they use to log in), the only requirement would be to restart the server pod, but that will happen because of the code (and with that the container image) changes.

yitsushi:
Tested workflow:
 - Used `main` while for `dev-cluster`
 - Log in as `dev/dev` and it shows `wego-admin`
 - Switched to this branch
 - Wait...
 - Log in as `dev/dev` and it shows `dev` and I see the same content

ozamosi:
> but that will happen because of the code (and with that the container image) changes.

No - merging this PR won't deploy core in any customer cluster. It will deploy the role binding. We can only do the chart change as part of a release.

yitsushi:
> No - merging this PR won't deploy core in any customer cluster. It will deploy the role binding. We can only do the chart change as part of a release.

That makes sense.

Callisto13:
_dejavu_

I think this is something we (someone) does not want. It was a deliberate choice to have the k8s object User to always be called wego admin. I had a PR very similar to this a while ago (cannot find it anywhere) and was told that we do not want this. Basically the argument was that if we let it be configurable to this level, people can create any number of static users and we don't want to become an identity provider.
So basically, even though it feels "broken" to you and me, there is only ever one "user", regardless of what people think they are setting.

yitsushi:
Closing this PR based on the discussion above.

Callisto13:
the fact that we keep having these discussions, because it looks to so weird to use and to implement, is... illuminating 😁

Callisto13:
This is the closest thing to a "summary issue" i can find. It has all the links to various discussions around this. I may write it up fully so we have an easy link in future for when this inevitably comes up again

ozamosi:
> Unfortunately I can't split it up code and chart changes. At least I don't see how can I do that.

Add 2 users right now, and add a ticket to remove the old one after the next release, I suppose.

> for when this inevitably comes up again

Given this has been reported about one time for every person who's tried to use this, could we just fix it instead? "We don't want this" isn't accurate - _we_ clearly all do.

Callisto13:
you'd have to take it up with @davidstauffer

yitsushi:
> Given this has been reported about one time for every person who's tried to use this, could we just fix it instead? "We don't want this" isn't accurate - _we_ clearly all do.

To be fair it's [kind of documented](https://docs.gitops.weave.works/docs/installation/#6-configure-password):
image

Maybe we just have to highlight the fact that the username + password pair is only for authentication, the logged in user will be `wego-admin`.

yitsushi:
> if we let it be configurable to this level, people can create any number of static users and we don't want to become an identity provider.

Technically with this approach they still can have only one user, because we read the user/pass from a single and hard-coded Secret, so no way they can create more than one user without changing the mechanism used to fetch auth credentials from Secret.

JamWils:
Alright, I think we should reopen this since it is a topic that keeps coming up.

1. Let's not break existing users so do what needs to be done.
2. The name can be configurable via helm chart values and it defaults to `wego-admin`.
3. This is the only role binding we provide in the chart.  This is effectively our one user.  We can't stop users from setting up more users then there already are, but that is on them.  They could do it now if they wanted to by uploading the correct yaml along side the `helm release`.

yitsushi:
Why the HelmChart test fails... updated the chart version :/

ozamosi (approved):
:tada: 

Not sure about the failed check - I added it, and it's clearly not doing what it should, so I need to debug it. I'm happy for you to ignore it and merge regardless.

ozamosi:
Fixed the chart check in #2094

Callisto13:
> They could do it now if they wanted to by uploading the correct yaml along side the helm release.

Should we add something in docs (in another pr) that if ppl want to add more users, it is at their own risk and we are doing (will always do) the bare minimum when it comes to being an identity provider, and that they should def use oidc?

PR #2094, opened 2022-05-04
ci: Fix check to see that chart version was bumped

ozamosi:
This fixes a typo - `head` is the latest commit (just like e.g. git
show HEAD), `base` is the base branch (which is also what github calls
it when you open a PR). So in effect, the code was comparing the most
recent version to itself and somehow kept finding that the version was
the same.


josecordaz (approved):
LGTM

PR #2100, opened 2022-05-05
fix: Missing deployments on Kustomization detail page

### LINKED ISSUE #2081: Missing deployments on Kustomization detail page when the Kustomization has set a specific targetNamespace different from flux-system
**Describe the bug**

The kustomization details page is not showing any k8s resources when its `targetNamespace` field is pointing to a namespace other than `flux-system`.

- [X] UI
- [X] Server

- [X] `severity/Major`: Weave Gitops functionality or a key feature is broken. There is a workaround, but the workaround requires significant effort.
- [ ] `severity/Minor`: A non-key feature or functionality is broken. There is a fairly straightforward workaround.
- [ ] `severity/Low`: Doesn’t affect primary flow/functionality but would be good to fix.

**Environment**
 - Weave-Gitops Version, commit sha or image tag: `868f753af24b04d6631119ecb33e024fe7f0c17b` [e.g. `v0.1.0`, `3499ba2a`, or `1651069883-3499ba2a2c78a7c04c47c56c58eb77eede4b21f4`]
 - How you deployed the Weave GitOps server: `Tilt` [e.g. Tilt, Helm Chart, etc]
 - Kubernetes version: `1.21.1` [e.g. 1.20.4]
 - Flux version: `0.29.5`
 - Where are you running your cluster?
    - [X] KinD - _version_
    - [ ] k3s - _version_
    - [ ] cloud [e.g., EKS, AKS]  _version_
 - Browser + version: Firefox 99.0.1, Chrome 100.0.4896.88 [e.g. chrome 74, safari 12, firefox 87]

**To Reproduce**
1. Having a kind cluster ready with flux installed(no bootstrap).
2. Enable `createClusterRole: true` on `tools/helm-values-dev.yaml` to give permissions over deployments.
3. Initialize tilt `make cluster-dev`
4. Run:
```
flux create source git podinfo-git-source \
  --url=https://github.com/stefanprodan/podinfo \
  --branch=master \
  --interval=30s
```
to create the git source
5. Run 
```
flux create kustomization podinfo \
  --target-namespace=podinfo-kustomization \
  --source=podinfo-git-source \
  --path="./kustomize" \
  --prune=true \
  --interval=5m
``` 
to create the kustomization
6. Confirm you have proper permissions with this command `kubectl get deployments -l kustomize.toolkit.fluxcd.io/name=podinfo -l kustomize.toolkit.fluxcd.io/namespace=flux-system -n podinfo-kustomization --as=wego-admin`
7. Go to Applications
8. Click on `podinfo`
9. See that there is no items under the Details table

**Expected behavior**
I should be able to see at least the podinfo deployment in the details table

**Screenshots**

**Additional context**
The problem seems to be that the BE is querying on the wrong namespace, it is using `flux-system` instead of using the targetNamespace value in the kustomization resource. We might want to return the targetNamespace value in the response of the `/kustomizations` end point so then it can be passed in to the `/reconciled_objects` end point.

yitsushi:
Fixes #2081

### DIFF
--- a/core/server/fluxruntime.go
+++ b/core/server/fluxruntime.go
@@ -94,13 +94,11 @@ func (cs *coreServer) GetReconciledObjects(ctx context.Context, msg *pb.GetRecon
 	switch msg.AutomationKind {
 	case pb.AutomationKind_KustomizationAutomation:
 		opts = client.MatchingLabels{
-			KustomizeNameKey:      msg.AutomationName,
-			KustomizeNamespaceKey: msg.Namespace,
+			KustomizeNameKey: msg.AutomationName,
 		}
 	case pb.AutomationKind_HelmReleaseAutomation:
 		opts = client.MatchingLabels{
-			HelmNameKey:      msg.AutomationName,
-			HelmNamespaceKey: msg.Namespace,
+			HelmNameKey: msg.AutomationName,
 		}
 	default:
 		return nil, fmt.Errorf("unsupported application kind: %s", msg.AutomationKind.String())
@@ -117,7 +115,7 @@ func (cs *coreServer) GetReconciledObjects(ctx context.Context, msg *pb.GetRecon
 			Version: gvk.Version,
 		})
 
-		if err := clustersClient.List(ctx, msg.ClusterName, &l, opts, client.InNamespace(msg.Namespace)); err != nil {
+		if err := clustersClient.List(ctx, msg.ClusterName, &l, opts); err != nil {
 			if k8serrors.IsForbidden(err) {
 				// Our service account (or impersonated user) may not have the ability to see the resource in question,
 				// in the given namespace.
--- a/core/server/fluxruntime.go
+++ b/core/server/fluxruntime.go
@@ -166,7 +166,7 @@ func (cs *coreServer) GetChildObjects(ctx context.Context, msg *pb.GetChildObjec
 		Kind:    msg.GroupVersionKind.Kind,
 	})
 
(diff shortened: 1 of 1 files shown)


yitsushi:
With access to the target Namespace:
image

If no access (clusterRole is not created) the list is still empty, no error message, but that's a different topic.

sympatheticmoose:
@yitsushi purely based on the screenshot, it looks like still missing replicaset and pods (at least). Does this PR purely seek to address deployments and subsequent PRs will solve for the other missing items?

yitsushi:
@sympatheticmoose : Did not notice they are not there, but indeed, they are missing. working on it.

yitsushi:
Now childObjects are there too:
image

yitsushi:
And they are on the Graph view too:
image

Callisto13 (approved):

core/server/fluxruntime.go
```diff
@@ -168,7 +166,7 @@ func (cs *coreServer) GetChildObjects(ctx context.Context, msg *pb.GetChildObjec
 		Kind:    msg.GroupVersionKind.Kind,
 	})
 
-	if err := clustersClient.List(ctx, msg.ClusterName, &l, client.InNamespace(msg.Namespace)); err != nil {
+	if err := clustersClient.List(ctx, msg.ClusterName, &l); err != nil {
```
josecordaz, line 169: @yitsushi This would only work if the user has access to list all namespaces, right?. I think this is missing the cross-namespace query logic.

josecordaz: NVM I see that `wego-system` doesn't have permissions to list namespace but this still works as you describe. I must be missing something.

PR #2101, opened 2022-05-05
Don't use quotes on flags to docker build action

SamLR:
The quotes around LDFLAGS were also being passed (although this was only
obvious in the docker output rather than the command being called). The
inclusion of these quote marks meant that `go build` was ignoring the
`-ldflags` as one giant string that it didn't understand.

Removing the quotes seems to work and has hopefully fixed the problem.


.github/workflows/pr.yaml
```diff
@@ -157,7 +157,7 @@ jobs:
           # Warning: the ' in build-args are fragile -- only use on LDFLAGS
```
ozamosi, line 157: I guess we should remove this comment too

ozamosi (approved):

.github/workflows/pr.yaml
SamLR: as I was writing the commit message I literally saw this and went 'must fix that' and yet, here we are (done now)

Issue #2032, opened 2022-04-28
Show container image for applications


sympatheticmoose:
**Short description**
As an application operator I want to view deployed images and their version so that I can verify which revision of my applications are actually running.

### Acceptance criteria
When I go to the detail page for a Kustomization or Helm Release I will see the following behavior on the detail tab.

- Images column Application Details view which shows `imageName:tag`. When multiple container images exist, for example a sidecar, these should be one image per line in the table.
- Table columns should be in order: name, type, namespace, status, message, images
- Shows on Deployment and Pod objects

### Out of Scope
- Reconciliation Graph also shows list of images
- Text within Reconciliation Graph is selectable

**Design materials**
Example from the Flux Runtime view running on the staging cluster
<img width="2328" alt="Screenshot 2022-04-28 at 13 02 36" src="https://user-images.githubusercontent.com/19860021/165747839-8d1e3c50-4bae-4cb9-afa8-94119a24b553.png">

Example from figma, note there have been changes since this in terms of other content, but shows an example of the images column and its format.
<img width="1810" alt="Screenshot 2022-04-28 at 12 56 55" src="https://user-images.githubusercontent.com/19860021/165746876-7f16667f-c14b-4c41-bd6e-f7f7981450ef.png">

**Open questions**
- Should we expose in some way on the Applications summary view? Initially left out as feel that view already has a larger number of columns than most screens real estate will allow; additionally complicated by likely large number of images associated with a given kustomization/helmrelease
- Are there scenarios where a secondary container may exist alongside the flux controllers?
- Column reordering - above proposal is based on some user testing where the assertion was that Status is ultimately the first thing to care about so have it first. Also open to swapping Name and Status, which would be comparable to GKE console for example

JamWils:
Per your open questions above:

1. No
2. Yes, there are cases that this could happen but none of the flux controllers do that.  I believe the controller needs to be labeled `part-of:flux`
3. I find namespace being last to be a bit odd.

sympatheticmoose:
@JamWils I understand your point on namespace having spoken, that it does 'pair' more with name/type, and is valuable upfront given scenarios where failing components in a dev environment are less worrying. 

Status and message similarly 'pair' in my mind, so I'm thinking in groups:
- [name, type, namespace, images] - identifiers
- [status, message] - status

Current implementation is: name, type, namespace, status, message. It still feels wrong having status so far along, which was my primary reason for wanting to reorder generally. So two other options:

[...]

PR #2102, opened 2022-05-05
feat: Show container image for applications [server side]

### LINKED ISSUE #2104: Return with used container images for reconciled and child objects
* Pod
* ReplicaSet
* Deployment

yitsushi:
This is only the server side implementation.

It will return with a list of images. Most of the time it will be only one, but it is possible to define more than one container in a Pod, so that's why it gives back a list instead. On the UI we can show only the first, or all of them, the data is available.

Related to #2032
Closes #2104


yitsushi:
image
image
image

jpellizzari:
> I have zero idea how to add it to the UI. Tried to extend the `ReconciledObjectsTable`, but that just goes through a lot of function calls and I don't know where do I have to add this extra field.

Adding another element to this array will add another column:
https://github.com/weaveworks/weave-gitops/blob/e26b706c8c018797076c99203498f5ef6ab0c3f2/ui/components/ReconciledObjectsTable.tsx#L58

Happy to leave that for another ticket, however.

yitsushi:
> Adding another element to this array will add another column:
Yes that's what I tried, but then I had to update `ReconciledObjectsTable` and `useGetReconciledObjects` and `getChildren` call and... well... that point I knew I don't know 😆

jpellizzari (approved):

PR #2111, opened 2022-05-05
docs: Add helm chart reference

### LINKED ISSUE #2002: documentation: Provide a reference of available helm values in the manual
There's a lot of configurable settings in our helm chart, but our manual doesn't provide a way for our users to discover what they are.

The manual should contain a reference of available helm settings. That way, pages that describe e.g. authentication, permissions, HTTPS management, etc, can all point the user at the reference and they can investigate further on their own.

The key point here is that it's a reference - all available knobs and settings, with an English sentence or two explaining what they do. We might be able to generate this from the values file itself, similar to the CLI reference. Guides, tutorials, and guided examples would all be good ideas to supplement a reference, but are out of scope for this ticket.

ozamosi:
This uses [helm-docs](https://github.com/norwoodj/helm-docs) to
generate a reference out of the values.yaml - the only thing you need
to do is to add `--` to the beginning of every block of comments that
is supposed to be used for the manual.

I decided to create a new directory `references` in the manual, and
store both the CLI reference and this helm reference next to each
other on equal footing.

I have not documented every single option, so there's a lot that just
say "type: object, default: {}, description: ''" which isn't ideal,
but those are mostly generic, auto-generated properties which I
frankly don't even know what they do.

This file will be automatically re-generated for the `main` version
whenever we make any type of release. When we tag a release, this will
be re-generated before we fork the documentation, so both main and the
forked documentation should contain the newest instructions.

In case something goes horribly wrong, I've added a comment at the top
spelling out which helm version we generated the reference for.

This fixes #2002


charts/gitops-server/values.yaml
```diff
@@ -11,48 +11,58 @@ image:
...
+  # -- If non-empty, this limits the resources that the service
+  # account can impersonate. It is strongly recommended that you
+  # configure groups that can be used to populate this.
   impersonationResourceNames: []
-  # Limit the type of principal that can be impersonated
+  # -- Limit the type of principal that can be impersonated
   impersonationResources: ["users", "groups"]
-  # Specify whether additional permissions should be granted to the service account
-  # in order to work with weave-gitops enterprise
+  # -- Specify whether additional permissions to view secrets should
```
SamLR, line 40: I think this is very wrong (I think I should have done a second pass on the docs when I added this). It should be something more like:
```
# -- Specify specific secrets the service account that can access, for example the
# enterprise entitlement.
```
we should probably rename `viewSecrets` to `viewSecretsResources` to make it clear it expects a list of resources and match naming with `impersonationResources`

charts/gitops-server/values.yaml
```diff
@@ -11,48 +11,58 @@ image:
 imagePullSecrets: []
```
SamLR, line 11: Can you add a comment @ the start of this file explaining what the `# --` format means, please?

charts/gitops-server/values.yaml
```diff
@@ -11,48 +11,58 @@ image:
...
   name: ""
 rbac:
-  # Specifies whether the clusterRole & binding to the service account should be created
+  # -- Specifies whether the clusterRole & binding to the service account should be created
   create: true
-  # List of resources that the service account can impersonate. It is strongly
-  # recommended that you configure groups that can be used to populate this.
+  # -- If non-empty, this limits the resources that the service
+  # account can impersonate. It is strongly recommended that you
+  # configure groups that can be used to populate this.
```
SamLR, line 36: We should probably add an example which is this value with the admin user added

SamLR (approved):
Looks good, a couple of tweaks if you feel like making them.

charts/gitops-server/values.yaml
ozamosi: It matches `impersonationResourceNames`, so I called it `viewSecretResourceNames` instead.

SamLR: thank you

charts/gitops-server/templates/role.yaml
```diff
@@ -22,6 +22,9 @@ rules:
     {{- with .Values.rbac.viewSecrets }}
     resourceNames: {{ . | toJson }}
     {{- end }}
+    {{- with .Values.rbac.viewSecretsResourceNames }}
```
SamLR, line 25: This should error if both `viewSecrets` and `viewSecretsResourceNames` are set (or we breaking-bump the version...)

SamLR (approved):

PR #2116, opened 2022-05-06
Supporting Partial success results for clustered list

### LINKED ISSUE #2067: If one cluster is failing to `List`, I should still be able to see resources in my other clusters
Right now if a `List` query on a leaf cluster fails (because the user does not have RBAC, for example) then the entire "batch" fails and I get an empty screen (eg I see nothing on my Sources page).

My other clusters are (may be) fine, so I should still see the resources from those.

luizbafilho:
Closes #2067

**What changed?**
Adding support to partial success results on clustered list calls. 

With this PR, `ClusteredList` calls will:
-  completely fail and return an error if **all** calls fail or there is any other failure unrelated to the actual cluster call, for instance, extracting the continue token from the request.
- when a call for a single cluster returns an error, we append that error into the `Errors` fields on the response. And returns a partial success

The UI will be responsible for checking that up and showing any errors presented there. 

**Why was this change made?**

unit and manual test

**Release notes**

**Documentation Changes**


jpellizzari (approved):
Nice work. Can you create a follow up issue with a `ui` tag so we remember to do this?

core/clustersmngr/client.go
```diff
@@ -42,8 +41,8 @@ type Client interface {
 	// ClientsPool returns the clients pool.
 	ClientsPool() ClientsPool
 
-	// RestConfig returns a rest.Config for a given cluster
-	RestConfig(cluster string) (*rest.Config, error)
+	// // RestConfig returns a rest.Config for a given cluster
+	// RestConfig(cluster string) (*rest.Config, error)
```
josecordaz, line 45: intentional comment?

josecordaz (approved):
LGTM

core/clustersmngr/client.go
luizbafilho: ops, I'll delete it. Thanks

PR #2123, opened 2022-05-06
Shrink JS dependencies

ozamosi:
Remove storybooks. Apparently we don't use them any longer.

Upgrade jest. Something about the old jest installation caused snyk to get upset once I removed storybooks. This does not upgrade to a new major version or anything, just a plain `npm upgrade` - I assume if tests still pass, the test framework didn't break.

Remove babel. We don't need to have a top level dependency on babel - it still gets pulled in, but this removes 10,000 lines of lock file and doesn't seem to change anything, so why not.  I've built the UI and clicked around "a bit" and can't see anything different - given the fundamental nature of babel, I'd assume I'd see something big if I broke something.

Remove parcel/transformer-svg-react - it's not used (we use transformer-inline-string). I still see SVGs after removing this.

In total, this seems to remove more than half of package-lock.json.


josecordaz (approved):
LGTM

PR #2124, opened 2022-05-06
Partial success flux response

josecordaz:
Closes

**What changed?**

**Why was this change made?**

**Release notes**

**Documentation Changes**


core/server/fluxruntime.go
```diff
@@ -49,7 +58,8 @@ func (cs *coreServer) ListFluxRuntimeObjects(ctx context.Context, msg *pb.ListFl
 		list := &appsv1.DeploymentList{}
 
 		if err := clustersClient.List(ctx, clusterName, list, opts, client.InNamespace(fluxNs.Name)); err != nil {
-			return nil, fmt.Errorf("could not list deployments in namespace %s: %w", fluxNs.Name, err)
+			respErrors = append(respErrors, &pb.ListError{ClusterName: clusterName, Namespace: fluxNs.Name, Message: fmt.Sprintf("%s, %s", ErrListingDeployments.Error(), err)})
```
josecordaz, line 61: @luizbafilho I tried to help in adding partial response to ListFluxRuntimeObjects but I couldn't find an easy way to make this line fail, would you have any suggestion?

core/server/fluxruntime_test.go
```diff
@@ -201,6 +204,30 @@ func TestListFluxRuntimeObjects(t *testing.T) {
 	g.Expect(res.Deployments[0].Name).To(Equal(fluxDep.Name))
 }
 
+func TestListFluxRuntimeObjectsResponseError(t *testing.T) {
```
josecordaz, line 207: @luizbafilho Do you see an easy way to make this test pass when running all tests at the same time? This only passes if it runs individually 😔

josecordaz: NVM I think I fixed this

josecordaz:
Trying to fix this, found it in testing enterprise env
image

core/server/fluxruntime.go
josecordaz: ok ok

PR #2127, opened 2022-05-09
Add exports

AlinaGoaga:
Added exports for EE and updated Data Table to work with both labels and label renderers as we have instances in which we need to pass through React Components for the table header 

Required as part of: https://app.zenhub.com/workspaces/weave-gitops-enterprise-6127680175876a0014fc5657/issues/weaveworks/weave-gitops-enterprise/504


ui/components/DataTable.tsx
```diff
@@ -165,8 +168,14 @@ function UnstyledDataTable({
           <TableHead>
             <TableRow>
               {_.map(fields, (f) => (
-                <TableCell key={f.label}>
-                  <SortableLabel field={f} />
+                <TableCell
+                  key={typeof f.label === "function" ? f.label(r) : r[f.label]}
```
foot, line 171: I'm not sure how `key` responds to receiving a React component? 🤔 ..

```ts
    type Key = string | number;
```

Maybe we still require a `label: string` that could be `"checkbox"` for example, but also provide a `labelRenderer` or something too that can return a react component like `labelRenderer ? labelRenderer(r) : label` wdyt?

AlinaGoaga: Sounds v good!

ui/components/DataTable.tsx
```diff
@@ -158,7 +159,6 @@ function UnstyledDataTable({
               maxWidth: f.maxWidth,
             }
           }
-          key={f.label}
```
foot, line 161: Oops, pop this back on?

foot (approved):
But otherwise LGTM!

PR #2131, opened 2022-05-09
Upgrade docusaurus to latest version

ozamosi:
This has an API change for algolia where the App ID needs to be
specified. See
https://docusaurus.io/blog/2021/11/21/algolia-docsearch-migration and
in particular note that they say these parameters are safe to put into
git. How did I know what parameters to enter? They're plain text
parameters as part of the request, I just grabbed them.

This also has an API change that causes the inlinetoc comoponent to be
a flat array instead of a tree - done in facebook/docusaurus#729 - so
this replaces the hard-coded index with a function that looks for the
start and end of the relevant section.

This gets rid of all security vulnerability warnings except for two,
which aren't relevant - see facebook/docusaurus#6394.

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -1,16 +1,8 @@
 on:
   push:
-    paths-ignore:
-      - 'doc/**'
-      - 'website/**'
-      - '*.md'
     branches:
       - main
   pull_request:
-    paths-ignore:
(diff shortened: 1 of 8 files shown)


mnowster (approved):
Thanks for resolving these. Please can I ask for some extra testing when this is in that our docs site still works correctly?

PR #2136, opened 2022-05-10
Add some more detail about k8s RBAC

SamLR:
**What changed?**

added pages to the docs:
* configuration/service account permissions
* configuration/user permission

Also removed some out of date permissions from the helm chart
* removed `patch` permissions from the cluster admin user
* removed `gitopsclusters` permissions from the service account (I think these were used by enterprise)
* removed `helmrepositories` permissions from the service account (I think any calls to check these should be going via impersonation)

**Why was this change made?**

To improve clarity in the docs: making sure that operators know why we're asking for the permissions we are

The permission changes also aim to keep the default cluster install secure by following the principal of least priviledge

Tested locally that things worked as expected.


SamLR:
If folks could check through things so far. I think these need to be back ported into the docs for `v0.7` and `v0.8`. 

I'm going to try and do a pass on the 'securing the dashboard' page as well

charts/gitops-server/templates/admin-user-roles.yaml
```diff
@@ -14,13 +14,13 @@ rules:
     verbs: [ "get", "list" ]
   - apiGroups: ["kustomize.toolkit.fluxcd.io"]
     resources: [ "kustomizations" ]
-    verbs: [ "get", "list", "patch" ]
+    verbs: [ "get", "list" ]
```
ozamosi, line 17: Does sync or suspend use patch on these resources? At the moment, sync is available in the UI, while suspend is in the TODO list.

website/docs/configuration/user-permissions.mdx
```diff
@@ -0,0 +1,60 @@
...
+
+These permissions exist because these are resources that, in order to understand,
+the status of Flux you need to be able to see. To understand if the change in
+a `kustomization` or `helmrelease` has worked you need to be able to see those
+resources, the source for them and the deployment that they have created.
+
+The `events` permissions are granted to view the history of resources that Flux
+manages (e.g. when a source is reconciled).
+
+It is important to note that read permissions are required for viewing helm release
```
ozamosi, line 59: This is supposed to say "read permissions for secrets", isn't it?

ozamosi (commented):
I did change the name of `rbac.viewSecrets` so that still needs updating here, and I wrote a couple of comments. Other than that, :+1:!

charts/gitops-server/templates/admin-user-roles.yaml
yitsushi: Yes, it uses Patch.

https://github.com/weaveworks/weave-gitops/blob/134f2580fb73441705f9228b25b1cf1ff5f8dcf2/core/server/sync.go#L128

website/docs/configuration/user-permissions.mdx
SamLR: yes

charts/gitops-server/templates/admin-user-roles.yaml
SamLR: Should that not use the users' permissions rather than the service account?

yitsushi: It used the user's permissions because it is possible to have users with read access but can't resync (i don't know if it's handled on the UI or not).

SamLR: ignore me just spotted this is the admin user, not the service account

SamLR:
@ozamosi @yitsushi I've added another page on a recommended set up. Could you both review that please? I believe it's a sensible direction but more eyes would be appreciated.

website/docs/configuration/recommended-rbac-configuration.mdx
```diff
@@ -0,0 +1,180 @@
...
+
+You could then create 3 groups:
+
+* `wego-admin`
+  - Bound to the `ClusterRole`, created by Helm, `wego-admin-cluster-role`
+  - Aisha is the only member
+* `wego-team-a-admin`
+  - Bound to a `Role`, using the same permissions as `wego-admin-role`, created
+    in Team's namespace
+  - Brian and Aisha are members
```
yitsushi, line 55: Why Aisha has to be a assigned to that group? Doesn't Aisha get's everything if their group/user is only one binding and that's the `wego-admin-cluster-role` ?

yitsushi: Or is it not membership, it's more like "has access to these resources"

website/docs/configuration/recommended-rbac-configuration.mdx
```diff
@@ -0,0 +1,180 @@
...
+* `wego-readonly`
+  - Bound to a `ClusterRole` that matches `wego-admin-cluster-role` but with
+    no `patch` permissions.
+  - Aisha, Brian, June & Jo are all members
```
yitsushi, line 59: same as above

website/docs/configuration/securing-access-to-the-dashboard.mdx
```diff
@@ -52,7 +52,7 @@ kubectl create secret generic oidc-auth \
   --from-literal=tokenDuration=<token-duration>
 ```
 
-Once the HTTP server starts, it will redirect unauthenticated users to the provider's login page to authenticate them. Upon successful authentication, the users' identity will be impersonated in any calls made to the Kubernetes API, as part of any action they take in the dashboard. At this point, the dashboard will fail to render correctly unless RBAC has been configured accordingly. Follow the instructions in the [RBAC authorization](#rbac-authorization) section in order to configure RBAC correctly.
+Once the HTTP server starts, it will redirect unauthenticated users to the provider's login page to authenticate them. Upon successful authentication, the users' identity will be impersonated in any calls made to the Kubernetes API, as part of any action they take in the dashboard. By default the Helm chart will configure RBAC correctly but it is recommended to read the [service account](docs/configuration/service-account-permissions) and [user](docs/configuration/user-permissions) permissions pages to understand which actions are needed for Weave GitOps to function correctly.
```
yitsushi, line 55: > Once the HTTP server starts, it will redirect unauthenticated users to the provider's login page to authenticate them.
I'm not sure if it's true. It will redirect to the login page of the weave-gitops UI. There they can click on the OIDC login button and they will be redirected to the provider's page. Or is there a case when the built-in login page is skipped automatically?

website/docs/configuration/recommended-rbac-configuration.mdx
SamLR: so in theory Aisha only needs the cluster admin role but it's good practice to define these things hierarchically. I should probably make that clear (and it probably means she wouldn't get the team-a admin role)
e.g. the rules would be "all members of engineering have read-only access", "team leads have admin perms to their ns", "cluster operators/SRE team get full admin"

SamLR: I'll add this to the text to make it clearer the goal

yitsushi (approved):

PR #2137, opened 2022-05-10
Continue processing and updating the clustermngr caches onError

foot:
- Right now we bail out if there's an error connecting to a leaf cluster
- This then requires a restart of the server to continue trying

We also tweak the userNamespace cache. This was previously 1h.
- When provisioning clusters the RBAC takes a minute to be configured
- We cache the permissions in an incomplete state for up to an hour
- This makes provisioning new clusters a bit awkward
- TODO: Follow up issue to discuss options

**Why was this change made?**

During cluster provisioning we run into many transitory errors like "cannot connect to cluster", we want to be resilient to these rough patches.

Tested in a MC environment


luizbafilho (approved):
LGTM. Sorry got hooked by another thing, completely forgot to getting this one out. Thanks.

PR #2139, opened 2022-05-11
server: Start logging all requests, and fix specifying log levels to application

ozamosi:
This changes 2 things: it starts to log all requests rather than just ones to `/v1/`, and makes it work to specify log levels in your config.

The log handler is now attached to the top HTTP server instead of just to the core handler, so you can see _all_ requests being served. Only non-OK requests are logged at above DEBUG level, so it shouldn't cause a massive increase in log sizes, but it does mean you can debug e.g. OIDC logins properly.

The log level specification is now correct - HTTP errors were logged as debug, debug messages were logged as errors, because zap and logr disagree on whether a bigger numuber is more important or more voluminous. This flips them so it works.

### DIFF
--- a/pkg/server/handler.go
+++ b/pkg/server/handler.go
@@ -37,16 +37,14 @@ type Config struct {
 
 func NewHandlers(ctx context.Context, log logr.Logger, cfg *Config) (http.Handler, error) {
 	mux := runtime.NewServeMux(middleware.WithGrpcErrorLogging(log))
-	httpHandler := middleware.WithLogging(log, mux)
-
-	if AuthEnabled() {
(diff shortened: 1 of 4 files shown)


pkg/server/handler.go
```diff
@@ -37,16 +37,16 @@ type Config struct {
 
 func NewHandlers(ctx context.Context, log logr.Logger, cfg *Config) (http.Handler, error) {
 	mux := runtime.NewServeMux(middleware.WithGrpcErrorLogging(log))
-	httpHandler := middleware.WithLogging(log, mux)
-
-	if AuthEnabled() {
```
ozamosi, line 42: Since this is being removed in #1983, I didn't reimplement this condition.

mnowster (approved):
Can we capture another task of having an ADR of our documented decision on our approach to logging if we don't already have one?

Issue #1905, opened 2022-04-12
Check function in `useIsAuthenticated` changes too frequently leading to continuous loops if added to dependencies


AlinaGoaga:
**Describe the bug**

`req: check` in the `useIsAuthenticated` hook is an anonymous function that keeps changing so we can't use it as a dependency without causing loops

**Is this a UI bug or a server bug?**
- UI

**What is the severity of the bug**

- `severity/Low`: Doesn’t affect primary flow/functionality but would be good to fix

**Environment**

Main prior to the removal of AddApplication.

**To Reproduce**

Add `check` as a dependency to `useEffect` in the `AddApplication` component. 

**Expected behavior**

We should we able to add the dependency without causing a loop. To solve this we'll be wrapping `req` in `useIsAuthenticated` in a `useCallback` to limit the amount of times the function changes.

PR #2142, opened 2022-05-11
 [UI] Lottie and check update

AlinaGoaga:
Closes https://github.com/weaveworks/weave-gitops/issues/1905
Closes https://github.com/weaveworks/weave-gitops-enterprise/issues/578

What changed and why?

- replaced SignIn svg background with Lottie animation (improved performance on Firefox)
- the Error page now uses a Lottie animation as well - improve user experience (we used to have text only)
- improve feedback to user whilst the FeatureFlags are being retrieved / if they error out


foot (approved):
💯

PR #2145, opened 2022-05-11
Simplify bug report template

jpellizzari:
Our bug report template had way too much boilerplate, leading to text-walls on bug reports.

Other notes:
* Bug reporters won't know if a bug is a server or UI bug; an eng will need to make that determination. Also better represented as a label, so that is removed.
* Severity is better represented as a label
* We really only care about k8s, flux, and wego versions: simplified

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,10 +1,9 @@
 ---
 name: Bug report
 about: Create a report to help us improve
-title: ''
-labels: 'bug'
-assignees: ''
-
+title: ""
+labels: "bug"
+assignees: ""
 ---
 
 <!--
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -17,39 +16,20 @@ Please read them :pray:
...
-- [ ] Server
 
-**What is the severity of the bug**
 <!--
-Please select a label for this ticket indicating the severity of this bug.
-A maintainer will add the label to the issue, taking your suggestion into account.
+A clear and concise description of what the bug is.
 -->
 
-- [ ] `severity/Critical`: Weave GitOps is crashing, the UI is inaccessible or a key feature is unusable. There is no known workaround.
```
josecordaz, line 34: Maybe instead of removing these we can put them inside a comment tag??. So we don't forget what are the rules in choosing one of the severity labels.

rokshana-b: I agree with @josecordaz - also, leaving it inside a comment tag might _at least_ prompt people to add labels.

PR #2147, opened 2022-05-11
Change table filtering logic to AND between categories, and OR within…

jpellizzari:
Closes https://github.com/weaveworks/weave-gitops-enterprise/issues/700

I was able to get a quick win with some sign-flipping, thanks to the solid test coverage we have here.

Some screen shots to explain the new behavior:

No filters, all rows:
Screenshot from 2022-05-11 14-04-45

`OR` filtering within a category
Screenshot from 2022-05-11 13-59-20

`AND` filtering; note that we have `Not Ready` selected, so only the single `Not Ready` element is appearing:

Screenshot from 2022-05-11 13-59-25

`AND + OR` to show "All Statuses, but only the types I specified":

Screenshot from 2022-05-11 13-59-33

### DIFF
--- a/ui/components/FilterableTable.tsx
+++ b/ui/components/FilterableTable.tsx
@@ -69,22 +69,22 @@ export function filterRows<T>(rows: T[], filters: FilterConfig) {
     return rows;
   }
 
-  return _.filter(rows, (r) => {
-    let ok = false;
+  return _.filter(rows, (row) => {
+    let ok = true;
 
-    _.each(filters, (vals, key) => {
+    _.each(filters, (vals, category) => {
       let value;
       //status
-      if (key === "status") {
-        if (r["suspended"]) value = "Suspended";
-        else if (computeReady(r["conditions"])) value = "Ready";
+      if (category === "status") {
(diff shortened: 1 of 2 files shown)


joshri (approved):
⭐  I CANNOT believe it was this easy ⭐

sympatheticmoose (approved):
LGTM - unable to test explicitly with multiple clusters, but confirmed cross-category filtering working as described.

bia (approved):
LGTM by your screenshot explanation, thanks for that.

When there is no content below the heading of each of these filter fields, perhaps the checkbox shouldn't appear?
Screenshot 2022-05-12 at 15 48 38

bia:
@mmoulian pointed out that the typography sizes look strange.
The column headings (Name, Type, Namespace..) should be `14px`, not `18px` and the table content (body text) should also be `14px` instead of `16px`.

There are 3 general sizes : `12px`, `14px`, and `20px`
Screenshot 2022-05-12 at 16 08 20

Should I create a separate issue for this?

jpellizzari:
> Should I create a separate issue for this?

Yes please :)

PR #2152, opened 2022-05-12
Remove offline_access from OIDC token requests

ozamosi:
We don't support using refresh_tokens, so stop requesting them.


ozamosi:
This was pointed out in #1949. It's also done in #2021 but it's got other changes and is in draft state, so I don't want to just undraft and merge it.

mnowster (approved):
🔒

PR #2155, opened 2022-05-12
580 UI need to improve the error message for wrong cluster user password

ahussein3:
Resolve this issue => https://github.com/weaveworks/weave-gitops-enterprise/issues/580

- Update error message at sign in instead of "401 Unauthorized" =>  Incorrect username or password.
- Making username is required as a extra layer of validation as password.


ui/pages/SignIn.tsx
```diff
@@ -88,7 +88,11 @@ function SignIn() {
           <AlertWrapper
             severity="error"
             title="Error signin in"
-            message={`${String(error.status)} ${error.statusText}`}
+            message={`${
+              error.status == 401
+                ? "Incorrect username or password."
+                : String(error.status) + error.statusText
```
jpellizzari, line 94: Won't this look like: `401Unauthorized`?

Maybe: 
```suggestion
                : ` ${error.status):  ${error.statusText}`
```

jpellizzari (approved):

ui/pages/SignIn.tsx
ahussein3: I made it like that ``${error.status} ${error.statusText}``

PR #2156, opened 2022-05-12
PoC: Ability to query Flagger resources

yitsushi:
Flagger is not guaranteed to be installed, it is optional to get those
resources. To get over this, a new lookupCRD function can return a CRD.
We can use that to check if we can query those resources or not.

A simple IsFlaggerAvailable endpoint can be useful on the UI to
determine if we add it to the menu or not.

That lookupCRD can be useful for other resources we may want to add, but
their existence is not guaranteed.

== List Canary objects

    :!./check.sh | jq
    {
      "canaries": [
      {
        "name": "hello-world",
        "clusterName": "Default",
        "provider": "traefik",
        "status": {
          "phase": "Succeeded",
          "failedChecks": 1,
          "canaryWeight": 0,
          "iterations": 0,
          "lastTransitionTime": "2022-05-11T13:54:51Z",
          "conditions": [
            {
            "type": "Promoted",
            "status": "True",
            "lastUpdateTime": "2022-05-11T13:54:51Z",
            "lastTransitionTime": "2022-05-11T13:54:51Z",
            "reason": "Succeeded",
            "message": "Canary analysis completed successfully, promotion finished."
            }
          ]
        }
      }
      ],
      "nextPageToken": "looooong token",
      "errors": []
    }

### DIFF
--- a/ui/components/AutomationDetail.tsx
+++ b/ui/components/AutomationDetail.tsx
@@ -4,6 +4,7 @@ import styled from "styled-components";
 import { AppContext } from "../contexts/AppContext";
 import { Automation, useSyncAutomation } from "../hooks/automations";
 import { AutomationKind } from "../lib/api/core/types.pb";
+import {AutomationType} from "../lib/types";
 import Alert from "./Alert";
 import EventsTable from "./EventsTable";
 import Flex from "./Flex";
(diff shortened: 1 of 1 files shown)


yitsushi:
> I have no intention to merge it without real plans, so it's really just a poc

PR #2157, opened 2022-05-12
Dynamically pass in the right automation kind

### LINKED ISSUE #2016: No deployment details are shown for Helm Releases

josecordaz:
Closes #2016 

**What changed?**
- Removed hardcoded kustomization kind value when hitting reconciled objects end point. This is now calculated dynamically based on the automation type.

**Why was this change made?**
- When hitting the reconciled objects end point, the parameter `automation kind` always had the value `Kustomization` even though the automation selected was a helm release.

- Manually and local using tilt

**Release notes**

**Documentation Changes**

### DIFF
--- a/ui/components/AutomationDetail.tsx
+++ b/ui/components/AutomationDetail.tsx
@@ -4,6 +4,7 @@ import styled from "styled-components";
 import { AppContext } from "../contexts/AppContext";
 import { Automation, useSyncAutomation } from "../hooks/automations";
 import { AutomationKind } from "../lib/api/core/types.pb";
+import {AutomationType} from "../lib/types";
 import Alert from "./Alert";
 import EventsTable from "./EventsTable";
 import Flex from "./Flex";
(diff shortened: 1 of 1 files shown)


ui/components/AutomationDetail.tsx
```diff
@@ -62,7 +63,7 @@ function AutomationDetail({ automation, className, info }: Props) {
         <SubRouterTabs rootPath={`${path}/details`}>
           <RouterTab name="Details" path={`${path}/details`}>
             <ReconciledObjectsTable
-              automationKind={AutomationKind.KustomizationAutomation}
+              automationKind={automation?.type === AutomationType.Kustomization ? AutomationKind.KustomizationAutomation: AutomationKind.HelmReleaseAutomation}
```
jpellizzari, line 66: This seems like a larger issue of `kind` and `type` being duplicates of each other. Can you create a follow up to fix that?

jpellizzari (approved):
Nice work on this fix. There is a larger problem here with the conflation of `type` and `kind` that we should fix.

ui/components/AutomationDetail.tsx
josecordaz: Done https://github.com/weaveworks/weave-gitops/issues/2162

PR #2167, opened 2022-05-16
refactor CoreClientCOntext to be more extensible

ahussein3:
- That is trying to extend the CoreClient to be able to used in enterprise as well


ui/contexts/CoreClientContext.tsx
```diff
@@ -15,11 +15,9 @@ export type CoreClientContextType = {
 export const CoreClientContext =
   React.createContext<CoreClientContextType | null>(null);
 
-export default function CoreClientContextProvider({ api, children }: Props) {
+export function UnAuthrizedInterceptor(api: any) {
```
foot, line 18:
```suggestion
export function UnAuthorizedInterceptor(api: any) {
```

PR #2168, opened 2022-05-17
Listing cluster scoped objects + allow specifying sheme for clients factory

alichaddad:
**What changed?**
Listing API objects using clusters manager requires specifying the scope of the requested object(cluster scoped VS namespace scoped).
You need to specify the scheme when getting a new clusters client factory or clients pool

To be able to list objects that are not bound to a namespace.

Added an extra parameter in `core.clustersmngr.Client.ClusteredList` to specify if the object is namespaced or not

**Release notes**

**Documentation Changes**


luizbafilho (changes requested):
Looks good, just a missing test and a minor change.

luizbafilho (approved):

PR #2176, opened 2022-05-18
Display backend version in footer

### LINKED ISSUE #2117: Display backend version in footer
When receiving a bug report, we want to know the exact version that the bug report is against. For releases, this is currently shown in the footer. For development versions, we show the last release version as described in the javascript. This means, if there's a long gap with lots of fixes between 0.8 and 0.9, we won't be able to easily spot if a problem report is newer or older than the fix that was supposed to fix it.

In #1953 we added a new backend endpoint that reports the exact build version.

Add a hook that calls the version endpoint using the generated API client, and use that hook to print the backend version in the footer component instead.

opudrovs:
Closes #2117

**Important**
- On @ozamosi 's advice, I've based the new code on the existing Feature Flags hook.

- Suggestions on naming, code style and so on are most appreciated.

- I've added simple data mapping to map the loaded version data to an object with standard cameCase keys. Maybe it's not necessary or the mapping should be performed in a different place.

**What changed?**
Adds a hook for loading the exactly app build version from the `/v1/version` endpoint.

If the version, loaded from the endpoint, does not match the release version in the `package.json` file, adds displaying this data in `branch-commit` format and points the link to the corresponding commit. 

Otherwise, if complete version data was not loaded from the API (if the branch and commit are empty etc., like what I see when testing the app locally) or if the "API version" and the `package.json` versions are the same, just displays the version from the `package.json` and link it to the corresponding release tag.

**Why was this change made?**
Please see the description of #2117 

Adds `useVersion` hook and code to the display the appropriate version in the `Footer` component.

Tested it locally, with the incomplete data provided by the endpoint when testing locally and with a mock data from the staging endpoint:

```
{"version":{"branch":"main", "buildtime":"2022-05-16_10:56:21", "git-commit":"7aa118f", "version":"7aa118f"}}
```

**Release notes**

**Documentation Changes**
None. But if there is any documentation which I should update please point me to it.


ui/components/Footer.tsx
```diff
@@ -16,7 +17,34 @@ const RightFoot = styled(Flex)`
...
+  };
+
+  const shouldDisplayApiVersion = !isLoading
+    && apiVersion.version !== p.version
+    && apiVersion.branch
+    && apiVersion.gitCommit;
+
+  const versionText = shouldDisplayApiVersion
+    ?
+    `${apiVersion.branch}-${apiVersion.gitCommit}`
```
jpellizzari, line 39: The formatting looks a little weird here. Are you using `prettier` and formatting on save? If not, we will get weird diffs when the next person saves this file.

ui/components/__tests__/__snapshots__/Footer.test.tsx.snap
```diff
@@ -0,0 +1,122 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`Footer snapshots default 1`] = `
```
jpellizzari, line 3: I would expect to see the version appearing in this snapshot. Let me know if you want some help on testing sanpshops + hooks together.

ui/hooks/version.ts
```diff
@@ -0,0 +1,25 @@
...
+};
+
+export function useVersion() {
+  const { api } = useContext(CoreClientContext);
+
+  return useQuery<GetVersionResponse, RequestError>(
+    "version",
+    () => api.GetVersion({}),
+    {
+      staleTime: Infinity,
```
jpellizzari, line 21: Does this prevent `react-query` from re-fetching? Good to know.

ui/components/Footer.tsx
```diff
@@ -16,7 +17,34 @@ const RightFoot = styled(Flex)`
...
 const LeftFoot = styled(Flex)``;
 
+const REPO_URL = "https://github.com/weaveworks/weave-gitops";
+
 function Footer({ className }: Props) {
+  const { data, isLoading } = useVersion();
+  const versionData = data?.version || {} as Version;
+  const apiVersion: Version = {
+    version: versionData.version,
+    gitCommit: versionData["git-commit"],
```
jpellizzari, line 27: I don't think this object is necessary.

We should change the backend to return a better `JSON` key: 

https://github.com/weaveworks/weave-gitops/blob/3d1af824228d8f2974003a8489be4940267a62d1/core/server/version.go#L16

We can enumerate the keys here and have `protoc` generate the types for us:

https://github.com/weaveworks/weave-gitops/blob/3d1af824228d8f2974003a8489be4940267a62d1/api/core/core.proto#L349

Happy to help on this.

jpellizzari (commented):
Looks good, but there are a couple of areas we can make some improvement to simplify things.

ui/components/Footer.tsx
opudrovs: Added running prettier on save and prettified the files changed.

ui/hooks/version.ts
opudrovs: Yes, AFAIK, the query should never go stale (copied the options from the feature flag hook).

jpellizzari:
@opudrovs I pushed a couple of commits to get things working:
Sets up the mock response to test the footer render logic: https://github.com/weaveworks/weave-gitops/pull/2176/commits/98d70d3e96639398196e8bc69838378107712a66

Simplifies the `withContext` helper func: https://github.com/weaveworks/weave-gitops/pull/2176/commits/9acbb0e32851ff9b09b16863c5421a9511f396f9

LMK if that makes sense!

ui/components/Footer.tsx
jpellizzari: Fixed this

jpellizzari (approved):
LGTM :+1:

Issue #2171, opened 2022-05-17
Web UI fails to redirect to login screen


rokshana-b:
**Describe the bug**

Our official doc site has an example of the helm chart that does not provide the smoothest experience. When I tried to use the example yaml file, my UI was giving me problems. I did not see any login page; I was taken directly to the applications page with an error on top. The version on the bottom of the page showed v0.8.0.

**Environment**

- Weave-Gitops Version = v0.8.0
- Flux Version = v0.30.2
- Kubernetes version = v1.23.4

**To Reproduce**

1. Go to official doc site
2. Navigate to Installation page: https://docs.gitops.weave.works/docs/next/installation/
3. Scroll down to weave-gitops helm chart example
4. Use the example file during flux bootstrap
5. After bootstrapping flux, see the ww-gitops-weave-gitops service up and running
6. Run `kubectl port-forward svc/ww-gitops-weave-gitops -n flux-system 9001:9001`
7. Go to dashboard UI via http://localhost:9001
8. See error on page

**Expected behavior**

The example on the official doc site should be working properly. If there's some field missing or the helm chart version needs to be explicitly provided, then the doc should reflect that. If this is an issue with weave-gitops v0.8.0, then it needs to be resolved asap.

**Actual Behavior With Screenshots**

![Screen Shot 2022-05-18 at 09.12.56.png](https://images.zenhubusercontent.com/606389351ead0c83798a9a73/5311c08e-5dc6-44aa-b708-6a7ca43359b4)

sympatheticmoose:
@rokshana-b can you confirm whether you were asked to login or whether you were able to get to this page without logging in?

rokshana-b:
That sounds critical - changing label

Issue #2051, opened 2022-04-29
Update install and getting started for Weave GitOps to pin to a version


sympatheticmoose:
We currently provide no guidance within the documentation for setting appropriate controls around which updates will be automatically applied to the Weave GitOps server. By default, users will get the latest chart version and the latest image. This is most likely fine for development environments and getting started, but users will likely want to pin to either a specific version/tag or semver range (i.e. patches only) in other environments. 

To do so, the HelmChart spec can be updated as per the Flux docs - https://fluxcd.io/docs/components/helm/helmreleases/ 

> The chart.spec.version can be a fixed semver, or any semver range (i.e. >=4.0.0 <5.0.0). It is ignored for HelmRelease resources that reference a GitRepository or Bucket source.

i.e:

Pin to chart version 2.0.7:
```
apiVersion: helm.toolkit.fluxcd.io/v2beta1
kind: HelmRelease
metadata:
  name: ww-gitops
  namespace: flux-system
spec:
  chart:
    spec:
      chart: weave-gitops
      reconcileStrategy: ChartVersion
      sourceRef:
        kind: HelmRepository
        name: ww-gitops
      version: 2.0.7
  interval: 1m0s
```

Could also do `version: 2.0.X` to accept patches, or `version: 2.X.X` to accept minor updates, or a range as described above.

For image version, each time the chart updates, it sets the default image version as the latest published https://github.com/weaveworks/weave-gitops/blob/main/charts/gitops-server/values.yaml#L14. As such - in most cases, pinning the chart version should be sufficient, however a user can also override the image as shown below, for example if they wanted to test with a release candidate:

pin to a specific version.
```
values:
    adminUser:
      create: true
      passwordHash: <value>
      username: <value>
    image:
      tag: 0.8.1-rc.4
```

Note we publish the images both with and without a `v` prefix.

In our documentation we should also describe that we try to follow SemVer with our Chart and Image versions.

Tasks:
- [ ] update example HelmRelease configuration in https://docs.gitops.weave.works/docs/installation/#install-the-helm-chart to pin to specific chart version. **This version should automatically be updated on a new release**.
- [ ] update Getting Started https://docs.gitops.weave.works/docs/installation/#install-the-helm-chart step 4 to include `--chart-version` specifying the latest version, as above this should automatically be updated on a new release.
- [ ] Add "Managing updates" section to Installation Guide

JamWils:
@sympatheticmoose is this still an issue or can I close?

[...]

PR #2178, opened 2022-05-18
Doc install fix

### LINKED ISSUE #2050: doc: update install document based on getting started guide
https://docs.gitops.weave.works/docs/installation/#install-the-helm-chart

_Think_ we just need to remove rbac.create=true

sympatheticmoose:
Closes #2050 
Partially addresses https://github.com/weaveworks/weave-gitops/issues/2171 by removing the configuration which causes the issue to appear 

**What changed?**
Removed rbac.create=true ~as was causing issues where user was not directed to login screen~ - turned out to be red herring, but still better to be consistent with getting started guide, and this is already default when adminUser is specified.
removed namespace as not required.

### DIFF
--- a/website/versioned_docs/version-0.8.0/installation.mdx
+++ b/website/versioned_docs/version-0.8.0/installation.mdx
@@ -44,6 +44,7 @@ spec:
       sourceRef:
         kind: HelmRepository
         name: ww-gitops
+      version: 2.0.6
   interval: 1m0s
   values:
     adminUser:


josecordaz (approved):
LGTM

sympatheticmoose:
@yitsushi I've got https://github.com/weaveworks/weave-gitops/issues/2051 which covers some aspects related to this, i'll add a comment in there.

ozamosi:
OK, but, is that scheduled for work _now_? Because this PR means that if we cut a release before we get to that issue, the installation instructions will still tell users to install the old version until someone remembers to manually upgrade the numbers.

sympatheticmoose:
@ozamosi If we cut a release now, the `main` docs are unchanged, only `0.8.0` docs will pin a version, which in theory should be the only version which had an issue.

ozamosi:
:see_no_evil: sorry, you are right, ignore me.

PR #2182, opened 2022-05-19
Provide auth compatibility for older images

### LINKED ISSUE #2171: Web UI fails to redirect to login screen

foot:
closes #2171 

If using the "new" helm chart (post https://github.com/weaveworks/weave-gitops/pull/1983) with v0.8.0, which still expects the env var to be set.

### DIFF
--- a/charts/gitops-server/Chart.yaml
+++ b/charts/gitops-server/Chart.yaml
@@ -14,7 +14,7 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 2.0.7
+version: 2.0.8
 
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application. Versions are not expected to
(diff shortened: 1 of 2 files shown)


sympatheticmoose (approved):
tested with custom helm repo and successfully redirects to login, from there was able to login with cluster user and view resources

PR #2185, opened 2022-05-20
Return Available Versions for Profiles in descending order

AlinaGoaga:
Changes the ordering returned for available versions for a given profile.
1. Sort available versions by semver instead of alphanumeric so 0.16.0 comes after 0.2.0
2. Sort descending, so that most interesting (recent) versions are shown first by ui/cli displays.


AlinaGoaga:
Fixed in https://github.com/weaveworks/weave-gitops-enterprise/pull/801/files

foot (approved):

ozamosi (approved):

PR #2188, opened 2022-05-23
Table doesn't rerender when data becomes available in Applications and Sources

AlinaGoaga:


foot:
Closing in favour of #2191

PR #2189, opened 2022-05-23
Handle empty filter dialog options + adjust font sizes

### LINKED ISSUE #2164: [UI] Change table font styles
Follow-up on #2147 

The column headings (Name, Type, Namespace..) should be `14px`, not `18px` and the table content (body text) should also be `14px` instead of `16px`.

There are 3 general sizes : `12px`, `14px`, and `20px`
Screenshot 2022-05-12 at 16 08 20

joshri:
Closes #2164

Filter dialog should not render a checkbox on categories with no options (from #2147) :
image

Also adjusts global font sizes according to #2164


opudrovs (approved):
LGTM. 👍

PR #2190, opened 2022-05-23
Fix helmChart error

opudrovs:
Based on @jpellizzari 's suggestion to add optional chaining.

Accessing properties of undefined objects breaks UI when testing locally and on many pages in staging.

Screenshot 2022-05-23 at 13 08 54


jpellizzari (approved):
LGTM 👍

PR #2191, opened 2022-05-24
Fixes an issue where initial filter is set to the initial data

foot:
- This Fixes a small issue in the filtering, where the initial filter state was derived from the initial data. E.g. if all Automations are "ready", the filtering is set to only show "ready" automations from then on.
- Also adds some sorting to the FilterDialog so the items don't jump around

Repro: 
1. load up automations page that should all be in Ready (important)
2. `flux reconcile helm-release weave-gitops`
3. Watch weave-gitops disappear as it transitions to Not Ready

### DIFF
--- a/ui/components/FilterDialog.tsx
+++ b/ui/components/FilterDialog.tsx
@@ -58,7 +58,11 @@ const FilterSection = ({ header, options, formState, onSectionSelect }) => {
         <ListItem>
           {options[0] && (
             <ListItemIcon>
-              <Checkbox checked={all} onChange={handleChange} id={header} />
+              <Checkbox
+                checked={Object.keys(formState).length === 0 ? false : all}
+                onChange={handleChange}
+                id={header}
+              />
             </ListItemIcon>
           )}
           <Text capitalize size="small" color="neutral30" semiBold>
--- a/ui/components/FilterDialog.tsx
+++ b/ui/components/FilterDialog.tsx
@@ -37,16 +37,12 @@ const SlideContent = styled.div`
 export const filterSeparator = ":";
 
 const FilterSection = ({ header, options, formState, onSectionSelect }) => {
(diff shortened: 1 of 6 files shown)


ui/components/FilterDialog.tsx
```diff
@@ -37,16 +37,12 @@ const SlideContent = styled.div`
...
-      .filter((key) => _.includes(key, header))
-      .every((key) => formState[key])
-      .value();
-    setAll(allChecked);
-  });
+  const all = _.chain(formState)
+    // get all relevant keys' current value
+    .keys()
+    .filter((key) => _.includes(key, header))
+    .every((key) => formState[key])
```
foot, line 44: `every` looks sus? 🤔

ui/components/FilterableTable.tsx
```diff
@@ -153,12 +156,15 @@ export function initialFormState(cfg: FilterConfig, initialSelections?) {
 
 function applySelections(
   filters: FilterConfig,
-  selections: FilterSelections
+  initialSelections: FilterSelections
 ): FilterConfig {
-  if (!selections) {
-    return filters;
```
foot, line 159: this was the bigger bug about filtering too heavily to the properties of the initial data set.

ui/components/FilterDialog.tsx
foot:
```
_.every([], () => true)
true
```

AlinaGoaga: Yeah... I've found it a bit difficult to get my head around it so rewrote it

AlinaGoaga:
Great stuff I see you also addressed the issue with the filters moving around and disabled those without options! Looking good :star2:

AlinaGoaga (approved):

PR #2197, opened 2022-05-24
Upgrade a couple of dependencies with security warnings

ozamosi:
I don't believe either is affecting us (we pulled in dgrijalva/jwt-go from mockoidc but that's test code, and docker/distribution had a vulnerability related to parsing OCI manifests which I don't think we do directly), but I also don't believe it hurts to stay up to date.


mnowster (approved):
Thank you @ozamosi , appreciate the effort, yes keeping up to date and not letting a backlog of these things grow again is 💯

PR #2200, opened 2022-05-24
Fix getting a helm release's inventory when using targetNamespace

### LINKED ISSUE #2132: /v1/helmreleases fails as HelmRelease release secret name is not always correct.
**Describe the bug**

Adding a `HelmRelease` to a repo with a `targetNamespace` **seems** to create a slightly different release secretName. I'm not sure if the targetNamespace is responsible for this.

Causing this pattern to fail:

https://github.com/weaveworks/weave-gitops/blob/862973dd3e98cc86ef569fcdbf0563f1daeb8f5f/core/server/helm_release.go#L100

and the entire response from https://localhost:8000/v1/helmreleases? to fail:

```json
{
"code": 2,
"message": "secrets \"sh.helm.release.v1.weave-policy-agent.v1\" not found",
"details": []
}
```

The secret is actually called `sh.helm.release.v1.policy-system-weave-policy-agent.v1`

The `policy-system` seems to be coming from the targetNamespace field?

```yaml
  targetNamespace: policy-system
```

- [ ] UI
- [x] Server

Severity
- [ ] `severity/Major`: Weave Gitops functionality or a key feature is broken. There is a workaround, but the workaround requires significant effort.
- [x] `severity/Minor`: A non-key feature or functionality is broken. There is a fairly straightforward workaround.
- [ ] `severity/Low`: Doesn’t affect primary flow/functionality but would be good to fix.

**Environment**
 - Weave-Gitops Version 0.8.1-rc.1

ozamosi:
This just uses flux's built-in methods to use the correct
namespace/release name, instead of us trying to get it right
ourselves.

This fixes #2132.

I've tested this by creating objects with various combinations of `targetNamespace`, `storageNamespace` and `releaseName` to see that it failed before and is succeeding now. There are no automated tests however, given that the logic now lives in flux.

### DIFF
--- a/core/server/helm_release.go
+++ b/core/server/helm_release.go
@@ -80,15 +80,9 @@ func (cs *coreServer) GetHelmRelease(ctx context.Context, msg *pb.GetHelmRelease
 }
 
 func getHelmReleaseInventory(ctx context.Context, helmRelease v2beta1.HelmRelease, c clustersmngr.Client, cluster string) ([]*pb.GroupVersionKind, error) {
-	storageNamespace := helmRelease.GetNamespace()
-	if helmRelease.Spec.StorageNamespace != "" {
-		storageNamespace = helmRelease.Spec.StorageNamespace
-	}
+	storageNamespace := helmRelease.GetStorageNamespace()
 
-	storageName := helmRelease.GetName()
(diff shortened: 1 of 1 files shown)


foot (approved):
Tested w/out the releaseName workaround mentioned in the issue and works great 👌

PR #2201, opened 2022-05-25
Fix listing Events for event types other than Kustomization

### LINKED ISSUE #2175: Helmreleases missing events
**Describe the bug**

Helm release objects are missing events in Applications sub page. Kubectl commands show events that are not reflected for helm releases on the UI

**Environment**

- Weave-Gitops Version = v0.8.1-rc.4
- Flux Version = v0.30.2
- Kubernetes version = v1.23.4

**To Reproduce**

1. Go to UI dashboard
2. Click on a helm release object
3. Navigate to subpage and click on `Events` tab
4. See empty list

**Expected behavior**

Events relating to the helm release object should be visible.

**Actual Behavior With Screenshots**

﻿

ozamosi:
The frontend hard-coded to only search for Kustomization events. That
was wrong.

Instead of hard-coding strings on the frontend to stuff into
    ObjectReference, this re-uses the SourceRef object that we use
    everywhere else. This works fine, except for 2 things: first,
    SourceRef is a confusing name for something that can now reference any
    object, so rename the message to ObjectRef. Second, the name is
    "KindKustomization" rather than "Kustomization", so remove the prefix - the transformation
    isn't ideal, but at least the interface between backend and frontend
    now uses types.

This fixes #2175


core/server/types/helmrepository_test.go
```diff
@@ -34,7 +34,7 @@ func TestHelmRepository(t *testing.T) {
 			&pb.HelmRelease{
 				HelmChart: &pb.HelmChart{
 					Name: "-",
-					SourceRef: &pb.SourceRef{
+					SourceRef: &pb.ObjectRef{
 						Kind: -1, // This is invalid?
```
chanwit, line 38: Not related to your changes, but this can be `pb.FluxObjectKind_KindHelmRepository` instead of -1:
```
				HelmChart: &pb.HelmChart{
					Name: "-",
					SourceRef: &pb.ObjectRef{
						Kind: pb.FluxObjectKind_KindHelmRepository,
					},
				},
```

ozamosi: Yeah, it's a bit weird - https://github.com/weaveworks/weave-gitops/blob/main/core/server/types/utils.go#L11 seems to generate schema-violating values if the input is "unexpected". I wonder if it'd be worth adding a FluxObjectKind value which is Unknown, so nothing crashes in these kind of cases :thinking:

PR #2203, opened 2022-05-25
Fix syncing sources

### LINKED ISSUE #2179: Helm release cannot be synced via sync button
**Describe the bug**

Helm release object cannot be synced from UI via sync button; the action produces an error message

**Environment**

- Weave-Gitops Version = v0.8.1-rc.4
- Flux Version =  v0.30.2
- Kubernetes version =  v1.23.4

**To Reproduce**

1. Go to UI and click on ww helm-release object
2. Navigate to sub page
3. Click on sync button

**Expected behavior**

Successful sync message should pop up

**Actual Behavior With Screenshot**

ozamosi:
This fixes two copy-paste errors, plus removes one API request that we
don't need to do.

There fixes two copy-paste errors (chart vs repository, name vs
    key).
    
This also removes a redundant extra request to read the source a
    second time - the first thing requestReconciliation does is to read
    the object again, and then use the result.
    
Finally, this does a bit of renaming for
    consistency. types.NamespacedName and client.ObjectKey are the same
    type, but we used both names, and thus we were using `name` for the
    source key, and `key` for the automation name. Use `client.ObjectKey`
    everywhere instead.

This fixes #2179.

### DIFF
--- a/core/server/internal/adapters.go
+++ b/core/server/internal/adapters.go
@@ -141,7 +141,7 @@ func (obj HelmRepositoryAdapter) AsClientObject() client.Object {
 }
 
 func (o HelmRepositoryAdapter) GroupVersionKind() schema.GroupVersionKind {
-	return sourcev1.GroupVersion.WithKind(sourcev1.HelmChartKind)
+	return sourcev1.GroupVersion.WithKind(sourcev1.HelmRepositoryKind)
 }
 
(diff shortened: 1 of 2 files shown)


chanwit (approved):
LGTM 
thank you so much @ozamosi :1st_place_medal:

PR #2205, opened 2022-05-25
Fix the template for TFTemplate

Skarlso:
The template name changed on the last commit, but we didn't reflect that in this PR.

This is a fix to change that to the more recent name of the template.


chanwit (approved):
LGTM
Thank you so much @Skarlso :1st_place_medal:

PR #2206, opened 2022-05-25
Fix snyk, fix build

ozamosi:
#2204 broke the merge pipeline, because it checked in incorrect javascript lock files - revert those bits.

However, it didn't break the PR pipeline, because snyk isn't properly distinguishing between "couldn't read" and "found issues" when run in `--all-projects` mode - replace that with spelling out all projects manually.


mnowster (approved):
for expediency of not having broken main, will raise seperate team issue for permissions of how this happened with a core maintainer being involved in a review/merge.

PR #2232, opened 2022-05-27
update `get clusters` to use the new endpoint

sarataha:
Closes #weaveworks/weave-gitops-enterprise#869

**What changed?**
Updated `RetrieveClusters()` to use `/v1/clusters` endpoint for listing clusters

**Why was this change made?**
Endpoint changed in EE 

Manually tested it + unit tests


foot:
Ah! Maybe we have to tweak it a little bit actually. pullRequest is no longer there for example. So we should remove that from the output too.

Might be some other slight changes to response types / field names.

foot (approved):
Tested locally too! Works great. 👌

PR #2238, opened 2022-05-28
Fix logging on cluster connection error

stefanprodan:
Remove the client config from the error log since the wrapped error already contains the cluster name for which the connection couldn't be established.


chanwit (approved):
LGTM
thank you @stefanprodan

PR #2245, opened 2022-05-30
ui: Display object Kinds properly

### LINKED ISSUE #2217: ui: Remove "Kind" prefix from object types
Flux objects have a type of e.g. GitRepository, Kustomization, etc. However, since a recent refactoring, the protobuf now calls them KindGitRepository, KindKustomization, etc.

In the UI, we should always display these types without the Kind prefix.

### LINKED ISSUE #2220: Object's type/kind has gone missing in some tables since the type -> kind rename
All automations used to have a type property set. Now the frontend sets a kind property whenever it receives a reply.

However, the main tables haven't been changed, so the Type column is now always empty.

Please put it back, and look for any other place where we're still referring to "type".

ozamosi:
There's a `Kind` prefix that showed up in all sorts of places that
    displays kinds as of recently. Wrap any place I can find in a helper
    that removes the prefix again.
    
Note that the backend expects the prefix to still be there, so we
    can't just remove it.
    
This resolves #2217.
This also resolves #2220 since those places was given a similar
    treatment.


opudrovs (approved):
LGTM!

PR #2251, opened 2022-05-31
use window location to refresh to reset application

ahussein3:
Resolve [#745](https://github.com/weaveworks/weave-gitops-enterprise/issues/745)


jpellizzari (commented):
You can instead do:

```ts
const queryClient = useQueryClient()
queryClient.invalidateQueries()
```

https://react-query.tanstack.com/guides/query-invalidation

ahussein3:
Not all APIs Use react query yet.

foot:
> Not all APIs Use react query yet.

Good point. The react-query apis are the only ones w/ cached responses.

But there is also the notification state, which seems to linger. Blowing away the whole app in-mem state is a bit safer just in case.

opudrovs (approved):

PR #2252, opened 2022-05-31
Move kube Client creation into the handlers and out of the middleware

foot:
**What changed?**

Moved client instantiation from the middleware to the handlers.

**Why was this change made?**

- If there is error creating any of the client it will prevent all endpoints running.
- There are some endpoints in EE that we'd like to continue to function even if some clusters cannot respond. Filtering certain routes from the middleware is tricky and doesn't feel quite right.

Notes:
- We could consider pushing client creation even further down so the failure to create 1 client would allow partial results from other successul clients
- Client instantiation is not "free", in that they may often build up CRD discovery caches etc which can sometimes take a second or two.

Needs an integration test, stay posted..


bigkevmcd:
I think we can probably provide a limited-time per-principal cache for clients, keying on the principal details.

ozamosi (approved):

PR #2272, opened 2022-06-06
Fix loading icon size inside buttons

### LINKED ISSUE #2228: SYNC additional options does not create a floating menu
**Describe the bug**

On the staging cluster I can see that when I click the arrow for the dropdown option "SYNC WITHOUT SOURCE" that it bumps the rest of the content down on the detail page.  This is not the expected behavior.  I would expect for that to be a floating menu option and for it to not adjust any of the content underneath.

**Environment**

- Weave-Gitops Version
main-26c1230

- Flux Version
N/A

- Kubernetes version
v1.22

**To Reproduce**

**Expected behavior**

I would expect for that to be a floating menu option and for it to not adjust any of the content on the detail screen.

**Actual Behavior**

When you click the down arrow the menu appears and pushes content down.  When it is dismissed the content moves back up.

joshri:
Closes #2228 

When font sizes were changed recently, the loading icon on buttons became too large, which would adjust button size when it appeared. It wasn't tied to the theme so I didn't catch it, but now it is with the `useTheme` hook from styled components. Hooray!

### DIFF
--- a/ui/components/Button.tsx
+++ b/ui/components/Button.tsx
@@ -2,7 +2,7 @@
 import { CircularProgress, PropTypes } from "@material-ui/core";
 import MaterialButton, { ButtonProps } from "@material-ui/core/Button/Button";
 import * as React from "react";
-import styled from "styled-components";
+import styled, { useTheme } from "styled-components";
 
 /** Button Properties */
 export interface Props extends ButtonProps {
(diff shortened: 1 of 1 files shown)


opudrovs (approved):
Tested locally, LGTM.

PR #2273, opened 2022-06-06
Add number lines to yaml tab

### LINKED ISSUE #2229: Add number line to YAML code block
As a user, I would like to be able to make better sense of large YAML blobs so I can pinpoint colleagues to the same area of YAML that I am looking at.  

When I am viewing any YAML within Weave GitOps I would like to see a number line down the left side of the yaml like I would in a traditional IDE.

### Acceptance Criteria

1. When I go to the YAML tab I should see an auto incrementing number line down the side that is aligned to the text on the right side.  See screenshot for detail.  
2. The YAML should be formatted properly as a Software Engineer would expect to see it formatted (proper indentation, spacing, line break, etc.)

### Out of Scope
1. The `kubectl` command at the top of the yaml box.

joshri:
Closes #2229 

Adds a counter to the yaml view with CSS, based on this stack overflow post: https://stackoverflow.com/questions/41306797/html-how-to-add-line-numbers-to-a-source-code-block

The wrapping was getting a bit crazy, so I changed the view to `overflow: scroll` as well.
image

### DIFF
--- a/ui/components/YamlView.tsx
+++ b/ui/components/YamlView.tsx
@@ -7,8 +7,6 @@ type Props = {
 };
 
 function YamlView({ yaml, className }: Props) {
-  console.log(yaml.split("\n"));
-
   return (
     <pre className={className}>
       {yaml.split("\n").map((yaml, index) => (
--- a/ui/components/YamlView.tsx
+++ b/ui/components/YamlView.tsx
@@ -41,6 +41,7 @@ export default styled(YamlView).attrs({
   }
 
   code::before {
+    width: 24px;
     color: ${(props) => props.theme.colors.primary10};
     content: counter(listing);
     display: inline-block;
--- a/ui/components/YamlView.tsx
+++ b/ui/components/YamlView.tsx
@@ -4,14 +4,16 @@ import styled from "styled-components";
 type Props = {
   className?: string;
   yaml: string;
-}
+};
 
 function YamlView({ yaml, className }: Props) {
+  console.log(yaml.split("\n"));
+
   return (
     <pre className={className}>
-      <code>
-        {yaml}
-      </code>
+      {yaml.split("\n").map((yaml, index) => (
+        <code key={index}>{yaml}</code>
+      ))}
     </pre>
   );
 }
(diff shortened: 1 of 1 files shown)


ozamosi:
Scrolling works well. Copy-paste works well.

However, in this screenshot, `generation` and `finalizers` are at the same level of indentation, but they look different due to the line number reaching double digits:
Screenshot from 2022-06-07 12-06-26

joshri:
UPDATE 2: It really looks like 28px will support 4 digits so I will leave it at that....FOR NOW

image

opudrovs (approved):
@joshri looks totallycool! 🎉 

Minor issues:
- Selecting individual lines works fine for me.

- When clicking on a line and dragginng the mouse pointer down, when hovering over empty space (not another line's text), the selection moves up, then if I continued dragging, the selection moves down, where it should be. Attaching a screenshot (on the screenshot I am still dragging down, but the mouse pointer is above blank area to the right of a line's text).

Screenshot 2022-06-07 at 21 48 36

- Sometimes the selection is blinking when selecting text by dragging even when trying to drag only over text (probably because of some events firing for individual line elements?).

- Right-clicking and clicking Select All also selects text above and below of the YAML area (footer), is it by design?

I'm not sure if it is worth it fixing these minor issues right now or at all in the near future, the main functionality is working, so LGTM.

joshri:
> I'm not sure if it is worth it fixing these minor issues right now or at all in the near future, the main functionality is working, so LGTM.

It might be easier to just have a button to copy the entire yaml code to clipboard than to fix these things - I have no idea how to control what's in a select all section (maybe it's the semantic html) but I agree it's something we can look into in a follow-up

opudrovs:
Yes, a Copy button should work fine. In theory, we could add a third-party React YAML editor component, but it's not the main functionality in our app, I think it's not worth the drawbacks. A new package with complex functionality would make the UI heavier (Material-UI isn't exactly very lightweight on its own) and would introduce more potential bugs and unneeded dependencies. So, a Copy button is a good compromise.

PR #2281, opened 2022-06-07
Detail Pages filter by Type

### LINKED ISSUE #2253: Detail Pages filter by Type
As a user, my Kustomizations and Helm Releases are going to create a lot of different objects.  I want to be able to filter based on `Type` quickly.

In the filter menu add the `Type` as another category to filter by.

### Acceptance Criteria

I have a Kustomization or Helm Release.  When I go to the detail page for either one I should be able to

1. Open the filter menu and see "Type" as another category in the filter menu.  Along with a list of all the different types that I can filter by.  For example, if I have `Pod`, `Service`, and `Namespace` then I should see all 3 listed in the menu.
2. When I select `Pod` from the example above I should see just the rows that have `Pod` as a type.  I should also see a chip above the table labeled `type:Pod` as well as `Clear All`.
3. When I select `Service` I should see all rows that are marked `Pod` or `Service`.  I should see three chips above the table labeled `type:Pod`, `type:Service`, and `Clear All`.
4. When I hit "x" on the chip `type:Pod` then I only see rows with `Service`.
5. When I hit "x" on the chip `Clear All` then I see all rows.
6. When I repeat steps 2-3 and then "uncheck" `Service` from the filter menu then I only see rows with type `Pod`.

opudrovs:
Closes #2253

Added a custom filtering config and added support for filtering rows based on type in the `filterRows` function.


ozamosi:
In `SourcesTable` and `AutomationsTable`, I hacked a similar thing into being by doing
```
  sources = sources.map((s) => {
    return { ...s, type: displayKind(s.kind) };
  });
```
so I could call `filterConfigForString(sources, "type")` after. Which, I'm not proud of. Also, I wish that if we introduce a `filterConfigForType`, then I could just ploink it in secretly and have everything be a little bit nicer.

So now I started thinking. In the `DataTable` itself, every value can either be a string, in which case we look it up in the object, or a function, in which case we call the function with the row as the value. When I first looked at this code my intuition was that if I can pass it to the table itself, I should be able to pass it to a filter function - then I learned it doesn't work this way. But, would it be hard to make it work the same? So, like a `filterConfigForCallback` that takes a function, or just a `filterConfig` that always works the same as `DataTable`?

joshri (approved):
LGTM

joshri:
> In `SourcesTable` and `AutomationsTable`, I hacked a similar thing into being by doing
> 
> ```
>   sources = sources.map((s) => {
>     return { ...s, type: displayKind(s.kind) };
>   });
> ```
> 
> so I could call `filterConfigForString(sources, "type")` after. Which, I'm not proud of. Also, I wish that if we introduce a `filterConfigForType`, then I could just ploink it in secretly and have everything be a little bit nicer.
> 
> So now I started thinking. In the `DataTable` itself, every value can either be a string, in which case we look it up in the object, or a function, in which case we call the function with the row as the value. When I first looked at this code my intuition was that if I can pass it to the table itself, I should be able to pass it to a filter function - then I learned it doesn't work this way. But, would it be hard to make it work the same? So, like a `filterConfigForCallback` that takes a function, or just a `filterConfig` that always works the same as `DataTable`?

If I'm understanding this correctly - yes it's hard because we usually want to return different things based on whether we're converting the data for table use or filter use. Sometimes we're returning a component, or a different string than the one we want displayed to the user in the filters. I think most of this issue was brought about simply because we made the sorting first, and the filtering second - there's definitely some overlap where things could be simplified....probably.

ozamosi:
To be clear: I don't want to block this PR until the whole filtering system has been rewritten. My specific question is "could we make this more general", which is what I think `filterConfigForCallback` would be - and as it's a question, the answer can be "no".

"Could we enable filtering of everything we can display" is 100% out of scope - that's me thinking out loud.

opudrovs:
@ozamosi thanks, an interesting suggestion, I need to think about.

Want to try your mapping hack first.

opudrovs:
@ozamosi @jpellizzari @joshri Added a filterConfig which accepts a key and optionally a callback.

Added tests to test filtering by type with callback.

ozamosi (approved):

PR #2282, opened 2022-06-07
Kustomizations and Helm Releases display images in tables

### LINKED ISSUE #2032: Show container image for applications

joshri:
Closes #2032 

The ReconciledObjectsTable now displays container images across multiple lines
image

### DIFF
--- a/ui/components/ReconciledObjectsTable.tsx
+++ b/ui/components/ReconciledObjectsTable.tsx
@@ -143,8 +143,11 @@ function ReconciledObjectsTable({
 export default styled(ReconciledObjectsTable).attrs({
   className: ReconciledObjectsTable.name,
 })`
-  td:nth-child(5) {
+  td:nth-child(5),
+  td:nth-child(6) {
     white-space: pre-wrap;
+  }
+  td:nth-child(5) {
     overflow-wrap: break-word;
     word-wrap: break-word;
   }
--- a/ui/components/ReconciledObjectsTable.tsx
+++ b/ui/components/ReconciledObjectsTable.tsx
@@ -9,7 +9,7 @@ import {
 } from "../lib/api/core/types.pb";
 import { formatURL, objectTypeToRoute } from "../lib/nav";
 import { NoNamespace } from "../lib/types";
-import { addKind, imageString, statusSortHelper } from "../lib/utils";
+import { addKind, makeImageString, statusSortHelper } from "../lib/utils";
 import { SortType } from "./DataTable";
 import FilterableTable, {
   filterConfigForStatus,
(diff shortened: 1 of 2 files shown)


ui/components/ReconciledObjectsTable.tsx
```diff
@@ -137,8 +143,11 @@ function ReconciledObjectsTable({
 export default styled(ReconciledObjectsTable).attrs({
   className: ReconciledObjectsTable.name,
 })`
-  td:nth-child(5) {
+  td:nth-child(5),
+  td:nth-child(6) {
```
opudrovs, line 147: These two selectors can be combined as a range (but your version is easier to read):
```
td:nth-child(n+5):nth-child(-n+6) {
    white-space: pre-wrap;
}
  ```

File: ui/lib/utils.ts [hunk 2]
```diff
@@ -79,3 +79,13 @@ export function removeKind(kind: string): string {
   }
   return kind;
 }
+
+export function imageString(images: string[]) {
```
opudrovs, line 83: It is common to start the name of a function with a verb (`make`, `compose`, `generate`, I don't know, whatever fits). But it's a subjective preference.

ui/lib/utils.ts
```diff
@@ -79,3 +79,13 @@ export function removeKind(kind: string): string {
...
+  let imageString = "";
+  if (!images[0]) imageString = "-";
+  else imageString += images[0];
+  if (images[1]) {
+    for (let i = 1; i < images.length; i++) imageString += `\n${images[i]}`;
+  }
+  return imageString;
+}
```
opudrovs, line 91: Would it be possible to use `reduce` or smth. like that for this?

opudrovs (approved):
LGTM, minor comments.

PR #2292, opened 2022-06-08
Replace `npm ci` with `npm i` in releases

ozamosi:
This is a follow-up to f3e7230ff - that commit made sure that `npm ci`
    stopped changing the lock file (as expected from that command), but
    that means that when running `npm i` right now, package-lock.json
    doesn't change, which means the lock file's version doesn't change.


opudrovs (approved):

ozamosi:
Argh, `npm install` was stabilising, but whenever I remove `node_modules` in between then it just enters a loop where it changes the file every commit. I've changed the PR to only re-generate it completely at release time, anxiety be damned.

PR #2294, opened 2022-06-08
js: Test that all dependencies are listed, add yaml dependency

ozamosi:
It seems I didn't add yaml to package.json - but everything worked
anyway, until we tried to publish a npm package, and that was the
point at which it broke. Let's run the command that discovered a
problem on every PR, to make sure I don't screw this up again.


ozamosi:
Breakage: https://github.com/weaveworks/weave-gitops/runs/6796694127?check_suite_focus=true

joshri (approved):

PR #2297, opened 2022-06-09
Adds multiline and link support to metadata

### LINKED ISSUE #2224: Multiline annotation support

### LINKED ISSUE #2280: Improve display of metadata
If you add metadata to your objects, there will be a metadata table on all detail pages as described in #2222 and #2223 - however before those stories are finished, we need to improve how values are displayed:

* When the value is a HTTP or HTTPS URL, we should wrap it in a clickable link.
* When the value contains special characters such as `<`, `>` etc, we should print it as such, not evaluate it as HTML tags.

These two things will cover point 4, 8 and 9 of #2223.

joshri:
Closes #2224 
Closes #2280 

A couple styling changes allow for multiline support in metadata annotations, as well as a quick loop that supports http or https links. I wanted to put the loop into the info list component instead but we're already passing down non-strings in the pair.
image

### DIFF
--- a/ui/components/Metadata.tsx
+++ b/ui/components/Metadata.tsx
@@ -17,10 +17,11 @@ function Metadata({ metadata, className }: Props) {
   }
 
   metadata.forEach((pair) => {
-    if (isHTTP(pair[1]))
+    const data = pair[1];
+    if (isHTTP(data))
       pair[1] = (
-        <Link newTab href={pair[1]}>
-          {pair[1]}
+        <Link newTab href={data}>
+          {data}
         </Link>
       );
   });
--- a/ui/components/Metadata.tsx
+++ b/ui/components/Metadata.tsx
@@ -1,8 +1,10 @@
 import * as React from "react";
 import styled from "styled-components";
-import Text from "./Text";
+import { isHTTP } from "../lib/utils";
 import Flex from "./Flex";
 import InfoList from "./InfoList";
+import Link from "./Link";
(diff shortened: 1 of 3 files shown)


ui/components/Metadata.tsx
```diff
@@ -1,22 +1,35 @@
...
   if (!metadata?.length) {
-    return (<></>);
+    return <></>;
   }
 
+  metadata.forEach((pair) => {
+    if (isHTTP(pair[1]))
+      pair[1] = (
+        <Link newTab href={pair[1]}>
+          {pair[1]}
```
opudrovs, line 23: It might be easier to read/more efficient to assign `pair[1]` to a constant in the three places where you are just reading the value instead of accessing the element by index.

joshri: good idea. will do.

joshri: WAIT but I have to reassign it - If I set it to a variable I'd just be changing the variable not the actual data right?

opudrovs: I mean smth. like this:

```
metadata.forEach((pair) => {
    const val = pair[1];

    if (isHTTP(val))
      pair[1] = (
        <Link newTab href={val}>
          {val}
        </Link>
      );
  });
```

ozamosi:
I think I've been aware of the `white-space` css property for about 15 years - I had _no idea_ it could do that. :exploding_head:

PR #2298, opened 2022-06-10
Display missing resources from a Kustomization's inventory

### LINKED ISSUE #2254: Kustomization Detail Page missing K8s Resources
As a platform operator using Flux I want to see a list of all my resources that are produced from a Flux Kustomization.  It is important that the Weave GitOps application gives me a full insight into everything that it is producing.

Today, when I go to the Kustomization detail and graph tabs I only see a partial representation of the objects that are created by my Kustomization.  This could give me quite a shock when I go to my Loki stack which shows no stateful sets, but they are there if you run `flux tree`.

### Acceptance Criteria

Create a Kustomization with various native k8s objects and custom CRDs as well such as

**Native Resources**
1. Namespace
2. ClusterRoleBinding
3. RoleBinding
4. Services
5. Config Maps
6. Jobs
7. Stateful Sets

**CRDs**
1. Kustomization
2. Helm Release
3. Git Repositories
4. Terraform (this require installing the tf-controller)
5. Ingress (this requires installing Contour)

Go to the detail page for that Kustomization.

1.  At a minimum I expect to see all of the objects listed on the `detail page` along with what is produced by `flux tree`.
2. If there is a status for those particular objects we should also show the status.

### Out of Scope
Fixing the graph tab on the detail view will be done in a separate story.
This does not address the HelmRelease detail page.

### Insights

From @stefanprodan 

The nice thing about Flux inventory is that you don’t need RBAC to list all things that Flux reconciles, you only need read access for Flux Kustomizations status. IMO we should list all the objects in the inventory, and if their kind matches what Weave GitOps has access to e.g. Deployment, then it should enrich the row with readiness info. This way people can use the UI to debug Flux, right now they must use the Flux CLI to see what things Flux manages in the cluster.

chanwit:
Fixes #2254

This PR
  - expanded role permissions to cover missing resources
  - fixed the UI behavior of a StatefulSet to query its Pod children. 
    This is similar to what we did for Deployment -> ReplicaSets -> Pods.
  - did some refactoring


ozamosi (changes requested):
This also needs to update website/docs/configuration/user-permissions.mdx

Could we make this configurable? Then the next time someone asks about more/less permissions they can just update their values file.

chanwit:
> This also needs to update website/docs/configuration/user-permissions.mdx
> 
> Could we make this configurable? Then the next time someone asks about more/less permissions they can just update their values file.

Via Helm values, it seems? Will look into it.

chanwit:
The early result. Please don't mind the red block - that's caused by a Kubernetes bug.
image

ozamosi (approved):

PR #2307, opened 2022-06-10
Update dev docs with build error due to tilt expanding docker image size

joshri:
Here's what I wrote - I'm not sure if everyone uses Docker Desktop but it's my reference point. I guess I can leave that bit out?

Is your local build failing with an error along the lines of "no space left on device"? This (likely) isn't the storage on your machine - it's Docker! You can go to Preferences -> Resources in Docker Desktop and pump up that image size a bit, but the size grows every time tilt rebuilds the image, so eventually you're going to want to run `docker system prune -a` to clear things out.


ozamosi (approved):
>I'm not sure if everyone uses Docker Desktop but it's my reference point. I guess I can leave that bit out?

Leave it in - I think that's the one you get when you click on the docker link at the top, right? So that's the one we're assuming. If you're using a different one, then it's now clear the exact steps won't apply to you, but the symptom and solution might be the same.

Issue #2284, opened 2022-06-08
Filtering does not work on a Source Details page


opudrovs:
**Describe the bug**

Using filtering on a Source Details page changes the page URL to an invalid string and displays an error (please see the attached screenshot).

**Environment**

- Weave-Gitops Version[main-1a17518](https://github.com/weaveworks/weave-gitops/commit/1a17518)
- Flux Version v0.29.5
- Kubernetes version 1.22.8-gke.201

**To Reproduce**

Navigate to Sources > flux-system or Sources > podinfo or Sources > weave-gitops (possibly any other Source Details page) and try applying any filters to the table.

**Expected behavior**
The URL in the browser address bar does not change (the URL-changing functionality has not been implemented yet), the table is filtered, no errors are displayed. See an automation details page for an example of expected behavior.

**Actual Behavior**
The page URL changes with possibly filtering the main object out of the URL and an error is displayed.

**Additional Context (screenshots, logs, etc)**

<img width="1242" alt="Screenshot 2022-06-08 at 12 09 01" src="https://user-images.githubusercontent.com/8824708/172597430-631b608b-b454-4a75-a90e-10038a2be4df.png">

PR #2308, opened 2022-06-10
Fix invalid URL caused by filters

joshri:
Closes: #2284 

Massively simplifies the `filterSelectionsToQueryString` function and fixes an error where it would replace the entire query string with just the filter part - `query-string` allows for us to isolate just the filter section of the query string and change it, then put it all back together again with `stringify()`. 

Adds a `URLAddressableTable` to the `ReconciledObjectsTable` component, meaning urls will actually show up on Kustomization or HelmReleaseDetail.

Fixes another bug I found while working on this - the table in the source detail page was using data from the `useListAutomations` hook without checking for a loading value, which would pass down no data to the filters, which would ruin the config and leave checkboxes empty. At least now we're accounting for if the hook is still running.

Definitely check out the branch and try to break everything so we can find out what else is wrong with these dang components please and thank you ❤️

### DIFF
--- a/ui/components/FilterableTable.tsx
+++ b/ui/components/FilterableTable.tsx
@@ -9,7 +9,7 @@ import FilterDialog, {
   FilterConfig,
   FilterSelections,
   filterSeparator,
-  selectionsToFilters
+  selectionsToFilters,
 } from "./FilterDialog";
 import Flex from "./Flex";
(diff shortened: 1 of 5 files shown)


opudrovs (approved):
Tried hard to make things break 😄 , but whatever I tested works as expected.

PR #2316, opened 2022-06-14
2269 Format metadata keys; 2301 Sort metadata key/value pairs by key

### LINKED ISSUE #2269: create function to format the key part of the metadata

### LINKED ISSUE #2301: When there are multiple metadata key/values they are sorted alphabetically
So even if the payload returns metadata.weave.works/foo before metadata.weave.works/bar. I would expect to see Bar then Foo on the detail page in the metadata section.

opudrovs:
Closes #2269 
Closes #2301 

- Added capitalizing words in metadata keys if needed.
- Added sorting metadata key/value pairs by key (by original key value, not formatted) alphabetically.
- Minor improvements:
-- now we are mutating a local copy of Metadata props instead of the props themselves, it is considered unsafe to mutate props in React and can lead to hard-to-detect bugs when component complexity increases.
-- if we check for the `http` substring in the `isHTTP` function we don't need to check for `https` after that.
-- instead of metadata keys the `ImageList` component now uses `item` with index as item keys, because some of the formatted keys can be the same (to prevent the duplicate key React error).
-- changed the metadata prop type to match the input type.
- Added snapshot tests for the `Metadata` component.
- Added tests for `lib/utils.ts` functions (they were just asking for it 😺 ).

### DIFF
--- a/ui/components/InfoList.tsx
+++ b/ui/components/InfoList.tsx
@@ -11,7 +11,7 @@ const InfoList = styled(
       <table className={className}>
         <tbody>
           {_.map(items, ([k, v], index) => (
-            <tr key={`key ${index}`}>
+            <tr key={`item ${index}`}>
               <td>
                 <Text capitalize semiBold color="neutral30">
                   {k}:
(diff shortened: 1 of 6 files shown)


joshri (approved):
Nice tests!!!!!!!!!! 🌞

chanwit (approved):

PR #2319, opened 2022-06-14
ADR 0014 Kubernetes Client API Fairness and Rate Limits

chanwit:
This PR proposes a new ADR for Kubernetes API Priority and Fairness in Kube 1.24 and re-adjusts values of QPS and Burst for Kube <= 1.23.

### DIFF
--- a/doc/adr/0014-k8s-client-api-fairness-and-rate-limits.md
+++ b/doc/adr/0014-k8s-client-api-fairness-and-rate-limits.md
@@ -7,24 +7,25 @@ Proposed
 ## Problem
 
 The UI part of Weave GitOps Core is an interactive Web application, which pulls the Kubernetes cluster periodically.
-The QPS configuration of the Kubernetes Client API is for controlling the maxium query per second to the API server from a client, which is Weave GitOps Core in this case.
+The QPS configuration of the Kubernetes Client API is for controlling the maximum query per second to the API server from a client, which is Weave GitOps Core in this case.
 Burst configuration is the maximum burst for throttle. When QPS is not specified, the default value is 5 query per second, and the default Burst value is 10.
 
-Before Kubernetes 1.23, QPS and Burst must be specifed only from the client side.
+Before Kubernetes 1.23, QPS and Burst must be specified only from the client side.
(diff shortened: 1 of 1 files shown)


SamLR (commented):
I think we should adopt this but I think we need to flesh out the consequences section a bit more.

Do we know what our average number of calls is and how it scales? As I understand it the number of calls to kube-api scales with namespaces (as resources have to be fetched for each one). If it is true that some of our customers have O(100s) of namespaces we may want to be careful as we could easily blow through those limits.

Assuming we don't have tons of head room on those limits, what happens then? Do we get 429s back from the kube-api and does our client handle those gracefully? I'm assuming this is something we'd mostly solve via documenting it's existence and how to adjust these limits?

chanwit:
> Do we know what our average number of calls is and how it scales?
@SamLR yes, we can measure this number from a single cluster usage. But for the scalability concern, I think it links to our performance and caching stories, which would be taken into account in the near future.

My concern is that we should not scale the number of QPS up, but we rather need to optimize the number of calls down to a certain number. what do you think?

SamLR:
> My concern is that we should not scale the number of QPS up, but we rather need to optimize the number of calls down to a certain number. what do you think?

I would say that if that's the approach we want to pursue (and I think we should) then there is an additional consequence that needs to be added around monitoring/optimising the number of calls.

That being said that's somewhat a cultural thing around how ADRs are used. When I've used them previously they're very focused on the state of the project when the ADR is written and used to drive further work and explain it (e.g. once this ADR is accepted I'd expect several issues to be created linking back to it, and the work to 'optimise calls/monitor the average number of calls' to be one of them).

All that being said my issue is wholly predicated on the assumption that the number of calls to the kube-api scales roughly linearly with number of namespaces (and even then this may be addressable by making QPS a parameter and adding some documentation). I'm not sure who's best placed to answer that (@ozamosi, maybe?)

SamLR (approved):

ozamosi (approved):
>there is an additional consequence that needs to be added around monitoring/optimising the number of calls.

This is a great point, and we should do more work regarding monitoring - here, and in other places.

PR #2322, opened 2022-06-14
Display oci helm repositories well

### LINKED ISSUE #2266: SPIKE: Behavior of OCI Helm Chart in Weave GitOps
As a user,
I would like Helm with OCI support in Weave GitOps
so that I could see which Helm Release was installed from which OCI image.

## Technical bits

The new version of HelmRepository will support ".spec.type: oci" to indicate that the source of a HelmRelease will be an OCI image.

image

ozamosi:
This upgrades the source controller, exports the helm repository type, and then displays it.

It also disables linking to oci URLs, and handles helm repositories that don't have a last updated time in both the source table and the details page - there's more details why in the UI commit.

End result:

Screenshot from 2022-06-14 18-33-06

Screenshot from 2022-06-14 18-33-20

This closes #2266


joshri (commented):
Do you know why the Breadcrumbs and Metadata snapshots changed? Seems like you didn't change anything that would effect those

opudrovs:
@joshri good catch! Fixed the tests and snapshots.

ui/components/HelmRepositoryDetail.tsx
```diff
@@ -45,6 +44,18 @@ function HelmRepositoryDetail({
   );
 }
 
+function tryLink(url: string) {
```
opudrovs, line 47: Just a suggestion, maybe we could start adding utility/helper function return types (if not tests). Smth. like `React.ReactElement<any, any> | string` could work.

opudrovs (approved):
The UI part looks good to me, tested it locally. Also looked through the backend part, looks good, but I don't know why the tests are failing.

ui/components/SourcesTable.tsx
```diff
@@ -108,8 +108,10 @@ function SourcesTable({ className, sources }: Props) {
                 return "-";
               case FluxObjectKind.KindHelmRepository:
                 text = (s as HelmRepository).url;
-                url = text;
-                link = true;
+                if ((s as HelmRepository).repositoryType != "OCI") {
```
chanwit, line 111: should this be a constant, please?
maybe `HelmRepositoryTypeOCI` ?

chanwit (approved):

PR #2343, opened 2022-06-17
release: Don't run make dependencies, don't set flux version

ozamosi:
Make dependencies doesn't exist, flux version is set by the make file already.


joshri (approved):
😈

PR #2345, opened 2022-06-21
Duplicates on detail pages table view

### LINKED ISSUE #2324: Duplicates on detail pages table view
**Describe the bug**

When I go to the detail page for a kustomization or helm release

**Environment**

- Weave-Gitops version: main-624342f
- Flux version: ???
- Kubernetes version: ???

**To Reproduce**

1. Go to the `flux-system` kustomization.
2. Make sure there are kustomizations on the page and you will see duplicates per the screenshot below.

**Expected behavior**

I would expect to not see duplicates.

**Actual Behavior**

There are duplicates.

opudrovs:
Closes #2324 

- Added de-duplicating reconciled objects by UID (based on @chanwit 's sample code).

- I haven't  been able to reproduce the same issue with child objects, so I've removed the de-duplication code I started adding for `GetChildObjects` and stored it. The current code fixes the original issue with the reconciled object table. If there is an additional issue with `GetChildObjects`, I can add the code back fast in another request.

- I thought of storing and sorting reconciled objects with the same ID within a pair to add only objects with bigger version numbers to the result (out of each pair of objects with the same UIDs), but noticed that v1beta2 objects come in first before v1beta1 objects, so there is no need to sort the objects in each pair. I hope this behavior is consistent.

### DIFF
--- a/core/server/fluxruntime.go
+++ b/core/server/fluxruntime.go
@@ -126,6 +126,8 @@ func (cs *coreServer) GetReconciledObjects(ctx context.Context, msg *pb.GetRecon
 
 	result := []unstructured.Unstructured{}
 
+	checkDup := map[types.UID]bool{}
+
 	for _, gvk := range msg.Kinds {
 		listResult := unstructured.UnstructuredList{}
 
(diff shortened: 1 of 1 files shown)


chanwit (approved):
LGTM thank you so much @opudrovs :1st_place_medal:

PR #2349, opened 2022-06-21
Make it easy to find out the k8s and flux version

### LINKED ISSUE #2285: Make it easy to find out the k8s and flux version
For reporting a bug, we expect users to provide k8s version and flux version. However, we don't really make those easy to find out at the moment.

We should add these to the version endpoint, so we can display them in the footer next to the gitops version.

The flux version is set in annotations on any flux object directly - you can see it, for example, in the `flux-system` namespace itself. It's theoretically possible for different components to come from different flux versions, but that's not normal and we don't consider that here.

The k8s version can be fetched by creating a [discovery client](https://pkg.go.dev/k8s.io/client-go/discovery) from an existing k8s client and calling `ServerVersion` on that. In a multi-cluster world these could be different, but as long as the `Default` cluster is queried, that's sufficient for this ticket.

### Acceptance Criteria
1. When I open Weave GitOps I see the version of Weave GitOps that is running on the footer with the same behavior as today.
2. To the left of the Weave GitOps version I will see the flux versions as `flux: v0.20.0`.
3. To the left of the Flux version I will see the `k8s: v1.23.0`.

There should be enough spacing to make them all legible.

### LINKED ISSUE #2328: Make it easy to find out the k8s and flux version - UI
The UI counterpart to #2285

See full acceptance criteria in the linked issue.

### Acceptance Criteria
1. When I open Weave GitOps I see the version of Weave GitOps that is running on the footer with the same behavior as today.
2. To the left of the Weave GitOps version I will see the flux versions as `flux: v0.20.0`.
3. To the left of the Flux version I will see the `k8s: v1.23.0`.

Design, provided by @JamWils 
Screen Shot 2022-06-14 at 8.47.00 PM.png
@opudrovs here is a quick design for you. I think we're going to need to make a dedicated settings page in the app moving forward for some of this data.

opudrovs:
Closes #2285
Closes #2328 

- ~~I've added a namespace to the `GetVersionRequest` object to make `TestGetVersion` pass because it could not find the `flux-system` namespace without it. So, for the test I create a new namespace or if the namespace is not provided default to `flux-system` (`wego.DefaultNamespace` constant). If anyone knows how to improve this and the issue with nested ifs (will add a comment), I'm all ears. 🙂~~

UPDATE: Removed the namespace field from `GetVersionRequest` (switched to listing namespaces) and fixed nested ifs based on @ozamosi 's suggestions.

- I've copy-pasted the test for `GetImpersonatedDiscoveryClient` from the test for `GetImpersonatedClient`. I am not sure what this API does and what exactly should be tested, maybe it should be modified.

Also removed the following from the copied test, maybe it should be put back (this part was failing because the number of found namespaces in the copy-pasted part was 0 instead of 1 or 2):

```
t.Run("checks all namespaces in the cluster when through the filtering", func(t *testing.T) {
	g.Expect(nsChecker.FilterAccessibleNamespacesCallCount()).To(Equal(1))

	_, _, nss := nsChecker.FilterAccessibleNamespacesArgsForCall(0)
	nsFound := 0
	for _, n := range nss {
		if n.Name == ns1.Name || n.Name == ns2.Name {
			nsFound++
		}
	}

	g.Expect(nsFound).To(Equal(2))
})
```


core/server/version.go
```diff
@@ -15,10 +21,66 @@ var (
...
+			cs.logger.Error(err, "error getting scoped client")
+		} else {
+			err = c.Get(ctx, key, u)
+			if err != nil {
+				cs.logger.Error(err, "error getting object")
+			} else {
+				fluxVersion = u.GetLabels()["app.kubernetes.io/version"]
+			}
+		}
+	}
```
opudrovs, line 61: @ozamosi @chanwit suggestions on how to simplify nested ifs here (looks similar to a "callback hell" in JS) welcome!

I don't want to return nil from `GetVersion` on error when getting additional values, because I want to display at least available values (at least the values set at the build time) in the app for debug purposes.

opudrovs:
@ozamosi @chanwit also, maybe I can use the `VersionLabelKey` from the flux.go instead of `app.kubernetes.io/version` to get the label value?

core/server/version.go
ozamosi: Maybe lift this bit into a new function, return an error from there any time there's an error, and then this function only has one error to check?

opudrovs: Thanks, good idea, will do it.

core/server/version.go
```diff
@@ -15,10 +21,66 @@ var (
...
+
+	if msg.Namespace != "" {
+		ns = msg.Namespace
+	} else {
+		ns = wego.DefaultNamespace
+	}
+
+	key := client.ObjectKey{
+		Name: ns,
+	}
```
ozamosi, line 42: So, you're right that this can be pretty much any namespace - but I'm not sure the frontend knows what namespace to use :thinking: 

What if instead of `Get`ing a specific namespace, we `List` filtered by `{"app.kuberenetes.io/part-of": "flux"}` and just grab the first result we find? After a bit of grepping - I think this helper does the filtering? https://github.com/weaveworks/weave-gitops/blob/main/core/server/server.go#L86

opudrovs: OK, can do it.

From frontend we don't provide any namespace. This param was added just for passing the test (because when running the test it could not find the "flux-system" namespace). If the namespace is not provided (like when we make the request via the API, `{}` with no args), we just default to "flux-system" (and while the app is running normally, the "flux-system" namespace is found).

I need to see if the list helper will return the flux-system/flux-whatever namespace while running the test.

ozamosi:
> @ozamosi @chanwit also, maybe I can use the `VersionLabelKey` from the flux.go instead of `app.kubernetes.io/version` to get the label value?

Sounds good!

core/server/version.go
```diff
@@ -14,11 +22,81 @@ var (
...
+	c, err := clustersClient.Scoped(clustersmngr.DefaultCluster)
+	if err != nil {
+		return defaultVersion, fmt.Errorf("error getting scoped client: %w", err)
+	}
+
+	err = c.Get(ctx, key, u)
+	if err != nil {
+		return defaultVersion, fmt.Errorf("error getting object: %w", err)
+	} else {
+		return u.GetLabels()["app.kubernetes.io/version"], nil
```
chanwit, line 53: u.GetLabels() may return `nil` here.

opudrovs: Thanks, will keep it in mind.

opudrovs:
@ozamosi @chanwit added grabbing a namespace from the list of namespaces. For testing, I create a new namespace with required key because the `flux-ns` namespace available during testing doesn't have a Flux version.

Also moved creating scoped client, getting Flux and K8s version to separate functions, based on @ozamosi 's suggestion.

@chanwit just in case added checking for nil when getting the version from labels.

@joshri I've added displaying Flux and Kubernetes versions in the footer. Also added `noWrap` to the Text component and  removed neutral30 color specified for some footer elements because it is already specified for the whole footer, it looks like, the elements already inherit this color from footer. If it is wrong, can move it back.

Spaced elements based on the draft design in the UI ticket description to make it consistent with current styling of the Footer component. Please let me know if styling/spacing/whatever is wrong.

opudrovs:
@ozamosi @chanwit also, would it be possible to filter for Version label key with any value? Or only filtering by specific values is possible?

```
opts := client.MatchingLabels{
    coretypes.PartOfLabel: FluxNamespacePartOf,
    flux.VersionLabelKey: <any value>
}
```

Right now I am filtering labels only by `coretypes.PartOfLabel: FluxNamespacePartOf,` and then loop through the result and find a label which has `flux.VersionLabelKey` key.

opudrovs:
@ozamosi OK, thank you, got it, will add the changes now.

ui/components/Footer.tsx
```diff
@@ -37,23 +37,54 @@ function Footer({ className }: Props) {
...
+      <>
+        <Text semiBold>Flux:</Text>
+        <Spacer padding="xxs" />
+        <Text>{fluxVersionText}</Text>
+        <Spacer padding="xxs" />
+      </>
+    ) : null;
+
+  const kubeVersionText = versionData.kubeVersion;
+  const kubeVersion =
```
joshri, line 52: fluxVersion and kubeVersion could potentially be combined into one component with props but I don't mind either way

ui/components/Footer.tsx
```diff
@@ -37,23 +37,54 @@ function Footer({ className }: Props) {
...
+        <Text>{kubeVersionText}</Text>
+        <Spacer padding="xxs" />
+      </>
+    ) : null;
+
   return (
     <Flex as="footer" wide between className={className} role="footer">
       <LeftFoot>
-        <Text color="neutral30">Need help? Contact us at</Text>
+        <Text>Need help? Contact us at</Text>
```
joshri, line 65: add the excellent `noWrap` prop here

ui/components/Footer.tsx
```diff
@@ -37,23 +37,54 @@ function Footer({ className }: Props) {
...
+            </Text>
+            <Spacer padding="xxs" />
+            <Link newTab href={versionHref}>
+              <Text semiBold>{versionText}</Text>
+            </Link>
+          </>
         )}
         <Spacer padding="xxs" />
-        <Text color="neutral30">© 2022 Weaveworks</Text>
+        <Text>© 2022 Weaveworks</Text>
```
joshri, line 87: same as above - add the excellent `noWrap` prop!!

ui/components/Footer.tsx
opudrovs: Added.

joshri (approved):
I had to update something...classic. Rebuilt and the footer looks good!

ui/components/Footer.tsx
opudrovs: Added the Version component with snapshot tests.

1) Not sure if adding `attrs` when exported is still required, like recommended in the style.md doc, so still added it like that:

```
export default styled(Version).attrs({ className: Version.name })``;
```

2) Footer versions are wrapped at narrow screens (see layout at 768 or 1000 px for example) , otherwise we would have to add a horiz scrollbar, preferably for the whole layout (or it looks ugly if added only to the footer).

Does this look OK?

Screenshot 2022-06-28 at 08 38 53

ozamosi (approved):

ui/components/Footer.tsx
joshri: Three options that come to mind:
1. Set everything to nowrap and just have it overflow the container 
2. Increase the minWidth of the container so we know we're always fitting the whole footer in
3. Decrease font size on the footer?
Maybe we do a mix of all three?

opudrovs: Used options 1 and 2: set the Version component text to nowrap and changed 768px in Layout.tsx from 768px to 900px (conveniently, a Material-UI breakpoints).

Everything I tried with scrollbars in the footer and footer containers looked ugly. So, I've decided not to add `overflow-x: auto`.

I think it's either fixes or we'll need to serious redesign the footer (for example, move it under the whole app content or move content info + email text somewhere from the footer), because its current text is really wide.

So, the current version looks best for now.

PR #2373, opened 2022-06-27
refactor: move cli options to package, refactor command args

jmickey:
**What changed?**

Move CLI arg options out of the root package and into a separate package. Refactor the command implementations to accept the new options type instead of individual args.

**Why was this change made?**

The args list for commands is becoming too long in many cases, and with the introduction of kubeconfig token passthru authentication coming soon it will probably get bigger.

### DIFF
--- a/cmd/config/options.go
+++ b/cmd/config/options.go
@@ -1,0 +1,11 @@
+package config
+
+type Options struct {
+	Endpoint              string
+	OverrideInCluster     bool
+	GitHostTypes          map[string]string
+	InsecureSkipTLSVerify bool
+	Username              string
+	Password              string
+	KubeconfigLocation    string
+}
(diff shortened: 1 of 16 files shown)


cmd/config/options.go
```diff
@@ -0,0 +1,11 @@
+package config
```
foot, line 1: Should this be a little bit further down inside of `gitops` too? 🤔

foot (approved):
🧹 🧹 🧹 ✨ 

Nice one! LGTM

cmd/config/options.go
ozamosi: There is a `cmd/internal` directory for some "CLI library code", this might go well in there.

jmickey: @foot @ozamosi I have moved the config package to `cmd/internal`. @ozamosi If you are happy to approve can you please also merge, thanks!

PR #2374, opened 2022-06-27
Link to images in detail pages

### LINKED ISSUE #2267: Make Image Containers links on Detail Page
We would want all docker images displayed to be links, but as discovered in #2326, different image repositories need to be linked to in different ways.

**Acceptance criteria**
Implement support for (these examples all come from #2326, look there for more of a description):
 * Dockerhub - all formats. E.g.
   * docker.io/weaveworks/eksctl -> https://hub.docker.com/r/weaveworks/eksctl
   * weaveworks/eksctl -> https://hub.docker.com/r/weaveworks/eksctl
   * docker.io/nginx -> https://hub.docker.com/r/_/nginx
   * nginx -> https://hub.docker.com/r/_/nginx
   * library/nginx -> https://hub.docker.com/r/_/nginx
   * docker.io/library/nginx -> https://hub.docker.com/r/_/nginx
  * Quay.io. E.g.:
    * quay.io/jitesoft/nginx -> https://quay.io/repository/jitesoft/nginx
   * GHCR. E.g.:
     * ghcr.io/weaveworks/charts/weave-gitops can just be used.
   * GCR. E.g.:
     * gcr.io/cloud-builders/gcloud can just be used.
 
This will be reused in more places, such as for OCI helmcharts and the flux runtime page, so please make sure it can be.

**Out of scope**
We do not want to link to the exact tag - just the image. Everything after the colon is always stripped out from the link

We will not right now link to image repositories we can't recognise - if it's not one of these cases, just print it as text.

We will not right now make this user-configurable.

joshri:
Closes #2267

Unsupported links display as text while we link to docker, quay.io, and google/github images

image


ui/components/ImageLink.tsx
```diff
@@ -0,0 +1,69 @@
...
+  const prefix = split.shift();
+  let url = "";
+
+  const makeUrl = (url: string, parts: string[]) => {
+    parts.forEach((part, index) => {
+      if (index === split.length - 1) url += part;
+      else url += `${part}/`;
+    });
+    return url;
+  };
```
opudrovs, line 17: It might be worth it moving this function to `lib/utils.ts` and adding tests for it (there are similar sample tests for the `isHTTP` function in `utils.test.ts`). It seems like a rather standalone utility function.

And besides adding tests would help to see how the resulting URL returned from the function is different from the input URL.

If it is moved to that file, instead of using local constants within function body directly, the values of these constants could be passed to the function as arguments.

opudrovs (approved):
LGTM! 🎉

PR #2379, opened 2022-06-29
Allow the oidc secret to be named other than 'oidc-auth' and improve logging

SamLR:
**What changed?**

* Added `--oidc-secret-name` flag to allow alternate secrets to be provided to the server
* Enabled creation of the `oidc-auth` secret as part of the helm chart
* Improved logging around OIDC start up
* Fixed some typos in the docs

**Why was this change made?**

Because I ran into problems relating to these issues several times whilst setting up Dex (particularly around the logging not providing feedback on why OIDC wasn't working). The additional flag was mostly a "while I'm here".

This should also mean that providing OIDC configuration via CLI actually works now (I believe it didn't previously because `featureflag.go` would fail to find a secret and then mark `OIDC_AUTH` as disabled).

Mostly through adding a new flag. The only debatable change was adding a global variable to the `pkg/server/auth/server.go` (`isOidcEnabled`) which replaces the `featureflag.go` secret lookup. I think this is a preferable method as the state of this flag is exposed by a function in the server.go file and should be the concern of it (rather than featureflag). Alternates involved threading a lot more state through the system as featureflag and the authserver are fairly disjoint. I think in theory the information could have been passed by context but that felt even messier.

I added some tests and ran it locally with dex in a variety of configurations (default `oidc-auth` secret, oidc passed by CLI and oidc from a different secret). All worked as expected.

The helm chart change was tested by inspection of the output of `helm template`

**Release notes**

No

**Documentation Changes**

Fixed some types which gave incorrectly capitalised oidc parameters in the table and listed the namespace to use as the legacy `wego-system`

### DIFF
--- a/cmd/gitops-server/cmd/cmd.go
+++ b/cmd/gitops-server/cmd/cmd.go
@@ -133,7 +133,15 @@ func runCmd(cmd *cobra.Command, args []string) error {
 		Namespace: v1alpha1.DefaultNamespace,
 		Name:      auth.OIDCAuthSecretName,
 	}, &secret); err == nil {
+		if len(options.OIDC.ClientSecret) > 0 && len(secret.Data["clientSecret"]) > 0 {
+			log.V(logger.LogLevelWarn).Info("OIDC client configured by both CLI and secret. CLI values will be overridden.")
+		}
 		oidcConfig = auth.NewOIDCConfigFromSecret(secret)
+
+	}
+
+	if len(oidcConfig.ClientSecret) > 0 {
+		log.V(logger.LogLevelDebug).Info("OIDC config", "IssuerURL", oidcConfig.IssuerURL, "ClientID", oidcConfig.ClientID, "ClientSecretLength", len(oidcConfig.ClientSecret), "RedirectURL", oidcConfig.RedirectURL, "TokenDuration", oidcConfig.TokenDuration)
 	}
 
 	tsv, err := auth.NewHMACTokenSignerVerifier(oidcConfig.TokenDuration)
--- a/cmd/gitops-server/cmd/cmd.go
+++ b/cmd/gitops-server/cmd/cmd.go
@@ -47,6 +47,7 @@ type Options struct {
 	HelmRepoName                  string
 	Path                          string
 	LogLevel                      string
+	OIDCSecret                    string
 	OIDC                          auth.OIDCConfig
 	NotificationControllerAddress string
 	TLSCertFile                   string
@@ -77,6 +78,8 @@ func NewCommand() *cobra.Command {
 	cmd.Flags().StringVar(&options.TLSKeyFile, "tls-private-key-file", "", "filename for the TLS key, in-memory generated if omitted")
 	cmd.Flags().BoolVar(&options.Insecure, "insecure", false, "do not attempt to read TLS certificates")
 	cmd.Flags().BoolVar(&options.MTLS, "mtls", false, "disable enforce mTLS")
+
+	cmd.Flags().StringVar(&options.OIDCSecret, "oidc-secret-name", auth.DefaultOIDCAuthSecretName, "Name of the secret that contains OIDC configuration")
 
 	cmd.Flags().StringVar(&options.OIDC.IssuerURL, "oidc-issuer-url", "", "The URL of the OpenID Connect issuer")
 	cmd.Flags().StringVar(&options.OIDC.ClientID, "oidc-client-id", "", "The client ID for the OpenID Connect client")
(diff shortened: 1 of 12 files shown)


pkg/server/auth/server.go
```diff
@@ -109,19 +112,30 @@ func NewAuthServer(ctx context.Context, cfg AuthConfig) (*AuthServer, error) {
...
+		// Set isOidcEnabled now we have a valid provider
+		isOidcEnabled = true
 	}
 
 	return &AuthServer{cfg, provider}, nil
 }
 
+// SetOidcEnabled is intended for test use only
+func SetOidcEnabled(newVal bool) {
+	isOidcEnabled = newVal
```
bigkevmcd, line 124: https://github.com/golang/go/wiki/CodeReviewComments#initialisms

pkg/server/auth/server.go
```diff
@@ -109,19 +112,30 @@ func NewAuthServer(ctx context.Context, cfg AuthConfig) (*AuthServer, error) {
 		if err != nil {
 			return nil, fmt.Errorf("could not create provider: %w", err)
 		}
+		// Set isOidcEnabled now we have a valid provider
+		isOidcEnabled = true
```
bigkevmcd, line 116: Why not call `SetOidcEnabled()` here?

cmd/gitops-server/cmd/cmd.go
```diff
@@ -127,13 +130,26 @@ func runCmd(cmd *cobra.Command, args []string) error {
...
 		Namespace: v1alpha1.DefaultNamespace,
-		Name:      auth.OIDCAuthSecretName,
+		Name:      options.OIDCSecret,
 	}, &secret); err == nil {
+		if len(options.OIDC.ClientSecret) > 0 && len(secret.Data["clientSecret"]) > 0 {
+			log.V(logger.LogLevelWarn).Info("OIDC client configured by both CLI and secret. CLI values will be overridden.")
+		}
 		oidcConfig = auth.NewOIDCConfigFromSecret(secret)
+	} else if err != nil {
+		log.V(logger.LogLevelDebug).Info("Could not read OIDC secret", "secretName", options.OIDCSecret, "error", err)
```
bigkevmcd, line 148: Maybe use `logger.LogLevelError` ?

cmd/gitops-server/cmd/cmd.go
```diff
@@ -127,13 +130,26 @@ func runCmd(cmd *cobra.Command, args []string) error {
...
+	}
+
+	if len(oidcConfig.ClientSecret) > 0 {
```
bigkevmcd, line 151: It's usually easier to read `oidcConfig.ClientSecret != ""`

It's easier to understand what ClientSecret is (and they compile to the same code internally), and this is the style used in Effective Go https://go.dev/doc/effective_go

bigkevmcd (commented):
Thanks for fixing the docs.

cmd/gitops-server/cmd/cmd.go
ozamosi: >We should also try and standardise on the log string format.

Sure. Out of scope for this PR.

pkg/server/auth/server.go
ozamosi: This is not how we give feedback.

We give a suggestion, and explain to the author why we prefer our way to theirs. We do not slam a "RTFM" in their heads.

cmd/gitops-server/cmd/cmd.go
ozamosi: >Maybe use `logger.LogLevelError` ?

It's not an error that you haven't set up OIDC login.

core/server/featureflags.go
```diff
@@ -36,20 +37,7 @@ func (cs *coreServer) GetFeatureFlags(ctx context.Context, msg *pb.GetFeatureFla
...
-	if err != nil {
-		if apierrors.IsNotFound(err) {
-			flags["OIDC_AUTH"] = "false"
-		} else {
-			cs.logger.Error(err, "could not get secret for oidc")
-		}
-	} else {
-		flags["OIDC_AUTH"] = "true"
-	}
+	flags["OIDC_AUTH"] = strconv.FormatBool(serverauth.OidcEnabled())
```
ozamosi, line 40: This makes me so happy to see :tada:

ozamosi (approved):

cmd/gitops-server/cmd/cmd.go
bigkevmcd: > > Maybe use `logger.LogLevelError` ?
> 
> It's not an error that you haven't set up OIDC login.

This depends on the nature of the error, yes, if the secret is not found, it's "Info", but if it's not, it's perhaps Error?

pkg/server/auth/server.go
SamLR: because `SetOidcEnabled` is intended for testing only and frankly shouldn't exist. I'd like to minimise it's usage to test files only so if we can remove it it's as easy to remove as possible

pkg/server/auth/server.go
SamLR: I'm happy to change to `isOIDCEnabled` but IMO it's less readable and doesn't match the existing method names

PR #2387, opened 2022-06-30
Removes tags from image links

### LINKED ISSUE #2386: Docker image links aren't stripping the tags away

joshri:
Closes #2386

Adds a check on the last piece of the image string to remove anything after and including the `:` character plus a test - the tag will still show in the table but it will not be a part of the link

### DIFF
--- a/ui/lib/utils.ts
+++ b/ui/lib/utils.ts
@@ -130,33 +130,25 @@ export const convertImage = (image: string) => {
   //Github GHCR or Google GCR
   if (prefix === "ghcr.io" || prefix === "gcr.io") return "https://" + noTag;
   //Quay.io
-  else if (prefix === "quay.io") {
+  if (prefix === "quay.io") {
     url = "https://quay.io/repository/";
     return makeUrl(url, split);
   }
   //complex docker prefix case
-  else if (prefix === "docker.io") {
+  if (prefix === "docker.io") {
     url = "https://hub.docker.com/r/";
     //library alias
     if (split[0] === "library") return url + "_/" + split[1];
     //global
-    else if (!split[1]) return url + "_/" + split[0];
+    if (!split[1]) return url + "_/" + split[0];
     //namespaced
-    else {
-      return makeUrl(url, split);
-    }
+    return makeUrl(url, split);
   }
   //docker without prefix
-  else if (prefix === "library")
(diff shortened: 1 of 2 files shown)


ui/lib/utils.ts
```diff
@@ -147,9 +154,9 @@ export const convertImage = (image: string) => {
     !prefix.includes("public.ecr.aws") &&
     !prefix.includes("amazonaws.com")
```
ozamosi, line 155: Sorry to be reacting to things that weren't part of this PR :see_no_evil: 

I think you need to handle this case by counting slashes - if there's 0 slashes then the link is `hub.docker.com/_/image_name` (no r!), if there's 1 slash then the link is `hub.docker.com/r/image_name`, if there's 2 slashes then the first part is a domain and domain checking applies.

ui/lib/utils.ts
```diff
@@ -108,6 +108,13 @@ export function formatMetadataKey(key: string): string {
 
 export const convertImage = (image: string) => {
```
opudrovs, line 109: This function is somewhat hard to read. I suggest to refactor top-level else-ifs into just top-level ifs with returns. For example, instead of:

```
if (prefix === "ghcr.io" || prefix === "gcr.io") return "https://" + noTag;
else if (prefix === "quay.io") {
  url = "https://quay.io/repository/";
  return makeUrl(url, split);
}
```

we could have

```
if (prefix === "ghcr.io" || prefix === "gcr.io") return "https://" + noTag;
if (prefix === "quay.io") {
  url = "https://quay.io/repository/";
  return makeUrl(url, split);
}
```

This would improve readability.

opudrovs: And probably some of the nested if-elses could be simplified like that too (replaces with nested ifs).

ui/lib/utils.ts
joshri: No worries happy you're looking at it! Didn't realize there were cases without the `/r/` - will change it

ui/lib/utils.ts
joshri: brb refactoring

opudrovs (approved):
LGTM.

ozamosi (approved):

PR #2389, opened 2022-06-30
Nav fixed vertically, but scrolls horizontally

### LINKED ISSUE #2383: When the browser is very narrow, the content overflows the menu

joshri:
Closes #2383 

Here's the new behavior - a cool trick with `transform: translate` allows for a half-fixed styling of the nav so the page overflows horizontally as expected. Also decreased top tool bar min width so that the user settings icon is visible more often

https://user-images.githubusercontent.com/65822698/176738109-7da2e54f-bd6c-46d6-89ed-a73df917631f.mov

### DIFF
--- a/ui/components/Layout.tsx
+++ b/ui/components/Layout.tsx
@@ -69,12 +69,12 @@ const navWidth = "200px";
 const topBarHeight = "60px";
 
 const NavContainer = styled.div`
-  position: fixed;
   width: ${navWidth};
-  //topBarHeight + correct margins of 36px
-  height: calc(100% - 84px);
+  min-width: ${navWidth};
+  height: calc(100% - 24px);
   margin-top: ${(props) => props.theme.spacing.small};
-  margin-bottom: ${(props) => props.theme.spacing.small};
+  //topBarHeight + small margin
+  transform: translateY(72);
 `;
 
 const NavContent = styled.div`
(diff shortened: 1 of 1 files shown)


opudrovs:
Hiding navigation elements (the sidebar) when scrolling horizontally may be confusing to some users.

ozamosi:
The menu should turn to a hamburger when the window is too narrow, so this problem goes away :see_no_evil:

The z-index video doesn't feel right - my mental model is that it should work like a good, ol table layout - the page chrome should all be fixed, it's the content cell that should scroll. So the grey background bit between the navigation and the content should remain. I don't know if there's more elements where we could put z indexes on to achieve that.

But, there's way too much white space already, and I did just propose adding more wasted space to "fix" the lack of space. So being able to get rid of the navigation sounds like a feature.

But, I have a computer mouse, so unlike a trackpad I can't move my scroll wheel horizontally. Modern designers hate scroll bars, so it's impossible to see if there is one. If I start to select some text, my browser will scroll right, and if that hides the navigation then I can't quite "get back" without reloading the page, or very, very carefully selecting text in the opposite direction, or by wiggling my pointer by the bottom of the window until the scroll bar decides to reveal itself.

Personally, I fixed the problem of scrolling horizontally by getting a 4ft wide monitor, so I don't know how much I can really be trusted with an answer.

ozamosi (approved):

PR #2391, opened 2022-07-01
Add optional prometheus metrics

### LINKED ISSUE #2347: Export metrics for request duration

ozamosi:
If we want to be able to confidently refactor and improve the
application, then we need to keep metrics on before and after. This
may or may not be useful to end-users, but I definitely want this in
our staging cluster, and turning it on might help debugging user
environments.

This pulls in a 3rd party middleware to gather RED style metrics about
all of our endpoints - I see a histogram
`http_request_duration_seconds` that is labeled by handler, method,
and response code, I see a `http_requests_inflight` gauge labeled by
handler, and a `http_response_size_bytes` histogram labeled by
handler, method, and response code. This is an excellent start - I'd
want the same RED metrics for k8s client requests themselves, but that
can be a later problem.

This resolves #2347.


opudrovs (approved):
LGTM.

Issue #2355, opened 2022-06-22
Create new gitops beta run command


JamWils:
As a platform operator, I want to be able to get started with the power of Flux’s reconciliation process quickly.  I don’t know what Flux is capable of.  I don’t know how to “bootstrap” a cluster; and I heard of “GitOps” but really don’t know how to get started.  I want a simple command such as `gitops beta run <filepath>` to get started.

### Acceptance Criteria

When I run `gitops beta run <filepath>`

1. If I don’t have any cluster in my kube config then I will be shown an error.
2. Given I have a valid connection to a cluster when I run the command I expect the command to connect to my currently selected cluster.
3. If I don’t have Flux installed on the cluster then the prerequisites will be installed to initiate the reconciliation process.
    1. This includes all default controllers to set up a reconciliation loop such as the notification-controller, helm-controller, kustomization-controller, and source-controller.
    2. This will also add all relevant CRDs from the controllers above such as Kustomizations, Helm Releases, Git Repository, Helm Repository, Bucket, Alerts, Providers, and Receivers.
4.  Given I have Flux installed on the cluster then we do not need to install flux. 
5. This would be able to work with local (kind and k3d) and remote clusters.

### Out of Scope
1. This just includes installing Flux onto the cluster but does not involve creating the prerequisite reconciliation.
2. This does not include image auto CRDs and controller from Flux.  These are not normally installed by default and will be dealt with in a future story.

chanwit:
I would suggest the command name be `gitops runtime install` or `gitops install-flux` instead of `gitops run`.

JamWils:
Do you think something like `gitops runtime` makes more sense @chanwit?  Also, you have the idea of `install` or `add`.  I was viewing this as a long running process such as `port-forward`.  Is that the wrong approach?  It seems hard to pull this off without some process running on the local machine as well as the cluster.

ozamosi:
See notes in #2363 for how to generate manifests.

chanwit:
I emulated the behavior of gitops run with Tilt and found a gap.
here's the scenario:
  - A Dev had an Nginx program. The dev also had Deployment yaml for nginx.
  - The dev put the deployment yaml into ./dev
  - The dev run `gitops run ./dev`
then the Dev would ask what's next?

The missing puzzle in the above scenario was the port-fwd element (cc @ozamosi)

`gitops run ./dev` should expect a port-fwd parameter for a resource (a Service, a Pod etc)

[...]

PR #2401, opened 2022-07-04
Help command for gitops beta run

### LINKED ISSUE #2356: Help command for gitops beta run

opudrovs:
Closes #2356

- Added templates for the `beta` command with `run` subcommand.
- Added a short description for the `beta` command.
- Added a short and a long description for the `run` command.
- Added a positional argument requirement (this argument will be used for `filepath`) to the `run` command.

**Possible improvements/questions:**

1. There are `[flags]` mentioned in the `run` command help in #2356 , but I don't see specific flags in #2355 Probably it will be clarified what flags the `run` command can support when the actual command is added.

That's why, I've added an empty placeholder structure for flags (with suppressing the linter warning and a todo).

If the `run` command is supposed to be run with global flags only, I can remove the placeholder.

2. I've added checking for the enterprise endpoint set up in the `run` command, similar to the `add cluster` command. Is this check needed?

3. If the sync/run is a priority feature, should "how to get help for the`run` command" info be added to the command usage examples of the `gitops` command, similar to the `add cluster` command? Current description:

```
Example: `
  # Get help for gitops add cluster command
  gitops add cluster -h
  gitops help add cluster

  # Get the version of gitops along with commit, branch, and flux version
  gitops version

  To learn more, you can find our documentation at https://docs.gitops.weave.works/
`,
```
4. For now, at least one positional argument is required for the `beta run` command and the `cobra.MinimumNArgs(1)` validator displays a default error message: `Error: requires at least 1 arg(s), only received 0`.

If we should require exactly one argument I can switch to the range argument validator. I could also add a custom error message, like `the filepath has not been set` instead of just complaining about the missing argument. What is preferred for validating that positional arguments are provided?

5. I haven't split the long description for the run command to strings (because it's a single line in the ticket description), but it might be easier to read if split to lines shorter than 80 characters. This is what it looks like right now:

Screenshot 2022-07-04 at 20 35 40

Should I keep it like that? Or if it should be split, then how exactly?

### DIFF
--- a/cmd/gitops/beta/cmd.go
+++ b/cmd/gitops/beta/cmd.go
@@ -1,0 +1,19 @@
+package beta
+
+import (
+	"github.com/go-resty/resty/v2"
+	"github.com/spf13/cobra"
+	"github.com/weaveworks/weave-gitops/cmd/gitops/beta/run"
+	"github.com/weaveworks/weave-gitops/cmd/internal/config"
+)
+
+func GetCommand(opts *config.Options, client *resty.Client) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "beta",
+		Short: "This component contains unstable or still-in-development functionality",
+	}
+
+	cmd.AddCommand(run.RunCommand(opts, client))
+
+	return cmd
+}
--- a/cmd/gitops/beta/run/cmd.go
+++ b/cmd/gitops/beta/run/cmd.go
@@ -1,0 +1,50 @@
+package run
+
+import (
+	"github.com/go-resty/resty/v2"
+	"github.com/spf13/cobra"
+	"github.com/weaveworks/weave-gitops/cmd/gitops/cmderrors"
+	"github.com/weaveworks/weave-gitops/cmd/internal/config"
+)
+
+type runCommandFlags struct{}
+
+// TODO: Add flags when adding the actual run command.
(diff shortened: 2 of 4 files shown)


ozamosi:
> That's why, I've added an empty placeholder structure for flags (with suppressing the linter warning and a todo).

Sounds good :+1: - we'll inevitably want something :joy: 

> I've added checking for the enterprise endpoint set up in the run command, similar to the add cluster command. Is this check needed?

Please remove - the cluster functionality only works in enterprise, but this functionality is specifically for core-only users too.

> If the sync/run is a priority feature, should "how to get help for therun command" info be added to the command usage examples of the gitops command, similar to the add cluster command

Excellent question. My gut feeling is "it's either beta, or on the front page, not both" - or rather, let's not put it on the front page until it works.

> What is preferred for validating that positional arguments are provided?

I say, let's write a custom error message the way you suggested.

I don't mind too much what it says if there's too many arguments, but when I try out a command I often just run it and see what happens, so I'm very likely to see the error that there's too few arguments as my first interaction with the command, so it's worth making that look nice.

> I haven't split the long description for the run command to strings (because it's a single line in the ticket description)

Fair - I see none of the global arguments are anywhere near 80 characters.

ozamosi (approved):

PR #2406, opened 2022-07-05
Introduce BE featureflags, make the auth system use them

### LINKED ISSUE #2392: Allow dynamically adding more Feature Flags for library consumers
**Short description**
As a weave-gitops-library-consumer I want to be able to add additional featureflags so that I can enable / disable parts of my system by piggy-backing on top of the weave-gitops featureflag system.

**Acceptance criteria**
- [ ] I can declare additional environment variables in some form e.g. `WEAVE_GITOPS_FEATURE_FLAG_FOO` and it will appear in the response from `/v1/featureflags`
- [ ] I can call some function in `golang` e.g. `featureflags.Get("WEAVE_GITOPS_FEATURE_FLAG_FOO")` to see whether a flag is set?

ozamosi:
There is a bit of godoc in the new featureflags module - basically,
it's a static, global map. It is initialised automatically from server
startup by setting an env var starting with
`WEAVE_GITOPS_FEATURE_whatever`, and you can also call the `Set()`
function directly. The frontend featureflag endpoint now literally
just copies this map, so the backend and frontend are in sync.

To test it out, dump the following in `tools/helm-values-dev.yaml`:
```
envVars:
  - name: WEAVE_GITOPS_FEATURE_FOO
    value: it worked
```
Then visit the sign in page, and you'll see the frontend request
feature flags, which (after the backend pod restarted) now includes
this flag.

This changes the auth system to use the same flags - which has a
behaviour change: the pod will now refuse to start if there's no login
methods configured. This "preserves" the behaviour that you can create
a secret after installing the pod and have that be picked up, but it
does mean the pod is red until you've created it.

This fixes #2392.

### DIFF
--- a/cmd/gitops-server/cmd/cmd.go
+++ b/cmd/gitops-server/cmd/cmd.go
@@ -13,6 +13,7 @@ import (
 	"os/signal"
 	"path"
 	"path/filepath"
+	"strings"
 	"syscall"
 	"time"
 
@@ -25,6 +26,7 @@ import (
 	"github.com/weaveworks/weave-gitops/core/logger"
 	"github.com/weaveworks/weave-gitops/core/nsaccess"
 	core "github.com/weaveworks/weave-gitops/core/server"
+	"github.com/weaveworks/weave-gitops/pkg/featureflags"
 	"github.com/weaveworks/weave-gitops/pkg/kube"
 	"github.com/weaveworks/weave-gitops/pkg/server"
 	"github.com/weaveworks/weave-gitops/pkg/server/auth"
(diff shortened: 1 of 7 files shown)


foot:
Awesome! 🙏 , I'll take it for a spin.

What would your ideal helmchart/values.yaml look like? Explicit flags section or just extraEnv?

ozamosi:
My thinking is
 * feature flags for things that are incomplete can use `extraEnv` - they are for our use.
 * feature flags for things that are "released" should be somewhere where they are documented properly. `flags: {}` is probably a good idea.
 * If an optional feature requires more configuration (see e.g. #2391 which doesn't but could use this system) then I would rather have a `metrics: {enabled: true, other_settings: present}` and have `metrics.enabled` set the flag, rather than force making changes in 2 places.

opudrovs (approved):
LGTM.

Tested locally
- that the new feature flag added via `tools/helm-values-dev.yaml` is present in the network response
- that I can call `featureflags.Get("WEAVE_GITOPS_FEATURE_FOO")` and receive the feature flag's value.

PR #2410, opened 2022-07-06
feat: kubeconfig token passthrough authentication for WGE commands in CLI

jmickey:
Closes weaveworks/weave-gitops-enterprise#692
Closes weaveworks/weave-gitops-enterprise#582

**What changed?**

- Move CLI `config` package out of `internal` as it needs to be accessed by higher level packages.
- Rename the `--kubeconfig` flag on the `gitops get clusters` command as this now causes a conflict with the root `--kubeconfig` flag. The root flag is required in order to set the `--kubeconfig` flag within the `flag` package so that `kubernetes-sigs/controller-runtime/client/config` can find the kubeconfig if a custom location is passed.
- Instantiate a `*rest.Config` from the `kubeconfig`, and generate a `http.RoundTripper` from the `client-go/rest` package. This `http.RoundTripper` will handle the authentication requirements, generating a `transport.BearerToken` and apply it to the request header.
- Add token passthrough capabilities to the server. We don't want or need to validate the token, as the Kubernetes cluster will do this for us. Therefore if an `Authorization` header exists, we store the header within the current context. The token will then be applied to the `KubeHTTPClient` when requests are made to the management cluster.
- Slight refactor of the usage of `adapters.HTTPClient`. The client is now instantiated with auth in main and passed to root, rather than passing a `resty.Client`. In the `CmdRunE` function for each command we configure the client with the `client.ConfigureClientWithOptions` method, which configures the client with the necessary information to complete a request (e.g. Endpoint, TLS config) and handles auth if enabled.

**Why was this change made?**

The `gitops` CLI doesn't currently support authentication and therefore all enterprise CLI commands fail.

When the user issues a CLI command that requires authentication to the Weave GitOps Enterprise server they are able to supply either a username/password, or a kubeconfig file that is valid to the Kubernetes management cluster. If no username or password is provided, then we default to the kubeconfig with the following prededence:

* --kubeconfig flag pointing at a file
* KUBECONFIG environment variable pointing at a file
* In-cluster config if running in cluster
* $HOME/.kube/config if exists.

If a valid kubeconfig is located the CLI will generate a `(client-go/rest) *rest.Config` which is then used to generate a `http.RoundTripper` that will apply the necessary authentication information to the request (see: https://pkg.go.dev/k8s.io/client-go/rest#TransportFor). 

Once the request is made and received by the server [we check if the incoming request contains an `Authorization` header](https://github.com/weaveworks/weave-gitops/pull/2410/files#diff-b5353235fe4ac169f1b5e6e07bd40e49834ae65b69d8ec8dc012bfd68551599fR112). If so, we store this token within the request `context.Context` and continue. This token is then extracted later when building the `*rest.Config` that will be used to make requests from the WGE server to the Kubernetes API on the management server.

Change validated locally by running the `cluster-service` pointing to the demo cluster and issuing a command from the CLI to the local service. The CLI command returned successfully.

**Release notes**

- BREAKING: The `--kubeconfig` flag on the `gitops get clusters <cluster_name>` command has been renamed to `--print-kubeconfig`
- Enterprise customers can now authenticate to the Weave GitOps Enterprise API server when running CLI commands using a valid `kubeconfig` context for the management cluster which supplies an Authorization token.
- `gitops` commands that require authentication now support the `--kubeconfig <location>` flag, allowing users to specify a custom kubeconfig location. The `KUBECONFIG` environment variable is also supported. If neither is specified `$HOME/.kube/config` will be used.


pkg/kube/config_getter.go
```diff
@@ -35,7 +35,14 @@ func NewImpersonatingConfigGetter(cfg *rest.Config, insecure bool) *Impersonatin
 func (r *ImpersonatingConfigGetter) Config(ctx context.Context) *rest.Config {
 	shallowCopy := *r.cfg
 
-	if p := auth.Principal(ctx); p != nil {
+	var hasToken bool
+
+	if t := auth.BearerToken(ctx); len(t) != 0 {
+		shallowCopy.BearerToken = t
+		hasToken = true
+	}
```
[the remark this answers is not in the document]
jmickey: > The way this is written, I think the insecure flag only works if you didn't provide a token, which doesn't seem right.

The insecure flag will still work, it only bypasses the Principal check, although I am about to fix that. I believe @bigkevmcd is also going to be removing the insecure check in a later PR.

> Finally, I'm a bit confused. I thought the code we use for impersonation was in clustersmngr.go

The client in clustermgr.go is, I believe, mainly used for communication with leaf clusters from the wego server.

pkg/server/auth/auth.go
```diff
@@ -66,6 +69,7 @@ func Principal(ctx context.Context) *UserPrincipal {
 type UserPrincipal struct {
 	ID     string   `json:"id"`
 	Groups []string `json:"groups"`
+	Token  string   `json:"-"`
```
foot, line 72: Small reservation here, wrt bundling the token into the Principal. The `-` json tricks works great for responses, but I guess the principal might be quite easy to accidentally include in logging etc, where `%s` seems to still include it.

Maybe there are other tricks? Like implementing `String()`?

jmickey: Ah yes, right you are. I will add a `String()` method to the `UserPrincipal` struct to avoid this.

foot:
Whoop! Works for me, except w/ `--insecure-skip-tls-verify` :D, then `SetTLSClientConfig` seems to fail as its a k8s transport not an http transport:

`2022/07/11 19:50:22.667805 ERROR RESTY current transport is not an *http.Transport instance`

pkg/server/auth/auth.go
jmickey: Fixed!

pkg/kube/config_getter.go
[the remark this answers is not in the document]
foot: > I wish there was an on-off switch, rather than something everybody can always do.

For sure, we could definitely enable / disable token passthrough via a flag etc here?

foot (approved):
All looks good to me. 👌

ozamosi (approved):

PR #2411, opened 2022-07-07
Add functionality to gitops beta run command

### LINKED ISSUE #2355: Create new gitops beta run command

opudrovs:
Closes #2355


opudrovs:
It's a WIP, will clean up error handling and vars later.

cmd/gitops/beta/run/cmd.go
```diff
@@ -48,8 +72,268 @@ func betaRunCommandPreRunE(endpoint *string) func(*cobra.Command, []string) erro
...
+		changeSet.Append(cs.Entries)
+	}
+
+	return changeSet.String(), nil
+}
+
 func betaRunCommandRunE(opts *config.Options, client *resty.Client) func(*cobra.Command, []string) error {
 	return func(cmd *cobra.Command, args []string) error {
+		// If there is no cluster in the kube config, return an error.
+		cfg, ctx, err := kube.RestConfig()
```
opudrovs, line 264: Does using `RestConfig` work equivalently/errs the same way to the `outOfClusterConfig` (@ozamosi suggested) in our case?

Or do I need to copy-paste/expose code from the `outOfClusterConfig`

How is `InClusterConfig` different from `outOfClusterConfig`, why int he `RestConfig` we first try to create the first and if it fails the second one?

cmd/gitops/beta/run/cmd.go
```diff
@@ -50,6 +74,266 @@ func betaRunCommandPreRunE(endpoint *string) func(*cobra.Command, []string) erro
...
+		listResult.SetGroupVersionKind(schema.GroupVersionKind{
+			Group:   "",
+			Version: "v1",
+			Kind:    "Namespace",
+		})
+
+		ctx2 := context.Background()
+
+		listOptions := ctrlclient.MatchingLabels{
+			coretypes.PartOfLabel: "flux",
```
opudrovs, line 102: Is it OK to use `coretypes` here in a CLI command?

cmd/gitops/beta/run/cmd.go
```diff
@@ -50,6 +74,266 @@ func betaRunCommandPreRunE(endpoint *string) func(*cobra.Command, []string) erro
...
+	}
+
+	content := []byte(manifest.Content)
+
+	err = os.MkdirAll(filePath, os.ModePerm)
+	if err != nil {
+		return fmt.Errorf("couldn't create file %+v", err)
+	}
+
+	err = os.WriteFile(filepath.Join(filePath, shortFilename), content, 0666)
```
opudrovs, line 194: Am I making the dir and writing the file with the correct permission?

And should I use the gitops command directory or some working directory. If the latter, how should I get it. (I think I heard smth. mentioned about the working directory during the last standup but it was about another issue.)

cmd/gitops/beta/run/cmd.go
```diff
@@ -50,6 +74,266 @@ func betaRunCommandPreRunE(endpoint *string) func(*cobra.Command, []string) erro
...
+	// without the `-s` short version. While we're no longer on par with kubectl's flags, we maintain backwards compatibility
+	// on the CLI interface.
+	apiServer := ""
+	kubeconfigArgs.APIServer = &apiServer
+	// rootCmd.PersistentFlags().StringVar(kubeconfigArgs.APIServer, "server", *kubeconfigArgs.APIServer, "The address and port of the Kubernetes API server")
+
+	// kubeclientOptions.BindFlags(rootCmd.PersistentFlags())
+
+	// rootCmd.RegisterFlagCompletionFunc("context", contextsCompletionFunc)
+	// rootCmd.RegisterFlagCompletionFunc("namespace", resourceNamesCompletionFunc(corev1.SchemeGroupVersion.WithKind("Namespace")))
```
opudrovs, line 232: What should I do with global flags (if any of those supported by the `gitops` command are provided) in the `run` command here? I've just copy-pasted and commented out code from the Flux instal command.

Which of this or original `gitops` flags do I need to add here?

cmd/gitops/beta/run/cmd.go
```diff
@@ -50,6 +74,266 @@ func betaRunCommandPreRunE(endpoint *string) func(*cobra.Command, []string) erro
...
+		defaults:     install.MakeDefaultOptions(),
+	}
+
+	var VERSION = "0.0.0-dev.0"
+
+	rf.defaults.Version = "v" + VERSION
+
+	var kubeconfigArgs = genericclioptions.NewConfigFlags(false)
+
+	var kubeclientOptions = new(runclient.Options)
```
opudrovs, line 210: I am not sure should configure `kubeconfigArgs` and `kubeclientOptions` here. What exactly should I configure here for our purpose?

cmd/gitops/beta/run/cmd.go
```diff
@@ -50,6 +74,266 @@ func betaRunCommandPreRunE(endpoint *string) func(*cobra.Command, []string) erro
...
+	if err != nil {
+		return fmt.Errorf("couldn't create file %+v", err)
+	}
+
+	err = os.WriteFile(filepath.Join(filePath, shortFilename), content, 0666)
+	if err != nil {
+		return fmt.Errorf("couldn't write flux manifests to file %+v", err)
+	}
+
+	rf := rootFlags{
```
opudrovs, line 199: Will remove everything related to root flags, copy-pasted and I think here this code is not needed, we already used called `MakeDefaultOptions` when generating the manifests.

cmd/gitops/beta/run/cmd.go
```diff
@@ -50,6 +74,266 @@ func betaRunCommandPreRunE(endpoint *string) func(*cobra.Command, []string) erro
...
+
+func apply(kubeClient ctrlclient.Client, ctx context.Context, rcg genericclioptions.RESTClientGetter, opts *runclient.Options, manifestPath string, manifestContent []byte) (string, error) {
+	objs, err := ssa.ReadObjects(bytes.NewReader(manifestContent))
+
+	if err != nil {
+		return "", err
+	}
+
+	if len(objs) == 0 {
+		return "", fmt.Errorf("no Kubernetes objects found at: %s", manifestPath)
```
opudrovs, line 293: Can be removed, probably. `manifestPath` here is used only for this error message. The generated manifest should be guaranteed to be valid?

cmd/gitops/beta/run/cmd.go
opudrovs: Besides, do we still need to write the file if we are not actually reading the file for applying it to the cluster and are just passing the manifest's content (which is in memory in a var) to `apply`.

Or do we need to read the manifest contents from file/will the file be read later?

cmd/gitops/beta/run/cmd.go
```diff
@@ -50,6 +74,266 @@ func betaRunCommandPreRunE(endpoint *string) func(*cobra.Command, []string) erro
 
 func betaRunCommandRunE(opts *config.Options, client *resty.Client) func(*cobra.Command, []string) error {
 	return func(cmd *cobra.Command, args []string) error {
+		// If there is no cluster in the kube config, return an error.
+		cfg, ctx, err := kube.RestConfig()
```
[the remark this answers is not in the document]
ozamosi: They're different in that InClusterConfig connects to the kube API as if it's running inside the cluster, and outOfClusterConfig will connect as if it's running outside the cluster. This is a legacy of how all code used to work whether you called it from the web API or the CLI.

We expect this to run outside the cluster, so as far as I know InClusterConfig won't do anything, but it might end up returning a different error message in the end? But since the error you return is ErrGetKubeClient, it doesn't matter.

This should work fine :+1:

cmd/gitops/beta/run/cmd.go
```diff
@@ -50,6 +74,266 @@ func betaRunCommandPreRunE(endpoint *string) func(*cobra.Command, []string) erro
...
+
+		listResult := unstructured.UnstructuredList{}
+
+		listResult.SetGroupVersionKind(schema.GroupVersionKind{
+			Group:   "",
+			Version: "v1",
+			Kind:    "Namespace",
+		})
+
+		ctx2 := context.Background()
```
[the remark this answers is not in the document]
chanwit: You can use `context.Background()` because we would not need to set its deadline or cancel it.

cmd/gitops/beta/run/cmd.go
chanwit: hmm. I think this behavour + the locaton of Flux's manifest need refinement.

cmd/gitops/beta/run/cmd.go
ozamosi: Oh god, do I have to make decisions :scream: 

So, I looked at what flags `flux install --help` and `gitops --help` prints out and tried to look at what's important Note: this is written without looking at the code, so if I say something that is not fairly straight-forward copy-paste, then the correct answer is "that's very interesting Robin, I suggest you make a ticket if you feel strongly about it, because that's not part of this ticket's description".

`--version` makes sense - especially as that'd let us set and print a default in the help text, so it's possible to figure out what we'd do. We should have a default set at compiletime in `pkg/version/version.go` already. Maybe it should be called `--flux-version` instead here, though. 

I think maybe `--components-extra` and `--components` would be useful as well, again as a bit of documentation in the help text what the thing will do.

There's already a global `--namespace` in gitops, so we should use that here - especially since I've come to grow opinions about namespaces and gitops recently :blush: 

We need a timeout set to _something_ or install doesn't work, and what I hard-coded isn't a good timeout. Let's just add a `--timeout` flag?

`--insecure-skip-tls-verify` is a global arg in both gitops and flux, is it easy to hook up?

The rest of the global gitops flags are irrelevant.

There's _loads_ of global flux flags which I'm sure are important to some people (personally, I really want `--context`), but they all require changes to the kube config/client bits, so I think they're out of scope for now.

opudrovs:
@ozamosi great points, thank you! 🎉  Will rework the code based on yours and @chanwit 's suggestions.

cmd/gitops/beta/run/cmd.go
```diff
@@ -50,6 +59,56 @@ func betaRunCommandPreRunE(endpoint *string) func(*cobra.Command, []string) erro
...
+		fmt.Println("fluxVersion:", fluxVersion)
+
+		// If Flux is not installed on the cluster then the prerequisites will be installed to initiate the reconciliation process.
+		// This includes all default controllers to set up a reconciliation loop such as the notification-controller, helm-controller, kustomization-controller, and source-controller.
+		// This will also add all relevant CRDs from the controllers above such as Kustomizations, Helm Releases, Git Repository, Helm Repository, Bucket, Alerts, Providers, and Receivers.
+
+		if fluxVersion == "" {
+			filePath := args[0]
+
+			err = run.InstallFlux(log, ctx2, kubeClient, filepath.Join(filePath, run.FluxDirectory), run.ShortManifestFilename)
```
chanwit, line 96: I think we have to discuss the semantics of this flux location.

For example,
When running `gitops beta run ./deploy/overlay/dev`, the `filePath` variable would be `./deploy/overlay/dev`.

Then the `InstallFlux` function would install Flux manifests to `./deploy/overlay/dev/flux-system`, which would not be the correct location.

Per Flux convention, an on-disk location of Flux installation would be something like 
`./cluster/dev/flux-system` - which is normally passed to the `flux bootstrap` command.

So I think we need to refine this behavior.

opudrovs:
@ozamosi @chanwit I've reworked code based on your suggestions.

- Reworked the error/warning messages to use  the logger when appropriate (sometimes a regular formatted error is returned to avoid duplicate error messages). Please let me know if message wording should be corrected or if I missed some places where it is better to use a logger alongside with returning an error.

- for kube config options and kube client options I copied code from Flux and simplified it. Maybe we don't need all the options.

- For now, I've removed everything related to global flags: I think it is better to add that functionality in a separate issue/PR in order not to hold @chanwit 's background part of the sync component PR and to check the basic functionality first. If you think that support for global flags should be added in this PR, please let me know.

- The same about tests for the run command: I haven't added them for now, haven't found a relevant sample test which can reworked fast, so I would need some time to think on how to add the tests because the command is rather complex.

ozamosi (approved):
Your description of what is missing and will be done later matches what I'd want. It's in beta, so it's not the end of the world if the code in main isn't perfect yet.

The code here looks like it'll make a good starting point for the rest of the work, so :shipit:

opudrovs:
@ozamosi great, thanks!

pkg/run/run.go
```diff
@@ -0,0 +1,236 @@
...
+	opts := install.Options{
+		BaseURL:      install.MakeDefaultOptions().BaseURL,
+		Version:      "v0.31.2",
+		Namespace:    "flux-system",
+		Components:   []string{"source-controller", "kustomize-controller", "helm-controller", "notification-controller"},
+		ManifestFile: "flux-system.yaml",
+		Timeout:      5 * time.Second,
+	}
+
+	manifest, err := install.Generate(opts, "")
```
chanwit, line 78: This variable would be `manifests` to be consistent with your error logs

chanwit (approved):
LGTM thank you @opudrovs :1st_place_medal:

pkg/run/run.go
opudrovs: Fixed, good catch!

opudrovs:
@chanwit thank you! 🎉

Issue #2333, opened 2022-06-16
Auth logging too noisy


bigkevmcd:
**Describe the bug**

```
{"level":"info","ts":1655376674.5460489,"logger":"gitops","caller":"middleware/middleware.go:90","msg":"missing or invalid token."}
{"level":"info","ts":1655376674.546245,"logger":"gitops.auth-server","caller":"auth/jwt.go:120","msg":"attempt to read token from cookie"}
{"level":"info","ts":1655376674.5475652,"logger":"gitops","caller":"templates/templates.go:133","msg":"Getting client from context"}
{"level":"info","ts":1655376674.5460567,"logger":"gitops","caller":"middleware/middleware.go:90","msg":"missing or invalid token."}
{"level":"info","ts":1655376674.5511372,"logger":"gitops.auth-server","caller":"auth/jwt.go:120","msg":"attempt to read token from cookie"}
{"level":"info","ts":1655376674.546089,"logger":"gitops","caller":"middleware/middleware.go:90","msg":"missing or invalid token."}
{"level":"info","ts":1655376674.5516546,"logger":"gitops.auth-server","caller":"auth/jwt.go:120","msg":"attempt to read token from cookie"}
{"level":"info","ts":1655376674.5970092,"logger":"gitops","caller":"middleware/middleware.go:90","msg":"missing or invalid token."}
{"level":"info","ts":1655376674.5971458,"logger":"gitops.auth-server","caller":"auth/jwt.go:120","msg":"attempt to read token from cookie"}
{"level":"info","ts":1655376674.5970187,"logger":"gitops","caller":"middleware/middleware.go:90","msg":"missing or invalid token."}
{"level":"info","ts":1655376674.599616,"logger":"gitops.auth-server","caller":"auth/jwt.go:120","msg":"attempt to read token from cookie"}
```
**Environment**

WGE

- Weave-Gitops Version
- Flux Version
- Kubernetes version

**To Reproduce**
Login, and look at the logs

**Expected behavior**
Expected behaviours should not be logged out, it becomes hard to separate important events from non-important events.

We should either log these as `debug` level, or possibly not at all.

These appear to be important messages (because they're in the logs, but they really don't mean much).

**Actual Behavior**

Too much logging.

**Additional Context (screenshots, logs, etc)**

PR #2413, opened 2022-07-07
reduce logging verbosity

makkes:
refs #2333

### DIFF
--- a/pkg/server/middleware/middleware.go
+++ b/pkg/server/middleware/middleware.go
@@ -87,7 +87,7 @@ func WithProviderToken(jwtClient auth.JWTClient, h http.Handler, log logr.Logger
 		tokenSlice := strings.Split(tokenStr, "token ")
 
 		if len(tokenSlice) < 2 {
-			log.Info("missing or invalid token.")
+			log.V(logger.LogLevelDebug).Info("missing or invalid token.")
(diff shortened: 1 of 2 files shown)


AlinaGoaga:
Hey Max! Kevin's off atm I think. Tagging @sarataha and @foot who work more on this part of the code. I deal primarily with the front end :)

pkg/server/middleware/middleware.go
```diff
@@ -87,7 +88,7 @@ func WithProviderToken(jwtClient auth.JWTClient, h http.Handler, log logr.Logger
 		tokenSlice := strings.Split(tokenStr, "token ")
 
 		if len(tokenSlice) < 2 {
-			log.Info("missing or invalid token.")
+			log.V(logger.LogLevelDebug).Info(fmt.Sprintf("missing or invalid token '%s'", tokenStr))
```
foot, line 91: I'm not sure if we should log out actual tokens. They might be slightly badly formatted or something but potentially still valid and a slight security risk.

ozamosi: I think @foot makes a good point. What if instead of the token, we log the value of GitProviderTokenHeader instead - then I can look up the header in my request if I want to.

PR #2417, opened 2022-07-08
CLI command to generate RBAC resources for tenant 

sarataha:
Closes https://github.com/weaveworks/weave-gitops-enterprise/issues/1002

**What changed?**

**Why was this change made?**

**Release notes**

**Documentation Changes**

### DIFF
--- a/cmd/gitops/create/cmd.go
+++ b/cmd/gitops/create/cmd.go
@@ -1,0 +1,20 @@
+package create
+
+import (
+	"github.com/spf13/cobra"
+	"github.com/weaveworks/weave-gitops/cmd/gitops/create/tenants"
+)
+
+func CreateCommand() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "create",
+		Short: "create new resources",
+		Example: `
+# Create a new tenant
+gitops create tenants --file tenants.yaml`,
+	}
+
+	cmd.AddCommand(tenants.TenantsCommand())
+
+	return cmd
+}
--- a/cmd/gitops/create/tenants/cmd.go
+++ b/cmd/gitops/create/tenants/cmd.go
@@ -1,0 +1,67 @@
+package tenants
+
+import (
+	"context"
+	"fmt"
+	"os"
+
+	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
+	"github.com/weaveworks/weave-gitops/pkg/kube"
+	"github.com/weaveworks/weave-gitops/pkg/tenancy"
+)
+
+type tenantCommandFlags struct {
+	filename string
+	export   string
+}
+
+var flags tenantCommandFlags
+
+func TenantsCommand() *cobra.Command {
+	tenantsCmd := &cobra.Command{
+		Use:   "tenants",
+		Short: "create and update tenant resources",
+		RunE:  createTenantsCmdRunE(),
+	}
+
+	tenantsCmd.Flags().String(
+		flags.filename,
+		"",
+		"the file containing the tenant declarations",
+	)
+	tenantsCmd.Flags().String(
+		flags.export,
+		"",
+		"the file to export the generated resources to",
+	)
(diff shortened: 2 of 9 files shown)


cmd/gitops/create/tenants/cmd.go
```diff
@@ -0,0 +1,56 @@
...
+var flags tenantCommandFlags
+
+var TenantsCommand = &cobra.Command{
+	Use:   "tenants",
+	Short: "create and update tenant resources",
+	RunE:  createTenantsCmdRunE(),
+}
+
+func init() {
+	TenantsCommand.PersistentFlags().StringVar(&flags.filename, "filename", "", "the file containing the tenant declarations")
```
jmickey, line 27: I wonder if we should make this `--from-file=<file-path>`, but also support creating a tenant without a tenant file - e.g. `gitops create tenant --name <name> --namespace <namespace>`

cmd/gitops/create/tenants/cmd.go
```diff
@@ -0,0 +1,56 @@
...
+func createTenantsCmdRunE() func(*cobra.Command, []string) error {
+	return func(cmd *cobra.Command, args []string) error {
+		ctx := context.Background()
+
+		kubeClient, err := kube.NewKubeHTTPClient()
+		if err != nil {
+			return fmt.Errorf("failed to create kube client: %w", err)
+		}
+
+		err = tenancy.CreateTenants(ctx, flags.filename, kubeClient)
```
jmickey, line 42: If I am reading this correct - the command will always create the resources within the cluster even if the user supplies the `--export` flag?

cmd/gitops/create/tenants/cmd.go
sarataha: Nice 🙂 I will add this too.

jmickey: I think accepting an array of namespaces would be good!

sarataha: Makes sense to me :)

cmd/gitops/create/cmd.go
```diff
@@ -0,0 +1,20 @@
...
+	"github.com/weaveworks/weave-gitops/cmd/gitops/create/tenants"
+)
+
+func CreateCommand() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "create",
+		Short: "create new resources",
+		Example: `
+# Create a new tenant
+gitops create tenants --file tenants.yaml`,
```
foot, line 14:
```suggestion
gitops create tenants --from-file tenants.yaml`,
```

pkg/tenancy/testdata/example.yaml.golden
```diff
@@ -0,0 +1,117 @@
+apiVersion: v1
```
foot, line 1: .golden? ✨

cmd/gitops/create/tenants/cmd.go
```diff
@@ -0,0 +1,87 @@
...
+
+var TenantsCommand = &cobra.Command{
+	Use:   "tenants",
+	Short: "create and update tenant resources",
+	RunE:  createTenantsCmdRunE(),
+}
+
+func init() {
+	TenantsCommand.PersistentFlags().StringVar(&flags.name, "name", "", "the file containing the tenant declarations")
+	TenantsCommand.PersistentFlags().StringSliceVar(&flags.namespaces, "namespace", []string{}, "the file containing the tenant declarations")
```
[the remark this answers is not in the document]
sarataha: Right. Thank you for pointing this out 🙂

pkg/tenancy/testdata/example.yaml.golden
bigkevmcd: That's it exactly.

This reduces the ongoing cost as we generate more resources...we don't care about the contents as much as we care that they are being generated...

We didn't go all-in https://speakerdeck.com/mitchellh/advanced-testing-with-go?slide=19 but this is a start on this.

PR #2418, opened 2022-07-08
Fix chart location in OCI publish step

ozamosi:
The chart repository is called `charts/weave-gitops`, not `charts`.

This should fix the error from https://github.com/weaveworks/weave-gitops/runs/7255134483?check_suite_focus=true

### DIFF
--- a/.github/workflows/chart.yaml
+++ b/.github/workflows/chart.yaml
@@ -14,7 +14,7 @@ on:
 
 env:
   REGISTRY: ghcr.io
-  CHART_LOCATION: weaveworks/charts
+  CHART_LOCATION: weaveworks/charts/weave-gitops
 
 permissions:
   contents: write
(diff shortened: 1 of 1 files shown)


chanwit:
Hmm I think the old setting is already correct. The chart location does not include the chart name, so that it's only `/charts`.
`weave-gitops` which is the chart name will be appended automatically by the push command.

Please try `helm push` locally to very this behavior.

The 403 error in the run would be something related to the write permission of the token maybe (I'm totally speculating anyway).

ozamosi:
Problem was that the weave-gitops user didn't have any access in the [package settiings](https://github.com/orgs/weaveworks/packages/container/charts%2Fweave-gitops/settings)

I've added write permissions so it ought to work - of course, when I tested it and I got it working I used the gitops bot token so I don't know for sure if main still works. It should work, though, so closing.

PR #2424, opened 2022-07-11
Set version flags when building release images, too

### LINKED ISSUE #2370: version endpoint returns 0.0.0
**Describe the bug**

Trying to get the version information from http://localhost:9001/v1/version

But the response is:

```json
{
"semver": "v0.0.0",
"commit": "",
"branch": "",
"buildTime": ""
}
```

Has this intentionally been changed? 

**Environment**

- Weave-Gitops Version (Chart 2.0.11, Image 0.9.0)
- Flux Version 0.31.0
- Kubernetes version 1.21.1

**Workaround**
- dashboard displays version information in the footer
- can get chart version from sources view and image version from looking at the images in the ww-gitops deployment

ozamosi:
When hitting `/v1/version` against a 0.9.0 image, it returns
`{"semver":"v0.0.0","commit":"","branch":"","buildTime":""}`. This is
incorrect.

This copy-pastes bits from the dev image build process that we use to
set these properly for the staging cluster.

The footer still prints the correct thing since the javascript does
have the right versions.

This fixes #2370

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -113,6 +113,10 @@ jobs:
           registry: ${{ env.REGISTRY }}
           username: ${{ github.actor }}
           password: ${{ secrets.GITHUB_TOKEN }}
+      - name: Set build-time flags
+        run: |
+          echo "LDFLAGS=$(make echo-ldflags)" >> $GITHUB_ENV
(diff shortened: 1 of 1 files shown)


opudrovs (approved):
LGTM! 🎉 Great that you've confirmed and fixed that issue.

PR #2434, opened 2022-07-13
Add flags to gitops run

### LINKED ISSUE #2426: Add flags to `gitops run`
We need to add more flags to `gitops run` to make it more useful

* --flux-version, to set the version of manifests we install.

* --components & --components-extra to specify the flux components

* Use global --namespace

* A --timeout to override the default flux timeout the same way as flux's --timeout

~~* Hook up --insecure-skip-tls-verify from gitops global flag to the flux flags~~
Based on @chanwit 's research: "it's been deprecated for a quite sometime and removed in 1.24" So, the `gitops run` command would not use this flag. But it is still available in the global flags of the `gitops` command and the decision if it should be removed from other commands requires more research.

* There's a bunch like --kubeconfig, --context ~~and --cluster~~ which would help if you have multiple contexts - see if they can be supported.

opudrovs:
Closes #2426 

- Added the following command-specific flags to the `run` command: `--flux-version`, `--allow-k8s-context`, `--components`, `--components-extra`, `--timeout`.
- Added using two global flags, `--namespace` and `--kubeconfig` by the `run` command.
- Hooked up the `--context` flag, auto-generated by `genericclioptions`, when creating `kubeConfigArgs`.
- Reworked creation of kube config args and kube client.
- Added passing the kube config to the `InstallBucketServer` and down the line.
- Renamed `kubeConfigOptions` to `kubeConfigArgs` back, for consistency with their purpose and  the Flux code.

Notes:
- If a `--kubeconfig` flag is provided a `--context` flag is required too, because in this case we are not able to call `RestConfig()` initially to get the current context name in this case. Added an error message if the `--context` flag is not provided together with the `--kubeconfig` flag.
- Have not added the `--cluster` and `--allow-k8s-cluster` flags yet. We need to decide first how to use them.
- I think the default values displayed in the command help for the `--components` and `--components-extra` flags (empty, fortunately) of type strings slice are somewhat misleading:

```
--components strings         The Flux components to install, as a comma-separated list: --components=component1,component2,component3 (default [source-controller,kustomize-controller,helm-controller,notification-controller])
--components-extra strings   Additional Flux components to install, as a comma-separated list: --components=component1,component2,component3
```

It looks like the default value should be passed with square brackets while it should be just a comma-delimited list. Thus, I added and example in the usage string. If the user passes in

`components=[component1,component2,component3]`
instead of
`components=component1,component2,component3`

the components will be auto-split like `[component1`, `component2`, and `component3]`.

I hope it's a well-known argument format and it will be easy to understand how to format the flag values.

Issues:
- When using the flag `--allow-k8s-context` with a context name with underscores (for example, `--allow-k8s-context=gke_denim-testing_us-central1-c_olga-testing-1`) without also providing `--context=gke_denim-testing_us-central1-c_olga-testing-1` there is an issue with the context name changed when sanitized.

When this remote cluster is selected, we get context name from the `kube.RestConfig()` call. And it calls the following function under the hood to sanitize the context name:

```
func sanitizeClusterName(s string) string {
	// remove leading email address or username prefix from context
	if strings.Contains(s, "@") {
		return s[strings.LastIndex(s, "@")+1:]
	}

	s = strings.ReplaceAll(s, "_", "-")

	return s
}
```

So, after all underscores get replaces with dashes, the context name received from the rest config becomes `gke-denim-testing-us-central1-c-olga-testing-1` and the following check does not pass:

```
contextName = kubeClient.ClusterName
if flags.AllowK8sContext == contextName {
	log.Actionf("Explicitly allow GitOps Run on %s context", contextName)
} else if !isLocalCluster(kubeClient) {
	return errors.New("allowed to run against a local cluster only")
}
```

Temp log values:

```
flags.AllowK8sContext:
gke_denim-testing_us-central1-c_olga-testing-1

contextName:
gke-denim-testing-us-central1-c-olga-testing-1

kubeClient.ClusterName:
gke-denim-testing-us-central1-c-olga-testing-1
```

### DIFF
--- a/cmd/gitops/beta/run/cmd.go
+++ b/cmd/gitops/beta/run/cmd.go
@@ -15,6 +15,7 @@ import (
 	"github.com/weaveworks/weave-gitops/cmd/internal"
 	"github.com/weaveworks/weave-gitops/pkg/kube"
 	"github.com/weaveworks/weave-gitops/pkg/run"
+	"github.com/weaveworks/weave-gitops/pkg/version"
 	"k8s.io/cli-runtime/pkg/genericclioptions"
 )
 
@@ -55,7 +56,7 @@ gitops beta run ./deploy/overlays/dev [flags]`,
 
 	cmdFlags := cmd.Flags()
 
-	cmdFlags.StringVar(&flags.FluxVersion, "flux-version", "v0.31.3", "The version of Flux to install")
+	cmdFlags.StringVar(&flags.FluxVersion, "flux-version", version.FluxVersion, "The version of Flux to install")
 	cmdFlags.StringVar(&flags.AllowK8sContext, "allow-k8s-context", "", "The name of the kubeconfig context to allow explicitly")
 	cmdFlags.StringSliceVar(&flags.Components, "components", []string{"source-controller", "kustomize-controller", "helm-controller", "notification-controller"}, "The Flux components to install, as a comma-separated list: --components=component1,component2,component3")
 	cmdFlags.StringSliceVar(&flags.ComponentsExtra, "components-extra", []string{}, "Additional Flux components to install, as a comma-separated list: --components=component1,component2,component3")
--- a/cmd/gitops/beta/run/cmd.go
+++ b/cmd/gitops/beta/run/cmd.go
@@ -152,11 +152,6 @@ func betaRunCommandRunE(opts *config.Options) func(*cobra.Command, []string) err
 		kubeClientOpts := run.GetKubeClientOptions()
 		kubeClientOpts.BindFlags(cmd.Flags())
 
-		_, err = kubeConfigArgs.ToRESTConfig()
-		if err != nil {
-			return fmt.Errorf("error converting kube config args to a restconfig: %w", err)
-		}
-
 		var contextName string
 		if flags.Context != "" {
 			contextName = flags.Context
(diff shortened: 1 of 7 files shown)


cmd/gitops/beta/run/cmd.go
```diff
@@ -33,6 +53,18 @@ gitops beta run ./deploy/overlays/dev [flags]`,
 		DisableAutoGenTag: true,
 	}
 
+	cmdFlags := cmd.Flags()
+
+	cmdFlags.StringVar(&flags.FluxVersion, "flux-version", "v0.31.3", "The version of Flux to install")
```
ozamosi, line 58: Could we pull this in from the binary itself? I think we have a good number in `pkg/version` we're not really using at the moment - then we just need to bump the version in `FLUX_VERSION` in the `Makefile` before releasing.

ozamosi:
I think most of your notes can be summed up as "the helpers provided by the pkg/kube module aren't doing any helping" - it'd be worth taking a critical look at that module, but not in this PR.

> I think the default values displayed in the command help for the --components and --components-extra flags (empty, fortunately) of type strings slice are somewhat misleading

Definitely misleading - it _kind of works_ because it's very common for command line help text to contain square brackets that you don't type, e.g.

```
Usage:
  gitops beta run [flags]
```
doesn't mean to type `[flags]`.

Your example helps, though.

ozamosi (approved):

cmd/gitops/beta/run/cmd.go
opudrovs: Added! But now we need to keep in mind that if we are testing from the source code (like `go run main.go beta run ./`) without running `make gitops` first, Flux will be installed and when running the command again there a message `Flux version "undefined" is found` will be displayed. When running the command from `bin/gitops`, the version is displayed properly.

ozamosi: Maybe we can have it default to the string "latest", and github will take care of redirecting us to the most recent release tag automatically :thinking:

opudrovs: OK, great! Should we make it default to "latest" only in the command source (with first checking if it's undefined) or should just replace "undefined" with "latest" in `pkg/version`?

ozamosi: Just replace "undefined", I think.

opudrovs:
@ozamosi thanks, yes, it is definitely out of scope for this PR. And good example with `[flags]`.

PR #2440, opened 2022-07-14
Allow adding new allowed kinds to some of the generic endpoints

ozamosi:
We have a couple of categories of endpoints: ones that are specific to one kind (like `ListKustomizations`), ones that are generic across "all" kinds (like `GetChildObjects`), and ones that are somewhat generic, in that they could be made generic, but they have been made to only work for "primary" event kinds. This focuses on the last kind.

I am slightly nervous about just allowing access to all kinds (e.g. secrets) without any kind of white/black list, but I also want these endpoints to be reusable outside of core. So my half-in-half-out solution is to make it possible for a downstream consumer to configure the kinds that count as "primary" - so by calling `cfg.PrimaryKinds.Add` at start-up, you can now add canaries or clusters or whatever you want to the list of primary kinds.

I don't foresee this being a long term solution, but I'm hoping it'll work around some current pain.


opudrovs (approved):

opudrovs (commented):
Just noticed the UI checks failing.

ui/hooks/events.ts
```diff
@@ -1,18 +1,18 @@
...
   const { api } = useContext(CoreClientContext);
 
-  return useQuery<ListFluxEventsResponse, RequestError>(
+  return useQuery<ListEventsResponse, RequestError>(
     ["events", obj],
     () =>
-      api.ListFluxEvents({
-        involvedObject: obj,
+      api.ListEvents({
+        involvedObject: {...obj, kind: removeKind(obj.kind),
```
opudrovs, line 15: `}` missing.

opudrovs: Haha, sounds familiar! 😅

PR #2459, opened 2022-07-19
docs: Add progressive delivery docs

yiannistri:
**What changed?**
Added documentation for setting up progressive delivery with Flagger/Linkerd.

**Why was this change made?**
To help WGE users setup Flagger.

N/A

N/A

**Release notes**
This is for an upcoming version of WGE.

**Documentation Changes**
N/A


website/docs/guides/delivery.mdx
```diff
@@ -0,0 +1,473 @@
...
+title: Progressive Delivery using Flagger
+hide_title: true
+sidebar_position: 6
+---
+
+import TierLabel from "../_components/TierLabel";
+
+# Progressive Delivery using Flagger <TierLabel tiers="enterprise" />
+
+Weave GitOps Enterprise includes the ability to view [Flagger](https://docs.flagger.app/) [canary](https://docs.flagger.app/usage/how-it-works#canary-resource) objects information in the UI through the Applications > Delivery view.
```
sympatheticmoose, line 11: The two hyperlinks next to each other may make this a little confusing, and we could probably sell this a little more.

 Suggested rework:
"[Flagger](https://docs.flagger.app/) is a progressive delivery operator for Kubernetes. It is designed to reduce risks when introducing new software versions and to improve time to delivery through automating production releases. Weave GitOps Enterprise's UI allows you to view the state of these progressive delivery rollouts, and how they are configured using Flagger's [canary](https://docs.flagger.app/usage/how-it-works#canary-resource) object, through the Applications > Delivery view."

website/docs/guides/delivery.mdx
```diff
@@ -0,0 +1,473 @@
...
+
+This guide uses Flux manifests to install Flagger and Linkerd. Using Flux allows us to manage our cluster applications in a declarative way through changes in a Git repository.
+
+## Installing Flagger using Flux
+
+For the Flagger installation, a Kustomization file will be used.
```
sympatheticmoose, line 17: > A Kustomization file will be used

Why? Can we explain this to the reader. 

I think we will also need to show a sample repository structure or at least talk about an assumed path which is already being reconciled with Flux.

website/docs/guides/delivery.mdx
```diff
@@ -0,0 +1,473 @@
...
+
+First, add a Namespace resource for Flagger:
+
+```yaml title="flagger/namespace.yaml"
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: flagger
```
sympatheticmoose, line 25: The Flagger docs talk about installing Flagger in the Linkerd namespace - _maybe_ worth being consistent with this approach.

website/docs/guides/delivery.mdx
```diff
@@ -0,0 +1,473 @@
...
+sidebar_position: 6
+---
+
+import TierLabel from "../_components/TierLabel";
+
+# Progressive Delivery using Flagger <TierLabel tiers="enterprise" />
+
+Weave GitOps Enterprise includes the ability to view [Flagger](https://docs.flagger.app/) [canary](https://docs.flagger.app/usage/how-it-works#canary-resource) objects information in the UI through the Applications > Delivery view.
+
+This guide uses Flux manifests to install Flagger and Linkerd. Using Flux allows us to manage our cluster applications in a declarative way through changes in a Git repository.
```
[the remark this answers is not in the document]
yiannistri: - No reason, Linkerd may be installed before Flagger
- I guess it could be skipped, the most important thing is to get Prometheus to scrape the right metrics otherwise Flagger won't be able to complete the analysis. This example uses the built-in Prometheus instance that comes with Linkerd but I don't expect many people to use this in production, they will likely have other Prometheus instances already running. Generally speaking Flagger only needs the prometheus URL.

website/docs/guides/delivery.mdx
yiannistri: In the case of Linkerd, the use of a `secretGenerator` makes it very easy to read files (certificates) and turn them into secrets.

website/docs/guides/delivery.mdx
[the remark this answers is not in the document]
yiannistri: I decided to go for a detailed guide because of the nature of Flagger: the webhooks section will need to refer to services within the cluster so having the right names/namespaces is necessary for Flagger to work correctly.

website/docs/guides/delivery.mdx
yiannistri: I went for a separate namespace to make the guide more flexible/less opinionated about where things go.

website/docs/guides/delivery.mdx
yiannistri: Updated 👍

website/docs/guides/delivery.mdx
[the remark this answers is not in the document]
sympatheticmoose: > I decided to go for a detailed guide because of the nature of Flagger: the webhooks section will need to refer to services within the cluster and having the right names/namespaces is necessary for Flagger to work correctly.

Agree with your approach 100%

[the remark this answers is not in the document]
sympatheticmoose: what I would suggest is (feel free to reword as you see fit):

- expand the outline - "This guide uses Flux manifests to install Flagger and Linkerd." -> "Flagger can work with a number of service meshes and ingress controllers, to support various progressive delivery [deployment strategies](https://docs.flagger.app/usage/deployment-strategies)."

keep your bit about Using Flux..., then add:

In this guide we will walk you through a full end-to-end scenario where you will:
  - Install the Linkerd service mesh (href)
  - Install Flagger (href)
  - Deploy a sample [application]((https://github.com/stefanprodan/podinfo)) using a [canary](https://docs.flagger.app/usage/deployment-strategies#canary-release) release strategy based on metrics provided through Linkerd's in-built Prometheus instance(href)
  
Then, add a pre-requisites section which states:
- This guide assumes you already have a Kubernetes cluster running and have bootstrapped Flux. To apply the manifests listed in this guide, you will need to commit them to a repository being reconciled with Flux. For help installing Flux, you can follow their [getting started](https://fluxcd.io/docs/get-started/) documentation.


---
These changes means you could remove (at least) "In this case, Flagger is configured to work with the Linkerd service mesh. [Other providers](https://docs.flagger.app/install/flagger-install-on-kubernetes) may also be used instead of Linkerd to integrate with Flagger."

website/docs/guides/delivery.mdx
sympatheticmoose: there's an inconsistency with the following resources, where they add "add the following manifest to the cluster repository". I'd either remove them and hope that its covered by my suggestions in the pre-req section, or make sure its added to all files

website/docs/guides/delivery.mdx
```diff
@@ -0,0 +1,473 @@
...
+- releases.yaml
+```
+
+## Installing Linkerd using Flux
+
+For the Linkerd installation, a Kustomization file will be used.
+
+In order to support mTLS connections between meshed pods, Linkerd requires a trust anchor certificate and an issuer certificate with its corresponding key. These certificates are automatically created when the `linkerd install` command is used but when using a Helm chart to install Linkerd, these certificates need to be provided.
+
+Follow [these instructions](https://linkerd.io/2.11/tasks/generate-certificates/#generating-the-certificates-with-step) to generate the certificates using the `step` CLI.
```
sympatheticmoose, line 100: I may be wrong... but i think you're summarising the steps below, so the reader doesnt have to go to this link right?

website/docs/guides/delivery.mdx
```diff
@@ -0,0 +1,473 @@
...
+      targetPath: identity.issuer.tls.crtPEM
+    - kind: Secret
+      name: linkerd-certs
+      valuesKey: issuer.key
+      targetPath: identity.issuer.tls.keyPEM
+  values:
+    installNamespace: false
+    identity:
+      issuer:
+        crtExpiry: "2023-07-18T20:00:00Z"
```
sympatheticmoose, line 177: Maybe add a comment along the lines of #change this value ?

website/docs/guides/delivery.mdx
```diff
@@ -0,0 +1,473 @@
...
+```
+
+To generate the issuer certificate run:
+```bash
+step certificate create identity.linkerd.cluster.local issuer.crt issuer.key \
+--profile intermediate-ca --not-after 8760h --no-password --insecure \
+--ca ca.crt --ca-key ca.key
+```
+
+Add the `ca.crt`, `issuer.crt` and `issuer.key` files to the cluster repository under a `linkerd` directory.
```
sympatheticmoose, line 115: maybe mirror this approach for Flagger and suggest the creation of a Flagger directory? Then after each stage we could have a pseudo repo structure shown like

/clusters/my-cluster
|-linkerd
  |- files
|- flagger
  |- files
|- flux-system

wdyt?

website/docs/guides/delivery.mdx
```diff
@@ -0,0 +1,473 @@
...
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+namespace: test
+resources:
+- namespace.yaml
+- deployment.yaml
+- canary.yaml
+```
+
+After a short time, the status of the canary object should be set to `Initialized`:
```
sympatheticmoose, line 445: stick some screenshots in here as alternatives as well. I can send you some if need

website/docs/guides/delivery.mdx
```diff
@@ -0,0 +1,473 @@
...
+podinfo   Progressing   35       2022-07-19T12:25:04
+```
+
+After a short time the rollout is completed and the status of the canary object is set to `Succeeded`:
+
+```bash
+> kubectl -n test get canary podinfo
+NAME      STATUS      WEIGHT   LASTTRANSITIONTIME
+podinfo   Succeeded   0        2022-07-19T12:28:04Z
+```
```
sympatheticmoose, line 473: Add something to end it maybe:

##Summary
Congratulations, you have now completed a progressive delivery rollout with Flagger and Linkerd :tada: 

Next steps:
- Explore more of what [Flagger](https://flagger.app) can offer
- Configure manual approving for progressive deliver deployments (link to guide)

website/docs/guides/delivery.mdx
```diff
@@ -0,0 +1,473 @@
...
+├── issuer.crt
+├── issuer.key
+├── kustomization.yaml
+├── kustomizeconfig.yaml
+├── namespace.yaml
+├── releases.yaml
+└── source.yaml
+```
+
+Once Flux reconciles this directory to the cluster, Linkerd should get installed.
```
sympatheticmoose, line 246: We should add guidance on how to check both Flagger and Linkerd have been installed before proceeding to deploy the canary

sympatheticmoose:
Can confirm have run through and everything worked :tada:

only additional issues (as discussed in slack):
- k8s cluster needs the `autoscaling/v2` API
- bootstrapping with everything defined in your cluster repo would fail as is, need the flagger CRDs installed prior to the Canary object being reconciled.

website/docs/guides/delivery.mdx
yiannistri: That's a good point, I'll remove these.

website/docs/guides/delivery.mdx
yiannistri: No, I think that would help the reader confirm that they have all the right files in place.

website/docs/guides/delivery.mdx
yiannistri: Yes ideally this needs to be set to the expiry date of the generated certificate. I'll add a comment.

website/docs/guides/delivery.mdx
yiannistri: Update 👍

yiannistri:
Thanks for the valuable feedback 🙏 @sympatheticmoose 

Can you take another look?

sympatheticmoose (approved):

sympatheticmoose:
@yiannistri I think this is looking great - well done and thank you for taking on the task. I've added some more explanation around what will happen with the podinfo deployment, which hopefully isn't too contentious!

PR #2469, opened 2022-07-20
Install GitOps Dashboard for gitops beta run

### LINKED ISSUE #2358: Install GitOps Dashboard for gitops beta run
Offer to install gitops dashboard using an interactive prompt if it's not installed already.

If the user answers yes, let them enter a password and use that for the installation.

opudrovs:
Closes #2358

- Moved functions in `pkg/run/run.go` to separate files, because the existing file was getting huge.
- Added checking if the GitOps Dashboard is installed, and if not, offering to install it.
- Added prompting user for their password (if the user agrees to install the dashboard) and generating the secret which will be used for logging into the dashboard. The prompt's behavior is similar to a prompt in Flux and everything which is not an explicit Yes is a no. No looping of user input was added, tried it, it was annoying.
- Added waiting for the dashboard reconciliation and deployment if the dashboard is installed.
- Added the `--dashboard-port` flag with default value 9001 and opening this port automatically for the dashboard.
- Refactored some messages and added comments.

Notes:
- The code is working in general but maybe i missed the wording in some log/error messages because there are so many of them. So, any comments on naming and wording are very welcome.
- I've copied and reworked @chanwit's code for opening ports and waiting for dashboard reconciliation and deployment. So, I hope I've used it correctly and in the right places, if not, please let me know.


chanwit (approved):
LGTM in general
thank you, @opudrovs :1st_place_medal:

pkg/run/install_dashboard.go
```diff
@@ -0,0 +1,293 @@
...
+
+	log.Successf("GitOps Dashboard has been installed")
+
+	fmt.Println(applyOutput)
+
+	return nil
+}
+
+// IsDashboardInstalled checks if the GitOps Dashboard is installed.
+func IsDashboardInstalled(log logger.Logger, ctx context.Context, kubeClient *kube.KubeHTTP) bool {
```
chanwit, line 88: A small glitch here is that someone might install Weave GitOps with a different method.

But we can fix it later when needed.

opudrovs: Would another method also require them to use the same HelmChart? Is it possible to install the dashboard without this HelmChart? Or will the chart be the same just its name will be different?

I mean if using a HelmChart is required for installing the dashboard can we just use something else (not the name) for detecting that this HelmChart is present reliably?

opudrovs:
@chanwit great, thank you! 🌞

PR #2470, opened 2022-07-20
Unify oci link handling

ozamosi:
This stops `oci://` links from showing up anywhere in the UI.

This does not use `isHTTP`, because we actually use `<Link/>` for `mailto:`

### DIFF
--- a/ui/components/Link.tsx
+++ b/ui/components/Link.tsx
@@ -1,6 +1,7 @@
 import * as React from "react";
 import { Link as RouterLink } from "react-router-dom";
 import styled from "styled-components";
+import { isAllowedLink } from "../lib/utils";
 import Text, { TextProps } from "./Text";
 
 type Props = {
@@ -24,6 +25,10 @@ function Link({
   textProps,
   ...props
 }: Props) {
+  if (href && !isAllowedLink(href)) {
+    return <Text {...textProps}>{children}</Text>;
+  }
+
   const txt = (
     <Text color="primary" {...textProps}>
       {children}
--- a/ui/components/SourcesTable.tsx
+++ b/ui/components/SourcesTable.tsx
@@ -5,7 +5,6 @@ import {
   FluxObjectKind,
   GitRepository,
   HelmRepository,
-  HelmRepositoryType,
 } from "../lib/api/core/types.pb";
 import { formatURL, objectTypeToRoute } from "../lib/nav";
 import { showInterval } from "../lib/time";
@@ -109,12 +108,8 @@ function SourcesTable({ className, sources }: Props) {
                 return "-";
               case FluxObjectKind.KindHelmRepository:
                 text = (s as HelmRepository).url;
-                if (
-                  (s as HelmRepository).repositoryType != HelmRepositoryType.OCI
-                ) {
-                  url = text;
-                  link = true;
-                }
+                url = text;
+                link = true;
                 break;
             }
             return link ? (
--- a/ui/components/__tests__/Link.test.tsx
+++ b/ui/components/__tests__/Link.test.tsx
@@ -1,0 +1,73 @@
+import React from "react";
+import { MemoryRouter } from "react-router-dom";
+import { render } from "@testing-library/react";
+import { withTheme } from "../../lib/test-utils";
+import Link from "../Link";
+
+describe("Link", () => {
+  it("doesn't create a link for oci links", () => {
+    const { container: link } = render(
+      withTheme(<Link href="oci://ghcr.io/some/chart">Text</Link>)
+    );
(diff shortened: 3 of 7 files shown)


ui/components/Link.tsx
```diff
@@ -3,6 +3,14 @@ import { Link as RouterLink } from "react-router-dom";
...
 import Text, { TextProps } from "./Text";
 
+function isAllowed(uri: string): boolean {
+  // Regex from https://github.com/cure53/DOMPurify/blob/cce00ac40d33c2aae6422eaa59e6a8aad5c73901/src/regexp.js
+  const regex = new RegExp(
+    /^(?:(?:(?:f|ht)tps?|mailto|tel|callto|cid|xmpp):|[^a-z]|[a-z+.\-]+(?:[^a-z+.\-:]|$))/i // eslint-disable-line no-useless-escape
+  );
+  return regex.test(uri);
+}
+
```
joshri, line 13: Is it possible/worth it to replace `isHTTP` from lib/utils line 30 with this? We're definitely using it in Metadata to see if we should render a link or text, but we don't need to do that anymore as this component now returns text if the link is invalid. It'd be nice to not have to run what's basically the same function twice - it's not a big deal since it's literally one regex but yeah.

ui/components/Link.tsx
```diff
@@ -24,6 +32,10 @@ function Link({
   textProps,
   ...props
 }: Props) {
+  if (href && !isAllowed(href)) {
+    return <Text {...textProps}>{children}</Text>;
+  }
+
```
joshri, line 38: I think it'd be useful to do the thing here where you also pass the textProps down - that way we can still style this Text component the way we want. Just spread the textProps thing this component already has like line 40. It was a quick fix I did a long time ago for something and forgot about but here it is! There might be a better way.

ui/components/SourcesTable.tsx
```diff
@@ -109,12 +108,8 @@ function SourcesTable({ className, sources }: Props) {
...
               case FluxObjectKind.KindHelmRepository:
                 text = (s as HelmRepository).url;
-                if (
-                  (s as HelmRepository).repositoryType != HelmRepositoryType.OCI
-                ) {
-                  url = text;
-                  link = true;
-                }
+                url = text;
+                link = true;
```
joshri, line 112: Just want to make sure that SourcesTable is the only one out of AutomationsTable, EventsTable, ReconciledObjectsTable, heck even FluxRuntime where link changes need to be made

ui/components/__tests__/Link.test.tsx
```diff
@@ -0,0 +1,73 @@
+import React from "react";
+import { MemoryRouter } from "react-router-dom";
+import { render } from "@testing-library/react";
+import { withTheme } from "../../lib/test-utils";
+import Link from "../Link";
+
+describe("Link", () => {
+  it("doesn't create a link for oci links", () => {
+    const { container: link } = render(
```
joshri, line 9: I have no idea how this works but I'm not mad at it. Might be worth adding isAllowed to the isHTTP tests in utils, or deleting the isHTTP tests altogether if we don't need to use it anymore

joshri (approved):
It's fire. Just minor comments about refactoring the isHTTP function that exists already.

ui/components/Link.tsx
```diff
@@ -3,6 +3,14 @@ import { Link as RouterLink } from "react-router-dom";
 import styled from "styled-components";
 import Text, { TextProps } from "./Text";
 
+function isAllowed(uri: string): boolean {
+  // Regex from https://github.com/cure53/DOMPurify/blob/cce00ac40d33c2aae6422eaa59e6a8aad5c73901/src/regexp.js
+  const regex = new RegExp(
+    /^(?:(?:(?:f|ht)tps?|mailto|tel|callto|cid|xmpp):|[^a-z]|[a-z+.\-]+(?:[^a-z+.\-:]|$))/i // eslint-disable-line no-useless-escape
+  );
+  return regex.test(uri);
+}
```
foot, line 12: Could we export this and add to index.ts?

ui/components/Link.tsx
ozamosi: This regex allows relative links, so if it was the same function then `/foo` would become a link, which I don't _think_ we want.

ui/components/Link.tsx
ozamosi: I shouldn't need you to prompt to do this, should I.

ui/components/Link.tsx
ozamosi: Moved it to `utils` and wrote some comments to explain what the difference is.

PR #2472, opened 2022-07-21
display error messages when dev-ks fails

### LINKED ISSUE #2462: GitOps Run - display error message when the reconciliation is failed
As a user, I want it clear that the reconciliation loop failed and what the issue is.  By reconciliation loop we are referring to the "dev kustomization" that we added for the live syncing.  

Users can see issues with other kustomizations via the GitOps Dashboard.

### Acceptance Criteria

1. When there is an issue with the "dev kustomization" I see that error in the logs of gitops run.
2. When there are issues with other workloads (i.e. Kustomizations and Helm Releases).  I don't see those printed out in the logs.
3. We print out some useful error messages when something isn't working

chanwit:
Fixes #2462

### DIFF
--- a/pkg/run/install_bucket_source_and_ks.go
+++ b/pkg/run/install_bucket_source_and_ks.go
@@ -2,20 +2,25 @@ package run
 
 import (
 	"context"
+	"fmt"
+	"strings"
 	"time"
 
 	kustomizev1 "github.com/fluxcd/kustomize-controller/api/v1beta2"
 	"github.com/fluxcd/pkg/apis/meta"
 	sourcev1 "github.com/fluxcd/source-controller/api/v1beta2"
-	"github.com/weaveworks/weave-gitops/pkg/kube"
 	"github.com/weaveworks/weave-gitops/pkg/logger"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"sigs.k8s.io/cli-utils/pkg/object"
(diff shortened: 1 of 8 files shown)


pkg/run/install_bucket_source_and_ks.go
```diff
@@ -164,3 +172,66 @@ func CleanupBucketSourceAndKS(log logger.Logger, kubeClient *kube.KubeHTTP, name
 
 	return nil
 }
+
+// FindConditionMessages finds the messages in the condition of objects in the inventory.
+func FindConditionMessages(kubeClient *kube.KubeHTTP, ks *kustomizev1.Kustomization) ([]string, error) {
```
ozamosi, line 177: I think it would be easier to understand this function with some tests serving as examples - could we add some?

ozamosi (approved):
I tried running it in a way that breaks, and the output looked like:

```
► Request reconciliation of dev-bucket, and dev-ks (timeout 30s) ... 
✗ Kustomization flux-system/flux-system: Source 'GitRepository/flux-system' not found
✗ HelmRelease flux-system/ww-gitops: install retries exhausted
✗ HelmRelease flux-system/ww-gitops: Helm install failed: rendered manifests contain a resource that already exists. Unable to continue with install: Secret "cluster-user-auth" in namespace "flux-system" exists and cannot be imported into the current release: invalid ownership metadata; label validation error: key "app.kubernetes.io/managed-by" must equal "Helm": current value is "tilt"; annotation validation error: missing key "meta.helm.sh/release-name": must be set to "ww-gitops"; annotation validation error: missing key "meta.helm.sh/release-namespace": must be set to "flux-system"

Last Helm logs:


✗ GitRepository flux-system/flux-system: failed to get secret 'flux-system/flux-system': Secret "flux-system" not found
✗ Error requesting reconciliation: timed out waiting for the condition
✔ Reconciliation is done.
```

So it's definitely discovering the sub-object's logs and showing them. There's some odd whitespace around `Last Helm Logs` but I don't think that comes from us?

pkg/run/install_bucket_source_and_ks.go
chanwit: sure.

PR #2476, opened 2022-07-21
Remove DagreD3 to make a solid foundation for the graph

### LINKED ISSUE #2437: SPIKE: Graph investigation
HERE WE GO AGAIN. 

The goal is to keep our graph the same while removing DagreD3. Why? DagreD3 did a lot of the graph making for us, which was good, but also leaves us without the ability to customize the graph as we see fit (link shape, node spacing, overall graph shape, etc. These elements are not exposed in our code...we just say createNode()). 

We need to get our data into a proper tree structure for d3.hierarchy(), create the nodes and links ourselves, and then get our styling, scaling, zoom, and drag back to where it is today. 

As of July 14 - Robin and I have already got the tree structure, nodes, and links made. I'm still working on the styling. The graph code is now based on this example from the d3 repo on github: https://observablehq.com/@d3/d3-hierarchy

joshri:
Closes #2437

The main goal here was to remove the DagreD3 library and create the graph nodes and paths ourselves. This gives us a lot more flexibility with the graph code. Not only should it be a lot easier to change, but I think it's easier to understand as well (maybe that's just because I wrote it). I also separated out the actual JSX for the nodes for readability - no longer will you have to scroll through 200 lines of node styling to get to the graph code. I'm starting VERY simple here (although I'm definitely close to what we have right now), and the graph is now ready for any new features or redesigns we want to throw at it!

I also made a small change in the `computeReady` function to distinguish between `ReadyType.NotReady` and `undefined` and followed up accordingly. 

image

Some particulars:
1. Online videos recommended NOT using the `useRef` hook and instead separating the data formation (`ReconciliationGraph`) and the rendering (`DirectedGraph`). We are now mapping out two arrays into `g` (group) elements - one of nodes (into `GraphNodes`), and one of links (into `paths`). We had to do a lot of code to get the `foreignObject` tag into DagreD3, and now we just throw it into JSX! This also allowed me to move the node html to it's own component, which I'm really happy about.

2. No more graph class - this is the only d3 code for the graph (FOR NOW):

```
  const root = d3.hierarchy(rootNode, (d) => d.children);
  const makeTree = d3
    .tree()
    .nodeSize([
      nodeSize.width + nodeSize.horizontalSeparation,
      nodeSize.height + nodeSize.verticalSeparation,
    ]);
  const tree = makeTree(root);
  const descendants = tree.descendants();
  const links = tree.links();
```

The rest is made using svg html tags and their attributes in JSX.

3. I learned something very useful about Typescript + styled components:

```
type StatusLineProps = {
  suspended: boolean;
  status: ReadyType;
};

const StatusLine = styled.div<StatusLineProps>
```

This lets you pass down additional custom props to your styled components without Typescript complaining about it. Hooray!

4. I tried to move all the relevant numbers into variables as high up as possible. My hope is that you (mostly) won't have to know d3 or svg stuff AT ALL in order to make small styling or sizing changes.

### DIFF
--- a/ui/components/DirectedGraph.tsx
+++ b/ui/components/DirectedGraph.tsx
@@ -15,17 +15,58 @@ type Props = {
   descendants: any[];
   links: any[];
   nodeSize: nodeSize;
+  zoomPercent: number;
 };
 
-function DirectedGraph({ className, descendants, links, nodeSize }: Props) {
+function DirectedGraph({
+  className,
+  descendants,
+  links,
+  nodeSize,
+  zoomPercent,
+}: Props) {
+  const zoomBox = 15000 - 14000 * (zoomPercent / 100);
+  const [mouseDown, setMouseDown] = React.useState<{
+    x: number;
+    y: number;
+    active: boolean;
+  }>({ x: 0, y: 0, active: false });
+  const [mouseMove, setMouseMove] = React.useState<{ x: number; y: number }>({
+    x: 0,
+    y: 0,
+  });
+
+  React.useEffect(() => {
+    // d3.select(draggable.current).call(
+    //   d3
+    //     .drag()
+    //     .subject(draggable.current)
+    //     .on("drag", (event) => {
+    //       d3.event.subject.fx = d3.event.x;
+    //       d3.event.subject.fy = d3.event.y;
+    //     })
+    // );
+  }, []);
   return (
-    <svg width="100%" height="100%" viewBox="0 0 5000 5000">
-      <g
-        stroke="#7a7a7a"
-        strokeWidth={5}
-        fill="none"
-        transform="translate(2500, 50)"
-      >
+    <svg
+      width="100%"
+      height="100%"
+      viewBox={`${mouseMove.x} ${mouseMove.y} ${zoomBox} ${zoomBox}`}
+      // onMouseDown={(e) =>
+      //   setMouseDown({ x: e.clientX, y: e.clientY, active: true })
+      // }
+      // onMouseMove={(e) => {
+      //   if (mouseDown.active)
+      //     setMouseMove({
+      //       x: e.clientX - mouseDown.x,
+      //       y: e.clientY - mouseDown.y,
+      //     });
+      // }}
+      // onMouseUp={(e) =>
+      //   setMouseDown({ x: e.clientX, y: e.clientY, active: false })
+      // }
+    >
+      <g stroke="#7a7a7a" strokeWidth={5} fill="none">
         {_.map(links, (l, index) => {
           // l is an object with a source and target node, each with an x and y value. M tells the path where to start, H draws a straight horizontal line, and V draws a straight vertical line
           return (
@@ -38,7 +79,7 @@ function DirectedGraph({ className, descendants, links, nodeSize }: Props) {
           );
         })}
       </g>
-      <g transform="translate(2500, 50)">
+      <g>
         {_.map(descendants, (d, index) => {
           //turn each descendant into a GraphNode
           return (
--- a/ui/components/DirectedGraph.tsx
+++ b/ui/components/DirectedGraph.tsx
@@ -6,8 +6,8 @@ import * as React from "react";
 import styled from "styled-components";
 import { muiTheme } from "../lib/theme";
 import {
-  calculateZoomRatio,
   calculateNodeOffsetX,
+  calculateZoomRatio,
   mapScaleToZoomPercent,
   mapZoomPercentToScale,
 } from "../lib/utils";
--- a/ui/components/DirectedGraph.tsx
+++ b/ui/components/DirectedGraph.tsx
@@ -43,10 +43,10 @@ function DirectedGraph({ className, descendants, links, nodeSize }: Props) {
           //turn each descendant into a GraphNode
           return (
             <foreignObject
-              width={750}
-              height={300}
+              width={nodeSize.width}
+              height={nodeSize.height}
               key={index}
-              transform={`translate(${d.x - 375}, ${d.y})`}
+              transform={`translate(${d.x - nodeSize.width / 2}, ${d.y})`}
               fill="white"
               strokeWidth={2}
               stroke={"#737373"}
--- a/ui/components/DirectedGraph.tsx
+++ b/ui/components/DirectedGraph.tsx
@@ -36,17 +36,17 @@ function DirectedGraph({
     y: 0,
   });
 
-  React.useEffect(() => {
-    // d3.select(draggable.current).call(
-    //   d3
-    //     .drag()
-    //     .subject(draggable.current)
-    //     .on("drag", (event) => {
-    //       d3.event.subject.fx = d3.event.x;
-    //       d3.event.subject.fy = d3.event.y;
-    //     })
-    // );
-  }, []);
+  // React.useEffect(() => {
+  //   d3.select(draggable.current).call(
+  //     d3
+  //       .drag()
+  //       .subject(draggable.current)
+  //       .on("drag", (event) => {
+  //         d3.event.subject.fx = d3.event.x;
+  //         d3.event.subject.fy = d3.event.y;
+  //       })
+  //   );
+  // }, []);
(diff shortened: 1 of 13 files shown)


joshri:
Hello one and all (@opudrovs @ozamosi @JamWils )

I'm bringing you into this draft PR because I wanted to show where the styling is at before spending the rest of today (and probs tomorrow) on the drag + zoom functionality. Also it's a LOT of changes so I wanted to get it out there early.

joshri:
Zoom is back

https://user-images.githubusercontent.com/65822698/180442190-8f690294-2532-4231-b773-5fbb29aefc6a.mov

joshri:
Drag is back - the scaling isn't perfect (your mouse doesn't move 1:1 with the graph) but it's close enough imo. 

https://user-images.githubusercontent.com/65822698/180485126-b86b6314-994a-4f70-a4dd-e59ffa43a4fa.mov

JamWils:
Great work @joshri

opudrovs:
Great job, @joshri ! The code is much simpler and easier to modify than with the Dagre library.

A couple of minor issues:

1. When the vertical scrollbar is not scrolled to top, there is flickering when zooming to 0% and back to a non-zero value. I was able to reproduce the issue consistently in Chrome.

https://user-images.githubusercontent.com/8824708/180670001-598f9ed8-f0fc-479b-a41f-43042d695747.mov

I suggest mapping the `zoomPercent` value received from the slider and passed as a prop to the `DirectedGraph` from the range of [0, 100] to [1, 100] and using this mapped zoom percent for all consequent calculations, smth. like that  (actually, I've copied the first range mapping function I found):

```
function mapValueToRange(value, inMin, inMax, outMin, outMax) {
  return ((value - inMin) * (outMax - outMin)) / (inMax - inMin) + outMin;
}

  const sanitizedZoomPercent = mapValueToRange(zoomPercent, 0, 100, 1, 100);

  //minimum zoomBox is 1000
  const zoomBox = 15000 - 14000 * (sanitizedZoomPercent / 100);
  //since viewbox is so large, make smaller mouse movements correspond to larger pan
  const panScale = 300 / sanitizedZoomPercent;
  
```

If this/a similar solution works, then `mapValueToRange` should be added to `lib/utils`, probably.

2. When first dragging the graph to a side and then changing the zoom values with the slider, the graph looks like it is moving on a curve:

https://user-images.githubusercontent.com/8824708/180670189-83c71ab3-df26-4ac5-9e4d-52d89276bdcc.mov

I haven't looked how to make the graph maintain its position while being zoomed after dragging and I'm not sure if it is important to fix right now, but movement "on a curve" should be related to this line: `const panScale = 300 / zoomPercent;`

joshri:
1. Zoom has been set to a minimum of 5 via a prop on the Mui Slider!
2. To fix the zoom curving - unfortunately I need a ref to access the `getBoundingClientRect` method on the svg. This gives us an accurate ratio for the panning rather than the one I just made up (`300/zoompercent`). The good news is that it doesn't need a `useEffect`. Also I promise this `<SVGSVGElement>` type is correct. Typescript complained until I put it in. The first SVG is the SVG category, and then the second one is the actual svg. Another example would be `<SVGGElement>` or `<SVGRectElement>`.
` const svgRef = React.useRef<SVGSVGElement>();
  let panScale = 1;
  if (svgRef.current) {
    panScale = zoomBox / svgRef.current.getBoundingClientRect().width;
  }`

opudrovs:
1. 

> Zoom has been set to a minimum of 5 via a prop on the Mui Slider!

Cool, much simpler fix than what I suggested. 🎉 

2.

> To fix the zoom curving - unfortunately I need a ref to access the getBoundingClientRect method on the svg.

Great fix, can confirm that the graph now maintains drag position on zoom.

opudrovs:
The code looks fine to me. I only see a few minor issues remaining.

The main issue is that, IMHO, the new lines/edges are somewhat misleading comparing to the old lines.

Old:

Screenshot 2022-07-25 at 20 35 47

New:

Screenshot 2022-07-25 at 20 36 04

With the old lines you could clearly see that all top non-root nodes are connected to the root object. In the new version, they are connected to each other.

opudrovs:
4. It looks like the borders of the node elements are more rounded in the new version. Is it by design?

Old:

Screenshot 2022-07-25 at 20 51 13

New:

Screenshot 2022-07-25 at 20 51 29

Besides, the green thing over the border looks a bit weird at large zoom:
Screenshot 2022-07-25 at 20 08 27

opudrovs:
5. If the text is long, the checkmark icons sometimes look a bit squished:

Screenshot 2022-07-25 at 19 49 02

I wonder if we need to make the nodes wider and/or make more space for text?

Maybe it all is not important to fix it all right now and I am nitpicking already.

joshri:
3. The intersection of the lines is SUPPOSED to be covered by the root node - small bug there I'll have to look into how the path is being drawn exactly. I think I have to add in the horizontal separation bc I added that in post-path.
4. You are right about the status line - I wasn't super focused on design this PR, just trying to get it close to what we have now, but I will decrease the border radius to get it closer

opudrovs:
> 3. The intersection of the lines is SUPPOSED to be covered by the root node - small bug there I'll have to look into how the path is being drawn exactly. I think I have to add in the horizontal separation bc I added that in post-path.

The intersection of the lines is fine with me. I don't like that top level non-root objects look like they are connected to each other (because the lines that connect each of these objects to the root object overlap).

opudrovs:
3. Under the root objects I mean the Kustomization, not GitRepository.

opudrovs:
Great! I have no other questions except for: @JamWils , @ozamosi , what do you think of the new lines here:

https://github.com/weaveworks/weave-gitops/pull/2476#issuecomment-1194472413

and the initial zoom:

Screenshot 2022-07-26 at 01 54 57

If both are OK with you, I am approving the PR.

opudrovs:
If we replace

```
return (
  <path
    key={index}
    d={`M${l.source.x} ${
      l.source.y + nodeSize.verticalSeparation
    }H${l.target.x}V${l.target.y + nodeSize.verticalSeparation}`}
  />
);
```

with

```
return l.source.x === l.target.x ? (
  <path
    key={index}
    d={`M${l.source.x} ${
      l.source.y + nodeSize.verticalSeparation
    }V${l.target.y + nodeSize.verticalSeparation}`}
  />
) : (
  <path
    key={index}
    d={`M${l.source.x} ${
      l.source.y + nodeSize.verticalSeparation
    }L${l.target.x} ${
      l.target.y - nodeSize.verticalSeparation / 2
    }V${l.target.y + nodeSize.verticalSeparation}`}
  />
);
```
(using only the second part without condition would work too, but I'm not sure if there could be any artifacts under some conditions)
we can make lines look like that:

Screenshot 2022-07-26 at 03 06 46

The only issue left is that the nodes are not centered relative to the root node, but it is not related to making the lines diagonal.

opudrovs:
@joshri , @ozamosi said that there was a ticket ( https://github.com/weaveworks/weave-gitops/issues/2484 ) about changing the shape of the lines (what I had not seen).

In this case the only issue remaining is centering non-root nodes horizontally (if there is an even number of columns) relative to the root node.

ozamosi:
Looking carefully at #2484, that ticket describes a path that goes straight down, straight to the side, straight down again, whereas this PR only out and then down. So it's not quite the "right" shape (but then this branch doesn't promise to implement that shape).

> The only issue left is that the nodes are not centered relative to the root node

I think the issue here is that the boxes have irregular spacing. It looks like if the child node has children of its own, then the gap before we draw the next child is 2 boxes' width, whereas if the child node has no children then the next node is drawn much closer. So in your screenshot, it's "centered" for an even number of boxes, but there's an empty box on the left side of the parent node.

Do I make any sense?

opudrovs:
Great observations, @ozamosi !

opudrovs (approved):
As we agreed, lines and node positioning can be handled in a separate ticket.

LGTM in general, removing the Dagre library and switching to pure D3.js is a huge step forward for the graph! 🎉

PR #2482, opened 2022-07-25
fix: stop auth passthrough from ignoring token in favour of SA

jmickey:
Fix a bug where the `BearerToken` is being ignored because the `BearerTokenFile` is being used to authenticate the client as the InCluster service account.

Additionally, we now validate the token when it is received by utilising the `TokenReview` API. This requires additional permissions on the service account:

```
kind: ClusterRole
name: system:auth-delegator
```

### DIFF
--- a/core/clustersmngr/clustersmngr.go
+++ b/core/clustersmngr/clustersmngr.go
@@ -76,12 +76,18 @@ func ClientConfigWithUser(user *auth.UserPrincipal) ClusterClientConfig {
 	return func(cluster Cluster) *rest.Config {
 		config := &rest.Config{
 			Host:            cluster.Server,
-			BearerToken:     cluster.BearerToken,
 			TLSClientConfig: cluster.TLSConfig,
-			Impersonate: rest.ImpersonationConfig{
+		}
+
+		if user.Token != "" {
+			config.BearerToken = user.Token
+			config.BearerTokenFile = ""
+		} else {
+			config.BearerToken = cluster.BearerToken
+			config.Impersonate = rest.ImpersonationConfig{
 				UserName: user.ID,
 				Groups:   user.Groups,
-			},
+			}
 		}
 
 		enabled, err := flowcontrol.IsEnabled(context.Background(), config)
--- a/core/clustersmngr/clustersmngr.go
+++ b/core/clustersmngr/clustersmngr.go
@@ -81,7 +81,6 @@ func ClientConfigWithUser(user *auth.UserPrincipal) ClusterClientConfig {
 
 		if user.Token != "" {
 			config.BearerToken = user.Token
-			config.BearerTokenFile = ""
 		} else {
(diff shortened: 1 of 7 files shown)


core/clustersmngr/clustersmngr.go
```diff
@@ -76,12 +76,18 @@ func ClientConfigWithUser(user *auth.UserPrincipal) ClusterClientConfig {
...
 		config := &rest.Config{
 			Host:            cluster.Server,
-			BearerToken:     cluster.BearerToken,
 			TLSClientConfig: cluster.TLSConfig,
-			Impersonate: rest.ImpersonationConfig{
+		}
+
+		if user.Token != "" {
+			config.BearerToken = user.Token
+			config.BearerTokenFile = ""
```
foot, line 84: Maybe don't need this one here now.

pkg/server/auth/passthrough.go
```diff
@@ -35,5 +41,32 @@ func (pg *BearerTokenPassthroughPrincipalGetter) Principal(r *http.Request) (*Us
 		return nil, nil
 	}
 
-	return &UserPrincipal{Token: extractToken(token)}, nil
+	token = extractToken(token)
+
+	config, err := kubecfg.GetConfig()
```
foot, line 46: Maybe we pass in authSrv.cfg.kubernetesClient instead of creating a new client wdyt?

jmickey: I don't think they're the same client. `authSrv.cfg.kubernetesClient` is a `sigs.k8s.io/controller-runtime/pkg/client.Client`, whereas the client in `passthrough.go‘ is a `k8s.io/client-go/kubernetes.ClientSet`

jmickey: We also pass the AuthServer as a param, so we don't have access to the private fields on the auth server object. We could make them public, or add a method to return the client.

pkg/server/auth/passthrough.go
```diff
@@ -35,5 +41,24 @@ func (pg *BearerTokenPassthroughPrincipalGetter) Principal(r *http.Request) (*Us
 		return nil, nil
 	}
 
-	return &UserPrincipal{Token: extractToken(token)}, nil
+	token = extractToken(token)
+
+	authv1.AddToScheme(pg.kubernetesClient.Scheme())
```
foot, line 46: We usually do all the `AddToScheme` at client creation. I don't know if `AddToScheme` is expensive etc, but would move if there just to keep that pattern, even though its harder to understand _why_ we're adding it all the way at the top.

jmickey: I only added it here for testing so I can pull the branch into WGE running in tilt. I will move it.

foot (approved):
Nice one 👌

ozamosi (approved):

PR #2487, opened 2022-07-25
List CRDs in Flux Runtime

### LINKED ISSUE #2241: Display CRDs on the Flux Runtime page
As a cluster admin I want to see the Flux CRDs and their version so that I know which `apiVersion` I must use in my manifests.

### Acceptance Criteria
- Create two tabs on the page "Controllers" and "CRD"
- By default when I click on "Flux Runtime" the controllers page should be the active page.
- When I click the "CRD" tab I am shown a different table outlined below.
- the CRDs table would contain these columns "Name" "Kind" "Version"

| Name                                           | Kind                  | Version |
|------------------------------------------------|-----------------------|---------|
| alerts.notification.toolkit.fluxcd.io          | Alert                 | v1beta1 |
| buckets.source.toolkit.fluxcd.io               | Bucket                | v1beta2 |
| gitrepositories.source.toolkit.fluxcd.io       | GitRepository         | v1beta2 |
| helmcharts.source.toolkit.fluxcd.io            | HelmChart             | v1beta2 |
| helmreleases.helm.toolkit.fluxcd.io            | HelmRelease           | v2beta1 |
| helmrepositories.source.toolkit.fluxcd.io      | HelmRepository        | v1beta2 |
| imagepolicies.image.toolkit.fluxcd.io          | ImagePolicy           | v1beta1 |
| imagerepositories.image.toolkit.fluxcd.io      | ImageRepository       | v1beta1 |
 | imageupdateautomations.image.toolkit.fluxcd.io | ImageUpdateAutomation | v1beta1 |
| kustomizations.kustomize.toolkit.fluxcd.io     | Kustomization         | v1beta2 |
| providers.notification.toolkit.fluxcd.io       | Provider              | v1beta1 |
| receivers.notification.toolkit.fluxcd.io       | Receivers             | v1beta1 |

The `Name` rows could have a link that would open in pop-up the YAML content on the CRD.

[Figma](https://www.figma.com/file/IVHnM9iyeFWpd11evtY8ux/Weave-GitOps?node-id=10539%3A59692)

joshri:
Closes #2241 

Adds a second data table with CRDs to the Flux Runtime page using a new api request - this pr also adds URL addresses to the Flux Runtime page and handles some *interesting* behavior with overflow of chips. There's a lot of whitespace at the end of the page because there are a trillion filters in the CRDs table. Fixing that particular strangeness would take a lot of effort that I didn't want to put in this pr - how many filters are we expecting per page in general? These pages weren't made with two tables in mind. 

https://user-images.githubusercontent.com/65822698/180865579-ee31b090-bacf-4136-9863-4ca903410183.mov

Why are they coming out at different widths? Depends on the length of the filters - maybe we impose a character limit on these? Ignore the fact that there's a name column and we shouldn't be filtering that - this could still be a problem in the future?
image
image


core/server/fluxruntime.go
```diff
@@ -93,6 +94,39 @@ func (cs *coreServer) ListFluxRuntimeObjects(ctx context.Context, msg *pb.ListFl
...
+	list := &apiextensions.CustomResourceDefinitionList{}
+
+	if err := clustersClient.List(ctx, msg.ClusterName, list, opts); err != nil {
+		respErrors = append(respErrors, &pb.ListError{ClusterName: msg.ClusterName, Message: fmt.Sprintf("%s, %s", ErrListingDeployments.Error(), err)})
+	}
+
+	for _, d := range list.Items {
+		r := &pb.Crd{
+			Name:        d.Spec.Names.Plural + "." + d.Spec.Group,
+			Version:     d.Spec.Versions[0].Name,
```
ozamosi, line 120: I worry `versions` might be empty. We should make sure we have tests for that.

Have a look at https://github.com/weaveworks/weave-gitops/blob/main/core/server/fluxruntime_test.go#L84 - it creates a few objects and plugs them into a `fake` client, creates a new core server using the fake client, and then calls the endpoint in the client.

core/server/fluxruntime.go
```diff
@@ -93,6 +94,39 @@ func (cs *coreServer) ListFluxRuntimeObjects(ctx context.Context, msg *pb.ListFl
...
+	respErrors := []*pb.ListError{}
+
+	opts := client.MatchingLabels{
+		coretypes.PartOfLabel: FluxNamespacePartOf,
+	}
+
+	list := &apiextensions.CustomResourceDefinitionList{}
+
+	if err := clustersClient.List(ctx, msg.ClusterName, list, opts); err != nil {
+		respErrors = append(respErrors, &pb.ListError{ClusterName: msg.ClusterName, Message: fmt.Sprintf("%s, %s", ErrListingDeployments.Error(), err)})
```
ozamosi, line 114: `ErrListingDeployments` talks about deployments - we need a new error for CRDs.

core/server/fluxruntime.go
```diff
@@ -93,6 +94,39 @@ func (cs *coreServer) ListFluxRuntimeObjects(ctx context.Context, msg *pb.ListFl
...
+	}
+
+	for _, d := range list.Items {
+		r := &pb.Crd{
+			Name:        d.Spec.Names.Plural + "." + d.Spec.Group,
```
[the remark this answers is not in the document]
joshri: Absolutely worth at least including the fields - gives us the flexibility to filter or not later

ozamosi (approved):

Issue #2471, opened 2022-07-21
Sort by "last updated" inconsistent


makkes:
**Describe the bug**

When navigating to the "Applications" page and clicking the "Last Updated" column header to sort by that value, the sorting is incorrect:

![2022-07-21_09-21](https://user-images.githubusercontent.com/387444/180157660-366fa0ee-5e6b-4ec2-95f3-600ca714522b.png)

**Environment**

- Weave-Gitops Version: 0.9.0 (chart version 2.2.2)
- Flux Version: 0.31.3
- Kubernetes version: 1.24.3

**To Reproduce**

see above

**Expected behavior**

Sorting should be consistent

**Actual Behavior**

Row ordering appears to be random

**Additional Context (screenshots, logs, etc)**

see above

PR #2493, opened 2022-07-26
Refactor SortWithType to use OrderBy from lodash so we will be able to Order/Sort Data by Fields to generate Unique Collection that is not Flick.

ahussein3:
-Fixes [#975](https://github.com/weaveworks/weave-gitops-enterprise/issues/975)
-Fixes https://github.com/weaveworks/weave-gitops/issues/2471 where sort by updatedAt wasn't working as it should. by exposing the sortValue instead of the react component.
-Also fixes https://github.com/weaveworks/weave-gitops/pull/2493#discussion_r931270004


ui/components/AutomationsTable.tsx
```diff
@@ -139,7 +147,7 @@ function AutomationsTable({ className, automations, hideSource }: Props) {
     <URLAddressableTable
       fields={fields}
       filters={initialFilterState}
-      rows={automations}
+      rows={sortAutomationsByDate(automations)}
```
jpellizzari, line 150: We (should) have sorting controls already on the `DataTable`. Can we set the default sort column to "Last Updated At" rather than sorting before passing them in.

@joshri Is there a way to select the "default" sort column?

joshri: There is a `defaultSort` prop on DataTable which is looking for the index of the field you want to sort by. @ahussein3 does this actually fix 975? Won't other sorting cause the reordering issue? It's something that's been present since the polling. I thought previously that if we add a key (or maybe all objects already have a key?) we could use that as a backup sort value for cases where objects have the same value (which could be one reason why the sort isn't stable). If this fixes the reordering problems though that would be magical!

jpellizzari: If they `key` prop is the index of the the loop that is being iterated over, this will cause issues, since the order might be changing on every poll. Something to look into.

foot: Ooops, yes, @ahussein3 and I were talking about how to address this and I didn't really think it through.

I think we want to sort by the selected column _and then_ at least `(name, namespace, kind, clusterName)` which should make a unique set and provide a stable sort.

`_.sortBy(data, [selectedColumn, name, namespace, kind, clusterName])` or something?

sympatheticmoose: as part of this change could we make the default sort by Status so that failing/not ready items are highlighted?

ahussein3: looks like sort by time is working as a key. I mean they all are sorted by time then by name. that's make it the rate of change is not that frequent like by name only. As data been called from different sources so they come in random shape so when sort by name it will get anyone first but as they are sorted by time so they are ordered and that item with that name will always be first. Sure this is not fixing the Issue but only reduce the rate of change. Hope I did explain that clearly. @joshri , @jpellizzari .
while we can set default sort I guess I can refactor that and go with the default sort instead.
As I can't see a final solution that can fix this issue. any suggestions ?

ui/components/AutomationsTable.tsx
```diff
@@ -141,6 +144,7 @@ function AutomationsTable({ className, automations, hideSource }: Props) {
       filters={initialFilterState}
       rows={automations}
       className={className}
+      defaultSort={5}
```
foot, line 147: Maybe a comment here that 5 = status

ui/components/DataTable.tsx
```diff
@@ -153,14 +140,10 @@ function UnstyledDataTable({
   defaultSort = 0,
   children,
 }: Props) {
-  const [sort, setSort] = React.useState(fields[defaultSort]);
+  const [sortField, setSort] = React.useState(fields[defaultSort]);
```
foot, line 143: maybe `setSortField` too?

ui/components/SourcesTable.tsx
```diff
@@ -42,6 +42,7 @@ function SourcesTable({ className, sources }: Props) {
 
   return (
     <URLAddressableTable
+      defaultSort={4}
```
foot, line 45: another comment about 4 = lastUpdated or is it status?

foot: There isn't really an `id` field or something, that would be nice, maybe we do in a follow up though..

ui/components/DataTable.tsx
```diff
@@ -75,31 +75,18 @@ const TableButton = styled(Button)`
...
-  const sortFn = sort.sortValue || defaultSortFunc(sort);
-  return (rows?.slice() || []).sort((a: Row, b: Row) => {
-    switch (sort.sortType) {
-      case SortType.number:
-        return sortFn(a) - sortFn(b);
-
-      case SortType.date:
-        return Date.parse(sortFn(a)) - Date.parse(sortFn(b));
-
-      case SortType.bool:
```
foot, line 94: I think we can remove `SortType` from everywhere now too actually?

ui/components/DataTable.tsx
```diff
@@ -75,31 +75,18 @@ const TableButton = styled(Button)`
...
-        if (sortFn(a) === sortFn(b)) return 0;
-        else if (sortFn(a) === false && sortFn(b) === true) return -1;
-        else return 1;
-
-      default:
-        return (sortFn(a) || "").localeCompare(sortFn(b) || "");
-    }
-  });
+export const sortWithType = (
```
foot, line 78: Maybe needs a new name now its not so much about the type.

`sortRows` ?

ui/components/__tests__/DataTable.test.tsx
```diff
@@ -62,7 +62,7 @@ describe("DataTable", () => {
   describe("sorting", () => {
     describe("sortWithType", () => {
       it("should handle sorting with case SortType.string", () => {
-        const nameSort = sortWithType(rows, {
+        const nameSort = sortWithType(rows, false, {
```
foot, line 65: We can probably remove these tests? They are just testing lodash functionality now, or? wdyt?

ui/components/SourcesTable.tsx
foot: We could also make it a property of the `Field`, like `defaultSort: true` on the one we wanna start sorting by.

E.g. Then it would become:
```
  const [sortField, setSort] = React.useState(fields.find(f => f.defaultSort) || fields[0]);
```

Feels a little less fragile if we add/rm/move columns. Could also be a follow up though.

ui/components/__tests__/DataTable.test.tsx
ahussein3: yes we can do that they are testing lodash :D

ui/components/DataTable.tsx
ahussein3: actually yes we can do this

ozamosi (changes requested):
Can we just pause for a second and decide what we actually are trying to do here.

As far as I can see, we've made Sources default sort by date - which doesn't change _all the time_ but does change, so it shouldn't be _too flickery_. And we've made Automation sort by status, of which there's about two different values, so it will be almost "as flickery" as it was before?

If the original problem is that multi-cluster requests come back in random order, then this doesn't solve either whenever you sort by anything that isn't the default sort. Or have I misunderstood?

The timestamp fix is great - I believe you've fixed #2471.

ahussein3:
@ozamosi  yes I do I had found the issue and yes I'm writing in the description that I had included the fix here.

ui/components/__tests__/DataTable.test.tsx
```diff
@@ -59,112 +58,6 @@ describe("DataTable", () => {
...
-        const numberSort = sortWithType(rows, {
-          label: "Last Synced At",
-          value: "lastSyncedAt",
-          sortType: SortType.number,
-          sortValue: ({ lastSyncedAt }) => lastSyncedAt,
-        });
-        expect(numberSort[0].lastSyncedAt).toBe(1000);
-      });
-    });
-    it("initially sorts based on defaultSort", () => {
```
foot, line 106: Sorry! It would still be good to keep the DataTable tests, just lose the `sortWithType` ones

foot:
> As far as I can see, we've made Sources default sort by date

There was a change of plan in end. We now sort by some preferred field (`Status` right now), and then _all_ the other columns in order.

You're right there will be churn in the table. If an automation/source transitions to NotReady (or Reconciling?) and back again, it will jump to the top of the table (while everything else remains in a stable sort), and then back down to its sorted place. If NotReady happens a lot this is not ideal.

cc @sympatheticmoose

ozamosi (approved):
> We now sort by some selected field (which is Status by default right now), and then all the other columns in order.

I was missing something - there it is! Thank you!

sympatheticmoose:
The intent with sorting by status is to help highlight the automation objects which the user should most likely care about, those which have issues or are currently in-flux. But if a more stable view is required, then I would expect the user to sort by name or other more stable field. I suggested the change since we were going in and changing the sort behaviour.

foot:
> The intent with sorting by status is to help highlight the automation objects which the user should most likely care about

Okay playing around with this I think it works quite well sorting first by Status.

ui/components/DataTable.tsx
```diff
@@ -15,37 +15,26 @@ import Icon, { IconType } from "./Icon";
...
   maxWidth?: number;
+  /** boolean for field to initially sort against. */
+  defaultSort?: boolean;
 };
 
 /** DataTable Properties  */
 export interface Props {
   /** CSS MUI Overrides or other styling. */
   className?: string;
-  /** A list of objects with four fields: `label`, which is a string representing the column header, `value`, which can be a string, or a function that extracts the data needed to fill the table cell, `sortType`, which determines the sorting function to be used, and `sortValue`, which customizes your input to the search function */
```
foot, line 43: Oops lets keep this comment but just remove the sortType bit

PR #2494, opened 2022-07-26
Add SubRouterTab Styles

### LINKED ISSUE #2412: Refine tab design
- [ ] Font size should be `12px`
- [ ] height of tab should be `32px`
- [ ] set `letter-spacing: 1px;` to add space between characters
- [ ] tab width should be `132px`
- [ ] text should be `bold`(`font-weight:800`) in active state and `semi-bold`(`font-weight:600`) in all other states.
- [ ] `border-bottom` should be `3px` for the active tab

Screenshot 2022-07-07 at 11 23 39

### References

[Figma](https://www.figma.com/file/IVHnM9iyeFWpd11evtY8ux/Weave-GitOps?node-id=9469%3A50766)

joshri:
Closes #2412

Tab styling now more consistent with Buttons

image

### DIFF
--- a/ui/components/SubRouterTabs.tsx
+++ b/ui/components/SubRouterTabs.tsx
@@ -123,7 +123,7 @@ export default styled(SubRouterTabs).attrs({ className: SubRouterTabs.name })`
   .MuiTabs-fixed {
     height: 32px;
   }
-  .PrivateTabIndicator-root-1 {
+  .MuiTabs-indicator {
     height: 3px;
   }
 `;
--- a/ui/components/SubRouterTabs.tsx
+++ b/ui/components/SubRouterTabs.tsx
@@ -124,6 +124,6 @@ export default styled(SubRouterTabs).attrs({ className: SubRouterTabs.name })`
     height: 32px;
   }
   .PrivateTabIndicator-root-1 {
-    height 3px;
+    height: 3px;
   }
 `;
--- a/ui/components/SubRouterTabs.tsx
+++ b/ui/components/SubRouterTabs.tsx
@@ -68,21 +68,30 @@ function SubRouterTabs({ className, children, rootPath, clearQuery }: Props) {
         indicatorColor="primary"
         value={routesToIndex(routes, window.location.pathname)}
       >
-        {_.map(routes, (route, i) => (
-          <Tab
-            component={ForwardedLink as typeof Link}
-            key={i}
-            to={formatURL(`${route.path}`, clearQuery ? "" : query)}
-            className={`${
-              window.location.pathname.includes(route.path) && "active-tab"
-            }`}
-            label={
-              <Text uppercase bold>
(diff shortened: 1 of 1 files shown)


opudrovs:
Tested the styles locally, everything is fine except for `border-bottom should be 3px for the active tab`

If border-bottom here means MuiTabs-indicator height, then its height is 2px, not 3px, will add a comment.

ui/components/SubRouterTabs.tsx
```diff
@@ -96,4 +105,24 @@ export default styled(SubRouterTabs).attrs({ className: SubRouterTabs.name })`
...
+  }
+  //trust me there's both tab and tabS
+  .MuiTabs-root {
+    min-height: 32px;
+  }
+  .MuiTabs-fixed {
+    height: 32px;
+  }
+  .PrivateTabIndicator-root-1 {
+    height 3px;
```
opudrovs, line 126: Missing colon. And I don't see that this setting has an effect even when adding the colon. Maybe we should use

```
.MuiTabs-indicator {
  height: 3px;
}
```

instead?

joshri: Wow...sorry not sure how the colon got messed up there. It seemed like PrivateTabIndicator was the correct class - I'm overriding the 2px value here. It looks different to me when I change the value but I could definitely be wrong.  
image

joshri: 2px screenshot:
image

3px screenshot:
image

opudrovs: How do you run the app? Via `npm start` or `tilt up`?

Because I'm using `tilt up` and don't see the `.PrivateTabIndicator-root-1` class in the inspector and the height of the blue border block for me is still 2px with the branch code.

But if I replace `.PrivateTabIndicator-root-1` with `.MuiTabs-indicator`, the height becomes 3px, as expected. 

Maybe `.PrivateTabIndicator-root-1` works only for `npm start`? Besides, here it says that it's a private class and should not be used

https://github.com/mui/material-ui/issues/21736#issuecomment-656402837

and I haven't found it in MUI docs:

With `.PrivateTabIndicator-root-1`:

Screenshot 2022-07-27 at 22 59 56

With `.MuiTabs-indicator`:

Screenshot 2022-07-27 at 23 08 43

opudrovs: I also see another advice of using some props for this:

https://stackoverflow.com/questions/61394435/how-do-i-override-material-uis-tab-selection-color

joshri: Changed the class to `.MuiTabs-indicator` - should be good to go!

opudrovs (approved):
LGTM! 🎉  Tested the branch locally, everything, including the "border" element height matches the acceptance criteria.

PR #2495, opened 2022-07-26
Update logo

### LINKED ISSUE #2435: Update logo
**Short description**
Update the top left product logo, related to #2377 
20220712new-bg

**Acceptance criteria**
- [ ] new logo

**Design materials**
[Logo](https://drive.google.com/file/d/13R_kYK5H2Ump73LrZ6Eq4WQSy6B15PEH/view?usp=sharing)

joshri:
Closes #2435

Updates Logo - There is an issue where the image is not centered unless it is adjusted one pixel downward. It's consistently one pixel down in Figma and doesn't look centered without this shift. Video below screenshot.

image
image

https://user-images.githubusercontent.com/65822698/181101249-fbbf7c92-6b6d-4a7a-a776-e3d51548d30e.mov


opudrovs:
There is no hand cursor (was it intentionally removed with the link?) when I hover over the logo and nothing happens when I click the logo (no redirect). Tested it in Chrome and FF.

opudrovs:
Maybe you could set an isRedirecting  flag in the Logo component state to true when the image is clicked and render the redirect instead of the image if it was clicked. It will work to redirect once.

But then the state of the Logo component will get stuck on isRedirecting = true because the Logo component is shared across the pages, so we'll need to add a useEffect hook, like that:

```
function Logo({ className }: Props) {
  const [isRedirecting, setIsRedirecting] = React.useState<boolean>(false);

  React.useEffect(() => {
    if (isRedirecting) {
      setIsRedirecting(false);
    }
  }, [isRedirecting]);

  return (
    <Flex className={className} start>
      <img src={images.logoSrc} onClick={() => setIsRedirecting(true)} />
      {isRedirecting && <Redirect to={V2Routes.Automations} />}
    </Flex>
  );
}
```

Is there a reason of not using a link for navigation like before?

joshri:
The Link component wrapping was really messing with the spacing on the new svg. I've played with it enough to fix it....ish. I have a janky value of 8px padding-top in there. I can't explain to you why it has to be 8px other than it looks right. 

image

opudrovs:
I think there are issues with whitespace in the exported image, it's just sticking out beyond the header, doesn't look right:

Screenshot 2022-07-28 at 00 33 50

I've added questions in the ticket.

joshri:
The svg definitely caused some issues - getting rid of that white space would be helpful for sure haha <|:^)

opudrovs:
Please update the  Logo component snapshot.

joshri:
> Please update the Logo component snapshot.

Updated!

Separate comment - I want to know what you think about not using our `Link` component here - I realized the reason it was throwing off the styling is that it automatically wraps its children in a Text component, and rather than changing that just imported `Link` from react-router separately.

opudrovs:
> Updated!

Thanks!

> Separate comment - I want to know what you think about not using our Link component here - I realized the reason it was throwing off the styling is that it automatically wraps its children in a Text component, and rather than changing that just imported Link from react-router separately.

Sure, will check it now.

ui/components/Link.tsx
```diff
@@ -58,6 +58,6 @@ function Link({
 export default styled(Link)`
   text-decoration: none;
   &.title-bar-button {
-    width: 250px;
+    width: 240px;
```
opudrovs, line 61: What is the reason for changing the styling of our custom Link component if we are directly importing the Link component from the React Router in the Logo component?

joshri: Great point - I think I'm able to delete that class entirely now as it *should* be unused. Will verify.

joshri: Update - definitely unused - pushed up new code removing the class and updating the snaps

opudrovs (approved):

PR #2499, opened 2022-07-27
fix: GetImpersonatedClient returns a multierror for use in API response

jmickey:
Closes weaveworks/weave-gitops-enterprise#1069

**What changed?**

For list requests, when a connection cannot be established to a leaf cluster we not longer fail the entire request, instead any errors are added to the response in the `response.Errors` field. This means that clusters that can be connected to, and a client is successfully created for, will still return valid data.

Additionally, get requests no longer create clients for all clusters and then filter to the required client based on the request. Rather, we only create a client for the relevant cluster as specified within the request query params.

**Why was this change made?**

When a connection to a leaf cluster could not be established while creating a client, the request would respond with an error and no data for any cluster would be returned, including the management cluster. 

Clients are now created in parallel and errors are returned to the calling function within a `multierror.Error` struct. A mutex has been add to the `ClientsPool` struct, which is used within the `Add()` method to allow for concurrent creation of clients, while maintaining thread safety when adding the client to the map stored within the pool.

The returned error is typecast back into a `multierror.Error` type, and errors are extracted from the errors list and added to the request response.

For get requests a new `ClientsFactory` method `GetImpersonatedClientForCluster` has been added which iterates through the available known clusters, locates the specific cluster required and a client is only created for the cluster.

**Release notes**

Fix: Applications, Sources, and Flux Runtime pages will now correctly display data for reachable clusters even when one or more clusters cannot be reached.


core/clustersmngr/clustersmngr.go
```diff
@@ -130,6 +132,24 @@ func (cp *clientsPool) Add(cfg ClusterClientConfig, cluster Cluster) error {
 	return nil
 }
 
+// Add adds a cluster client to the clients pool with the given user impersonation
```
chanwit, line 135: I think the comment of `AddSync` is somehow taken from `Add`?

core/clustersmngr/clustersmngr.go
```diff
@@ -130,6 +132,24 @@ func (cp *clientsPool) Add(cfg ClusterClientConfig, cluster Cluster) error {
...
+	config := cfg(cluster)
+
+	leafClient, err := client.New(config, client.Options{
+		Scheme: cp.scheme,
+	})
+	if err != nil {
+		return fmt.Errorf("failed to create leaf client: %w", err)
+	}
+
+	mutex.Lock()
```
chanwit, line 146: is it possible that `mutex` happens to be `nil` ?
if so, it would be nice to check
`if mutex == nil` and return an error before this line.

core/clustersmngr/factory.go
```diff
@@ -188,13 +203,61 @@ func (cf *clientsFactory) GetImpersonatedClient(ctx context.Context, user *auth.
...
+			defer wg.Done()
+
+			if err := pool.AddSync(ClientConfigWithUser(user), cluster, mutex); err != nil {
+				errChan <- &ClientError{ClusterName: cluster.Name, Err: fmt.Errorf("failed adding cluster client to pool: %w", err)}
+			}
+		}(cluster, pool, &mutex, errChan)
+	}
+
+	wg.Wait()
+	close(errChan)
```
chanwit, line 225: You close `errChan` here before looping over it on line 229.
It means that you will already get `nil` out of `errChan`.

I'm guessing (correct me if I'm wrong) that this `errChan` would be needed to close via `defer`?

chanwit (changes requested):

chanwit:
Please also address the failed unit test here:
https://github.com/weaveworks/weave-gitops/runs/7538664517?check_suite_focus=true#step:6:472

chanwit:
@jmickey, do we have some examples of the error messages sent back to the front end by this PR?
I just want to make sure that we don't leak any sensitive information via this change.

core/clustersmngr/factory.go
jmickey: Hi @chanwit 

Because the channel is buffered it means we can place messages onto the channel and then close it, and still receive the messages later when we `range` over the channel. See here for an example: https://gobyexample.com/range-over-channels. Here is a playground example in action: https://go.dev/play/p/_tN7jc6tTz2

core/clustersmngr/clustersmngr.go
ozamosi: I'm not sure I'm a fan of this API - it _can_ be used securely, but now any time anybody uses `AddSync`, everybody always has to use `AddSync`. Could we instead just put the mutex in the pool itself, and always lock?

core/clustersmngr/factory.go
chanwit: Thank you for sharing the playground example. It makes sense to me now.

chanwit (approved):
LGTM Thank you so much @jmickey

PR #2502, opened 2022-07-27
Change graph line path

### LINKED ISSUE #2484: Graph lines to extend at lines when branching out horizontally
As a user, I would like to see the lines of the graph extend out horizontally from the lines instead of from the sides of the node.

In addition, there should be equal spacing between child nodes.

image

joshri:
Closes #2484

New paths:
image

A lowercase v or h in the d attribute of a path means that the number given will draw the line relative to the current position vs. using absolute coordinates. This was key in separating the y distance into two parts.


opudrovs:
Could you please also remove the dagre package because it is graph-related and we missed it in the previous graph PR? We don't want to keep unneeded dependencies.

joshri:
> Could you please also remove the dagre package because it is graph-related and we missed it in the previous graph PR? We don't want to keep unneeded dependencies.

YES. Forgot again. Classic me.

opudrovs:
@joshri thank you for removing dagre, at last we cleaned up everything!

opudrovs:
The new lines look fine to me, if there are some vertical spacing adjustments required in the future, we'll just need more detailed specs. For now it should work fine.

There is the second acceptance criterion in the ticket:

> In addition, there should be equal spacing between child nodes.

Is it technically hard to implement?

opudrovs:
Just sae you comment on Slack, so this ticket really covers both issues.

ui/components/DirectedGraph.tsx
```diff
@@ -49,13 +49,17 @@ function DirectedGraph({
...
             return (
               <path
                 key={index}
-                d={`M${l.source.x} ${
-                  l.source.y + nodeSize.verticalSeparation
-                }H${l.target.x}V${l.target.y + nodeSize.verticalSeparation}`}
+                d={`M${l.source.x}, ${l.source.y + nodeSize.verticalSeparation} 
+                v${(l.target.y - l.source.y) / 2} 
+                H${l.target.x} 
+                v${(l.target.y - l.source.y) / 2}`}
```
opudrovs, line 62: Can we add a constant before return for `(l.target.y - l.source.y) / 2` to prevent doing the same calculation twice?

opudrovs (approved):
LGTM! 🎉

joshri:
As discussed in standup I will implement the new node spacing in a follow up issue https://github.com/weaveworks/weave-gitops/issues/2506

PR #2504, opened 2022-07-28
2489 gitops run port forwarding error follow up

### LINKED ISSUE #2489: Port forwarding error immediately after installing the GitOps Dashboard not on the same run when installing Flux
If I run the `gitops run` on a clean cluster and refuse to install the GitOps Dashbord furing the first command run (when installing Flux) and only install it during a subsequent run, I get the following error:

```
► Port forwarding to pod dev-bucket/dev-bucket-f649c7d6c-x5xs7 ...
Forwarding from 127.0.0.1:9000 -> 9000
Forwarding from [::1]:9000 -> 9000
✔ Port forwarding for dev-bucket is ready.
✗ Error getting pod from specMap: no running pods found for service
Error: dashboard pod not found
```

During the next runs the dashboard works fine.

Probably, the pod readiness is not detected properly during that second run.

opudrovs:
Closes #2489 , finally. I tested it and am unable to reproduce the error anymore.

- Removed passing pod status phase to `GetPodFromSpecMap` (added in the previous PR). Now the code is simpler.
- Rolled back error messages to previous versions (without pod phase). Replaced "service" in deployment-related error messages in `GetPodFromSpecMap` with "deployment".
- Added properly waiting for getting the dashboard pod and making sure its pod Ready condition is true.
- Added constants for getting pod errors which are re-used in code or similar errors which could be re-used (also to make checking for specific errors easier). Other `run` pkg errors can be added later.
- Since I was on this, added passing namespace to some dashboard-related functions instead of using `wego.DefaultNamespace`.
- Added a utility function to test pod conditions `IsPodStatusConditionPresentAndEqual`.

Notes:
- The updated fix does not use these constants for checking the returned error, so they are only used once in the `GetPodFromSpecMap` function. So, I can remove them if needed. Just thought that maybe we could start adding some constants for run command error messages already.
- There is a static checks error because of reported vulnerability in the `"@docusaurus` packages:
https://github.com/weaveworks/weave-gitops/runs/7600157647?check_suite_focus=true
I could update them from version `^2.0.0-beta.20` to `^2.0.0-rc.1` (it might fix it, but I have not checked the `node-fetch` version the new version requires) but it is a core package (site generator).

So, it might be better if someone who works on the website checked that this update does not break anything.

### DIFF
--- a/cmd/gitops/beta/run/cmd.go
+++ b/cmd/gitops/beta/run/cmd.go
@@ -30,7 +30,6 @@ import (
 	"github.com/weaveworks/weave-gitops/pkg/logger"
 	"github.com/weaveworks/weave-gitops/pkg/run"
 	"github.com/weaveworks/weave-gitops/pkg/version"
-	corev1 "k8s.io/api/core/v1"
 	apimeta "k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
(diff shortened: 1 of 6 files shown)


ozamosi (approved):
Looks good!

(I think the static check fixes itself if you just rerun it, there's just a race condition between this branch and main that should've sorted itself out)

chanwit (approved):
LGTM

PR #2509, opened 2022-07-29
change docs to refer to the OCI registry for chart storage

makkes:
**What changed?**

Revised all public-facing documentation to refer to the ghcr.io OCI registry for chart storage.

**Why was this change made?**

OCI is more efficient in terms of resource usage by Flux (no index to
be cached etc.) and in terms of maintenance as no index has to be
created and published.

The existing HTTPS Helm repository may also have led to issues with
caching as hinted to by the discussion in
https://github.com/fluxcd/flux2/discussions/2939.

Preview: https://staging.docs.gitops.weave.works/docs/next/installation

### DIFF
--- a/website/docs/getting-started.mdx
+++ b/website/docs/getting-started.mdx
@@ -122,7 +122,7 @@ cd fleet-infra
 ```
 flux create source helm ww-gitops \
   --interval=60m \
-  --url=https://helm.gitops.weave.works \
+  --url=oci://ghcr.io/weaveworks/charts \
   --export > ./clusters/my-cluster/weave-gitops-source.yaml
 ```
 
@@ -136,7 +136,8 @@ metadata:
   namespace: flux-system
 spec:
   interval: 1h0m0s
-  url: https://helm.gitops.weave.works
+  type: oci
+  url: oci://ghcr.io/weaveworks/charts
 ```
 
 3. Commit and push the `weave-gitops-source.yaml` to the `fleet-infra` repository
--- a/website/docs/getting-started.mdx
+++ b/website/docs/getting-started.mdx
@@ -13,7 +13,7 @@ If you haven't already, be sure to check out our [introduction](./intro.md) to W
 ## Before you begin
 
 We will provide a complete walkthrough of getting Flux installed and Weave GitOps configured. However, if you have:
-- an existing cluster bootstrapped Flux 🎉
+- an existing cluster bootstrapped Flux version >= 0.31.0 🎉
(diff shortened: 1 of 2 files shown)


stefanprodan (approved):
LGTM

Thanks @makkes 🏅

sympatheticmoose:
@makkes a couple of points:
- doesn't this change bump the minimum required version of Flux to `0.31.0`? Which is fine - we should be recommending people upgrade, and we have previous versions of the docs to fall back on as well. But we should make this clear in https://docs.gitops.weave.works/docs/installation/#install-flux 
- given the way our docs work, this change will only be visible when we next cut a release. I recommend also making this change against `/website/versioned_docs/version-0.9.1`

ozamosi:
Changing the minimum system requirements seems like the kind of change we should actually hold off publishing until we cut a release.

chanwit (approved):
LGTM thank you @makkes

makkes:
Converting this to draft to buy us some more time to discuss this change and prevent accidental merging.

makkes:
Back in the game after verifying that the changes to the docs do work as expected. I'll leave this open for now so we can decide on the time to merge this given @ozamosi's and @sympatheticmoose's comments above.

sympatheticmoose:
@makkes I'm fine with Robin's suggestion of it waiting till the next release. But would still prefer that we explicitly call this new version requirement out otherwise folks may hit unexpected errors trying to follow our install/getting started guides.

sympatheticmoose:
There's also the slack thread on updating intervals etc :)

makkes:
@sympatheticmoose @ozamosi I added info on Flux version requirements with regards to OCI. The installation instructions still say we've tested WEGO against older versions. Not sure what to do about that as that's still correct but in the following paragraph we say Flux 0.31.0 is required which might confuse users.

website/docs/getting-started.mdx
```diff
@@ -121,7 +127,7 @@ cd fleet-infra
 
 ```
 flux create source helm ww-gitops \
-  --url=https://helm.gitops.weave.works \
+  --url=oci://ghcr.io/weaveworks/charts \
```
stefanprodan, line 130:
```suggestion
  --url=oci://ghcr.io/weaveworks/charts \
  --interval=60m0s \
```
Same interval for the HelmRelease please

sympatheticmoose: should we change intervals in a different PR against live docs? wdyt @ozamosi ?

makkes: I created a follow-up PR and went through all intervals I could find: https://github.com/weaveworks/weave-gitops/pull/2513

PR #2513, opened 2022-07-29
increase Flux resource intervals

makkes:
**What changed?**

All reconciliation intervals mentioned in the docs for Flux-managed resources are increased.

**Why was this change made?**

The previous intervals were much too low and did nothing really
outside of putting load on the cluster.

### DIFF
--- a/website/docs/getting-started.mdx
+++ b/website/docs/getting-started.mdx
@@ -121,6 +121,7 @@ cd fleet-infra
 
 ```
 flux create source helm ww-gitops \
+  --interval=60m \
   --url=https://helm.gitops.weave.works \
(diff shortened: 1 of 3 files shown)


website/docs/getting-started.mdx
```diff
@@ -149,6 +150,8 @@ git push
 
 ```
 flux create helmrelease ww-gitops \
+  --interval=50m \
+  --chart-interval=65m \
```
stefanprodan, line 154: we can keep only interval here set to 60m.

stefanprodan (approved):

makkes:
@ozamosi can we get this merged, please? 🙏🏻

PR #2515, opened 2022-07-29
Remove enterprise code from core

ozamosi:
This removes enterprise commands that didn't work in core, as well as APIs that weren't available.

Outstanding work:
 * Document how to get the enterprise commands back.


makkes:
This PR is stale, closing.

PR #2516, opened 2022-08-01
docs: Use new algolia to re-enable search

ozamosi:
We didn't follow this [migration](https://docusaurus.io/blog/2021/11/21/algolia-docsearch-migration) when it happened, so our search has been completely broken for a bit.

This seems to fix it.

### DIFF
--- a/website/docusaurus.config.js
+++ b/website/docusaurus.config.js
@@ -81,8 +81,8 @@ module.exports = {
       copyright: `Copyright © ${new Date().getFullYear()} Weaveworks`,
     },
     algolia: {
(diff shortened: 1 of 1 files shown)


ozamosi:
The staging version doesn't work right, which I think is because it's not querying from the right URL - it is now connecting and finding _some_ documentation.

yiannistri (approved):

yiannistri:
I was able to get some results from http://staging.docs.gitops.weave.works/

opudrovs (approved):

PR #2517, opened 2022-08-01
expose errors coming from sources API calls

ahussein3:
- Dependency to resolve #[1069](https://github.com/weaveworks/weave-gitops-enterprise/issues/1069)


ui/hooks/sources.ts
```diff
@@ -33,25 +34,33 @@ export function useListSources(
...
-          })),
-          ..._.map(charts, (ch) => ({
-            ...ch,
-            kind: FluxObjectKind.KindHelmChart,
-          })),
+        const ErrorList: ListError[] = [
+          ...repoRes.errors,
+          ...helmReleases.errors,
+          ...bucketsRes.errors,
+          ...chartRes.errors,
```
foot, line 41: It would be cool to include `kind` here just in case we do want to use that in the error messages.

```
const ErrorList: { [kind: string]: ListError } = {
  [FluxObjectKind.KindGitRepository]: repoRes.errors
  ...
};
```

Or merging it into the ListError somehow wdyt?

ui/hooks/sources.ts
```diff
@@ -33,25 +34,33 @@ export function useListSources(
...
+            ..._.map(buckets, (b) => ({
+              ...b,
+              kind: FluxObjectKind.KindBucket,
+            })),
+            ..._.map(charts, (ch) => ({
+              ...ch,
+              kind: FluxObjectKind.KindHelmChart,
+            })),
+          ],
+          errors: [..._.uniqBy(ErrorList, "clusterName")],
```
foot, line 62: Maybe do the uniqBy in the rendering component too just so we're not losing info at this point. (number of errors etc).

foot: Then we can also pretend clusterName doesn't exist as an attribute too. Maybe the core error renderer doesn't even look at the `clusterName` key and just print out the `message` attr.

ui/hooks/sources.ts
ahussein3: I can merge but we will lose the type of `ListError `that we are using. will go with your suggestion.

ahussein3: I have a better idea I'm thinking about merging as we merge the response will refactor that part and make you rereview

ui/hooks/automations.ts
```diff
@@ -39,17 +40,28 @@ export function useListAutomations(namespace = NoNamespace) {
...
+            })),
+            ..._.map(helmReleases, (h) => ({
+              ...h,
+              kind: FluxObjectKind.KindHelmRelease,
+            })),
+          ],
+          errors: [
+            ..._.map(kustRes.errors, (e) => ({
+              ...e,
+              kind: FluxObjectKind.KindKustomization,
```
foot, line 57: We might need to do a new type here, above there is a

```
export type Automation = Kustomization & HelmRelease & { kind: FluxObjectKind };
```

So something similar for the `ListError` (in some util / types file I guess)? Otherwise trying to read err.kind might give a type error.

chanwit:
Hi @ahussein3. Thank you for this PR !!

Do you have some examples of the errors exposed by this PR, please?
Just to make sure that we don't show any sensitive information by accident.

cheers!

ahussein3:
Hi @chanwit,
Here is an example of error being sent back from API calls `clusterName: "default/demo120-broke"
message: "failed adding cluster client to pool: failed to create leaf client: Get \"https://172.18.0.123:6443/api?timeout=32s\": dial tcp 172.18.0.123:6443: connect: no route to host"
namespace: ""`
- We are not currently showing those Errors here in Core we are just Exposing the full response back so that EE can make usage of those errors.

ui/pages/v2/Automations.tsx
```diff
@@ -10,10 +10,9 @@ type Props = {
 
 function Automations({ className }: Props) {
   const { data: automations, error, isLoading } = useListAutomations();
-
   return (
     <Page error={error} loading={isLoading} className={className}>
```
foot, line 14: this should maybe still be something like?
```suggestion
    <Page error={error || automations?.errors} loading={isLoading} className={className}>
```

ui/pages/v2/Automations.tsx
```diff
@@ -10,10 +10,9 @@ type Props = {
 
 function Automations({ className }: Props) {
   const { data: automations, error, isLoading } = useListAutomations();
```
foot, line 12:
```suggestion
  const { data: automations, error: requestError, isLoading } = useListAutomations();
```

To differentiate the diff types of errors we're juggling now too?

ahussein3: resolved the type here
 https://github.com/weaveworks/weave-gitops/blob/c6436dc5ec747b1fec208a6e7937b0341bc1bf45/ui/components/Page.tsx#L19

ui/lib/types.ts
```diff
@@ -55,3 +55,10 @@ export interface Syncable {
   namespace?: string;
   clusterName?: string;
 }
+
+export interface MultiRequestError {
+  kind?: FluxObjectKind;
+  clusterName?: string;
+  namespace?: string;
+  message?: string;
+}
```
foot, line 64: Could we implement this as `ListError & { kind: FluxObjectKind; }` just to keep it sync'd up w/ the pb.ts files in case the BE response changes?

foot (approved):
Nice ⭐

PR #2518, opened 2022-08-01
Use getObject for all sources details

ozamosi:
The `getObject` endpoint already contains everything we need to display any details page - at least for source kinds (there's a complication related to helmreleases and secrets, but that's a problem we can solve when we get to it). So we can just unpack that response, and fill in all the data.

This commit simply ports some of the `core/server/types` code to the frontend. It uses getters in order to allow the display code to behave as if it's still operating on a plain object similar to the GRPC one, but all the fields are lazy-loaded so we're only paying for the fields that we use.

I've left the server-side code alone, and I've left every page other than the source details pages alone - this ports a significant chunk of the types code, but it's definitely not all of it. However, my goal is to apply this approach to all pages.

The reason I started with the sources pages is that they had no `get` endpoints, so the code had to fetch every source and then filter in the frontend. We already had to load the `getObject` endpoint, so as a result of this little architectural experiment the UI should end up a little bit snappier.

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
   "scripts": {
     "build": "parcel build --target default",
     "build:lib": "parcel build --target lib",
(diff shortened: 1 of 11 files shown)


opudrovs (approved):
Cool! Tested it manually, LGTM. 🎉

PR #2519, opened 2022-08-02
docs: Removes the word "test" from the example WGE ClusterRoleBinding

foot:
- After some feedback from users who were unsure if they needed it /
  should use it.
- Also clean up a lot of unused assets accidentally copy-pasted over
  when we introduced the policy user-guide docs


bigkevmcd (approved):
💯 for deletions!

ozamosi (approved):

PR #2521, opened 2022-08-02
Only release helm chart when the Chart version has changed

### LINKED ISSUE #2180: Publish helm charts only when the chart version is changed, rather than every time the chart is changed
Our current GH action https://github.com/weaveworks/weave-gitops/blob/main/.github/workflows/chart.yaml publishes a new version of the Helm Chart to our Helm Repository after a merge on main to the charts directory. 

Issues can occur when there is a dependency on updates between the server and chart, as one will be published before the other and can result in breakages such as #2171.

Publishing the helm chart should not be absolutely tied to the overall release process - it's possible to upgrade and downgrade the image using the same chart, which is a simple way to try out RCs - use the latest released chart (which is shipped with the most recent stable release), and override the image to the latest version. Therefore, we must be able to make a chart which is compatible with both the current and next release.

At the same time, we shouldn't release-on-merge the chart every single time - we should batch up changes and release them in a orderly and controlled manner.

**AC**
* When the helm chart is changed, it is not automatically released
* When the helm chart version is changed, it is automatically released
* As a developer, when I make a PR that will release the chart, I receive a warning telling me that/what I'm about to release.

ozamosi:
Before this, every time you changed `charts`, you had to bump the
chart version. Every time you bumped the chart version, the chart was
released. Thus, we effectively had deploy-on-merge for the chart,
straight to the users' systems.

This changes it so that you can choose whether you want to bump the
version or not - if you don't change it, it won't be released. The
action also posts a comment warning you that your merge is about to be
released as well, as it can be surprising.

This fixes #2180

### DIFF
--- a/.github/workflows/chart.yaml
+++ b/.github/workflows/chart.yaml
@@ -22,34 +22,63 @@ permissions:
   packages: write
 
 jobs:
-  helm-verify:
+  helm-new-version:
     runs-on: ubuntu-latest
-    if: github.event_name == 'pull_request'
+    outputs:
+      old-version: ${{ steps.old-version.outputs.version }}
+      new-version: ${{ steps.new-version.outputs.version }}
     steps:
       - uses: actions/checkout@v3
         with:
           fetch-depth: 0
       - name: Find new version
-        id: new_version
+        id: new-version
         run: |
           NEW_VERSION=$(yq e '.version' charts/gitops-server/Chart.yaml)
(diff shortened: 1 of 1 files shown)


ozamosi:
I did some testing in #2503 - as you can see, it'll nag every push, but I think that's going to be acceptable.

opudrovs (approved):

PR #2522, opened 2022-08-02
Make authentication method selection explicit

SamLR:
**What changed?**

Make authentication method selection explicit.

**Why was this change made?**

In general (and especially around security) explicit is to be preferred to implicit. This makes the selection of which authentication method is to be used explicit. The methods are chosen by setting the `--auth-methods` flag (e.g. `--auth-methods oidc,user-account`) to preserve backwards compatibility oidc and user-auth are the defaults but this could be further tightened. This also means that, e.g. `passthrough` auth (which is not supported outside of enterprise) can be disabled. This also means that, if further authentication methods are added, they can be made opt-in reducing the potential attach surface.

This also means that operators don't have to be so wary of potential gotchas around other configuration (e.g. it's not unreasonable to think that an operator may have another secret called `oidc-auth` or `cluster-user-auth`).

The change was implemented by adding a new type `AuthMethod` which enumerates the available options. Then, at start-up, only the methods that are implemented are added to the `MultiAuthPrincipal` (which is used by the `WithApiAuth` middleware to authenticate calls). To further harden the system only configuration for the selected methods are loaded, to aid this a new `InitAuthServer` function was added to wrap the common pattern of 'read config -> read secret -> create config object -> create HMAC verifier -> create server'

Unit tests were added and updated as appropriate. The application was also manually tested in oidc-only, user-only and oidc and user modes.

**Release notes**

This introduces the `--auth-methods` CLI flag which users should be aware of.

**Documentation Changes**


cmd/gitops-server/cmd/cmd.go
```diff
@@ -78,28 +84,29 @@ func NewCommand() *cobra.Command {
...
-	cmd.Flags().StringVar(&options.Port, "port", server.DefaultPort, "UI port")
-	cmd.Flags().StringVar(&options.Path, "path", "", "Path url")
+	cmd.Flags().StringVar(&options.LogLevel, "log-level", logger.DefaultLogLevel, "log level")
 	cmd.Flags().StringVar(&options.NotificationControllerAddress, "notification-controller-address", "", "the address of the notification-controller running in the cluster")
-
-	cmd.Flags().StringVar(&options.TLSCertFile, "tls-cert-file", "", "filename for the TLS certificate, in-memory generated if omitted")
-	cmd.Flags().StringVar(&options.TLSKeyFile, "tls-private-key-file", "", "filename for the TLS key, in-memory generated if omitted")
+	cmd.Flags().StringVar(&options.Path, "path", "", "Path url")
+	cmd.Flags().StringVar(&options.Port, "port", server.DefaultPort, "UI port")
+	cmd.Flags().StringSliceVar(&options.AuthMethods, "auth-methods", auth.DefaultAuthMethodStrings(), "Which auth methods to use, valid values are 'oidc' and 'user-account'")
```
ozamosi, line 93: I kinda feel this should be set through a set of env vars rather than through one command line arg. Main reason being, it composes easily, and doesn't require any changes to the CLI interface.

SamLR: I think in general I agree but I think that should be part of a piece of work to move most of the configuration to envars rather than our current CLI heavy system. I can add it for these if you want but it does feel a little out of scope?

ozamosi: I didn't mean to add env vars to these command line args - I meant to remove the command line args. `NewAuthServerConfig` could call a function that does `os.GetEnv` three times, and thus avoid having to change the API as much.

ozamosi (approved):
Could you please give the PR a more descriptive title before merging? :pray:

PR #2523, opened 2022-08-02
docs: Delivery UI permissions

yiannistri:
**What changed?**
Documentation

**Why was this change made?**
To explain how to setup permissions for Delivery UI

N/A

N/A

**Release notes**
N/A

**Documentation Changes**
N/A


sympatheticmoose:
@SamLR - if you have a chance to take a look at this PR that would be great - in particular, now that we the user permissions doc as well, whether you agree this is the right place for this content.

website/docs/configuration/recommended-rbac-configuration.mdx
```diff
@@ -178,3 +180,142 @@ roleRef:
   apiGroup: rbac.authorization.k8s.io
 ---
 ```
+
+
+## Weave GitOps Enterprise RBAC considerations  <TierLabel tiers="enterprise" />
+
+Weave GitOps Enterprise extends Weave GitOps permissions by adding more roles. These roles may need to be extended further in order to support certain use cases. Some of the most common use cases are described below.
+
+### Viewing Flagger-generated resources
```
sympatheticmoose, line 189: Given we've pitched these as "use cases" - suggest rewording as "Progressive delivery with Flagger"

website/docs/configuration/recommended-rbac-configuration.mdx
```diff
@@ -178,3 +180,142 @@ roleRef:
...
+
+Weave GitOps Enterprise integrates with Flagger in order to provide a view on progressive delivery deployments. Part of that view, is the ability to view all the resources that Flagger manages during its operation. The default ClusterRole `gitops-canaries-reader`, used for viewing canary details, includes the minimum permissions necessary for a user to be able to view canary object details, metric template object details and canary related events. When a canary relies on a service mesh such as Linkerd or Istio for the rollout, then this ClusterRole needs to be extended so that it can read the additional service mesh resources being generated by Flagger. Note that this functionality relies on each canary resource explicitly specifying the provider that it uses. The following table provides a list of all the custom resources that Flagger generates grouped by provider:
```
sympatheticmoose, line 191: nit: heavy repetition of the word view. Suggest replacing "Part of that view, is" with "This includes"

sympatheticmoose: ", used for viewing canary details" is redundant as explained later.

sympatheticmoose: > When a canary relies on a service mesh such as Linkerd or Istio for the rollout,

Suggestion: When Flagger is configured to integrate with a service mesh such as Linkerd or Istio,

sympatheticmoose: > Note that this functionality relies on each canary resource explicitly specifying the provider that it uses.

Note - currently in order to display service mesh or ingress related resources, we require `spec.provider` to be set in each canary resource.

sympatheticmoose: This is a fairly large block of text, suggest new lines before:
-  "When a canary relies..." 
- "The following table"

website/docs/configuration/recommended-rbac-configuration.mdx
```diff
@@ -178,3 +180,142 @@ roleRef:
...
+  resources:
+  - trafficsplits
+  verbs:
+  - get
+  - list
+```
+
+### Setting up remote cluster permissions
+
+In order to be able to view remote cluster information, such as canaries in a remote cluster, from the management cluster, you need to take into account the following:
```
sympatheticmoose, line 247: > In order to be able to view remote cluster information, such as canaries in a remote cluster, from the management cluster,

Simplify - In order to view canaries in a remote cluster from the management cluster

website/docs/configuration/recommended-rbac-configuration.mdx
```diff
@@ -178,3 +180,142 @@ roleRef:
...
+- The service account that is used to access the remote cluster, needs to be able to list the namespaces and custom resource definitions of its cluster. In addition to those permissions, it also needs to be able to impersonate users and groups.
```
sympatheticmoose, line 248: These points feel half way between two formats:

Option A
- The **service account** used to access the remote cluster - needs to be able to list namespaces and custom resource definitions; and impersonate users and groups.

Option B
- The service account used to access the remote cluster needs to be able to list namespaces and custom resource definitions in the given cluster. It additionally needs to be able to impersonate users and groups.


Either approach would be fine, it just reads a little odd to me currently

website/docs/configuration/recommended-rbac-configuration.mdx
```diff
@@ -178,3 +180,142 @@ roleRef:
...
+    namespace: default
+roleRef:
+  kind: ClusterRole
+  name: user-groups-impersonator
+  apiGroup: rbac.authorization.k8s.io
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: canary-reader
```
sympatheticmoose, line 285: what's the reason for having `gitops-canary-reader` and then `canary-reader`?

website/docs/configuration/recommended-rbac-configuration.mdx
```diff
@@ -178,3 +180,142 @@ roleRef:
...
+  - list
+```
+
+### Setting up remote cluster permissions
+
+In order to be able to view remote cluster information, such as canaries in a remote cluster, from the management cluster, you need to take into account the following:
+- The service account that is used to access the remote cluster, needs to be able to list the namespaces and custom resource definitions of its cluster. In addition to those permissions, it also needs to be able to impersonate users and groups.
+- The user or group that logs in to the management cluster, needs appropriate permissions to certain resources of the remote cluster.
+
+For example, the following manifest ensures that the `wego-admin` user will be able to view canary information from a remote cluster:
```
sympatheticmoose, line 251: Assume this manifest would need to be applied on the remote cluster? If Enterprise has a preferred way the user does this we should advise on the approach

sympatheticmoose (dismissed):
Overall I think this is looking very good - the points I've made are mostly formatting. It would be good to get Sam's take on whether this is indeed the right page of the 3 that we have around this topic, also I would prefer we copy this change over to the live `0.9.1` docs given directly impacts users with the latest release.

SamLR:
The intention of this page wasn't so much in-depth discussion of which permissions were needed to run gitops and why (that's left to the service-account/user permission pages) but to cover a high level view of how to configure OIDC and kubernetes RBAC to work well with gitops.

Frankly revising this page I'm (almost) tempted to delete most of the code block at the end.

With that said I think this should probably be split between either service-account and user pages or, possibly, a dedicated enterprise permissions page (as we should probably have similar documentation for the CAPI & policy features of enterprise)

website/docs/configuration/recommended-rbac-configuration.mdx
yiannistri: Not sure if there is a preferred way, cc @foot for input. This approach was tested with non-CAPI clusters, maybe there's a way to apply it for CAPI clusters.

website/docs/configuration/recommended-rbac-configuration.mdx
yiannistri: Updated 👍

website/docs/configuration/recommended-rbac-configuration.mdx
yiannistri: Updated 👍

website/docs/configuration/recommended-rbac-configuration.mdx
yiannistri: Updated 👍

website/docs/configuration/recommended-rbac-configuration.mdx
yiannistri: I was assuming `gitops-canary-reader` does not exist in remote cluster because WGE may not be installed there. Is that a wrong assumption?

yiannistri (commented):
@sympatheticmoose I've moved this to a separate page and applied your suggestions, can you take another look when you get a chance? 🙏

website/docs/configuration/recommended-rbac-configuration.mdx
yiannistri: Sure 👍

sympatheticmoose:
I can try to review by e.o.d Wednesday, but please don't hold on my account if someone else is able to review beforehand 👍

website/docs/enterprise/user-permissions.mdx
```diff
@@ -0,0 +1,148 @@
...
+  - list
+```
+
+### Setting up remote cluster permissions
+
+In order to view canaries in a remote cluster from the management cluster, you need to consider the following:
+- The service account used to access the remote cluster needs to be able to list namespaces and custom resource definitions in the given cluster. It additionally needs to be able to impersonate users and groups.
+- The user or group that logs in to the management cluster, needs appropriate permissions to certain resources of the remote cluster.
+
+For example, the following manifest ensures that the `wego-admin` user will be able to view canary information from a remote cluster:
```
sympatheticmoose, line 78: Just to be explicit on where this needs to be applied, something like:

For example, apply the following manifest on the remote clusters ensures that the `wego-admin` user will be able to view canary information from within the Weave GitOps UI on the management cluster:

website/docs/enterprise/user-permissions.mdx
```diff
@@ -0,0 +1,148 @@
...
+
+## Progressive delivery with Flagger
+
+Weave GitOps Enterprise integrates with Flagger in order to provide a view on progressive delivery deployments. This includes the ability to view all the resources that Flagger manages during its operation. The default ClusterRole `gitops-canaries-reader` includes the minimum permissions necessary for a user to be able to view canary object details, metric template object details and canary related events. 
+
+When Flagger is configured to integrate with a service mesh such as Linkerd or Istio for the rollout, then this ClusterRole needs to be extended so that it can read the additional service mesh resources being generated by Flagger. Note that currently, in order to display service mesh or ingress related resources, we require `spec.provider` to be set in each canary resource. 
+
+The following table provides a list of all the custom resources that Flagger generates grouped by provider:
+
+| Provider          | API Group                 | Resource        |
```
sympatheticmoose, line 20: non-blocking - but is there a source of truth for this? Whilst useful my only concern is whether it stays current as flagger or the technologies it is compatible with evolve.

sympatheticmoose (approved):
No blockers from me. Would be good for another pair of eyes with particular focus on the exact RBAC being suggested.

sympatheticmoose:
additionally, a suggested update would be to reference this page from here as well  https://docs.gitops.weave.works/docs/configuration/recommended-rbac-configuration/. In particular the note:

> These recommendations are for Weave GitOps Core. A similar system can be used within Enterprise but it would need to be adapted to account for multi-cluster configurations and is beyond the scope of this document.

website/docs/enterprise/user-permissions.mdx
```diff
@@ -0,0 +1,148 @@
...
+    resources: [ "canaries", "metrictemplates"]
+    verbs: [ "get", "list", "watch" ]
+  - apiGroups: [ "helm.toolkit.fluxcd.io" ]
+    resources: [ "helmreleases" ]
+    verbs: [ "get", "list" ]
+  - apiGroups: [ "kustomize.toolkit.fluxcd.io" ]
+    resources: [ "kustomizations" ]
+    verbs: [ "get", "list" ]
+  - apiGroups: [ "source.toolkit.fluxcd.io" ]
+    resources: [ "buckets", "helmcharts", "gitrepositories", "helmrepositories" ]
```
sympatheticmoose, line 133: ocirepositories?

rparmer: 💯

website/docs/enterprise/user-permissions.mdx
```diff
@@ -0,0 +1,148 @@
...
+    verbs: [ "get", "list" ]
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: read-canaries
+subjects:
```
rparmer, line 140: might be worth calling out the need to add additions users/groups to fit the clients needs

website/docs/enterprise/user-permissions.mdx
rparmer: It could be implied with other roles, but better to call it out

website/docs/enterprise/user-permissions.mdx
yiannistri: Not in this table form, this was extracted by looking into Flagger source code i.e. https://github.com/fluxcd/flagger/tree/main/pkg/apis which contains the list of the different integration types used.

website/docs/enterprise/user-permissions.mdx
yiannistri: Added 👍

PR #2531, opened 2022-08-03
Write unit tests for GitOps Run: Flux and GitOps Dashboard installation Part 1

opudrovs:
Partial PR for 2512 (Write unit tests for GitOps Run: Flux and GitOps Dashboard installation).

- Added unit tests for simpler functions. Will cover `InstallFlux` and `InstallDashboard` in the second PR.

Notes:
- Had to pass test config `k8sEnv.Rest` to `GetKubeClient` in the test instead of the actual config created with `kubeConfigArgs.ToRESTConfig()` (which is used when running the `gitops run` command) because I was getting an "error initializing kube client". So, for the actual config I am only checking that no error occurs when calling `kubeConfigArgs.ToRESTConfig()`.

### DIFF
--- a/pkg/run/get_kube_client_test.go
+++ b/pkg/run/get_kube_client_test.go
@@ -1,0 +1,38 @@
+package run_test
+
+import (
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	wego "github.com/weaveworks/weave-gitops/api/v1alpha1"
+	"github.com/weaveworks/weave-gitops/pkg/logger/loggerfakes"
+	"github.com/weaveworks/weave-gitops/pkg/run"
+)
+
+var _ = Describe("GetKubeClient", func() {
+	var fakeLogger *loggerfakes.FakeLogger
+
+	BeforeEach(func() {
+		fakeLogger = &loggerfakes.FakeLogger{}
+	})
+
+	It("returns kube client", func() {
+		kubeConfigArgs := run.GetKubeConfigArgs()
+
+		namespace := wego.DefaultNamespace
+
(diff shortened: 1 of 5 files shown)


pkg/run/get_kube_client_test.go
```diff
@@ -0,0 +1,57 @@
...
+)
+
+var _ = Describe("GetKubeClient", func() {
+	var fakeLogger *loggerfakes.FakeLogger
+
+	BeforeEach(func() {
+		fakeLogger = &loggerfakes.FakeLogger{}
+	})
+
+	It("returns kube client", func() {
```
ozamosi, line 23: This test makes me feel funny, because it's struggling with its scope - it's the only bit of code we have that uses `pflag` (directly, at least) which raises one of my eyebrows, and it specifies the dashboard port which is unrelated to kubeclients.

Does it still pass without the `kubeConfigArgs.AddFlags`? If there's a specific flag you need, can we add just those flags instead of going through all RunCommandFlags?

This suggestion removes what I think is a little bit of a hidden side-effect of this test - it also tests that "our command line arguments don't clash horribly with those understood by kubeconfig", which is valuable but I think that test ought to live in the main command. Perhaps we need a `gitops beta run --dry-run` that sets up kubeclients but doesn't do anything, and we could write a test in `cmd/beta/run` that uses that functionality.

opudrovs: Yes, I just copied all flags from the gitops run command to mock them with pflag in order not to call cmd.Flags() to get them.

For the current test I can remove the flags completely. It should pass.

OK, will remove the flags. I thought of adding a direct test for the run command later just have not looked yet at how to do it.

ozamosi (approved):

PR #2533, opened 2022-08-04
use proper numbered lists

makkes:
With this change, the resulting HTML page will use proper "ol" tags so
it's easier to add/remove an item without taking care for the order
(as all items are just preceded with "1.").

Before:

After:


ozamosi (approved):
Thank you!

PR #2536, opened 2022-08-04
Don't allow login method unless _both_ cli arg _and_ feature flag is set

ozamosi:
We now seem to have 2 systems for specifying allowed login methods,
and they were clashing - if you don't have an issuerURL but you
haven't removed oidc from the legal auth types, the server crashes
with a nil pointer exception.


opudrovs (approved):
Fixes the nil pointer exception server crash.

PR #2538, opened 2022-08-05
Extend automation details with tabs

TheGostKasper:
**Related to** https://github.com/weaveworks/weave-gitops-enterprise/issues/898
- Exports the ` ReconciledObjectsTable &  ReconciliationGraph`  component for use in EE just in case we will re-create the automation-details in EE.
- Adding `customTabs` as a prop to `AutomationDetail`  as we need to add `violations tab` to it 

> For simplicity we chooses to add it as an extra prop alongside exporting all components from Core to unblock EE


TheGostKasper:
@jpellizzari Is there any other components to export ?

jpellizzari (approved):
LGTM 👍

PR #2539, opened 2022-08-05
Install & wait for extra components when specified

### LINKED ISSUE #2423: Extra Components install for Flux
After applying Flux via SSA, the `gitops run` command needs to wait for all Flux controllers to be ready before proceeding to the next steps.

We can do this by polling and checking the ready replica for each Flux deployment object.

ozamosi:
This moves the waiting for components so it only happens when we install flux, rewires the installation of extra components so it works, and then proceeds to wait for the components the user specified.

This fixes #2423


opudrovs (approved):
Tested it locally, works great! 🎉

PR #2542, opened 2022-08-05
Fix local development

ozamosi:
When I removed backwards compatibility in 75fbce6d, I also removed the
thing that made dev-mode work, since we were always overriding the
username with wego-admin in tilt.

This removes the dev-user altogether - it just replaces the username,
but we can just log in with a different username instead if we want -
and reimplements dev-mode to mean "use cookie, disable verification"


opudrovs (approved):

opudrovs:
Tested it locally, everything works fine for me.

PR #2547, opened 2022-08-08
Write unit tests for GitOps Run: Flux and GitOps Dashboard installation Part 2

### LINKED ISSUE #2512: Write unit tests for GitOps Run: Flux and GitOps Dashboard installation

opudrovs:
Closes #2512 

- Moved the `NewManager` function to `apply_manifests` and added passing the mocked `ssa.ResourceManager` to the `Apply` (and related `applySet` and `waitForSet`), `InstallDashboard`, and `InstallFlux` functions to make them more testable.
- Moved generating secret to a separate function (to simplify testing of the `InstallDashboard` function).
- Added very basic tests for the `Apply`, `InstallFlux`, and `InstallDashboard` functions (more like test prototypes for now).
- Added tests for the `MakeValues`, `MakeHelmRepository`, `MakeHelmRelease`, and `GenerateDashboardManifests` functions. I am very open to suggestions on how to improve these tests.
- Split `IsDashboardInstalled` functions to `GetDashboardHelmChart` and `IsDashboardInstalled` functions to be able to test `GetDashboardHelmChart`. Added very basic tests for the `GetDashboardHelmChart` function (not sure if the first test is needed, because it's based on a mocked client response completely).
- Changed test value strings from `"some-` to `"test-` for consistency.

Notes:
- Newly added tests are supposed to be a foundation for testing. I wanted to make the functions easier to test in principle, to test the basic flow, have tests as basic documentation (as @ozamosi said), and if someone is working in this functions in the future and thinks of a useful test to add (to expand testing `InstallDashboard`, for example), it will be easier to do.

- What is the recommended way of dealing with tests with the same names for different functions? Tests are almost the same for a few functions (like `Apply`, `InstallDashboard`, and `InstallFlux`, see `should return an apply all error if the resource manager returns an apply all error`, for example). The basic prototype flow is the same.

### DIFF
--- a/pkg/run/apply_manifests_test.go
+++ b/pkg/run/apply_manifests_test.go
@@ -1,4 +1,4 @@
-package run_test
+package run
 
 import (
 	"context"
@@ -9,14 +9,13 @@ import (
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 	"github.com/weaveworks/weave-gitops/pkg/logger/loggerfakes"
-	"github.com/weaveworks/weave-gitops/pkg/run"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"sigs.k8s.io/cli-utils/pkg/object"
 )
 
 // mock ssa.ResourceManager
 type mockResourceManagerForApply struct {
-	run.ResourceManagerForApply
+	ResourceManagerForApply
 	state stateApply
 }
 
@@ -51,7 +50,7 @@ func (man *mockResourceManagerForApply) WaitForSet(set object.ObjMetadataSet, op
 	}
 }
 
-var _ = Describe("Apply", func() {
+var _ = Describe("apply", func() {
 	var fakeLogger *loggerfakes.FakeLogger
 	var fakeContext context.Context
 	var fakeInstallOptions install.Options
@@ -70,14 +69,14 @@ var _ = Describe("Apply", func() {
 	It("should apply manifests successfully", func() {
 		man := &mockResourceManagerForApply{}
 
-		_, err := run.Apply(fakeLogger, fakeContext, man, fakeManifestsContent)
+		_, err := apply(fakeLogger, fakeContext, man, fakeManifestsContent)
 		Expect(err).NotTo(HaveOccurred())
 	})
 
 	It("should return an apply all error if the resource manager returns an apply all error", func() {
 		man := &mockResourceManagerForApply{state: stateApplyAllReturnErr}
 
-		_, err := run.Apply(fakeLogger, fakeContext, man, fakeManifestsContent)
+		_, err := apply(fakeLogger, fakeContext, man, fakeManifestsContent)
 		Expect(err).To(HaveOccurred())
 		Expect(err.Error()).To(Equal(applyAllErrorMsg))
 	})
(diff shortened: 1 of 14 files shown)


pkg/run/apply_manifests_test.go
```diff
@@ -0,0 +1,92 @@
...
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	"github.com/weaveworks/weave-gitops/pkg/logger/loggerfakes"
+	"github.com/weaveworks/weave-gitops/pkg/run"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"sigs.k8s.io/cli-utils/pkg/object"
+)
+
+// mock ssa.ResourceManager
+type mockResourceManagerForApply struct {
```
ozamosi, line 18: Most of our mocks use counterfeiter - is there a reason it was easier here to write your own mock?

pkg/run/apply_manifests_test.go
```diff
@@ -0,0 +1,92 @@
...
+
+		fakeManifests, err := install.Generate(fakeInstallOptions, "")
+		Expect(err).NotTo(HaveOccurred())
+		fakeManifestsContent = []byte(fakeManifests.Content)
+	})
+
+	It("should apply manifests successfully", func() {
+		man := &mockResourceManagerForApply{}
+
+		_, err := run.Apply(fakeLogger, fakeContext, man, fakeManifestsContent)
```
ozamosi, line 73: We're totally inconsistent on this, and I think we use this pattern more often than not - but we don't have to make `Apply` (and several other examples) public just to be able to test it, we could just call the test package `run` instead of `run_test`, and then the tests have access to private functions, too.

pkg/run/install_dashboard.go
```diff
@@ -32,29 +31,33 @@ const (
...
 		log.Failuref("Error generating secret from password")
-		return err
+		return "", err
 	}
 
 	log.Successf("Secret has been generated:")
-	fmt.Println(string(secret))
 
+	secretStr := string(secret)
+	fmt.Println(secretStr)
```
ozamosi, line 50: Could we please stop printing the secret? It doesn't matter too much in practice, but even data derived from passwords should be kept private.

pkg/run/install_dashboard_test.go
```diff
@@ -38,3 +45,191 @@ var _ = Describe("InstallDashboard", func() {
...
+				"create":       true,
+				"passwordHash": "test-secret",
+				"username":     "admin",
+			},
+		}
+
+		values, err := json.Marshal(valuesMap)
+		Expect(err).NotTo(HaveOccurred())
+
+		helmRelease := &helmv2.HelmRelease{
```
ozamosi, line 207: These tests would probably require less maintenance if they were turned "inside out", so rather than serialising the expected data and comparing it exactly, we could de-serialize the generated data and comparing "some" of it to what we expect. I think that would make it lighter to add more tests so we could test e.g. namespace and other flags, but wouldn't have to change it every time the chart version changed.

ozamosi (approved):

pkg/run/apply_manifests_test.go
opudrovs: Followed @chanwits's example with the mocking client methods, just reworked it a bit for the ResourceManager.

Thanks for your suggestion, will have a look at the counterfeiter.

pkg/run/install_dashboard_test.go
opudrovs: Thanks, good idea, will try it!

pkg/run/install_dashboard.go
opudrovs: Removed printing the secret.

pkg/run/apply_manifests_test.go
opudrovs: Changed all instances of `package run_test` to `package_run`.

Made only required functions public.

PR #2548, opened 2022-08-08
document installing agent with WGE and update agent docs

alichaddad:
Closes

**What changed?**
Fixed values used to set up policy profile in the documentation and documented how to install policy agent with WGE

**Why was this change made?**
To explain how to add the agent on the management cluster in a WGE installation 

Updated documentation

### DIFF
--- a/website/docs/policy/getting-started.mdx
+++ b/website/docs/policy/getting-started.mdx
@@ -45,7 +45,7 @@ You should then configure the `values.yaml`. You can find more about the policy 
 ```
 policySource:
   url: URL of the repo where your policies exist
-  branch: Branch name on the policies repo
+  tag: tag name on the policies repo
   path: Path to the policies dir - or a kustomization.yaml that selects some policies - in the repo
   secretRef (if the repo is private): Name of the K8s secret with private repo credentials (leave empty if the repo is public)
 ```
--- a/website/docs/policy/getting-started.mdx
+++ b/website/docs/policy/getting-started.mdx
@@ -33,6 +33,14 @@ resources:
 Enterprise customers should have access to fork policy library repo into their local repositories. 
 :::
 
+## Install agent on management cluster
+
+The agent comes packaged with the WGE chart. To install it you need to set the following values:
+
+- `values.policy-agent.enabled`: set to true to install the agent with WGE
+- `values.policy-agent.accountId`: organization name, used as identifier
+- `values.policy-agent.clusterId`: unique identifier for the cluster
+
 ## Install Policy Profile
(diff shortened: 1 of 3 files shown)


website/docs/policy/getting-started.mdx
```diff
@@ -45,7 +45,7 @@ You should then configure the `values.yaml`. You can find more about the policy
 ```
```
foot, line 45: Maybe another note here about how to install/enabled the policy agent in the mgmt cluster. (See the WGE installation instructions on how to enable policy).

website/docs/policy/weave-policy-profile.mdx
```diff
@@ -37,13 +37,12 @@ resources: # specifies the path to each required policy
...
 
-```yaml
-policyLibraryConfig:
-  url: ssh://git@github.com/myorg/policy-library # repo url
-  branch: agent-profile # which branch to use
-  kustomizationPath: ./ # directory containing the kustomization file
-  private: true # set to true if a private repo
-  secretRef: policy-library-auth # secret containing cred to authenticate with a private repo
+```
+policySource:
```
foot, line 41: Should this also be namespaced under `policy-agent`?

website/docs/policy/weave-policy-profile.mdx
```diff
@@ -69,9 +68,10 @@ spec:
 To make use of this policy set in the profile:
 
 ```yaml
-config:
-  AGENT_ADMISSION_POLICY_SET: admission-policy-set # name of policy set to be used for admission
-  AGENT_AUDIT_POLICY_SET: audit-policy-set # name of policy set to be used for audit
+policy-agent:
```
foot, line 71: This works in the WGE chart context but perhaps not when configuring the profile for leaf cluster (not-namespaced), but I think thats pretty okay. Could add an [admonition](https://docusaurus.io/docs/markdown-features/admonitions) calling out the difference in usage?

website/docs/policy/weave-policy-profile.mdx
alichaddad: I think it makes more sense for them to be separate, since policy source is its own concept and its values have no bearings on the agent chart at all

website/docs/policy/weave-policy-profile.mdx
alichaddad: This should work in both situations since both WGE chart and the profile depend on the agent chart.

foot: Good point! 🎉

foot (approved):
Nice one! Thanks for the quick fix

PR #2558, opened 2022-08-08
Add support for OCI Repositories

### LINKED ISSUE #2520: Add support for OCI sources
The upcoming release of flux supports OCI sources as a separate source kind (as opposed to the OCI helm repository support that's already there).

We need to support that in gitops.

There's instructions for how to test the support before it's released in https://github.com/fluxcd/source-controller/pull/788 and there's documentation for how it's supposed to work in https://github.com/fluxcd/flux2/tree/main/rfcs/0003-kubernetes-oci

**Acceptance Criteria**
 * There's an endpoint to list OCI sources.
 * OCI sources are included in the sources view.
 * There's a details page for sources that works the same as the other source details pages with similar fields.
 * Where automations reference their sources (links, the graph), OCI sources work the same way other sources work.

ozamosi:
This provides super-basic support for OCI repositories:
 * Upgrades source-controller to latest release to add support for oci repositories
 * They show up in the source list
 * They can be seen as sources for kustomizations
 * You can click on them and get a detail page
 * You can sync them
 * You can sync a kustomization "with sources" and it syncs the oci repository
 * You can suspend them

Version numbers/references would be useful, but are all based on
metadata annotations so I left them out in this.

There's no support for anything that's exciting about OCI
repositories - no signed image verification, no extra annotations,
nothing. Alas, this means they'll show up and look "OK".

This resolves #2520.

Screenshot from 2022-08-08 17-17-00
Screenshot from 2022-08-08 17-17-11
Screenshot from 2022-08-08 17-17-28


opudrovs (approved):
LGTM! 🎉

Issue #2268, opened 2022-06-06
QPR and Burst setting for Kubernetes client


chanwit:
Weave GitOps got many connection errors because the client's configuration does not have a rate limiter.

```
I0603 09:54:37.501060       1 request.go:665] Waited for 1.036691978s due to client-side throttling, not priority and fairness, request: GET:[https://10.60.0.1:443/apis/authentication.k8s.io/v1?timeout=32s](https://10.60.0.1/apis/authentication.k8s.io/v1?timeout=32s)
2022-06-03T13:17:07.063Z	INFO	gitops	middleware/middleware.go:61	request error	{"uri": "/v1/flux_runtime_objects?clusterName=Default", "status": 401}
2022-06-03T13:17:07.260Z	INFO	gitops	middleware/middleware.go:61	request error	{"uri": "/oauth2/userinfo", "status": 400}
I0603 13:17:08.259404       1 request.go:665] Waited for 1.043454555s due to client-side throttling, not priority and fairness, request: GET:[https://10.60.0.1:443/apis/helm.toolkit.fluxcd.io/v2beta1?timeout=32s](https://10.60.0.1/apis/helm.toolkit.fluxcd.io/v2beta1?timeout=32s)
```

We should set QPR and Burst values like the following for our Kube client.
```
	config := ctrl.GetConfigOrDie()
	enabled, err := flowcontrol.IsEnabled(context.Background(), config)
	if err == nil && enabled {
		// A negative QPS and Burst indicates that the client should not have a rate limiter.
		// Ref: https://github.com/kubernetes/kubernetes/blob/v1.24.0/staging/src/k8s.io/client-go/rest/config.go#L354-L364
		config.QPS = -1
		config.Burst = -1
		return config
	}
	config.QPS = opts.QPS
	config.Burst = opts.Burst
	return config
```

chanwit:
A Flux controller uses QPS=50 and Burst=100 at the moment.
While WEGO uses QPS=1000 and Burst=2000.

What would be the good values for applications like WEGO?

JamWils:
This is on hold for Kubernetes upstream library (kube-openapi) to merge a change.  Once that is done this can be reviewed and merged.

chanwit:
The CVE that blocked this card has now been published. It's CVE-2022-1996.
We can continue to work on this card.

PR #2560, opened 2022-08-09
fix: add client timeout and fix namspace update

jmickey:
The function returned from `ClientConfigWithUser` can now include an error. We found through testing that there was a 30 second timeout that was delaying the call to `client.New`, causing requests that timeout to take 38 seconds (30 seconds + `kubeClientTimeout`.

This PR adds error handling that will cancel the client creation early if the call made in `flowcontrol.IsEnabled` fails and returns an error. A 404 to `livez/ping` (added in k8s 1.23) will not cause an error from `flowcontrol.IsEnabled`.


core/clustersmngr/clustersmngr.go
```diff
@@ -72,14 +74,18 @@ type clientsPool struct {
...
 func ClientConfigWithUser(user *auth.UserPrincipal) ClusterClientConfig {
-	return func(cluster Cluster) *rest.Config {
+	return func(cluster Cluster) (*rest.Config, error) {
 		config := &rest.Config{
 			Host:            cluster.Server,
 			TLSClientConfig: cluster.TLSConfig,
 			Timeout:         kubeClientTimeout,
+			Dial: (&net.Dialer{
+				Timeout:   kubeClientDialTimeout,
+				KeepAlive: 15 * time.Second,
```
foot, line 87: Maybe a comment about 15s. Is it the default in client-go?

core/clustersmngr/clustersmngr.go
```diff
@@ -92,20 +98,27 @@ func ClientConfigWithUser(user *auth.UserPrincipal) ClusterClientConfig {
 			}
 		}
 
+		// flowcontrol.IsEnabled makes a request to the K8s API of the cluster stored in the config.
+		// It does a HEAD request to /livez/ping which uses the config.Dial timeout. We can use this
+		// function to error early rather than wait to call client.New.
 		enabled, err := flowcontrol.IsEnabled(context.Background(), config)
```
foot, line 104: Just needed to double check this but:

From https://pkg.go.dev/net/http
> A non-2xx status code doesn't cause an error.

So the behaviour here shouldn't change even for < 1.23 clusters 👍 .

core/clustersmngr/clustersmngr.go
```diff
@@ -92,20 +98,27 @@ func ClientConfigWithUser(user *auth.UserPrincipal) ClusterClientConfig {
...
-		if err == nil && enabled {
+		if err != nil {
+			return nil, err
```
foot, line 106: Maybe wrap w/ a message about flow control?

I know that might make the error in the UI a bit wild but I think swinging back in the direction of overly verbose error reporting wouldn't be the worst.

core/clustersmngr/clustersmngr.go
```diff
@@ -120,7 +133,10 @@ func NewClustersClientsPool(scheme *apiruntime.Scheme) ClientsPool {
 
 // Add adds a cluster client to the clients pool with the given user impersonation
 func (cp *clientsPool) Add(cfg ClusterClientConfig, cluster Cluster) error {
-	config := cfg(cluster)
+	config, err := cfg(cluster)
+	if err != nil {
+		return err
```
foot, line 138: Wrap error w/ message?

foot (approved):
LGTM! 

Flowcontrol intro'd in https://github.com/weaveworks/weave-gitops/issues/2268

@chanwit looks good to you?

core/clustersmngr/clustersmngr.go
foot: > A 404 to livez/ping (added in k8s 1.23) will not cause an error from flowcontrol.IsEnabled.

🤦 . I need to read better.

core/clustersmngr/clustersmngr.go
jmickey: Oh, this should be 30s, which is the default in client-go

core/clustersmngr/clustersmngr.go
```diff
@@ -72,14 +74,19 @@ type clientsPool struct {
...
 
 func ClientConfigWithUser(user *auth.UserPrincipal) ClusterClientConfig {
-	return func(cluster Cluster) *rest.Config {
+	return func(cluster Cluster) (*rest.Config, error) {
 		config := &rest.Config{
 			Host:            cluster.Server,
 			TLSClientConfig: cluster.TLSConfig,
 			Timeout:         kubeClientTimeout,
+			Dial: (&net.Dialer{
+				Timeout: kubeClientDialTimeout,
```
foot, line 86: Presumably this ends up in https://pkg.go.dev/net#DialTimeout

Under the hood I wonder if this is effecting every request? `roundtripper.IsEnabled` doesn't use dial explicitly, its just doing a HEAD etc.

Maybe 3s is a bit low unfortunately?

core/clustersmngr/clustersmngr.go
```diff
@@ -72,14 +74,19 @@ type clientsPool struct {
 	mutex   sync.Mutex
 }
 
-type ClusterClientConfig func(Cluster) *rest.Config
+type ClusterClientConfig func(Cluster) (*rest.Config, error)
```
chanwit, line 77: I spent time trying to understand this change and finally realized that this type is a function.

Could you use this opportunity to rename this type to `ClusterClientConfigFunc`, please?
In Golang we have similar conventions like context's `CancelFunc` and http's `HandlerFunc`, for example.

core/clustersmngr/clustersmngr.go
```diff
@@ -120,7 +134,10 @@ func NewClustersClientsPool(scheme *apiruntime.Scheme) ClientsPool {
 
 // Add adds a cluster client to the clients pool with the given user impersonation
 func (cp *clientsPool) Add(cfg ClusterClientConfig, cluster Cluster) error {
-	config := cfg(cluster)
+	config, err := cfg(cluster)
```
chanwit, line 137: it would be nice to rename the `cfg` parameter to `cfgFunc` or something too.

core/clustersmngr/clustersmngr.go
```diff
@@ -72,14 +74,19 @@ type clientsPool struct {
 	mutex   sync.Mutex
 }
 
-type ClusterClientConfig func(Cluster) *rest.Config
+type ClusterClientConfig func(Cluster) (*rest.Config, error)
 
 func ClientConfigWithUser(user *auth.UserPrincipal) ClusterClientConfig {
-	return func(cluster Cluster) *rest.Config {
+	return func(cluster Cluster) (*rest.Config, error) {
```
chanwit, line 80: It would be nice to have GoDoc for this `ClientConfigWithUser` function to explain what it does.

core/clustersmngr/clustersmngr.go
```diff
@@ -72,14 +74,19 @@ type clientsPool struct {
...
 		config := &rest.Config{
 			Host:            cluster.Server,
 			TLSClientConfig: cluster.TLSConfig,
 			Timeout:         kubeClientTimeout,
+			Dial: (&net.Dialer{
+				Timeout: kubeClientDialTimeout,
+				// KeepAlive is default to 30s within client-go.
+				KeepAlive: 30 * time.Second,
```
chanwit, line 88: would you like to declare a constant for this one as well, please?

core/clustersmngr/factory.go
```diff
@@ -400,3 +407,9 @@ func restConfigFromCluster(cluster Cluster) *rest.Config {
 		Timeout:         kubeClientTimeout,
 	}
 }
+
+func restConfigFromClusterWrapper() ClusterClientConfig {
```
chanwit, line 411: Having a GoDoc for this function would be great too.

chanwit (changes requested):
Thank you so much @jmickey for the PR!

It's my first time reviewing the codes in `clustersmngr` and found that it uses high-order functions in many places.

It would be great for us to use this opportunity to make the code more readable, especially for someone new to it.

For the flowcontrol's part, that looks good to me.

jmickey:
@chanwit I have pushed changes to address your comments.

chanwit (approved):
Thank you @jmickey, LGTM :1st_place_medal:

PR #2570, opened 2022-08-09
Add options parameter for hooks

jpellizzari:
Part of https://github.com/weaveworks/weave-gitops-enterprise/issues/1243

Adds a way for consumers of the Core hooks to control how and when they get run


joshri (approved):
All requests are still polling as usual for me locally. Hooray!

PR #2571, opened 2022-08-09
docs: Aligning creation docs for existing cluster with CAPI

steve-fraser:
**What changed?**
Documentation

**Why was this change made?**

Aligning existing secret creation with CAPI, so the same bootstrap configuration can b e used 

N/A

**Release notes**

**Documentation Changes**


ozamosi:
There is another case of `--from-file=value.yaml` higher up in the file - does that need to be updated too?

steve-fraser:
> There is another case of `--from-file=value.yaml` higher up in the file - does that need to be updated too?

Yes, it does! Thanks for catching that!

ozamosi (approved):

PR #2572, opened 2022-08-10
fix: profiles api json tags can't json.Unmarshal

jmickey:
Closes weaveworks/weave-gitops-enterprise#1234

**What changed?**

Due to an issue with the proto3 generator the `json:""` tags that are generated when a proto field name includes an underscore (e.g. `available_versions`) also includes an underscore. However, the `protobuf:json=` definition does not. This means that when the server marshals the response into JSON the fields are marshalled with snakeCase. 

For an minimal example:

`profiles.proto`
```proto
massage Profile {
    repeated string available_versions;
}
```

Generated `profiles.pb.go`
```go
type Profile struct {
    AvailableVersions []string `protobuf:"bytes,11,rep,name=availableVersions,json=availableVersions,proto3" json:"available_versions,omitempty"`
}
```

This is causing an issue with the CLI which uses `https://github.com/go-resty/resty` to Unmarshal API responses, causing the affected fields to not correctly Unmarhsal into the `GetProfilesResponse` struct. 

This is a known issue with proto3, and no fix is coming. See the following issues:

https://github.com/golang/protobuf/pull/578
https://github.com/golang/protobuf/issues/183
https://github.com/golang/protobuf/issues/1388

Rather than customise the `protobuf:json=` name using `[json_name = "available_version"]`, which would break the UI, it is simpler just to update the field names in the proto definition.


ozamosi (commented):
I assume this breaks compatibility between CLI 0.9.2 and enterprise 0.9.1?

jmickey:
@ozamosi 

> I assume this breaks compatibility between CLI 0.9.2 and enterprise 0.9.1?

The compatibility is already broken, as this issue already exists between CLI 0.9.1 and enterprise 0.9.1

ozamosi (approved):
:facepalm: Of course.

PR #2573, opened 2022-08-10
[UI,BE] Filter resources by tenant

AlinaGoaga:
Closes https://app.zenhub.com/workspaces/team-pesto-6127680175876a0014fc5657/issues/weaveworks/weave-gitops-enterprise/1005

**What changed?**
- Included tenant in responses from `/v1/helmrelease` `/v1/kustomizations` `/v1/buckets` etc (all apps and sources, including requests for individual objects)
- Refactored https://github.com/weaveworks/weave-gitops/pull/2406/files#diff-06dc83544702564c0e9d6c8521c5cd04effb124e4b5bb5f4c691e4daf9a68a36R108-R121 into a helper function in https://github.com/weaveworks/weave-gitops/pull/2406/files#diff-c5bd0de5d6df572f5d3cc9c7529a43a7d79c64fa73b6bc82ecdbef1e4c658eafR42 called SetFromEnv that we can also use in EE
- Added tenant to Sources and Applications as column and filter if the `WEAVE_GITOPS_FEATURE_TENANCY` flag is true. This will be by default false in Core and true in EE. Left True for now in `helm-values-dev.yaml` for testing.


ui/components/AutomationsTable.tsx
```diff
@@ -140,7 +153,17 @@ function AutomationsTable({ className, automations, hideSource }: Props) {
 
   return (
     <URLAddressableTable
-      fields={fields}
+      fields={
+        flags.WEAVE_GITOPS_FEATURE_TENANCY === "true"
```
AlinaGoaga, line 157: I've put this at the end as having something like: 
```
flags.WEAVE_GITOPS_FEATURE_TENANCY === "true" &&  { label: "Tenant", value: "tenant" }
```
in the middle of the list resulted in an empty visible column when the flag is switched off. Any ideas on how to manage better welcomed

AlinaGoaga:
Hello! @foot @jmickey I've tagged you guys even though this is still in draft to get your thoughts so far + I'll also need some help writing tests for the new stuff I've added when anyone is available please :)

ui/components/AutomationsTable.tsx
foot: The magic voodoo might be something like:
```js
[
   { label: "namespace" },
   ...(flags.WEAVE_GITOPS_FEATURE_TENANCY === "true" ? [{ label: "tenant" }] : []),
   { label: "updated at" },
]
```

foot:
Adds a bit of boilerplate innit. I can't think of any other solutions though.

Cool! Will take it for a spin.

AlinaGoaga:
Thank you! I've updated EE also, just waiting to test once the upgrade is done https://github.com/weaveworks/weave-gitops-enterprise/pull/1284

foot:
Did we add the `tenant` field to the detail page too? I forgot what the decision there was though.

Testing it out now,.. 🐌

foot (approved):
⭐

joshri (approved):
UI side looks good! 🌞

AlinaGoaga:
Thank you both! Nope, not added yet for some reason I thought we were still discussing it. Will add it shortly :)

pkg/featureflags/featureflags.go
```diff
@@ -40,3 +49,21 @@ func Get(key string) string {
 func GetFlags() map[string]string {
 	return flags
 }
+
+// SetFromEnv sets the feature flags from the environment variables
+func SetFromEnv(envVars []string) {
```
foot, line 54: @ozamosi any reservations here?

core/server/types/kustomization.go
```diff
@@ -55,6 +55,7 @@ func KustomizationToProto(kustomization *kustomizev1.Kustomization, clusterName
 		Suspended:             kustomization.Spec.Suspend,
 		ClusterName:           clusterName,
 		ApiVersion:            version,
+		Tenant:                tenant,
```
jpellizzari, line 58: We are starting to build up a set of fields that exist on every object. I wonder if it is type for them all to "inherit" some base fields so we don't have to change this on every object.

ui/components/AutomationsTable.tsx
```diff
@@ -22,13 +23,20 @@ type Props = {
...
+
   automations = automations.map((a) => {
     return { ...a, type: removeKind(a.kind) };
   });
+
   const initialFilterState = {
     ...filterConfig(automations, "type"),
     ...filterConfig(automations, "namespace"),
     ...filterConfig(automations, "clusterName"),
+    ...(flags.WEAVE_GITOPS_FEATURE_TENANCY === "true"
```
jpellizzari, line 37: This can be cleaned up by removing the ternary from this block:

```suggestion
if (flags.WEAVE_GITOPS_FEATURE_TENANCY) {
  initialFilterState = {...initialFilterState, ...filterConfig(automations, "tenant") }
}
```
`inifialFilterState` needs to be a `let` instead for this to work.

ui/components/HelmReleaseDetail.tsx
```diff
@@ -47,6 +48,9 @@ function helmChartLink(helmRelease: HelmRelease) {
 }
 
 function HelmReleaseDetail({ helmRelease, className, customTabs }: Props) {
+  const { data } = useFeatureFlags();
```
jpellizzari, line 51: :thinking: This strikes me as a bit odd. Since it appears `data` can be `undefined` here (probably while we are loading), it means things will pop in after the feature flag is loaded.

Not a blocker, but something to think about.

jpellizzari (approved):
No blockers for me. Like I mentioned, we are starting to build up some universal fields that exist on all objects `clusterName`, `tenant`. It would be good to try and use a base class or some equivalent for these, much like how `k8s` objects have embedded structs for common fields: `TypeMeta`, `ObjectMeta`, etc

ozamosi (commented):
This is missing the tenant field on source detail pages. I think it's supposed to be there, too?

> This will be by default false in Core and true in EE

@JamWils asked for this to be default true in core (though, he didn't do so in the ticket, tssk tssk), so please turn it on here, too.

AlinaGoaga:
Yup yup, it's coming. We were still discussing it when I opened the PR - I'm just adding it as we speak ;)

AlinaGoaga:
@ozamosi Just checking -  are there any other points to address before we merge? Ta

ozamosi:
Just to make the feature flag default to true directly in the charts/gitops-server helm chart.

ozamosi:
Not any more, @weave-gitops-bot. Not any more.

ozamosi (approved):

PR #2580, opened 2022-08-10
Publish NPM package each merge

ozamosi:
This publishes a package each time we hit merge, so it's easier to try out unreleased frontend changes. It's pushed to a
`-main` package name to distinguish it from the regular repo.

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -237,6 +237,36 @@ jobs:
         aws s3 cp bin/gitops s3://weave-gitops/gitops-${{matrix.os}}-${{steps.gitsha.outputs.sha}}
         aws s3 cp s3://weave-gitops/gitops-${{matrix.os}}-${{steps.gitsha.outputs.sha}} s3://weave-gitops/gitops-${{matrix.os}}
 
+  ci-publish-js-lib:
+    name: Publish js library
+    runs-on: ubuntu-latest
+    needs: [ci-js]
+    if: github.event_name == 'push'
+    permissions:
+      packages: write
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - uses: actions/setup-node@v2
+        with:
+          node-version: "16.X"
+          registry-url: "https://npm.pkg.github.com"
+          scope: "@weaveworks"
+      - run: npm install
+      - run: make ui-lib
+      - name: Update package version
+        run: |
+          GITOPS_VERSION=$(git describe)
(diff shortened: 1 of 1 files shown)


opudrovs (approved):

.github/workflows/pr.yaml
```diff
@@ -237,6 +237,36 @@ jobs:
...
+        with:
+          node-version: "16.X"
+          registry-url: "https://npm.pkg.github.com"
+          scope: "@weaveworks"
+      - run: npm install
+      - run: make ui-lib
+      - name: Update package version
+        run: |
+          GITOPS_VERSION=$(git describe)
+          jq '.version = "'$GITOPS_VERSION'" | .name = "@weaveworks/weave-gitops-main"' < dist/package.json > dist/package-new.json
```
foot, line 263: Neato bandito, this will change the package name I think and we might have to alias in our build system (vs. changing all the imports from `import { foo } from @weaveworks/weave-gitops` -> `import { foo } from @weaveworks/weave-gitops-main`? Will do a quick test now..

ozamosi: Oh, right. Didn't think about that :disappointed: 

If you want to test, there's a renamed package here: https://github.com/ozamosi/weave-gitops/packages/1590415

foot: This syntax seems to work

```
    "@weaveworks/weave-gitops-foo": "npm:@weaveworks/weave-gitops@v0.9.1-rc.1",
```

Will test it out with your repo. 🔌

foot: I heard you like `@`s..

```console
 yarn add @weaveworks/weave-gitops@npm:@ozamosi/weave-gitops-main@0.5.0-rc0.4.1-1708-g3568a017
```

ozamosi:
To spell it out rather than keep it in my head: part of the reason I created a separate package is that there is a max limit of releases in one github npm repo (~300 I think?), and when that happens, I expect we can just delete all old releases in this package. If it was in the main package I would not want to delete stable releases.

.github/workflows/pr.yaml
foot: It seems to work! 🎉 , things are a bit out of sync right now so I'll just try and patch things up the interfaces and double check it all builds now..

foot (approved):
Amazing 💯 , very excited for this!

.github/workflows/pr.yaml
```diff
@@ -237,6 +237,36 @@ jobs:
...
+          mv dist/package-new.json dist/package.json
+          cp .npmrc dist
```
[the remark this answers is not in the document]
ozamosi: I'm not sure why it works in releases, but the second I don't do that copy I get errors from registry.npmjs.org, but when I add this it suddenly starts to talk to npm.pkg.github.com and therefore works.

PR #2584, opened 2022-08-11
Make OCI repositories opt-in, fix OCI repository permissions, fix rendering errors for sources

ozamosi:
This fixes a whole bunch of errors:
 * Adds question marks to the source detail page. I reset my cluster and was swamped in errors, and went through them all.
 * Fix the rendering of errors on the source detail page. This was broken by a recent commit that got the parenthesis backwards.
 * Fix merge error in permissions for OCI repositories - they would only have worked if gitops was installed inside the flux-system namespace.
 * Add a feature flag that is turned off by default, that unless enabled makes the list OCI repositories endpoint just return nothing. This stops it returning errors when flux is < 0.32 - since this is not released yet, I've opted to make it a manual opt in/out, in case something changes that would break this completely.

### DIFF
--- a/charts/gitops-server/templates/admin-user-roles.yaml
+++ b/charts/gitops-server/templates/admin-user-roles.yaml
@@ -28,7 +28,7 @@ rules:
     resources: ["events"]
     verbs: ["get", "watch", "list"]
   - apiGroups: ["source.toolkit.fluxcd.io"]
-    resources: [ "buckets", "helmcharts", "gitrepositories", "helmrepositories" ]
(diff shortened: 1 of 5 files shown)


opudrovs:
I still see this on the Sources page in this branch:

Screenshot 2022-08-11 at 15 38 42

opudrovs:
Or is this PR not related to that error on the Sources page?

ozamosi:
I tested with the flag on and off with both new and old flux - and then I committed with it set to on :see_no_evil: Fixed.

opudrovs (approved):
Tested it locally, LGTM! 🎉

Issue #2561, opened 2022-08-09
docs: search result links don't work consistently


sympatheticmoose:
For example, searching for `support@weave.works` to update potential references, and clicking to see all results: 
[https://docs.gitops.weave.works/search/?q=support%40weave.works](https://docs.gitops.weave.works/search/?q=support%40weave.works)

For me returns 6 results, but there seems to be a strange duplication going on and the duplicate is an invalid location
- https://docs.gitops.weave.works/docs/next/help-and-support/#weaveworks-materials ✅ 
- https://docs/next/help-and-support/#weaveworks-materials ❌ 
- https://docs.gitops.weave.works/docs/next/help-and-support/#community ✅ 
- https://docs/next/help-and-support/#community ❌ 
- https://docs.gitops.weave.works/docs/next/help-and-support/#commercial-support ✅ 
- https://docs/next/help-and-support/#commercial-support ❌

ozamosi:
Fixed in the referenced PR

PR #2589, opened 2022-08-12
website: Remove extra slash from website URLs

ozamosi:
Algolia has indexed our whole docs website twice, and is returning all
search results twice (#2561). I suspect this is related to how the
generated source code contains an incorrect double-slash:

```
<link data-rh="true" rel="canonical" href="https://docs.gitops.weave.works//docs/installation/">
```

This change appears to make the double-slash go away but still leave
links working, which should fix the problem. Maybe.


opudrovs (approved):
Interesting bug!

ozamosi:
Staging docs site still seems to work and have the right URLs.

PR #2590, opened 2022-08-12
Raise an error if an empty `UserPrincipal` is passed to `ClientConfigWithUser`

SamLR:
**What changed?**

Raise an error if an empty `UserPrincipal` is passed to `ClientConfigWithUser`. 

**Why was this change made?**

Most callers of `ClientConfigWithUser` check that the user principal they're about to pass isn't `nil` but this doesn't protect against the case when both the `Token` and `ID` are empty strings. If this happens then any calls to the Kubernetes API are made with the application's service account rather than the (incorrectly configured) user.

This is (unlikely) to pose a significant risk as other layers ensure that a principal is set correctly but this provides strength in depth against potential miss-configuration or future extensions.

A check for empty strings.

Extended the unit tests and running locally

### DIFF
--- a/core/clustersmngr/clustersmngr.go
+++ b/core/clustersmngr/clustersmngr.go
@@ -90,7 +90,9 @@ func ClientConfigWithUser(user *auth.UserPrincipal) ClusterClientConfigFunc {
 			}).DialContext,
 		}
 
-		if user.Token != "" {
+		if user.Token == "" && user.ID == "" {
+			return nil, fmt.Errorf("No user ID or Token found in UserPrincipal.")
+		} else if user.Token != "" {
 			config.BearerToken = user.Token
 		} else {
 			config.BearerToken = cluster.BearerToken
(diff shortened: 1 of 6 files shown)


core/clustersmngr/clustersmngr.go
```diff
@@ -90,7 +90,9 @@ func ClientConfigWithUser(user *auth.UserPrincipal) ClusterClientConfigFunc {
 			}).DialContext,
 		}
 
-		if user.Token != "" {
+		if user.Token == "" && user.ID == "" {
```
bigkevmcd, line 93: Maybe we could move this to `.Validate()` on the Principal?

SamLR: We could but as far as I'm aware nothing else uses this and it feels like it makes more sense to keep it inline where it's clear what's being checked for and why.

SamLR:
@bigkevmcd are you happy for me to merge this?

bigkevmcd:
> @bigkevmcd are you happy for me to merge this?

Sure, I would prefer `.Valid()` because it makes it more clear what it's doing ("Does it only need to check these fields in this case?"), saves someone else adding the same checking again, and makes it clear where to add other validation items, but I don't feel that this should be a blocker.

SamLR:
Thinking about it I was just being lazy (and this wasn't actually much of a change) so I've added a `valid` method and some tests

bigkevmcd (approved):
Thank you!

ozamosi (approved):

PR #2604, opened 2022-08-15
Return Added and Removed Clusters in Clusters.Set()

sarataha:
Part of: https://github.com/weaveworks/weave-gitops-enterprise/issues/1282

**What changed?**
Updated Set() method to return added and removed clusters.

**Why was this change made?**
To allow `GetProfiles` and `GetProfileValues` to work with any HelmRepository CR and on any cluster, in order to allow a user to select a Chart from a HelmRepository on a leaf cluster.

Added unit tests.

### DIFF
--- a/core/clustersmngr/factory_caches.go
+++ b/core/clustersmngr/factory_caches.go
@@ -9,18 +9,53 @@ import (
 	"github.com/cheshir/ttlcache"
 	"github.com/weaveworks/weave-gitops/pkg/server/auth"
 	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/util/sets"
 )
 
 type Clusters struct {
 	sync.RWMutex
 	clusters []Cluster
+	names    map[string]Cluster
 }
 
-func (c *Clusters) Set(clusters []Cluster) {
+// Set updates Clusters.clusters, and returns the newly added clusters and removed clusters.
+func (c *Clusters) Set(clusters []Cluster) (added, removed []Cluster) {
 	c.Lock()
 	defer c.Unlock()
 
+	currentClustersSet := sets.NewString()
+	for _, cluster := range c.clusters {
+		currentClustersSet.Insert(cluster.Name)
+	}
+
+	newClustersSet := sets.NewString()
+	clusterNames := map[string]Cluster{}
+
+	for _, cluster := range clusters {
+		newClustersSet.Insert(cluster.Name)
+		clusterNames[cluster.Name] = cluster
+	}
+
+	addedNames := newClustersSet.Difference(currentClustersSet)
+	added = appendClusters(clusterNames, addedNames.List())
+
+	removedNames := currentClustersSet.Difference(newClustersSet)
+	removed = appendClusters(c.names, removedNames.List())
+
 	c.clusters = clusters
+	c.names = clusterNames
+
+	return added, removed
+}
+
+func appendClusters(clustersMap map[string]Cluster, names []string) []Cluster {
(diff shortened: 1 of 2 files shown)


core/clustersmngr/factory_caches.go
```diff
@@ -8,19 +8,25 @@ import (
...
 )
 
 type Clusters struct {
 	sync.RWMutex
-	clusters []Cluster
+	clusters sets.New[Cluster]
 }
 
-func (c *Clusters) Set(clusters []Cluster) {
+func (c *Clusters) Set(clusters sets.New[Cluster]) (sets.New[Cluster], sets.New[Cluster]) {
```
foot, line 20: Lets keep the public interface the same for now while exploring this

```suggestion
func (c *Clusters) Set(clustersList []Cluster) ([]Cluster, []Cluster) {
```

And create the `set.New(clustersList)` in here and return `added.List(), removed.List()` etc

core/clustersmngr/factory_caches.go
```diff
@@ -9,18 +9,52 @@ import (
...
 )
 
 type Clusters struct {
 	sync.RWMutex
 	clusters []Cluster
+	names    map[string]Cluster
 }
 
-func (c *Clusters) Set(clusters []Cluster) {
+func (c *Clusters) Set(clusters []Cluster) ([]Cluster, []Cluster) {
```
bigkevmcd, line 21: We can godoc the exported `Set` function.

And, we can name the arguments for this to make it easier to understand...

https://github.com/golang/go/wiki/CodeReviewComments#named-result-parameters

> if a function returns two or three parameters of the same type, or if the meaning of a result isn't clear from context, adding names may be useful in some contexts.

it's returning two `[]Cluster` slices, we can name them `added` and `removed` which would help folks reading it (especially in Godoc) to understand what is returned?

core/clustersmngr/factory_caches_test.go
```diff
@@ -80,3 +81,39 @@ func TestClustersNamespaces(t *testing.T) {
...
+
+	clusters := []clustersmngr.Cluster{cluster1, cluster2}
+
+	added, removed := cs.Set(clusters)
+	if diff := cmp.Diff([]clustersmngr.Cluster{cluster1, cluster2}, added); diff != "" {
+		t.Fatalf("failed to calculate added:\n%s", diff)
+	}
+
+	if diff := cmp.Diff([]clustersmngr.Cluster{}, removed); diff != "" {
+		t.Fatalf("failed to calculate added:\n%s", diff)
```
bigkevmcd, line 98: this should be `removed` ?

core/clustersmngr/factory_caches.go
```diff
@@ -9,18 +9,53 @@ import (
 	"github.com/cheshir/ttlcache"
 	"github.com/weaveworks/weave-gitops/pkg/server/auth"
 	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/util/sets"
 )
 
 type Clusters struct {
 	sync.RWMutex
 	clusters []Cluster
+	names    map[string]Cluster
```
ozamosi, line 18: Does this name contain a namespace? Or would this become funny if both the `data-science` and the `frontend` namespaces defined a cluster `dev`?

sarataha: It's only the name of the cluster, since Cluster object doesn't include a namespace 🤔 but it contains the name and server, could we make use of them for uniqueness?

ozamosi (approved):

PR #2605, opened 2022-08-15
Add Nav Styles

### LINKED ISSUE #2265: [UI] Add hover effect and click behaviour to sidebar
**Short description**
As any Weave Gitops user, I want to highlight links (in this case only for the sidebar, but eventually all links) so that our application feels more responsive.

**Acceptance criteria**
See figma page [here](https://www.figma.com/file/IVHnM9iyeFWpd11evtY8ux/Weave-GitOps?node-id=10544%3A59510) for selected and hover states.

**Design materials**

Clusters [icon](https://fonts.google.com/icons?selected=Material%20Symbols%20Outlined%3Adns%3AFILL%400%3Bwght%40400%3BGRAD%400%3Bopsz%4048)
App [icon](https://fonts.google.com/icons?selected=Material%20Icons%20Outlined%3Asettings_applications%3A)
Flux [icon](https://drive.google.com/file/d/1A-lWhjhrpYFhpy4ZPMXNegpkkVAkmFg1/view?usp=sharing)
Policies [icon](https://fonts.google.com/icons?selected=Material%20Icons%20Outlined%3Apolicy%3A)
Docs [icon](https://fonts.google.com/icons?selected=Material%20Icons%20Outlined%3Aarticle%3A)

image

joshri:
Closes #2265

https://user-images.githubusercontent.com/65822698/184693616-933a3efc-043e-4edc-8c39-526b50ecda57.mov

As I started adding styles to the Nav things started getting REALLY bulky - I decided to separate it out into a new component. Hopefully this makes customization easier in the future


opudrovs:
A sidebar menu item can is highlighted when the mouse pointer is over it, but it does not navigate to links if the mouse is not over the text:

https://user-images.githubusercontent.com/8824708/184703203-87e3a970-7683-4e02-b568-c80afdc17184.mov

It leads to the cursor "flickering" (changing shape from arrow to pointer/hand and back, when moving over  menu item and between menu items, for example, from `Applications` to `Sources` because the `Sources` text is shorter.

Also, I think it's misleading to the user that an item is highlighted but is not clickable if the pointer is not exactly over the text.

joshri:
Mmmmm....I initially couldn't get things going with just the Link component but I think with where I ended up if I add an Icon prop to Link I can get the whole thing clickable. On it.

opudrovs:
Right now links are wrapped around the text and do not occupy the whole box of the items. So, it might require additional fixes.

Just discovered that the menu items are wider than the side bar, is it expected?

Screenshot 2022-08-15 at 21 43 18

joshri:
I've arrived at some sort of fix - allows you to pass a component down to `Link`. Also got rid of that extra width.

ui/components/Flex.tsx
```diff
@@ -28,7 +30,15 @@ const Styled = (component) => styled(component)`
 class Flex extends React.PureComponent<Props> {
   render() {
     const { className, children } = this.props;
```
opudrovs, line 32: Since you are already using a destructuring assignment here, why not to assign destructure all needed props in order not to use `this.props.` later?

```
const { className, children, onMouseEnter, onMouseLeave } = this.props;
```

ui/components/Text.tsx
```diff
@@ -42,6 +42,8 @@ const Text = styled.span<TextProps>`
 
   ${(props) => props.noWrap && "white-space: nowrap;"};
   ${(props) => props.titleHeight && "line-height: 1.75"};
+  //matches MuiIcon
+  transition: color 200ms cubic-bezier(0.4, 0, 0.2, 1) 0ms;
```
opudrovs, line 46: Do we need transitions on ALL instances of the `Text` component? Can we use styled `Text` with transitions only where it's needed instead (nav links or whatever).

Otherwise it can lead to side-effects in the future if we forget about this default transition and I _suppose_  it can affect performance (I _suppose_ , transitions are  calculated under the hood even if not active).

ui/components/Link.tsx
```diff
@@ -12,7 +13,19 @@ type Props = {
...
+};
+
+const SpacedIcon = ({ icon }: { icon: JSX.Element }) => {
+  return (
+    <>
+      {icon}
+      <Spacer padding="xxs" />
+    </>
+  );
 };
```
opudrovs, line 29: For similar components which only return a value and don't assign any local vars/don't use local functions, we could use a simpler return format for arrow functions:

```
const SpacedIcon = ({ icon }: { icon: JSX.Element }) => (
  <>
    {icon}
    <Spacer padding="xxs" />
  </>
);
```

ui/components/Nav.tsx
```diff
@@ -0,0 +1,126 @@
...
+      >
+        {_.map(navItems, (n) => (
+          <Tab
+            navItem={n}
+            key={n.label}
+            label={n.label}
+            value={n.link.value}
+            component={LinkTab}
+          />
+        ))}
```
opudrovs, line 120: Here it is possible to use the longer arrow function return format with a destructuring assignment to avoid using `n.`:

```
{_.map(navItems, (n) => {
          const { label, link } = n;

          return (
            <Tab
              navItem={n}
              key={label}
              label={label}
              value={link.value}
              component={LinkTab}
            />
          );
        })}
```

joshri:
Thanks for the great comments @opudrovs - consider them ADDRESSED!

opudrovs (approved):
LGTM! 🎉 Only fixing the border on top nav item focus is remaining, but it's optional.

Screenshot 2022-08-17 at 00 38 07

PR #2606, opened 2022-08-15
Remove Helm v2 dependency

devmage:
(No issue associated with this work. It was something I noticed while working elsewhere with the project.)

**What changed?**

Weave GitOps is remarked in the [development process](https://github.com/weaveworks/weave-gitops/blob/main/doc/development-process.md) document as requiring Helm v3.

However, two components of the project still depend on Helm v2. This commit updates those two components to use Helm v3 objects, and removes the Helm v2 dependency.

The APIs of `chartutil` and `strvals` as used in GitOps appear to be unchanged between Helm v2 and v3, so hopefully this is a drop-in update.

Also noticed, there are several references to [helm/charts](https://github.com/helm/charts) which should probably also be moved, however that is outside the scope of this work.

**Why was this change made?**

[Helm v2 is deprecated](https://helm.sh/blog/helm-v2-deprecation-timeline/) and should no longer be used.

Find-and-replace, in essence. Again, the two APIs appeared to be drop-in compatible between Helm v2 and v3.

`make gitops` builds cleanly. `make unit-tests` failed in `auth ./pkg/services/auth`, however I can reproduce the failure on `main` as well, so I suspect it's an issue of my setup. The PR build should reveal any lingering issues with this change. There are no new tests added.

**Release notes**

Removes Helm v2 dependency. _Long live Helm v2!_

**Documentation Changes**

As the documentation already states GitOps to be a Helm v3 project, I believe no further docs changes are needed.


devmage:
Opened https://github.com/weaveworks/weave-gitops/issues/2607 about that `helm/charts` bit.

ozamosi:
Thank you for the fix!

Could you just please run `make fmt` against this branch, so the updated import order follows `gofmt`'s import order opinions?

devmage:
> Thank you for the fix!
> 
> Could you just please run `make fmt` against this branch, so the updated import order follows `gofmt`'s import order opinions?

Happily! Please see the latest commit, and thanks.

devmage:
Also, I noticed the `exclude from release notes` tag. It would actually help me for the purpose I had coming in here, if there could be a small message for this change. The Helm v2 package was spotted by a dependency scanner that I use, so being able to clearly point to a release with the package removed would be cool. Ultimately your decision, of course.

ozamosi:
> It would actually help me for the purpose I had coming in here, if there could be a small message for this change. The Helm v2 package was spotted by a dependency scanner that I use, so being able to clearly point to a release with the package removed would be cool.

That is a good point - thank you for the feedback!

ozamosi (approved):

PR #2611, opened 2022-08-16
auth: Revert to the old auth method order

ozamosi:
We used to always try first local user, then passthrough, then
oidc. The OIDC methods have therefore grown accustomed to returning
errors which shuts down the auth flow completely, even when it wasn't strictly an error.

This effectively reverts to the hard-coded order, so that OIDC can be
made to behave in peace and quiet.


foot:
Code looks good, taking it for a quick spin..

foot (approved):

SamLR (approved):
Can we add a comment or something to make it clear that the order is important?

opudrovs (approved):

PR #2616, opened 2022-08-17
ui: Make source details do http requests from the top, too

ozamosi:
We always try to resolve HTTP requests at the page, so that we
can wrap them in the Page component, and that will handle displaying
loading state and errors so we don't have to.

We do this for automation details, and then we pass in the info prop
as a map of label and value.

However, sources was different - instead of an object and a map of
info, we passed down information about the object we wanted, and a
bunch of callbacks.

This makes source pages behave more like automation
pages, and like the table views.

While I'm at it, set an empty object of the right type as initial
    data, and that way we don't need question marks.

### DIFF
--- a/ui/components/BucketDetail.tsx
+++ b/ui/components/BucketDetail.tsx
@@ -11,36 +11,33 @@ import { InfoField } from "./InfoList";
 
 type Props = {
   className?: string;
-  name: string;
-  namespace: string;
-  clusterName: string;
+  bucket: Bucket;
 };
 
-function BucketDetail({ name, namespace, className, clusterName }: Props) {
+function BucketDetail({ className, bucket }: Props) {
   const { data } = useFeatureFlags();
   const flags = data?.flags || {};
+
+  const tenancyInfo: InfoField[] =
+    flags.WEAVE_GITOPS_FEATURE_TENANCY === "true" && bucket.tenant
+      ? [["Tenant", bucket.tenant]]
+      : [];
 
   return (
     <SourceDetail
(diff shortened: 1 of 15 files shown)


opudrovs:
Great, the code looks cleaner and more consistent! 🎉 

Two issues found during testing:

- I ran into an error at a `GitRepository` details page:

Screenshot 2022-08-17 at 23 31 02

Other source details pages I tested (`HelmChart` and `HelmRepository`) work fine, because the `convertGitURLToGitProvider` fucntion (which tries to convert an empty string and it causes the error) is used only on the  `GetRepository` detail page.

It looks like, now `isLoading` from the `useGetObject` hook is `false` initially, unlike the `main`  branch, where it is `true` at first. So, the details page probably "thinks" that the data was loaded already and tried to parse the URL. And because of the error the actual data is never loaded.

- Besides, the `AutomationsTable` on the source details pages is empty for some reason, unlike the one in the `main` branch.

ozamosi:
> It looks like, now `isLoading` from the `useGetObject` hook is false initially, unlike the `main` branch, where it is `true` at first. 

Aaaaah, that explains it! Thank you - I'd actually spotted the problem on error pages (where there's also no object), but I figured it was irrelevant to this change and was going to submit that separately. But no, of course. Yeah.

> The `AutomationsTable` on the source details pages is empty for some reason, unlike the one in the `main` branch.

Ugh, missing `source.` prefix to the comparison.

opudrovs (approved):
LGTM! 🎉

PR #2621, opened 2022-08-18
security documentation added for vulnerability management process

enekofb:
Adding an initial iteration of the security website 

- Added a security page to the website out of vulnerability management piece

- To add public security advisories

- tested locally

Screenshot 2022-08-28 at 20 01 57

**Release notes**
- added security page to the documentation 
- section for cves for oss and enterprise added

**Documentation Changes**


website/docs/security/_category_.json
```diff
@@ -0,0 +1,4 @@
+{
+  "label": "Security",
```
enekofb, line 2: how can we hide categories under security?

ozamosi: So I gave it a go, and it's being annoying. It looks like the only way to do that is to describe the whole navigation by hand.

enekofb:
can we avoid having the page versioned so we don't have  one per release but just a single one?

website/docs/security/_category_.json
enekofb: moved as page, now it looks better

website/docs/security/cve/enterprise/GHSA-378h-hpjp-gxfc.md
```diff
@@ -0,0 +1,39 @@
...
+apiVersion: gitops.weave.works/v1alpha1
+kind: GitopsCluster
+metadata:
+  name: demo-02
+  namespace: default
+  annotations:
+    metadata.weave.works/dashboard.hellothere: "javascript:alert('hello there ' + window.localStorage.getItem('name'));"
+```
+
+The attacker does not need access to the Weave GitOps UI to craft an attack, access to a Git repository that contains resources that are displayed is all that is needed.
```
bigkevmcd: This should also mention modifying K8s resources per https://github.com/weaveworks/weave-gitops/pull/2621/files/23fef237f72a9725ec84ae3a24a14d919b9da9f2#diff-e35735e4a39b31caa23ea95bcf1cb292668e76a8922663baddca716c67382032R30

enekofb: added https://github.com/weaveworks/weave-gitops/pull/2621/commits/d84e0c1c09f18b439bc5f2dbc83265e2b4dab076

bigkevmcd (approved):

enekofb:
thanks @sympatheticmoose  for your input! 

My only concerns are:
- whether it warrants being a header item (given thus far we’ve only provided quick links to install and getting started)
- the UI oddity as mentioned by robin on the issue where security shows in black rather than orange
- whether it is possible to add a link in the left menu items, as currently we have nothing which is only visible in the header

as suggested in slack, 
1. lets merge this
1. lets open a thread to seek next iteration

PR #2622, opened 2022-08-18
implement installer and exec packages for flux cli

### LINKED ISSUE #2592: Implement flux-exec package
To consume Flux CLI properly, we would follow the way used by HashiCorp's `tf-exec` for installing and executing their binaries.

What is HashiCorp's `tf-exec`?
It's a proven Go package responsible for installing and executing a certain version of Terraform binary. It's battle-tested and used by many projects and products.

We would use the design patterns of this package to manage Flux CLI.

### LINKED ISSUE #2610: implement flux-installer package
To support the `flux-exec`, another package, `flux-installer` needs to be implemented.

It's going to 1. download the binary, 2. install the binary to a cache directory, 3. return an executable path for `flux-exec` to use.

chanwit:
This PR implements two packages, Flux's installer and Flux's exec.

Flux installer is a package for installing a specific version of Flux. It downloads a Flux binary to a cached directory. It checksums during the process to ensure that we get the correct binary.

Flux exec is a type-safe wrapper to execute Flux binary. It has an option that implements each flag one by one.

Fixes #2610 
Fixes #2592

### DIFF
--- a/pkg/fluxexec/bootstrap_github_test.go
+++ b/pkg/fluxexec/bootstrap_github_test.go
@@ -1,0 +1,96 @@
+package fluxexec
+
+import (
+	"context"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("bootstrapGitHubCmd", func() {
+	It("should be able to generate correct install commands", func() {
+		By("generating the default command", func() {
+			flux, err := NewFlux(".", "/path/to/flux")
+			Expect(err).To(BeNil())
+
+			initCmd, err := flux.bootstrapGitHubCmd(context.TODO())
+			Expect(err).To(BeNil())
+			Expect(initCmd.Args[1:]).To(Equal([]string{
+				"bootstrap",
+				"github",
+				"--author-name",
+				"Flux",
+				"--branch",
+				"main",
+				"--cluster-domain",
+				"cluster.local",
+				"--components",
+				"source-controller,kustomize-controller,helm-controller,notification-controller",
+				"--log-level",
+				"info",
+				"--network-policy",
+				"--registry",
+				"ghcr.io/fluxcd",
+				"--secret-name",
+				"flux-system",
+				"--ssh-ecdsa-curve",
+				"p384",
+				"--ssh-key-algorithm",
+				"ecdsa",
+				"--ssh-rsa-bits",
+				"2048",
+				"--watch-all-namespaces",
+				"--hostname",
+				"github.com",
+				"--interval",
+				"1m0s",
+				"--private",
+				"--read-write-key",
+				"--reconcile",
+			}))
+		})
+
+		By("generating the command without network-policy, without private, but with token-auth", func() {
+			flux, err := NewFlux(".", "/path/to/flux")
+			Expect(err).To(BeNil())
+
+			initCmd, err := flux.bootstrapGitHubCmd(context.TODO(),
+				WithBootstrapOptions(
+					TokenAuth(true),
+					NetworkPolicy(false)),
+				Private(false))
+			Expect(err).To(BeNil())
+			Expect(initCmd.Args[1:]).To(Equal([]string{
+				"bootstrap",
+				"github",
+				"--author-name",
+				"Flux",
+				"--branch",
+				"main",
+				"--cluster-domain",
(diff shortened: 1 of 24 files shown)


opudrovs:
It would be really nice to have a general PR description of added files (which changes pertain to which functionality), types and functions, how they work and what they do. It's easy to get lost in 2000+ changed lines. 😊

pkg/fluxexec/fluxexec_test.go
```diff
@@ -0,0 +1,32 @@
...
+	"context"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	"github.com/weaveworks/weave-gitops/pkg/fluxinstall"
+	"github.com/weaveworks/weave-gitops/pkg/fluxinstall/src"
+)
+
+var _ = Describe("FluxExec", func() {
+	It("should be able to install Flux into the cluster", func() {
+		By("run flux install", func() {
```
ozamosi, line 14: This is an integration test - it does end to end testing of downloading the binary, running the binary, and observing the side effects of doing that. But it will run as though it's a unit test - this test suite should not do network requests, and shouldn't really depend on the flux binary.

Both fluxinstall and flux.Install are both already tested. If this is meant to test that the binary invocation works, add e.g. a shell script that echoes something known and execute that.

pkg/fluxexec/install_test.go
```diff
@@ -0,0 +1,92 @@
...
+		})
+		By("generating the command to install a wrong controller", func() {
+			path, err := exec.LookPath("flux")
+			Expect(err).To(BeNil())
+
+			flux, err := NewFlux(".", path)
+			Expect(err).To(BeNil())
+
+			initCmd, err := flux.installCmd(context.TODO(), Components("abc-controller"))
+			Expect(err).To(BeNil())
```
ozamosi, line 81: Is this really good behaviour? Neither flux nor gitops will be able to report that it won't do what you (thought you) asked it to do. I always think image automation is enabled with "image-automation-controller" and this wouldn't give me any indication that it won't.

ozamosi (commented):
fluxexec is both large, and seems fairly opinionated. I know it's kinda the functional options pattern-ish, but the way those options interact with commands, and how commands extend each other really takes quite a bit of reading. If I tried to extend it, I worry I would do it wrong, and then it would be weirdly somewhat different.

Could you please add a quick readme on general design principles? Just a bit of, how does it work, where do I need to make changes to add another command, and where do I need to make changes to add another flag to an existing command. Not massive amounts of text, but a bit of "where do I start".

pkg/fluxinstall/installer.go
```diff
@@ -0,0 +1,116 @@
...
+	i.removableSources = make([]src.Removable, 0)
+
+	for _, source := range sources {
+		if s, ok := source.(src.Removable); ok {
+			i.removableSources = append(i.removableSources, s)
+		}
+
+		switch s := source.(type) {
+		case src.Findable:
+			execPath, err := s.Find(ctx)
```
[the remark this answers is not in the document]
chanwit: We cannot do that. This is a problem I'd like to discuss too.
Checksum values are provided only for .tar.gz, but not for the binaries.

pkg/fluxexec/install_test.go
chanwit: The package should be able to filter out unsupported controllers. I'll double check this behavior.

pkg/fluxexec/bootstrap_github_test.go
```diff
@@ -0,0 +1,104 @@
...
+	"os/exec"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("bootstrapGitHubCmd", func() {
+	It("should be able to generate correct install commands", func() {
+		By("generating the default command", func() {
+			path, err := exec.LookPath("flux")
```
ozamosi, line 14: There's some more `exec.LookPath` - could you change the rest, too?

ozamosi (approved):

pkg/fluxexec/bootstrap_github_test.go
chanwit: Ha. Sorry about that @ozamosi 
I removed all of those lookups.

PR #2624, opened 2022-08-18
Passthrough token

bigkevmcd:
Closes: enterprise/1029

**What changed?**
This adds a slightly different OIDC JWT Passthrough, that extracts the token, verifies it by extracting it, but passes it onto the upstream Kubernetes, rather than impersonating

It also makes the Token on the Principal harder to leak (by making it a `*string`)

**Why was this change made?**
A customer has asked for this specific behaviour.

They appreciate that this will not work if they have multiple client secrets across different clusters and apparently they are sharing the same one across all clusters just now.

When OIDC is enabled and the feature flag is enabled, this feature is enabled.

You will need only the `oidc` method enabled, and `WEAVE_GITOPS_FEATURE_OIDC_AUTH_PASSTHROUGH` feature enabled for now.

**Release notes**

**Documentation Changes**


pkg/server/auth/passthrough.go
```diff
@@ -58,5 +57,44 @@ func (pg *BearerTokenPassthroughPrincipalGetter) Principal(r *http.Request) (*Us
...
+	if err == http.ErrNoCookie {
+		return nil, nil
+	}
+
+	principal, err := parseJWTToken(r.Context(), pg.verifier, cookie.Value)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse for passthrough: %w", err)
+	}
+
+	pg.log.V(4).Info("passing through token")
```
ozamosi, line 96: I can never remember which direction the verbosity numbers work in logr. This is a "tracing-ish" verbosity level, right? So it's not supposed to show up?

pkg/server/auth/auth.go
```diff
@@ -119,11 +164,16 @@ func WithAPIAuth(next http.Handler, srv *AuthServer, publicRoutes []string) http
 
 		switch method {
 		case OIDC:
-			if featureflags.Get(FeatureFlagOIDCAuth) == FeatureFlagSet {
+			if srv.oidcEnabled() {
 				// OIDC tokens may be passed by token or cookie
-				headerAuth := NewJWTAuthorizationHeaderPrincipalGetter(srv.Log, srv.verifier())
-				cookieAuth := NewJWTCookiePrincipalGetter(srv.Log, srv.verifier(), IDTokenCookieName)
-				multi.Getters = append(multi.Getters, headerAuth, cookieAuth)
+				multi.Getters = append(multi.Getters, NewJWTAuthorizationHeaderPrincipalGetter(srv.Log, srv.verifier()))
```
ozamosi, line 169: Would you happen to know what this method is for?

pkg/server/auth/passthrough.go
bigkevmcd: Yeah, the numbers reflect "verbosity" for things that accept `-v=0` style command-line options.

I _think_ it's probably derived from older commands that had `-vvv` for "more verbose" as cmd-line options.

And yes, that's basically not supposed to show up, unless you really want it to, because it would be logged on each request.

pkg/server/auth/auth.go
bigkevmcd: @ozamosi the `JWTAuthorizationHeader` one?

ozamosi: Yes

bigkevmcd: https://github.com/weaveworks/weave-gitops/blob/7be4190c671f86936ad87caa3fb788687a18f70c/pkg/server/auth/jwt.go#L48-L50

ozamosi: Excellent - so that's what's used by all kinds of oidc proxy/API gateway style workflows.

What is the reason you chose not to support that for this feature?

bigkevmcd: It does support it, the real difference in this code is that it stores the token on the principal, as well as validating the token as a JWT token.

There's definitely an argument to be made that we could update that to optionally store the token, but this is an odd feature.

ozamosi (approved):

PR #2627, opened 2022-08-18
ui: PageStatus should display Failed when state is bad

ozamosi:
This checked if the return value was truthy or falsy to determine
whether to show a green or a red icon - however, the return value
these days is an enum, so it's always truthy.

Use a good, ol' fashioned switch case, because we need to test every
case it can return.

### DIFF
--- a/ui/components/PageStatus.tsx
+++ b/ui/components/PageStatus.tsx
@@ -1,6 +1,7 @@
 import * as React from "react";
 import styled from "styled-components";
 import { Condition } from "../lib/api/core/types.pb";
+import { colors } from "../typedefs/styled.d";
 import Flex from "./Flex";
 import Icon, { IconType } from "./Icon";
 import { computeMessage, computeReady, ReadyType } from "./KubeStatusIndicator";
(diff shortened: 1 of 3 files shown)


ui/components/PageStatus.tsx
```diff
@@ -17,13 +17,21 @@ function PageStatus({ conditions, suspended, className }: StatusProps) {
...
+      case ReadyType.Reconciling:
+        iconType = IconType.ReconcileIcon;
+        break;
+      case ReadyType.Ready:
+        iconType = IconType.CheckCircleIcon;
+        break;
+      case ReadyType.NotReady:
+        iconType = IconType.FailedIcon;
+        break;
+    }
```
opudrovs, line 33: Usually, `switch` should have the default case to make sure that all possible cases are covered.

ozamosi: Oh. Usually in typed languages, a switch without a default against an enum means the compiler will make sure your matches are exhaustive - I assumed typescript was the same, but it seems I need type level trickery.

opudrovs: Yes, just checked if TypeScript catches it: left only two specific cases without a default case, but nothing was reported by the editor or by lint.

opudrovs (approved):

joshri (approved):

PR #2632, opened 2022-08-19
Populate the groups field in the oauth2/userinfo API response

alexandermarston:
This adds the groups scope from the user principal to the response of the /oauth2/userinfo endpoint.

Currently, the /oauth2/userinfo will return `null` as the value of the groups key. When debugging user permission issues, it's useful to have an endpoint to quickly see what groups have been mapped to the current authenticated user.

The change is minor. I have introduced a new variable called `userPrincipal` which is of type `UserPrincipal`. We can then extract the claims from the user provider into this struct to then populate the response JSON.

The changes have been validated by deploying this to a functioning Kubernetes environment with Flux installed, and testing it with a known-good OIDC provider.

I don't believe this is notable for a release.

**Documentation Changes**
N/A

### DIFF
--- a/pkg/server/auth/server.go
+++ b/pkg/server/auth/server.go
@@ -407,8 +407,17 @@ func (s *AuthServer) UserInfo() http.HandlerFunc {
 			return
 		}
 
+		var userPrincipal UserPrincipal
+
+		// Extract custom claims
+		if err := info.Claims(&userPrincipal); err != nil {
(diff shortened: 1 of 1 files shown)


ozamosi (approved):
Thank you for the PR!

I had a moment's thought about whether returning an error in this case would cause unexpected changes, but we already do that when you log in so the behaviour should be unchanged.

PR #2637, opened 2022-08-19
Update `UpdateClusters` to notify watchers with the added and removed clusters

sarataha:
Part of: https://github.com/weaveworks/weave-gitops-enterprise/issues/1282

**What changed?**
* Updated clientFactory to include a slice of watchers.
* Created a clientFactory method to create/remove watcher.
* Updated `UpdateClusters` to notify watchers with the added and removed clusters.

**Why was this change made?**
This is part of the changes made to allow `GetProfiles` and `GetProfileValues` to work with any HelmRepository CR and on any cluster, in order to allow a user to select a Chart from a HelmRepository on a leaf cluster.

Added unit tests.


core/clustersmngr/factory_caches_test.go
```diff
@@ -117,3 +132,51 @@ func newTestCluster(name, server string) clustersmngr.Cluster {
 		Server: server,
 	}
 }
+
+type ChannelDemo struct {
```
bigkevmcd, line 136: This looks like it can be removed?

core/clustersmngr/factory_caches_test.go
```diff
@@ -83,6 +83,9 @@ func TestClustersNamespaces(t *testing.T) {
 }
 
 func TestClusterSet_Set(t *testing.T) {
```
bigkevmcd, line 85: Feels like there's a test missing here, when there are _no changes_ to the set?

core/clustersmngr/factory.go
```diff
@@ -89,6 +90,25 @@ type clientsFactory struct {
...
+	Removed []Cluster
+}
+
+// ClustersWatcher watches for cluster list updates and notifies the registered clients.
+type ClustersWatcher struct {
+	updates chan ClusterListUpdate
+	cf      *clientsFactory
+}
+
+// Notify updates the clients pool with the new clusters list.
```
bigkevmcd, line 109: How about...

```go
// Notify publishes cluster updates to this subscriber.
```

core/clustersmngr/factory_caches_test.go
```diff
@@ -99,6 +102,18 @@ func TestClusterSet_Set(t *testing.T) {
 		t.Fatalf("failed to calculate removed:\n%s", diff)
 	}
 
+	cw.Stop()
+
+	go func() {
+		for u := range cw.updates {
+			if diff := cmp.Diff(
```
bigkevmcd, line 109: This should probably check for the number of messages received over the channel?

bigkevmcd (commented):
Nothing major, some small changes, and we can improve the testing.

core/clustersmngr/factory_test.go
```diff
@@ -175,3 +175,56 @@ func TestUpdateUsers(t *testing.T) {
...
+
+		g.Expect(notifications).To(HaveLen(2))
+	})
+
+	t.Run("watcher shouldn't be notified when there are no updates", func(t *testing.T) {
+		g.Expect(clientsFactory.UpdateClusters(ctx)).To(Succeed())
+
+		notifications := watcher.Updates
+
+		g.Expect(notifications).To(HaveLen(3))
```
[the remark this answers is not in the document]
sarataha: I was confused about this but then thought it'll always send a notification every time we call UpdateClusters?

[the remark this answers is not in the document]
bigkevmcd: It should only notify when there are changes?

core/clustersmngr/factory_test.go
```diff
@@ -175,3 +175,78 @@ func TestUpdateUsers(t *testing.T) {
...
+
+		g.Expect(clientsFactory.UpdateClusters(ctx)).To(Succeed())
+
+		update := <-watcher.Updates
+
+		g.Expect(clusterNames(update.Added)).To(BeEmpty())
+		g.Expect(clusterNames(update.Removed)).To(Equal(clusterNames([]clustersmngr.Cluster{c2})))
+	})
+
+	t.Run("watcher shouldn't be notified when there are no updates", func(t *testing.T) {
```
foot, line 236: Maybe 1 more test for Unsubscribe?

foot (approved):
Makes sense to me 💯

core/clustersmngr/factory.go
```diff
@@ -104,7 +135,28 @@ func NewClientFactory(fetcher ClusterFetcher, nsChecker nsaccess.Checker, logger
 		initialClustersLoad: make(chan bool),
 		scheme:              scheme,
 		newClustersPool:     clusterPoolFactory,
+		watchers:            []*ClustersWatcher{},
+	}
+}
+
+// Subscribe returns a new ClustersWatcher.
+func (cf *clientsFactory) Subscribe() *ClustersWatcher {
+	cw := &ClustersWatcher{cf: cf, Updates: make(chan ClusterListUpdate, 5)}
```
alichaddad, line 144: It is not clear why this capacity was chosen for the buffer. Perhaps it can be configured?

core/clustersmngr/factory.go
```diff
@@ -104,7 +135,28 @@ func NewClientFactory(fetcher ClusterFetcher, nsChecker nsaccess.Checker, logger
...
+	cf.watchers = append(cf.watchers, cw)
+
+	return cw
+}
+
+// RemoveWatcher removes the given ClustersWatcher from the list of watchers.
+func (cf *clientsFactory) RemoveWatcher(cw *ClustersWatcher) {
+	watchers := []*ClustersWatcher{}
+	for _, w := range cf.watchers {
+		if cw != w {
```
alichaddad, line 154: We are only comparing the channel's value here. Is that enough or we can maybe generate an id for each watcher?

alichaddad (commented):
LGTM, just a couple of comments

core/clustersmngr/factory.go
bigkevmcd: We're comparing the `ClustersWatcher` not the channel.

core/clustersmngr/factory.go
bigkevmcd: Yeah, this could probably be `1` rather than `5`.

core/clustersmngr/factory.go
alichaddad: Comparing structs is done through the struct fields and since `cf` will be the same for all watcher in a cluster factory, that's what we are essentially checking on or am I missing something?

foot: @sarataha do you wanna write maybe one more test :)

Where we add 2 watchers, remove 1, update, make sure the one that is still subscribed gets the update?

foot: > Comparing structs is done through the struct fields

So I think we're comparing pointers here. They only get auto-dereferenced if you use a _selector_, https://go.dev/ref/spec#Selectors

foot:
> subscribing to a factory just feels… incorrect?

Point, like you should be getting an update when something has been constructed?

`ClustersManager` is a bit of a classic.

foot:
> [Rename ClientsFactory to ClustersManager to reflect new changes.](https://github.com/weaveworks/weave-gitops/pull/2637/commits/04a65ced7f5240df790571b0816647faf3ab1bae)

Zounds! Nice one. WDYT @ozamosi 

> Is it a… ClustersPool? ClientPool?

I have heard _but its not really a pool_, thrown around a lot recently in reference to other things so would avoid that one. As this is also not really a pool.

ozamosi (approved):

PR #2638, opened 2022-08-19
GetProfiles and GetProfileValues to accept HelmRepo details

AlinaGoaga:
Closes https://app.zenhub.com/workspaces/team-pesto-6127680175876a0014fc5657/issues/weaveworks/weave-gitops-enterprise/1206


foot (approved):
🔥

api/profiles/profiles.proto
```diff
@@ -80,6 +80,10 @@ message Profile {
 }
 
 message GetProfilesRequest {
+  // The name of the HelmRepository
+  string HelmRepoName = 1; 
+  // The namespace of the HelmRepository
+  string HelmRepoNamespace = 2;
```
ozamosi, line 86: I think you need to make these names lowerCamelCase, or they won't work in the CLI.

@jmickey last fixed these in #2572, which references golang/protobuf#578 which references golang/protobuf#183 which contains a table which suggests UpperCamelCase introduces these incompatibilities again. Is that relevant here?

foot: Wow.

Yes, according to that table it looks like it might be an issue one day. Good catch. Let's lowerCase, will be consistent w/ the rest of the file at least.

ozamosi (approved):

PR #2645, opened 2022-08-22
Make feature flags part of context

ozamosi:
This way they are loaded at page load time, so they're instantly
available at any page. They won't change dynamically, so they don't
need to be refetched. Contexts are "slow", but network requests are
slower.

This should fix optional frontend features jumping out after the rest
of the page has already loaded.

I've implemented this in a "funny" way - I left the API from
    `useFeatureFlags` the same, so it looks a bit silly. It's failing
    silently - if feature flags are broken, you get no feature flags. I
    think I'd rather that happened than a crash.

### DIFF
--- a/ui/components/__tests__/Page.test.tsx
+++ b/ui/components/__tests__/Page.test.tsx
@@ -18,7 +18,7 @@ describe("Page", () => {
           withTheme(
             withContext(
               <CoreClientContext.Provider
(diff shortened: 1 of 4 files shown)


opudrovs (approved):
Cool! 🎉 Tested it, works fine for me.

AlinaGoaga (approved):
Thanks Robin! Tested on Core and EE, all looks good ;)

Issue #2317, opened 2022-06-14
As a user I would like to be able to suspend/resume multiple flux objects


mtparet:
**Short description**
In the gitops way we could have some big changes impacting lot of resources and we like to control the application instead of applying 10000 changes at one time.

Opening a new PR for each change is not always wanted or possible so this should be possible by using gitops tools.

Once the synchronisation is suspended for all resources we could sync again each resource individually or all the resources.

**Acceptance criteria**
~- [ ] _does suspend synchronisation for all resources_~
~- [ ] _does reactivate synchronisation for all resources_~
~- [ ] _does reactivate synchronisation individually for each resource_~

The criteria applies to the application and source pages within Weave GitOps.  This functionality should extend to the enterprise dashboard as well.

- [ ] Given I have no checkboxes selected on the list view, when I take no action, then I should see both buttons "sync" and "suspend" disabled at the top.
- [ ] Given I have no checkboxes selected on the list view, when I hit the checkbox on one or more line items, then I should see both buttons "sync" and "suspend" enabled at the top.
- [ ] Given I have one or more checkboxes selected on the list view, when I uncheck the checkboxes for those line items, then I should see both buttons "sync" and "suspend" disabled at the top.
- [ ] Given I have checkboxes selected, when I hit the "sync" button on the page, then I should see those items run a "reconciliation" in the app.
- [ ] Given I have checkboxes selected, when I hit the "pause" button on the page, then I should see those items turn to suspended state in the app.
- [ ] Given I have checkboxes selected, when I hit the "play" button on the page, then I should see those items turn to green or red depending on their healthy state.
- [ ] Given I have checkboxes selected, when I hit the "play" button on the page and some of those items cannot be resumed, then I should see an error indicated that they were unable to be resumed.
- [ ] Given I have checkboxes selected, when I hit the "pause" button on the page and some of those items cannot be resumed, then I should see an error indicated that they were unable to be suspended.

**Design materials**
- Screenshots

<img width="1007" alt="Screen Shot 2022-07-18 at 9 24 24 PM" src="https://user-images.githubusercontent.com/1825111/179644401-76453185-a021-4f2f-92c5-33beee3920e0.png">

https://www.figma.com/file/IVHnM9iyeFWpd11evtY8ux/Weave-GitOps?node-id=10323%3A59752

- Design diagrams
- WEP 
- PR/FAQ

JamWils:
[...]

PR #2649, opened 2022-08-23
Suspend/Resume All

joshri:
Closes: #2317 

Video showing multi suspend / resume:

https://user-images.githubusercontent.com/65822698/186517860-4a2ba9b9-ecd4-4928-94ed-26418aacf243.mov

Screenshot showing tooltips:
image

Sync all can happen in a follow up PR. This PR got WAY too big as we completely restructured the data tables so checkboxes were easy like they should be instead of passing them down through seventy layers. First step of what I'm sure is a larger data table restructuring.

What's done:

backend -
- suspend.go changed to allow multiple objects to be submitted and checked in one request, potentially returning multiple errors.
- Most objects now have a Uid field so we don't have to deep compare things in the frontend

frontend -
- URLAddressableTable and FilterableTable removed, functionality slapped into DataTable, tests fixed.
- Enterprise imports changed :(
- checkboxes prop added to DataTable, rendering checkboxes for use with new CheckboxActions component
- CheckboxActions component handles play and pause button, receiving array of checked IDs and filtered rows. 
- checkbox and play/pause button styling and tooltips
- handle suspend success/error similar to sync action

🌞


opudrovs:
- Could you please point the latest design to me? On the last screenshots @JamWils provided, the shape of the Play and Pause buttons is closer to square. When I run the app, the buttons are more rectangular:

Screenshot 2022-08-24 at 00 29 30

And do we already need the multi-Sync button there?

- It would be nice to have tooltips on buttons.

- I think we need to show some notifications while this multi-suspend/resume is working.

JamWils:
@opudrovs I like the buttons that way they look in your screenshot.  Lets keep it that way.  I agree on the tooltips as well.  It would be nice to also have the "sync" button there, but that can also be done in a later story.

Tooltip for play button: "Resume selected automations"
Tooltip for pause button: "Suspend selected automations"

opudrovs:
@joshri the tooltips I see right now are "Resume Selected" and "Suspend Selected". Were they edited by @JamWils ? Or should they still be
>Tooltip for play button: "Resume selected automations"
>Tooltip for pause button: "Suspend selected automations"

And what about tooltips on the Sources page? Are they considered "automations" in the context of resume/suspend functionality or should this be changed?

opudrovs:
@ozamosi are Uids added in the current PR compatible with Uids added in the listObjects PR?

https://github.com/weaveworks/weave-gitops/pull/2658

ozamosi:
> @ozamosi are Uids added in the current PR compatible with Uids added in the listObjects PR?

That's the intention - should be same field, same value. I was planning to let this get merged, rebase #2658, and actually test it though.

ui/components/AutomationDetail.tsx
```diff
@@ -49,10 +49,14 @@ function AutomationDetail({ automation, className, info, customTabs }: Props) {
 
   const suspend = useToggleSuspend(
     {
-      name: automation?.name,
-      namespace: automation?.namespace,
-      clusterName: automation?.clusterName,
-      kind: automation?.kind,
+      objects: [
+        {
+          name: automation?.name,
```
opudrovs, line 54: I don't think it's related to the current PR, but question marks should be probably used consistently in the `AutomationDetail` component.

In `useSyncFluxObject` and `useToggleSuspend` here we access `automation` fields with optional chaining. In `useGetObject` above them we just write `automation.name` and so on.

@ozamosi should question marks be added to the `useGetObject` args or does `useGetObject` always return a valid object, even if empty, and in this case should the question marks be removed from `useSyncFluxObject` and `useToggleSuspend` args?

opudrovs: In the `SourceDetail` the `source` prop is not optional.

ozamosi: Question marks should be added, because `AutomationDetail` still deals with responses from `GetKustomization` and `GetHelmRelease`. They can return an error, in which case we'll still render the whole component but with `automation = undefined`, which would be unfortunate.

With `useGetObject`, the return value is always a valid object, so in `SourceDetail` it doesn't need all those question marks. But they're still needed here.

ui/lib/objects.ts
```diff
@@ -1,13 +1,13 @@
...
-  GitRepositoryRef,
   Condition,
   FluxObjectKind,
   FluxObjectRef,
-  Object as ResponseObject,
+  GitRepositoryRef,
   Interval,
+  Object as ResponseObject,
 } from "./api/core/types.pb";
+import { addKind } from "./utils";
```
opudrovs, line 10: Changes in `objects ts` are reported by github a bit weirdly, I hope it will clear out after rebase.

ui/components/FilterDialog.tsx
```diff
@@ -138,8 +138,8 @@ type sectionSelectObject = { [header: string]: boolean };
 function UnstyledFilterDialog({
```
opudrovs, line 138: I know it was not added in this PR, but can we just remove `=== true` on line 55 of `FilterDialog`? Lines 54+55 can be replaced with:

```
compoundKeys.length > 0 && compoundKeys.every((key) => formState[key]);
```

Type of formState is already `{ [inputName: string]: boolean }`

opudrovs: BTW, I'm getting an unrelated warning ` Warning: An update to CoreClientContextProvider inside a test was not wrapped in act(`, can you please re-check the UI tests after rebase?

joshri:
> @joshri the tooltips I see right now are "Resume Selected" and "Suspend Selected". Were they edited by @JamWils ? Or should they still be
> 
> > Tooltip for play button: "Resume selected automations"
> > Tooltip for pause button: "Suspend selected automations"
> 
> And what about tooltips on the Sources page? Are they considered "automations" in the context of resume/suspend functionality or should this be changed for the Sources page?

This is why I did a little edit to James' text - as far as I know these checkboxes should also be on the sources page and it doesn't make sense to say automations there....or does it?

ui/components/FilterDialog.tsx
joshri: That error in the UI tests doesn't give a lot of info about where it's coming from besides that it pops up in the terminal right after the Breadcrumbs tests and before the Footer tests. I ran DataTable and DataTableFilter tests just to make sure it wasn't there, then I went into Breadcrumbs and made a small change in the context to match other tests. It's possible that the `createMockClient({})` was causing the issue because changing it to just `{}` fixes the error. The tests still pass but I might have broken something you were doing so let me know if that's a terrible change 🌞

ui/lib/objects.ts
joshri: Looks like a linting thing - is my formatting on save messed up? Or has this file just not gone through it yet?

core/server/suspend_test.go
```diff
@@ -57,34 +57,60 @@ func TestSuspend_Suspend(t *testing.T) {
...
 
+	requestObjects := []*api.SuspendReqObj{}
+
 	for _, tt := range tests {
 		t.Run(tt.kind.String(), func(t *testing.T) {
 			g := NewGomegaWithT(t)
 			g.Expect(k.Create(ctx, tt.obj)).Should(Succeed())
-
-			req := &api.ToggleSuspendResourceRequest{
+			requestObjects = append(requestObjects, &api.SuspendReqObj{
```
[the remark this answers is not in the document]
joshri: Tests split into a couple different t.Run moments in latest changes

ui/components/FilterDialog.tsx
opudrovs: Looking into it now.

There was another warning

```
      TypeError: api.GetFeatureFlags is not a function
          at Object.FeatureFlags.react_query_1.useQuery.staleTime [as queryFn] (/Volumes/Macintosh_HD_Data/_job_ww/_src/weave-gitops/ui/contexts/CoreClientContext.tsx:51:15)
```
while running the UI tests.
@ozamosi is it OK to add typeof Core to this function `function FeatureFlags(api: typeof Core) {` ?

If I do it, this warning disappears.

opudrovs: Checked it: wrapping everything with `act` and `async` (with switching to `render` instead of `renderer`) does not fix the `GetFeatureFlags` warnings (new warnings appear in other components). And I see the `not wrapped with act` warnings in the `main` branch. It looks like these test issues are not directly related to the current PR. We need to ask @ozamosi on Tuesday what can be done about it and which mocks are required and where.

ui/components/CheckboxActions.tsx
```diff
@@ -0,0 +1,81 @@
...
+function CheckboxActions({ className, checked = [], rows = [] }: Props) {
+  //TODO add multi sync
+  const [suspendReqs, setSuspendReqs] = React.useState([]);
+
+  React.useEffect(() => {
+    if (checked.length && rows.length)
+      setSuspendReqs(makeObjects(checked, rows));
+  }, [checked, rows]);
+
+  const resume = useToggleSuspend(
```
opudrovs, line 43: Just a suggestion (it does not affect the functionality): In order not to repeat the code, which creates button handlers, twice, it is possible to use smth. like:

```
function createSuspendHandler(suspend: boolean) {
    const result = useToggleSuspend(
      {
        objects: suspendReqs,
        suspend: suspend,
      },
      suspendReqs[0] ? suspendReqs[0].kind : ""
    );

    return () => result.mutateAsync();
  }
```

and pass it to buttons like

```
<Button disabled={enabled} onClick={createSuspendHandler(true)}>
```

true for suspend, false for resume

ui/components/CheckboxActions.tsx
```diff
@@ -0,0 +1,81 @@
...
+      suspend: true,
+    },
+    suspendReqs[0] ? suspendReqs[0].kind : ""
+  );
+
+  return (
+    <Flex start align className={className}>
+      <Tooltip title="Suspend Selected" placement="top">
+        <div>
+          <Button disabled={!checked[0]} onClick={() => suspend.mutateAsync()}>
```
opudrovs, line 62: You can define a constant `const enabled = !checked[0];` or smth. like that and pass it instead of `!checked[0]`.

ui/components/DataTable.tsx
```diff
@@ -38,10 +55,11 @@ export interface Props {
   fields: Field[];
   /** A list of data that will be iterated through to create the columns described in `fields`. */
   rows?: any[];
-  /** for passing pagination */
-  children?: any;
+  filters?: FilterConfig;
+  dialogOpen?: boolean;
+  checkboxes?: boolean;
```
opudrovs, line 60: I think the `checkboxes` prop name is a bit misleading. `rows` and `filters` are passed as data objects (lists, configs) and `checkboxes` is just a boolean value. When I looked at it first I expected to see an array of checkboxes data passed instead of just a flag.

It might be worth it to rename this prop to `hasCheckboxes` or `shouldDisplayCheckboxes` to avoid confusion.

ui/components/DataTable.tsx
```diff
@@ -176,81 +437,147 @@ function UnstyledDataTable({ className, fields, rows, children }: Props) {
...
-          <Text>{typeof f.value === "function" ? f.value(r) : r[f.value]}</Text>
-        </TableCell>
-      ))}
-    </TableRow>
-  ));
+  const [checked, setChecked] = React.useState([]);
 
+  const r = _.map(sorted, (r, i) => {
+    return (
+      <TableRow key={i}>
```
opudrovs, line 451: Just noticed that keys in the DataTable component rows are numbers. This is usually not recommended if there is any sorting/filtering in the table because it can lead to had to detect bugs with wrong rows displayed. I haven't noticed any similar issues with DataTable, fortunately, but as a rule it is better to have keys as unique IDs on data elements or some other unique values:
https://reactjs.org/docs/lists-and-keys.html#keys

If there is no unique stings on rows we can use as keys, well, it seems to work fine with indices as is.

ui/components/CheckboxActions.tsx
joshri: Nice - something like this was my initial idea but I couldn't get it all the way there. Returning the function was the piece I was missing!

ui/components/CheckboxActions.tsx
opudrovs: or `disabled`, don't remember, whatever is appropriate.:)

ozamosi (approved):

PR #2652, opened 2022-08-23
Initial pendo integration

### LINKED ISSUE #2451: Install Pendo agent for Weave GitOps Dashboard

ozamosi:
This first of all turns dev-mode into a feature flag, because I wanted it available on the front end to tag traffic from developer machines separately from end-users - anyway, we might want to flash something if someone has turned off cookie verification that they really shouldn't.

Then, it adds an inital pendo integration for extra telemetry data. It is completely, 100% turned off by default - the javascript isn't even loaded unless the user opts in.

This turns on the feature in tilt, so we can see that it actually works - but I expect we'll turn it off eventually.

The frontend integration is done using a "component" that does nothing - I don't think this is good react, and I'm happy to try something else, but I need it to run after contexts have been initialized so the feature flags are in place, which this seems to accomplish.

There's a backend bit to it to get a unique cluster value. This completely abuses the feature flag system to smuggle out a unique string value using this mechanism we already have. Again, this is not enabled or active unless the user opts in.

There's two "secret-looking" things - one is the pendo application id, the other is the key I've hashed all user data with. Both are committed in the open:
 * The pendo application ID is write-only - the worst you could do is to push fake data.
 * While the key are used together with user info (emails), it is only there to make sure we can't use the hash against a password dump to cross-reference users - there is no way to figure out who is who other than complete brute force.

Both the visitor ID and the account ID have been mnemonic encoded, so that they're easier to remember - they're probably unnecessarily long, but we can always make changes later. 

This resolves #2451.

### DIFF
--- a/cmd/gitops-server/cmd/cmd.go
+++ b/cmd/gitops-server/cmd/cmd.go
@@ -100,8 +100,6 @@ func NewCommand() *cobra.Command {
 	cmd.Flags().StringVar(&options.OIDC.IssuerURL, "oidc-issuer-url", "", "The URL of the OpenID Connect issuer")
 	cmd.Flags().StringVar(&options.OIDC.RedirectURL, "oidc-redirect-url", "", "The OAuth2 redirect URL")
 	cmd.Flags().DurationVar(&options.OIDC.TokenDuration, "oidc-token-duration", time.Hour, "The duration of the ID token. It should be set in the format: number + time unit (s,m,h) e.g., 20m")
-	// Dev mode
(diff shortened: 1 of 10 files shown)


opudrovs:
This is what I see when going to the login page in this branch (doesn't happen in main). Do I need to enable smth. else?

I think it didn't happen if I was already logged in from main.

Screenshot 2022-08-24 at 17 12 39

ozamosi:
No, no that's not what's supposed to happen.

opudrovs (approved):

PR #2654, opened 2022-08-24
ui: Sort events by newest first

### LINKED ISSUE #2534: sort events by newest first
Screenshot 2022-08-04 at 14 38 48

Makes it easier to know which events are the ones to care about and also helps when there are a large number of events, where the latest events may be outside of the visible part of the screen

### Acceptance Criteria
* When I go to any events tab it should be sorted by timestamp in descending order by default.
* Sorting by timestamp should also be a secondary sort if we are sorting by another column.  E.g. if I am sorting by `name`, I would expect rows with the same name to be in descending order via timestamp.
* This should be for all `event` tabs on detail pages such as sources, kustomizations, and helm releases.

opudrovs:
Closes #2534 

- Added support for the secondary sort.
- Added sorting the Events table by timestamp by default (earliest events first).
- Added secondary sort by timestamp to the Events table. If sorting on another column where some values are the same, values with the earlier timestamp will come first (no matter what direction the first column is sorted on).

### DIFF
--- a/ui/components/DataTable.tsx
+++ b/ui/components/DataTable.tsx
@@ -26,6 +26,8 @@ export type Field = {
   maxWidth?: number;
   /** boolean for field to initially sort against. */
   defaultSort?: boolean;
+  /** boolean for field to implement secondary sort against. */
+  secondarySort?: boolean;
 };
 
 /** DataTable Properties  */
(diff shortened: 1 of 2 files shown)


opudrovs:
It's best to test this PR right after adding msdemo and podinfo workloads, you'll have plenty of events. Testing on the same values with the secondary sort in the Events table works well on the Component column.

ozamosi (approved):

PR #2655, opened 2022-08-24
clustersmngr: Reduce the number of ways we create cluster configs

ozamosi:
This tries to make clustersmngr slightly more consistent with
    itself.
    
 * Every client now always uses the same base settings (with timeouts
       and rate limiting). It used to be that user connections used the
       "good" settings, and server connections usually didn't.
 * Then, every code path now creates clients in parallel. Everything
       already did except GetServerClient.
 * Then, remove any workarounds so all code here uses the same few
       code paths. There used to be private functions in factory that
       duplicated functionality that there was already public functions
       for - no more. Even namespace caching uses the cluster client.
    
This still leaves additional implementations in pkg/kube, but one
    thing at a time.

### DIFF
--- a/core/clustersmngr/clustersmngr.go
+++ b/core/clustersmngr/clustersmngr.go
@@ -3,13 +3,10 @@ package clustersmngr
 import (
 	"context"
 	"fmt"
-	"net"
 	"sync"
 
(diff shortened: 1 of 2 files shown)


chanwit (approved):
LGTM

PR #2658, opened 2022-08-25
Add listObjects, make sources table use it

ozamosi:
This means that the list view and the detail view for sources is
driven by the same code. To make it happen, I had to:
 * Add type directly as a property. My hacks to transform
   KindKustomization to Kustomization using a spread operator came back
   to haunt me when this made the object be not-the-right-object. But we can
   just add it on the object.
 * Make `sourceRef` actually return undefined when it's not set. We
   actually use `?.` intentionally here and I was breaking it.
 * Add UIDs - they're needed by #2649.

`listObjects` is really just copied from one of the many existing list
methods, and then changed to use the same type converter as
getObject.

This means that to add new `kind`s - e.g. for notifications - the only backend change we should need is to put them in `core/server/primarykinds.go` and listing, getting, and getting events should all work.


joshri (approved):
WOW

This is clearly a huge improvement - even though I'm not a hundred percent sure how it works, I checked out the branch and it definitely DOES work, so that's great. Most of my comments are questions so I can understand this a little better - is the next step to apply this to automations as well?

core/server/objects.go
```diff
@@ -2,15 +2,84 @@ package server
...
+	if err != nil {
+		return nil, err
+	}
+
+	clustersClient, err := cs.clientsFactory.GetImpersonatedClient(ctx, auth.Principal(ctx))
+	if err != nil {
+		if merr, ok := err.(*multierror.Error); ok {
+			for _, err := range merr.Errors {
+				if cerr, ok := err.(*clustersmngr.ClientError); ok {
+					respErrors = append(respErrors, &pb.ListError{ClusterName: cerr.ClusterName, Message: cerr.Error()})
```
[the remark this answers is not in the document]
opudrovs: > the request can still continue even if there're errors involved?

I suppose, if I got how the errors are handled here, that yes, it should return all results and all errors:

```
return &pb.ListObjectsResponse{
  Objects: results,
  Errors:  respErrors,
}, nil
```

(At least those errors which should get special treatment (ListErrors) other errors are returned as usual):

ui/lib/objects.ts
```diff
@@ -82,6 +83,11 @@ export class FluxObject {
     }
   }
 
+  // TODO: this should be merged with kind
```
opudrovs, line 86: Could you please clarify what this comment means?

opudrovs:
BTW, there is also an old branch named `list-objects`, is it relevant to this branch and is it still needed?

api/core/core.proto
```diff
@@ -342,6 +264,16 @@ message GetObjectResponse {
     Object object = 1;
 }
 
+message ListObjectsRequest {
+    string namespace   = 2;
```
opudrovs, line 268: Nothing important, but what's the purpose of starting the numbering with 2 here? Are you planning to add some fields before `namespace` later?

ui/lib/api/core/core.pb.ts
```diff
@@ -132,6 +87,16 @@ export type GetObjectResponse = {
   object?: Gitops_coreV1Types.Object
 }
 
+export type ListObjectsRequest = {
+  namespace?: string
+  kind?: string
```
opudrovs, line 92: I know, this file is auto-generated and see that all fields are generated as optional. So it's just a fantasy, but it would be nice if we could distinguish between non-optional and optional fields in requests in generated code.

For example, when listing objects for `ListObjectsRequest`, there is a kind lookup performed, and if it is not value/known, an error is returned. So, it would be nice to be able to generate kind as a non-optional value.

Or is it the Golang way: "everything can have a zeroed value", just throw an error if the value is not provided?

opudrovs (approved):
LGTM! 🎉 Cool enhancement/refactoring, now the code is more consistent and simpler. Tested it locally, everything works fine, I can see expected objects in network request responses with `uid`s.

core/server/objects.go
```diff
@@ -2,15 +2,84 @@ package server
...
 	pb "github.com/weaveworks/weave-gitops/pkg/api/core"
 	"github.com/weaveworks/weave-gitops/pkg/server/auth"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+func (cs *coreServer) ListObjects(ctx context.Context, msg *pb.ListObjectsRequest) (*pb.ListObjectsResponse, error) {
+	respErrors := []*pb.ListError{}
+
+	gvk, err := cs.primaryKinds.Lookup(msg.Kind)
```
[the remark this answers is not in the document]
ozamosi: So. Kinda.

You have to specify a single type when you make the k8s request. But there's nothing stopping us making multiple requests against the backend in the same HTTP request, and it'd be a bit more efficient.

ui/components/SourceDetail.tsx
```diff
@@ -79,17 +79,11 @@ function SourceDetail({ className, source, info, type }: Props) {
       return false;
     }
 
-    if (
-      type == FluxObjectKind.KindHelmChart &&
-      isNameRelevant(a?.helmChart?.name)
-    ) {
-      return true;
+    if (type == FluxObjectKind.KindHelmChart) {
+      return isNameRelevant((a as HelmRelease)?.helmChart?.name);
```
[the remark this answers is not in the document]
ozamosi: The one I meant was in `ui/components/HelmChartDetail.tsx`  - you're pointing to automations, and they are still the same. Though, the same thing becomes a problem with them too.

In general, `sourceRef` should return objects of the right type (name, namespace, kind) or nothing. It was returning an empty object which didn't correspond to the type - and if I changed the type I'd have to change all e.g. `sourceRef?.name` to `sourceRef.name?` for no real reason - being able to express "this object has no source" is actually useful.

api/core/core.proto
ozamosi: No purpose, just copy paste :tired_face:

ui/lib/api/core/core.pb.ts
ozamosi: It's actually not go, but the protobuf way - protobuf2 had a distinction between mandatory and optional fields, but protobuf3 removed it. The reason has to do with compatibility between versions of the protobuf schema.

In proto2, if the version of the client and the server are different, then one end might think a field is mandatory and the other end might not know about the field, and then they just refuse to talk to each other. Every time I've experienced this, it was because something that was once mandatory was no longer needed, but deploying that change would make the services no longer talk, and then I'd spend the rest of the day carefully coordinating multi-step deployments.

In proto3, every field is optional, every empty field is just even included in the message (so null, empty and missing are all the same thing), and every unknown field is just ignored. So as long as you don't re-use the numbers for different purposes between different versions, it will always work to parse a message. It might not be a useful result any more if the API has changed enough, but it will work.

This doesn't really apply to our use because we ship and deploy the server and client together, so they should always have the same version. But that's why.

opudrovs: Thanks for the detailed explanation, good to know!

ozamosi:
> is the next step to apply this to automations as well?

Yes. It adds a little inventory-related wrinkle, and this PR is large enough as it is, but yeah.

> there is also an old branch named list-objects, is it relevant to the current branch and is it still needed?

It was a hacky and untested prototype - it is not still needed, I'll remove it.

ui/lib/objects.ts
ozamosi: Good shout. Done.

ozamosi:
Rebased on top of multi-suspend/resume, have confirmed I didn't break it - merging.

core/server/objects.go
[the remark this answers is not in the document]
foot: `msg.Kinds`, cool thought.

PR #2662, opened 2022-08-29
Fixes apps/sources working for clusters that are reachable

foot:
- Don't bail if a single cluster is unreachable for some reason
- Also add in some logging to begin tracking down why we get empty
  responses from apps/sources sometimes (that recover on next poll)

**What changed?**

Instead of returning an err, continue updating namespace for other cluster we are able to connect to

**Why was this change made?**

So apps/sources work if a single cluster is temporarily unreachable

- Unit tests
- Manually testing by `replace`ing in WGE


ozamosi (approved):
Thank you!

PR #2663, opened 2022-08-29
Fix titles overflowing graph nodes

### LINKED ISSUE #2614: Long captions in graph nodes
Long captions go beyond graph node boundaries:

Screenshot 2022-08-17 at 01 29 31

Possible solutions:
- Add wrapping to text in graph nodes (and make nodes taller). The text has no spaces, maybe it should be broken on a dot instead?
- Nodes could be wider, but it will lead to too nodes being to wide for shorter text. Maybe we could check the text length, and if it is longer than a cutoff, should use a bigger node for this text without affecting other nodes?

### Acceptance Criteria

* We will truncate name by `...` using overflow.
* As a user when I hover over the node, I will see the full name as a tooltip.

joshri:
Closes #2614

"The text-overflow property only affects content that is overflowing a block container element in its inline progression direction (not text overflowing at the bottom of a box, for example)."
I shamefully broke this rule when carrying over my old node styling to the new graph node component
https://developer.mozilla.org/en-US/docs/Web/CSS/text-overflow
 
adding spans was also convenient for the MUI tooltip which expects a component that's not customized

image

### DIFF
--- a/ui/components/GraphNode.tsx
+++ b/ui/components/GraphNode.tsx
@@ -91,9 +91,8 @@ function GraphNode({ className, object }: Props) {
           {getStatusIcon(computeReady(object.conditions), object.suspended)}
           <div style={{ padding: 4 }} />
           <Tooltip
-            open={true}
             placement="top"
-            title={object.name.length > 20 ? object.name : ""}
+            title={object.name.length > 25 ? object.name : ""}
           >
             <span>{object.name}</span>
           </Tooltip>
--- a/ui/components/GraphNode.tsx
+++ b/ui/components/GraphNode.tsx
@@ -91,6 +91,7 @@ function GraphNode({ className, object }: Props) {
           {getStatusIcon(computeReady(object.conditions), object.suspended)}
           <div style={{ padding: 4 }} />
           <Tooltip
+            open={true}
             placement="top"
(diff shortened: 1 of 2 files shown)


opudrovs:
- Now the tooltips are displayed for all nodes, event those with short captions, where they are of no use. Can we somehow display them only after some caption length cut-off.
- Is it possible to make the tooltip text bigger? It is <sub>tiny</sub>

joshri:
Changed from .675rem (mui default) to 1rem what do you think?

image

opudrovs:
I think it looks better, the old version was almost unreadable on my laptop.

joshri:
Great - tooltips won't show unless the title is longer than 20 characters (might still get some cases where there's an unnecessary tooltip but I wanted to be safe) as well

opudrovs:
- I think you can safely make the check 25 characters, there is still some margin.
- I can see all of the tooltips at once. I thought only one if the element is hovered is supposed to be displayed?

Screenshot 2022-08-29 at 22 25 27

opudrovs:
From the acceptance criteria: "As a user when I hover over the node, I will see the full name as a tooltip."

opudrovs (approved):
LGTM! 🎉

PR #2665, opened 2022-08-29
Exit more gracefully in ListHelmReleases when a HelmRelease revision is missing

### LINKED ISSUE #2664: Missing Helm release secrets cause ListHelmReleases to return an error, and to list no releases
Hey friends! @amulyakish and I encountered what we're pretty sure is a corner-case in WGO's rendering of deployed HelmReleases.

The behavior is similar to what was seen in #2016 and a few other issues. The associated fix for #2016 , #2042 , introduced behavior that requires all Helm releases in `ListHelmReleases` to be "complete", i.e. to have a Helm [release secret](https://codeengineered.com/blog/2020/helm-secret-storage/) created for a given release revision. Without the Helm release secret, WGO is not able to read the details of the release. Without the details of the release, WGO does not render _any_ HelmReleases, renders a stack trace instead, and yields an error about a missing Helm secret.

**Describe the bug**

If a Flux HelmRelease receives a new Helm revision number, but the associated Helm release at that revision did not complete and did not create a release secret, then there is no release to read. WGO therefore does not render the *entire list* of HelmReleases. We also see a stack trace, and a console message like `"secrets \"sh.helm.release.v1.myapp8v9np.v1\" not found"`.

**Environment**

- Weave-Gitops Version: 0.9.3 and `main` `HEAD`
- Flux Version: v0.28.5
- Kubernetes version: 1.22

**To Reproduce**

Introduce a Flux HelmRelease with a Helm revision number that doesn't exist. This may be due to a Helm upgrade that is attempted, but that fails due to Kubernetes validation rules.

We saw an example of this in our own cluster. Following one cluster's upgrade to Kubernetes 1.22, the Ingress object at `extensions/v1beta1` was [deprecated](https://kubernetes.io/docs/reference/using-api/deprecation-guide/#ingress-v122). One application had not updated their Helm chart for this, causing a Helm upgrade failure. Some sample outputs:

```
$ kubectl get hr -n this-namespace
NAME                     AGE    READY   STATUS
this-service   355d   False   Helm upgrade failed: current release manifest contains removed kubernetes api(s) for this kubernetes version and it is therefore unable to build the kubernetes objects for performing the diff. error from kubernetes: unable to recognize "": no matches for kind "Ingress" in version "extensions/v1beta1"
```

```yaml
$ kubectl describe hr -n this-namespace this-service
kind: HelmRelease
metadata:
  creationTimestamp: "2021-09-08T01:29:15Z"
  finalizers:
  - finalizers.fluxcd.io
  generation: 50
  labels:
    kustomize.toolkit.fluxcd.io/name: applications
    kustomize.toolkit.fluxcd.io/namespace: internal-system
  managedFields:
  ...
spec:
  ...
status:
  conditions:
  - lastTransitionTime: "2022-08-25T21:22:42Z"
    message: 'Helm upgrade failed: current release manifest contains removed kubernetes
      api(s) for this kubernetes version and it is therefore unable to build the kubernetes
      objects for performing the diff. error from kubernetes: unable to recognize
      "": no matches for kind "Ingress" in version "extensions/v1beta1"'
    reason: UpgradeFailed
    status: "False"
    type: Ready
  - lastTransitionTime: "2022-08-25T21:22:42Z"
    message: |-
      Helm upgrade failed: current release manifest contains removed kubernetes api(s) for this kubernetes version and it is therefore unable to build the kubernetes objects for performing the diff. error from kubernetes: unable to recognize "": no matches for kind "Ingress" in version "extensions/v1beta1"

      Last Helm logs:

      preparing upgrade for this-service
      resetting values to the chart's original version
      performing update for this-service
    reason: UpgradeFailed
    status: "False"
    type: Released
  failures: 1
  ...
  lastReleaseRevision: 35 ## <----- NOTICE HERE!
  observedGeneration: 50
  upgradeFailures: 1
```

Compare the above `lastReleaseRevision` to the following `REVISION`:

```
$ helm list -n this-namespace
NAME          NAMESPACE                 REVISION        UPDATED                                      STATUS          CHART                    APP VERSION
this-service  this-namespace            34              2022-07-21 16:29:31.442591761 +0000 UTC      deployed        this-service-1.0.0       1.0.0
...
```
**Expected behavior**

WGO Applications lists all HelmReleases, regardless of deployment state.

**Actual Behavior**

WGO Applications lists *no* HelmReleases, regardless of deployment state. WGO instead shows an error message.

image

What might we do about this? In the short term, there is at least one relatively simple fix, and I'll have a PR open shortly to show. The logic in `ListHelmReleases` may be amended to append `err` to `respErrors` on failure to `getHelmReleaseInventory`, instead of returning early: https://github.com/weaveworks/weave-gitops/blob/main/core/server/helm_release.go#L58-L67

This will cause the Helm Release listing in WGO to *omit* the failing release in its list view. Not great, but it permits the Helm Release list view to render overall in the case identified in this issue, thereby satisfying this issue. Users will not see entries that do not have releases at their specified revision. *This is confusing* for users, and users will need to a) notice that their release is missing, and then b) manually dig into `kubectl` and `helm` in their clusters to figure out what happened to their missing entry. It is however arguably better than one failing release causing WGO to not render _any_ releases, which is the current state.

In the long-term, however, I'm not sure what you all might want to do with error handling and messaging, and I don't mean to presume at this stage. The inner-loop error handling logic in most of the other List methods looked fine for now. I saw a couple of spots in both List and Get methods that seemed like they would push stack traces back to the user (`return nil, err` being the indicator). All of those are out of scope of this Issue.

There is at least one other List endpoint that might warrant a similar review as here, in `ListKustomizations`: https://github.com/weaveworks/weave-gitops/blob/main/core/server/kustomization.go#L66-L69

Thank you for your consideration here! PR for `ListHelmReleases` coming shortly.

devmage:
Fixes #2664 (kind of, see below)

In #2664, @amulyakish and I show how a missing Helm release causes WGO to render _no_ applications.

This patch will cause the HelmRelease listing in WGO Applications to omit the missing release in its list view.

As explained in #2664 , this is an error view that should not be shown to users.

We disabled an early-return in `ListHelmReleases`, and instead send errors to the `respErrors` accumulator. Note again that this has the effect of *hiding* releases that are misconfigured. This patch, therefore, is a stop-gap in hope of a more comprehensive review of error handling, also noted in #2664 .

We included a unit test that simulates the behavior that we saw in #2664 .

**Release notes**

This is Bugfix level, in SemVer terms. Amulya and I would appreciate a short bugfix note somewhere, so that we know when a fix is made.

**Documentation Changes**

None that I would suggest.

### DIFF
--- a/core/server/helm_release_test.go
+++ b/core/server/helm_release_test.go
@@ -138,6 +138,84 @@ func TestGetHelmRelease(t *testing.T) {
 	g.Expect(err).To(HaveOccurred())
 }
 
+/**
+ * This test demonstrates the behavior of ListHelmReleases.
+ * The third HR created in this test is managed by the Helm Release Controller,
+ * but does not have a corresponding secret. Because of this the getHelmReleaseInventory
+ * call throws an error. This should not cause the entire list call to error out, and
+ * instead skip just the one HR that is in error state.
+ **/
+func TestListHelmReleases_withInventory(t *testing.T) {
+	g := NewGomegaWithT(t)
+	ctx := context.Background()
+
+	c, _ := makeGRPCServer(k8sEnv.Rest, t)
+
+	scheme, err := kube.CreateScheme()
+	g.Expect(err).To(BeNil())
+
+	k, err := client.New(k8sEnv.Rest, client.Options{
+		Scheme: scheme,
+	})
+	g.Expect(err).NotTo(HaveOccurred())
+
+	appName := "myapp" + rand.String(5)
+	nsA := newNamespace(ctx, k, g)
+	nsB := newNamespace(ctx, k, g)
+	nsC := newNamespace(ctx, k, g)
+
+	newHelmRelease(ctx, appName, nsA.Name, k, g)
+	newHelmRelease(ctx, appName, nsB.Name, k, g)
+
+	res1, error1 := c.ListHelmReleases(ctx, &pb.ListHelmReleasesRequest{})
+
+	g.Expect(error1).NotTo(HaveOccurred())
+
+	releasesFound := len(res1.HelmReleases)
+
+	// Create helm release without a corresponding secret.
+	helmRelease := helmv2.HelmRelease{
+		TypeMeta: metav1.TypeMeta{
+			Kind:       helmv2.HelmReleaseKind,
+			APIVersion: helmv2.GroupVersion.String(),
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      appName,
+			Namespace: nsC.Name,
+		},
+		Status: helmv2.HelmReleaseStatus{
+			LastReleaseRevision: 2,
+		},
+		Spec: helmv2.HelmReleaseSpec{
+			Chart: helmv2.HelmChartTemplate{
+				Spec: helmv2.HelmChartTemplateSpec{
+					SourceRef: helmv2.CrossNamespaceObjectReference{
+						Kind: "GitRepository",
+						Name: "somesource",
+					},
+				},
+			},
+		},
+	}
+
+	opt := []client.PatchOption{
+		client.ForceOwnership,
+		client.FieldOwner("helmrelease-controller"),
+	}
+
+	g.Expect(k.Patch(ctx, &helmRelease, client.Apply, opt...)).To(Succeed())
+
+	helmRelease.Status.LastReleaseRevision = 1
+	helmRelease.ManagedFields = nil
+
+	g.Expect(k.Status().Patch(ctx, &helmRelease, client.Apply, opt...)).To(Succeed())
+
+	res2, error2 := c.ListHelmReleases(ctx, &pb.ListHelmReleasesRequest{})
+
+	g.Expect(error2).NotTo(HaveOccurred())
+	g.Expect(res2.HelmReleases).To(HaveLen(releasesFound))
+}
+
 func TestGetHelmRelease_withInventory(t *testing.T) {
 	g := NewGomegaWithT(t)
 	ctx := context.Background()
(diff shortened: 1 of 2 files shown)


devmage:
I would also point out that several `go fmt` changes appear to be needed on the repository at present. I ran `make fmt` and discarded all of the changes that were not relevant to the files in this PR.

core/server/helm_release_test.go
```diff
@@ -138,6 +138,84 @@ func TestGetHelmRelease(t *testing.T) {
...
+	nsC := newNamespace(ctx, k, g)
+
+	newHelmRelease(ctx, appName, nsA.Name, k, g)
+	newHelmRelease(ctx, appName, nsB.Name, k, g)
+
+	res1, error1 := c.ListHelmReleases(ctx, &pb.ListHelmReleasesRequest{})
+
+	g.Expect(error1).NotTo(HaveOccurred())
+
+	releasesFound := len(res1.HelmReleases)
```
devmage, line 174: There was an error on this line, before I added a newline between it and the `g.Expect` before it. Curiously, this was not caught by `make fmt`.

ozamosi:
Are you by any chance using go 1.19? I haven't gotten around to changing the version we use, and `go fmt` in 1.18 and 1.19 have a bunch of different opinions about formatting. It's on my list of things, I think I need to move it up the list.

As for the change itself, as you note it is a step forward while still not being ideal, so I'll merge this while leaving the bug report open (which is what you said as well - look at me, being able to read).

Thank you so much for both the bug report, the patch, and the clear description of the problem!

ozamosi (approved):

devmage:
> Are you by any chance using go 1.19?

Ah, yes, that's probably it then. Admittedly my Go development environment is not very well configured. 😅 

> As for the change itself, as you note it is a step forward while still not being ideal, so I'll merge this while leaving the bug report open

Thank you! Glad we could contribute, looking forward to the next release.

PR #2667, opened 2022-08-30
refresh OIDC tokens after expiry

alichaddad:
**What changed?**

- Added a refresh OIDC token mechanism when current token in the session is expired
- Remove the "profile" scope, we don't use it right now.

**Why was this change made?**
There were issues with tokens with a short expiration that was less than the session time which led to the user being forced to log in multiple times in a single session

- In the auth middleware when a a request is unauthenticated. The code attempts to refresh the tokens before reporting to the user that he is unauthenticated. The refresh token is added to a cookie and used to send a refresh request to the OIDC server.

To support token passthrough we also have to:
- Update the token in the principal
- Expire the any cached clients using the old token, we do this by including a hash of the token in the client cache key

- I ran an OIDC server using `dex` and configured the expiration to be 2min and checked that the refresh flow was triggered.
- Ran Auth0 instance with 2min expiry and configure k8s-api-server to use same OIDC and enabled #3218, clients correctly expire and the refresh is quite smooth and invisible to the user.


pkg/server/auth/server.go
```diff
@@ -186,6 +187,11 @@ func (s *AuthServer) oauth2Config(scopes []string) *oauth2.Config {
 		scopes = append(scopes, scopeGroups)
 	}
 
+	// Ensure "offline_access" scope is always present for refresh tokens.
+	if !contains(scopes, oidc.ScopeOfflineAccess) {
```
bigkevmcd, line 191: One thing we could check is, do all providers support offline access?

Maybe we make a quick list of possibles (other than Dex) and check whether or not they allow it.

Some examples, Keycloak (we have customers using it), Google and anybody else you can think of?

What happens if the provider does not allow it?

pkg/server/auth/server.go
```diff
@@ -429,6 +436,35 @@ func (s *AuthServer) UserInfo() http.HandlerFunc {
 	}
 }
 
+func (s *AuthServer) Refresh(rw http.ResponseWriter, r *http.Request) error {
+	ctx := oidc.ClientContext(r.Context(), s.client)
+
+	refreshTokenCookie, err := r.Cookie(RefreshTokenCookieName)
+	if err != nil {
+		return errors.New("couldn't fetch refresh token from cookie")
```
bigkevmcd, line 444: Is this an error?

Or is it just "not refreshed" ?

pkg/server/auth/server_test.go
```diff
@@ -625,6 +633,69 @@ func TestUserInfoOIDCFlow(t *testing.T) {
...
+	tf := tokens["refresh_token"].(string)
+
+	req := httptest.NewRequest(http.MethodGet, "https://example.com/test", nil)
+	req.AddCookie(&http.Cookie{
+		Name:  auth.RefreshTokenCookieName,
+		Value: tf,
+	})
+
+	w := httptest.NewRecorder()
+	cookies := make(map[string]*http.Cookie)
```
bigkevmcd, line 655: Nitpick: this could be moved to above line 659 because you don't need it to determine whether or not it succeeds?

From https://wiki.c2.com/?DeclareVariablesAtFirstUse
> It's [good] to declare variables when you first use them to ensure that they are always initialized to some valid value and that their intended use is always apparent.

pkg/server/auth/server_test.go
```diff
@@ -625,6 +633,69 @@ func TestUserInfoOIDCFlow(t *testing.T) {
...
+		if c.Name == auth.IDTokenCookieName || c.Name == auth.AccessTokenCookieName || c.Name == auth.RefreshTokenCookieName {
+			cookies[c.Name] = c
+		}
+	}
+
+	g.Expect(cookies).To(HaveKey(auth.IDTokenCookieName))
+	g.Expect(cookies).To(HaveKey(auth.AccessTokenCookieName))
+	g.Expect(cookies).To(HaveKey(auth.RefreshTokenCookieName))
+
+	_, err = m.Keypair.VerifyJWT(cookies[auth.IDTokenCookieName].Value)
```
bigkevmcd, line 669: These tests are a bit complex to understand, can we either split them, or at least comment to explain why the three _almost_ identical calls are made?

pkg/server/auth/server_test.go
```diff
@@ -625,6 +633,69 @@ func TestUserInfoOIDCFlow(t *testing.T) {
...
+	g := gomega.NewGomegaWithT(t)
+	s, _ := makeAuthServer(t, nil, nil, []auth.AuthMethod{auth.OIDC})
+
+	w := httptest.NewRecorder()
+	req := httptest.NewRequest(http.MethodGet, "https://example.com/test", nil)
+	req.AddCookie(&http.Cookie{
+		Name:  auth.RefreshTokenCookieName,
+		Value: "invalid",
+	})
+	g.Expect(s.Refresh(w, req)).To(HaveOccurred())
```
bigkevmcd, line 696: I think this should be `MatchError` rather than just `HaveOccurred` ?

foot: Yesterday I discovered you can pair w/ `MatchRegexp` which makes things a bit easier sometimes.

`.To(MatchError(MatchRegexp("failed adding cluster client to pool.*cluster: %s.*", clusterName3)))`

pkg/server/auth/server.go
alichaddad: There is a list of all oauth providers who support OIDC: https://en.wikipedia.org/wiki/List_of_OAuth_providers, we can make a shortlist from that list as a start?

alichaddad:
> To quote from the Dex docs...
> 
> > Depending on the connectors limitations in protocols can prevent dex from issuing [refresh tokens](https://dexidp.io/docs/custom-scopes-claims-clients/#scopes) or returning [group membership](https://dexidp.io/docs/custom-scopes-claims-clients/#scopes) claims.
> 
> So, we might need to think about what happens if the upstream doesn't allow it, do we get an error?

I think this will depend on the provider used, ideally the server should check the `grant_type` and if it isn't supported(in that case `refersh_token`) it should return an error, but I am not sure if we can depend on this

ozamosi:
It's not just a matter of whether the provider technically support it - a `refresh_token` is intended to be used "forever", so an administrator might not want us to have `refresh_tokens`.

If anything, if the system administrator has decided to limit lifetimes of tokens to very, very short times, then I would think it's pretty likely that the same system administrator would also refuse to issue `refresh_tokens`.

> There were issues with tokens with a short expiration that was less than the session time which led to the user being forced to log in multiple times in a single session

Can you please provide some more information about this? I'm not clear on what type of time windows you're talking about, which expiration, or where it expires (OIDC server or k8s API server).

In particlar, are you sure that the problem is not that we require the user to specify `tokenDuration` as a hard-coded value in the secret, instead of using `expires_in` from the token itself?

alichaddad:
@foot  @saeedfazal  can you provide more context here?

foot:
> If anything, if the system administrator has decided to limit lifetimes of tokens to very, very short times, then I would think it's pretty likely that the same system administrator would also refuse to issue refresh_tokens.

Interesting. I'm unfamiliar with setting this up. We've had reports from people (cc @makkes and others) trying to do demos that their IdP's have configured id_token to only last 15mins (even 5mins sometimes!). I would have guessed that the refresh_token would have been the way to support 5/15mins id_token expirations and still keep an okay UX.

> In particlar, are you sure that the problem is not that we require the user to specify tokenDuration as a hard-coded value in the secret, instead of using expires_in from the token itself?

This is 1h by default I believe and so not the issue we're seeing, but as you say the cookie expiration should really match the expiration provided by the IdP.

>  a refresh_token is intended to be used "forever", so an administrator might not want us to have refresh_tokens.

@alichaddad have you noticed an expiration coming back w/ the refresh_tokens?

ozamosi:
> their IdP's have configured id_token to only last 15mins (even 5mins sometimes!).

Thank you - that is way less indeed!

Is the `id_token` and the `access_token` configured to the same duration?

alichaddad:
@foot refresh tokens have expiry but it is generally for a long duration. If it is configured for a short time then there isn't much we can do about it. Although certain providers will reset the expiry every time the refresh token is used.

@ozamosi they can have different expiration, a lot of providers allow giving them different values, so it depends on the way they were configured.
The way it is done now there shouldn't be a redirect. If the user becomes unauthenticated we attempt to send a request with the values in the cookies to refresh the tokens. I am not sure if that's the best approach though. Initial thought was making the UI refresh the token and retry the failed request but this involved a refactor in the way API calls are made in the UI

foot:
> @foot refresh tokens have expiry but it is generally for a long duration

Is that like 1d or 14d, does dex have a default here?

> I get redirected to google, I already have a valid session there so they immediately issue a new token.

Yeah that sounds like it would fly every day or so, every 5mins might still be a bit rough (maybe we don't have a choice though I'm not sure). We don't do this automatically, but would be kind of neat.

alichaddad:
For dex it depends on what's in the configuration file, otherwise no expiry is set.

@bigkevmcd there is way to know if an OIDC provider supports refreshing the tokens by checking the openid configuration that should be be provided by all providers.

alichaddad:
@bigkevmcd  Yes and `grant_types_supported`

foot:
> Is the id_token and the access_token configured to the same duration?

Both seem to be 1h by default https://github.com/weaveworks/weave-gitops/blob/44cbabd49c37c81b21c4881cd98e7aeff3273e9e/pkg/server/auth/server.go#L84-L87

makkes:
This is a great discussion (and it shows how unspecific much of the OIDC spec actually is). One thing that would greatly improve the UX would be if Wego would automatically redirect the user to the IdP because there the user might still be logged in. However, as a user I always have to hit "Login with OIDC Provider" as soon as the token expires.

makkes:
As much as I want to see this resolved, this PR hasn't seen any movement for over 3 months. I'm closing this now but not without stating that we should continue to strive towards a better OIDC UX. A [discussion](https://github.com/weaveworks/weave-gitops/discussions) might be a better fit at this point, though.

bigkevmcd:
Let's keep the branch around just now, as it's likely that we'll need to do this work (and some further OIDC) work sooner rather than later.

pkg/server/auth/server.go
foot: User can now omit offline_access from scope if they need to

pkg/server/auth/server.go
foot: This is a fair point if offline_access is not a requested scope.. 🤔

pkg/server/auth/server.go
```diff
@@ -488,6 +491,42 @@ func (s *AuthServer) UserInfo(rw http.ResponseWriter, r *http.Request) {
 	toJSON(rw, ui, s.Log)
 }
 
+// Refresh is used to refresh the access token and id token. It updates the cookies on the response
+// with the new tokens. It returns the new user principal.
+func (s *AuthServer) Refresh(rw http.ResponseWriter, r *http.Request) (*UserPrincipal, error) {
```
foot, line 496: if oidc is not enabled we shouldn't attempt a refresh either..

foot: If both oidc/user-account is enabled, and we try to refresh the user-account token, it will fail and log the user out anyway though.

pkg/server/auth/server.go
foot: A new method shouldAttemptRefresh that checks?
- refresh_token even exists
- oidc is enabled
- offline-scope is enabled
?

bigkevmcd: Probably check for refresh_token?

Would we have a refresh_token if oidc was not enabled/offline-scope was not enabled?

foot: true, fair.

foot (approved):
Thanks for the PR @alichaddad !

kingdonb:
I think this issue has been reactivated as [weaveworks/weave-gitops-enterprise#2383](https://github.com/weaveworks/weave-gitops-enterprise/issues/2383)

PR #2669, opened 2022-08-30
Fix error logs in UI tests

ozamosi:
This fixes a few errors that were logged but didn't cause failures in the UI tests. The cause was that by making feature flags part of the context, I triggered requests to happen where they weren't allowed in tests. My solution is to make the tests initialize the parameters directly (e.g. use the `CoreContext.Provider` directly instead of the `CoreContextProvider` helper). While I was messing with types, I also removed a couple of `any`.

This also makes the UI tests fail the next time this happens, so I don't make this mistake again.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -15,6 +15,7 @@
         "history": "^5.0.0",
         "http-proxy-middleware": "^2.0.3",
         "jest-canvas-mock": "^2.4.0",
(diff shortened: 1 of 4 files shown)


opudrovs (approved):

opudrovs:
I ran the UI tests locally, don't see any warnings! 🎉

PR #2677, opened 2022-08-31
Add Notification Providers

### LINKED ISSUE #2564: Notifications UI table view
As a platform operator, if I have notifications set up, I will see a list with the providers, namespace, cluster and severity so that I can quickly understand where notifications are being sent 
Figma Link: https://www.figma.com/file/IVHnM9iyeFWpd11evtY8ux/Weave-GitOps?node-id=4908%3A9125

joshri:
Closes #2564 

As always, it was a little more complicated than just adding providers as a resource. Channel is an optional value of the Provider spec, and if it wasn't set, I would expect users to be able to filter by undefined, or `-`. Allowing for undefined values in the DataTable and filtering by `-` led me to discover that searching by text was not working as expected in Sources and the new Notifications tables (possibly due to the restructuring of objects in the ListObjects request). When I dove into that code...it didn't click with me - we were checking every field before even checking if the `textSearchable` boolean was true! I rewrote it in a way that clicks with my brain. The tests for text search also were testing for OR values in filtering rather than AND, as in all text filters didn't have to apply for a row to be included, so I rewrote that test to match as well. 

Also - a couple props were removed from the Page component as they were unused

image


ui/components/ChipGroup.tsx
```diff
@@ -17,6 +17,7 @@ function ChipGroup({ className, chips = [], onChipRemove, onClearAll }: Props) {
   return (
     <Flex className={className} wide align start>
       {_.map(chips, (chip, index) => {
+        if (chip.slice(-1) === ":") chip += " -";
```
opudrovs, line 20: Why is this line required? It adds ` -` to any chip text ending with `:`?

opudrovs:
- What is required to see notifications data when testing this branch locally? I see no notifications data.

- When going to the settings page, the first sidebar item (Applications) stays selected, unlike the screenshot. Is it expected.

Screenshot 2022-09-01 at 22 35 26

opudrovs:
@joshri so, we do not have a Settings page with `settings` route yet, only Notifications with `notifications` route? (I see your comment in the ticket, just want to confirm the route.)

ui/components/DataTable.tsx
```diff
@@ -161,31 +161,27 @@ function filterText(
...
   return _.filter(rows, (row) => {
     let matches = false;
-    for (const colName in row) {
-      const value = row[colName];
 
-      const field = _.find(fields, (f) => {
-        if (typeof f.value === "string") {
-          return f.value === colName;
+    fields.forEach((field) => {
+      if (field.textSearchable) {
```
opudrovs, line 166: You could try adding

```
if (!field.textSearchable) {
  return;
}
```

at the beginning to remove this large `if` wrapping the whole code. It _should_ make `forEach` move to the next element, but please check it. 🙂

ui/components/DataTable.tsx
```diff
@@ -161,31 +161,27 @@ function filterText(
...
+        let value;
```
opudrovs, line 167: An empty line before this line might make the code more readable.

ui/components/DataTable.tsx
```diff
@@ -161,31 +161,27 @@ function filterText(
...
+          value = field.sortValue(row);
+        } else {
+          typeof field.value === "function"
+            ? (value = field.value(row))
+            : (value = row[field.value]);
         }
-
-        if (f.sortValue) {
-          return f.sortValue(row) === value;
+        for (let i = 0; i < textFilters.length; i++) {
```
opudrovs, line 175: And an empty line before this line would make reading the code easier.

ui/components/FilterDialog.tsx
```diff
@@ -78,20 +78,19 @@ const FilterSection = ({
           </Text>
         </ListItem>
         {options.sort().map((option: string, index: number) => {
```
opudrovs, line 80: You can use a shorter form here, because you are only returning the ListItem without using any local variables here:

```
{options.sort().map((option: string, index: number) => (
          <ListItem key={index}>
            <ListItemIcon>
              <FormCheckbox
                label=""
                name={`${header}${filterSeparator}${option}`}
              />
            </ListItemIcon>
            <Text color="neutral40" size="small">
              {_.toString(option) || "-"}
            </Text>
          </ListItem>
        ))}
```

joshri:
@opudrovs try using this yaml to get some providers going. Ripped it from the flux docs here https://fluxcd.io/flux/components/notification/provider

`apiVersion: notification.toolkit.fluxcd.io/v1beta1
kind: Provider
metadata:
  name: slack
  namespace: default
spec:
  type: slack
  channel: general
  # webhook address (ignored if secretRef is specified)
  address: https://hooks.slack.com/services/YOUR/SLACK/WEBHOOK
  # HTTP(S) proxy (optional)
  proxy: https://proxy.corp:8080
  # secret containing the webhook address (optional)
  secretRef:
    name: webhook-url`
    
`apiVersion: notification.toolkit.fluxcd.io/v1beta1
kind: Provider
metadata:
  name: msteams
  namespace: flux-system
spec:
  type: msteams
  address: <webhook-url>`

joshri:
@opudrovs in regards to the applications tab still being highlighted - the mui tabs component must always have a valid active tab value otherwise you get long console errors. I can look into overriding this more if it's a big issue

opudrovs:
@joshri 

> try using this yaml

Thanks, will try it.

> in regards to the applications tab still being highlighted - the mui tabs component must always have a valid active tab value otherwise you get long console errors. I can look into overriding this more if it's a big issue

I would prefer nothing to be selected in the sidebar if there is no Notifications item there, because might look misleading to the user (like we are still in the Applications page). But if it's hard to do it's not a big issue, we can wait until someone actually complains. @ozamosi do you agree or would you prefer to have all sidebar items deselected while we are at the Notifications page?

ozamosi:
> But if it's hard to do it's not a big issue, we can wait until someone actually complains. @ozamosi do you agree

Yeah, sounds good. All else being equal I'd not highlight it, but I can also make up excuses for why this behaviour is technically correct (the notifications relates to your flux resources e.g. applications/sources, so we highlight that section header).

opudrovs:
@joshri clicking on added chip with `-` does not remove it. Clicking on any other chip removes it, as expected

opudrovs:
Filtering by this hard to read channel displays empty rows:

Screenshot 2022-09-02 at 21 47 34

Screenshot 2022-09-02 at 21 47 47

opudrovs:
And when selecting all channels only the `general` channel is displayed:

Screenshot 2022-09-02 at 21 53 40

opudrovs:
Please consider adding tests for chips with hyphens/undefined values.

opudrovs:
I don't think that it is related to this PR but the chips when they appear slightly push content down (no matter if FilterDialog pushes content to the left or is displayed over it). I think some space should be reserved for chips:

https://user-images.githubusercontent.com/8824708/188232247-da84920e-799a-4c69-819d-b339cf5b8b2b.mov

joshri:
Inability to remove chips with '-' was bc I changed the value of the actual chip in ChipGroup. Added the filterSeparator.

The long channel string included the filterSeparator which was `:`. Changed it to something that no one will have in there stuff: ` --- `

I also added the filterSeparator variable to all the tests so we can actually change it without breaking everything.

ozamosi:
Could we please pause and take a step back.

We should be able to filter by empty value. This is a really good change. However, this seems like a separate enhancement, so we can make sure it works well together with e.g. tenants.

Among these filtering fixes there's also an important regression fix. This would also be good to have outside of this PR, so we can call it out in our changelog (i.e. "filtering fixes" as a bug fix PR that calls out the regression - not several bug fix PRs).

As for how to filter by empty value, it seems to me that the quick solution didn't work, and we need to solve the problem by structuring the data, so that the representation doesn't matter - as in, if a chip had a property "field" and a property "value", then whether it's rendered as "field:value" or "value@field" or even "I'm feeling lucky" wouldn't matter, because when determining what filter to remove we'd be able to just read the properties.

joshri:
As discussed in standup:

Filtering changes removed from this branch - this leaves some bugs with search and filtering remaining (searching does not work due to ListObjects object restructuring in Sources and Notifications tables, filterDialog displays nothing for undefined values but has a working(?) checkbox). I moved the work I had done on these into the branch `filter-undefined` in case we want to use them later.

The Applications tab being highlighted on Notifications is also something I don't want to address in this PR. The only ways I can see to override the required `value` prop are either creating a tab in the nav for notifications and overriding all the styling on it to make it invisible (ew), or to not use the Tabs component from Mui, both of which are large changes. 

@opudrovs @ozamosi take another look at the code and the notifications table locally for me and make sure I didn't break it with my lovely interactive rebase THANKS

opudrovs (approved):
LGTM! 🎉

opudrovs:
Tested it, works fine, I didn't test filtering because it will be added in the another PR.

ozamosi (approved):

joshri:
> Can we add delays to the tooltip in `UserSettings` to prevent tooltip from blinking when the user knows well what they are doing and clicks on the icon fast. Smth. like:
> 
> ```
> <Tooltip title="Account settings" enterDelay={500} enterNextDelay={500}>
> ```

Yep that exists as a prop in the mui component as `enterDelay`

opudrovs:
@joshri I guess `enterNextDelay` is required too. Otherwise it will appear too fast if it already appeared once.

PR #2679, opened 2022-09-01
Fix displaying errors with useListSources

ozamosi:
It seems I tried to be more clever than I am, and as a result when
there was an error all you'd see was the header "Error" with nothing
in it.

Replace the code with less clever code, so we can now once again see
errors in sources.


opudrovs (approved):
Tested it with some errors which were luckily returned because of some issues with the cluster state, hope those were the right ones. Tested it with the new cluster without errors. LGTM.

PR #2685, opened 2022-09-01
Add Sync All functionality 

### LINKED ISSUE #2661: Add Sync Button to CheckboxActions component with sync all functionality
https://github.com/weaveworks/weave-gitops/pull/2649 got quite large. Sync All functionality has been split out into a separate story - this one! 

AC:
- The backend request should be modified similar to suspend.go's modifications for suspend all
- The frontend needs the SyncButton component added to CheckboxActions as the farthest left button
- Will probably have to add a couple tests in the backend similar to the new suspend_test.go tests
- Add padding on the right side of CheckboxActions to space it from ChipGroup

joshri:
Closes #2661 

Sync request now accepts an array of objects to allow for sync all functionality, and the sync button has been added to the CheckboxActions component. The toasts have been moved to the `onSuccess` and `onError` of the query. I had the requests time out one or two times...couldn't really link it to anything in particular so I'm curious to see what happens for y'all <|:^|

https://user-images.githubusercontent.com/65822698/187982993-1e94ce75-556a-46ec-a9a5-102676da23d2.mov


api/core/core.proto
```diff
@@ -318,11 +318,8 @@ message ListEventsResponse {
 }
 
 message SyncFluxObjectRequest {
-    string         name        = 1;
-    string         namespace   = 2;
-    FluxObjectKind kind        = 3;
-    string         clusterName = 4;
-    bool           withSource  = 5;
+    repeated SuspendReqObj objects    = 1;
```
opudrovs, line 321: Why does `SyncFluxObjectRequest` use `SuspendReqObj` too? I think it should have its own object type, even if fields are the same? Or if it is necessary to share this object type between both `ToggleSuspendResourceRequest` and `SyncFluxObjectRequest` the type name should be more generic? @ozamosi , what do you think about it?

joshri: Honestly I just didn't come up with a name I liked - SyncSuspendReqObj is just....bleh you know? If you have a good name it's an easy change

ozamosi: So. I looked what we have.
* This is a `FluxObjectRef`, except with `clusterName`. `FluxObjectRef` is used together with a `clusterName` field in both kustomizations and helmreleases. So it shouldn't really be added there, instead the pattern with that one is to have one field with the object ref and a different field with the clusterName.
* That in turn is basically the same as `ObjectRef` except with the kind as an enum instead of a string. `ObjectRef` is only used by events, which do not specify cluster at all, which is incorrect.

`listObjects` spirits be willing, `FluxObjectRef` won't need to be a thing for long, so that duplication will be got rid of.

`ObjectRef` needs to grow a `clusterName` regardless of what we do here so events work properly in multi cluster.

I don't like the `FluxObjectKind` enum because it's not extensible, so I'd rather not use it.

So, this should just be an (improved version of) `ObjectRef`. Done, right?

However, the enummy bits are quite ingrained in the suspend/sync code, so that would grow the scope of this ticket quite a bit. So, no, out of scope.

Honestly, let's call it `ClusteredObjectRef` and put it next to the other object refs, thereby maximising the chance that it will nerd snipe someone into merging them all.

joshri: Great work Robin thanks for looking into it. Will implement your change shortly

opudrovs: @ozamosi great breakdown, Robin. ⭐

opudrovs (approved):
LGTM! Great job! 🎉

PR #2686, opened 2022-09-02
Makes the UI UnAuthorizedInterceptor TS generic

foot:
- So that consuming libs can more easily use it with additional API providers

**What changed?**

Switch to some generic types in the UnAuthorizedInterceptor so it can be used by other consuming projects which also use the Auth middleware in the BE.

**Why was this change made?**

To avoid having to:

```ts
  const wrapped = UnAuthorizedInterceptor(api as unknown as typeof Core) as unknown as typeof MyService;
```


ozamosi (approved):
Could we maybe move this to e.g. `AuthContext` if it's supposed to be generic? Otherwise, I'll make this more specific again the minute I forget why these types are here.

makkes:
@foot would you like to still merge this PR?

foot:
No its fine for now

PR #2714, opened 2022-09-06
Add filtering by undefined and fix searching by name

### LINKED ISSUE #2711: Searching by name does not work on Sources page
**Describe the bug**
The search function is not working in the SourcesTable component. I suspect it's because the structure of the Source object has changed recently. Our tests are more unit than integration, and so the tests have been passing with no issues, as they are expecting for the input to be formatted a certain way. Is this something that can be improved? Some existing work on this is living in the branch `filter-undefined`

MY PODINFOS
image

SEARCHING "P"
image

**Environment**

image

**To Reproduce**
Search by the name of an existing source in the Sources page. It's MISSING.

**Expected behavior**

Searching by fragments or whole string of the name field on all tables should show matches correctly.

**Actual Behavior**

Searching by any string filters away all resources

### LINKED ISSUE #2712: Filter dialog + chips do not display anything for undefined values
Filtering wasn't implemented with undefined values in mind. It........looks like it works? The problem is in the actual display, where you get nothing after the `:` in the chip, and nothing after the checkbox in the dialog. Fixing this quickly spirals into potentially rebuilding filtering from the ground up, as it uses these string values to accurately filter the data. It's all very integrated and not easily customizable. Existing work on this is living in the branch `filter-undefined`

image

**Environment**

image

joshri:
Closes #2711 
Closes #2712

- Fixed a bug where filter categories would be disabled if their first value was undefined
- Fixed display of undefined values in data across the app (in all tables, filter dialogs, and chips). Empty data is now represented by `-`
- Fixed a bug in the `filterText` function that didn't allow searching by name with the new object structure (name can be a function instead of a string). The function now finds the row's value as other filtering functions do - with `field.value(row)` or `row[field.value]`, which supports both strings and functions.
- Fixed a styling bug that adjusted content when chips appeared. ChipGroup is now set at 40px 
- Changes in tests allow for us to change what the filter separator is without breaking everything. I also added a test in `ChipGroup` to test if changing chips that end with the filter separator add a `-` to properly display undefined.

Quick vid showing filtering and searching on the notifications page:

https://user-images.githubusercontent.com/65822698/188746495-ef1d15b4-91bf-4134-b7fd-e2d9f9778a95.mov


ui/components/DataTable.tsx
```diff
@@ -161,31 +161,28 @@ function filterText(
...
-      const field = _.find(fields, (f) => {
-        if (typeof f.value === "string") {
-          return f.value === colName;
-        }
-
-        if (f.sortValue) {
-          return f.sortValue(row) === value;
-        }
-      });
+    fields.forEach((field) => {
```
ozamosi, line 165: I think this needs to be simplified further. Let's say that both name and namespace (in that order) is text searchable. If you search for 1 text filter that matches the name, then I think the namespace would set matches back to false. So I think the matches filter needs to be more carefully passed from "layer to layer", instead of just building up one value as we loop.

The text filters seem like they should be ANDed based on the tests, so I'm going to assume that's correct.

To make this work properly, I think we need to:
 * First (inner-most) OR each field together - any field that matches a single text filter means that the whole row matches that filter.
 * Then, AND the filters together.
 * And only then do we decide if the row as a whole matches the filters as a whole.

We're currently merging the filters before we're merging the columns, which I think doesn't work properly with multiple textSearchable columns - if again name and namespace are searchable, you wouldn't get a hit if one filter matches name and one filter matches namespace, only if all filters matches either name or namespace.

So I would change this whole loop to:

```
  return rows.filter((row) => {
    return textFilters.every((filter) => {
      return fields.some((field) => {
        if (!field.textSearchable) return false;

        let value;
        if (field.sortValue) {
          value = field.sortValue(row);
        } else if (typeof field.value === "function") {
          value = field.value(row);
        } else {
          value = row[field.value];
        }

        return value.includes(filter);
      });
    });
  });
```

Perhaps we could even turn that `value` extraction into some helper function. 

I like chaining together `every` and `some` because they help separate "checking the condition" from "combining the outputs" so they take care of the `matches` malarkey for us - we only return a boolean, and the right thing comes out, and they even handles short-circuiting for us.

ui/components/ChipGroup.tsx
```diff
@@ -19,7 +20,14 @@ function ChipGroup({ className, chips = [], onChipRemove, onClearAll }: Props) {
       {_.map(chips, (chip, index) => {
         return (
           <Flex key={index}>
-            <Chip label={chip} onDelete={() => onChipRemove([chip])} />
+            <Chip
+              label={
+                chip.slice(-filterSeparator.length) === `${filterSeparator}`
```
ozamosi, line 25: Either do something like `chip.indexOf(filterSeparator) == chip.length - filterSeparator.length` (I probably have an off-by-one in there, I haven't tried it) so it 100% only manipulates empty chip values, or just drop this.

ui/components/DataTable.tsx
joshri: I don't think it was ever intended to text search any column other than name - I took a look through and no tables have more than one column searchable. I'm down to add this because why not - but to me it'd be weird to have something be filterable AND text searchable.

joshri: Should I add textSearchable on all Namespace and Cluster columns? Any other columns that might be useful?

opudrovs:
How the user determine which fields are searchable? I see the textSearchable prop, but only as a developer, is there a visual indication for the fields you can search? The first thing I tried searching was by channel and type, it didn't work. Then I read that search is enabled only for names.

ui/components/ChipGroup.tsx
opudrovs: @joshri why are you using a template literal

```
`${filterSeparator}`
```

here instead of just `filterSeparator`? Was the comparison not working?

opudrovs:
If using the Chip component imported from Material-UI causes such problems, can we just create our own Chip component which has two props, `label` and `value` (or maybe not `label` to avoid confusion with the original Chip component), and combines them as the label of the original Chip  component it renders with a filterSeparator const, like <OriginalChildComponent label={`${OurChipComponentLabelProp}${filterSeparator}${OurChipComponentValueProp}`}>? So, we can always know that we are not messing up a real value.

joshri:
Created some follow-up issues - for now I've removed the hack in `ChipGroup` and left the `filterByText` function as is

ozamosi (approved):

opudrovs:
The UI tests are failing, including locally for me.

ui/components/DataTable.tsx
```diff
@@ -161,31 +161,28 @@ function filterText(
...
+      if (!field.textSearchable) return matches;
 
-      if (!field || !field.textSearchable) {
-        continue;
+      let value;
+      if (field.sortValue) {
+        value = field.sortValue(row);
+      } else {
+        typeof field.value === "function"
```
opudrovs, line 172: Can we re-write this ternary as:

```
value =
          typeof field.value === "function"
            ? field.value(row)
            : row[field.value];
```

It's a standard form of an assignment with ternary instead of having multiple assignments.

ui/components/DataTable.tsx
```diff
@@ -161,31 +161,28 @@ function filterText(
...
+          ? (value = field.value(row))
+          : (value = row[field.value]);
       }
 
-      // This allows us to look for a fragment in the string.
-      // For example, if the user searches for "pod", the "podinfo" kustomization should be returned.
-      for (const filterString of textFilters) {
-        if (_.includes(value, filterString)) {
+      for (let i = 0; i < textFilters.length; i++) {
+        if (value.includes(textFilters[i])) {
```
opudrovs, line 178: This can be refactored as

```
matches = value.includes(textFilters[i]);

        if (!matches) {
          break;
        }
```

It would be easier to read.

opudrovs:
Is it expected behavior that an empty filter dialog is displayed for the events table? I don't think it's specific to this PR.

I remember fixing sorting in the events table but don't remember working with filters in it. Is the events table supposed to have filters or should we add hiding the filter dialog button for it?

Screenshot 2022-09-12 at 20 36 38

opudrovs (approved):
Tested it, have not found a way to break anything! :tada:

joshri:
@opudrovs 

"Not Ready" wrapping - I don't believe it's intentional that it's wrapping, but I also don't see any limits on the width of the status column 

Events Table empty filters - we aren't providing a filter state to the table so it makes sense that it's empty. We should probably disable search and filtering on this table with a boolean prop on DataTable

joshri:
We decided to make the filter dialog optional on the data table since the events table has never had any filtering functionality

PR #2720, opened 2022-09-07
Enable stylecheck lint rule

ozamosi:
This lint rule has some pretty good checks - for example the "error strings shouldn't be capitalized", so we can always log errors properly, or "duplicate imports" that make the code look more complex than it is.

However, to enable it required me to search replace all sorts of things to capitalize them: Url with URL, Uri with URI, Http with HTTP, Json with JSON, etc.

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -5,5 +5,6 @@ run:
 linters:
   enable:
     - wsl
+    - stylecheck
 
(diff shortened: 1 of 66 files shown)


opudrovs (approved):

opudrovs:
LGTM, great job! 🎉

PR #2722, opened 2022-09-08
gitops run: Create empty kustomization if target directory doesn't exist

### LINKED ISSUE #2698: When gitops run is started, create an empty entrypoint kustomization if the target directory doesn't exist
If the workload directory (the directory the user specifies on the CLI) doesn't exist, create it, and put an empty kustomization in it.

If a user:
* Creates a new initialized repository in github (initialized in order to remove a couple of steps)
* Clones it.
* cd into it, and runs `gitops beta run ./workload` even though that directory does not exist.

Then:
* By the time gitops run is started, the directory `./workload` should exist.
* It should contain an empty placeholder k8s kustomization file (e.g. similar to [this](https://github.com/ozamosi/scenario-2-gitops-run-dotdot-workload/blob/main/workload/kustomization.yaml)).
* When the user adds files to it, they should be automatically updated.

ozamosi:
If I run `gitops beta run ./this-is-a-new-directory`, I will now end up with a kustomization that turns green in flux. The same thing happens if I run `mkdir directory && gitops beta run ./directory`.

I chose to hard-code the kustomization as text instead of as an object to marshal, because I wanted to put comments in there.

This closes #2698

### DIFF
--- a/cmd/gitops/beta/run/cmd.go
+++ b/cmd/gitops/beta/run/cmd.go
@@ -332,6 +332,10 @@ func betaRunCommandRunE(opts *config.Options) func(*cobra.Command, []string) err
 			}
 		}
 
+		if err := run.InitializeTargetDir(targetPath); err != nil {
+			return fmt.Errorf("couldn't set up against target %s: %w", targetPath, err)
+		}
+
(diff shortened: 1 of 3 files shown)


bigkevmcd (commented):
Looks mostly ok to me.

opudrovs:
Tested it, the directory with the empty kustomization is created.

opudrovs (approved):

PR #2723, opened 2022-09-08
Add Provider Detail view with Yaml and Alerts

### LINKED ISSUE #2566: Allow users to click into notifications to get more details
As a platform operator, I can click into a notification using the name so that I can see details on how that notification is set up on a new page
    - The new page will show me information on Alerts, Provider, and Events as individual tabs that can be clicked into
Figma Link: https://www.figma.com/file/IVHnM9iyeFWpd11evtY8ux/Weave-GitOps?node-id=4908%3A9125

### LINKED ISSUE #2567: Add Alerts tab details
As a platform operator, I can click into the Alerts tab to see the alert objects that are used by this provider with the details of how events are filtered by severity and involved object, and what provider to use for dispatching so that I can make sure the notification is configured correctly 
Figma Link: https://www.figma.com/file/IVHnM9iyeFWpd11evtY8ux/Weave-GitOps?node-id=4908%3A9125

image

### LINKED ISSUE #2568: Add Provider tab details
As a platform operator, I can click into the `YAML` tab to see a yaml file of  how events are encoded and the webhook address where they are being sent so that I can make sure the notification is configured correctly

Figma Link: https://www.figma.com/file/IVHnM9iyeFWpd11evtY8ux/Weave-GitOps?node-id=4908%3A9125
Note: screenshot is alert data below but it should display provider data.
image

joshri:
Closes #2566 
Closes #2567 
Closes #2568 

- Providers in the `NotificationsTable` are now clickable links to the new `ProviderPage` page
- This page fetches the individual provider with `useGetObject`, and passes it down to the new `ProviderDetail` containing `YamlView` and `AlertsTable` tabs.
- The new `AlertsTable` takes the Alerts from an individual provider using the new `useListAlerts` as rows. This hook grabs all Alerts and filters them by `providerRef` to grab the correct ones for the table.
- Added `Alert` as an extension of `FluxObject`.
- Fixed a styling bug that's probably been in `YamlView` for a while - a 5px blank spot was reserved on the right side of the view, likely for a scrollbar. I removed horizontal scrolling on the view (it already pre-wraps), and extended the header all the way across. 

The Events tab is impossible at the moment due to the flux notification-controller not recording kubernetes events. There is a remaining issue in the notification epic for a special display where no notifications are visible, and I believe that there were plans to add alert yaml on hover but I don't see an issue for it. 

Clickable providers:

image

Alerts Table:

image

Provider yaml:

image


api/core/types.proto
```diff
@@ -19,6 +19,8 @@ enum FluxObjectKind {
     KindHelmRelease    = 5;
     KindCluster        = 6;
     KindOCIRepository  = 7;
+    KindProvider       = 8;
+    KindAlert          = 9;
```
ozamosi, line 23: For my personal notes, do you remember why you needed to add this?

ui/lib/objects.ts
```diff
@@ -199,3 +206,15 @@ export class Provider extends FluxObject {
     return this.obj.spec.channel || "";
   }
 }
+
+export class Alert extends FluxObject {
+  get providerRef(): string {
+    return this.obj.spec.providerRef.name || "";
```
ozamosi, line 212: Shouldn't this have a namespace as well? 

Most refs in flux (see: sourceRef) have an optional namespace, where no value means "same as current" so to make references correct I keep finding myself having to add if statements to discover the correct namespace - I would expect to see something similar here.

ui/components/AlertsTable.tsx
```diff
@@ -0,0 +1,105 @@
...
+      value: "severity",
+    },
+    {
+      label: "Event Sources",
+      value: (a) => {
+        return (
+          <ul className="event-sources">
+            {a?.eventSources?.map((obj: CrossNamespaceObjectRef) => (
+              <li className="event-sources" key={obj.name}>
+                {obj.kind}: {obj.name}
```
ozamosi, line 53: Should this not (also, I wrote more in the other comment) have a namespace?

api/core/types.proto
joshri: Great question. I had to go back to look - the `kind` getter in the flux object class uses FluxObjectKind - so everything will crash when you try to use Provider and Alert as Objects unless they're in here - for whatever it's worth it also crashes the YamlView which I think is the only place where the Provider and Alert kinds will actually be looked at

` get kind(): FluxObjectKind | undefined {
    if (this.obj.kind) {
      return FluxObjectKind[addKind(this.obj.kind)];
    }
  }`

ui/lib/objects.ts
joshri: The docs led me here as the type for `providerRef` - is this the same thing like the uid where every object actually has one?
https://pkg.go.dev/github.com/fluxcd/pkg/apis/meta?utm_source=godoc#LocalObjectReference

ozamosi: Oh, you're right, it's always local. Nevermind, then.

Still valid for event sources, though.

api/core/types.proto
ozamosi: Ah. Hmpf. Of course. Thank you!

ui/components/AlertsTable.tsx
joshri: This can definitely have namespace - how do you feel about:
`{obj.kind}: {obj.name}, {obj.namespace}`
Was it supposed to be namespace instead of name? Trying to follow the original design:
image

joshri:
It absolutely was related to this PR - good spot. Had the wrong value in the filterConfig function for event severity!! Fixed.

image

opudrovs:
Could you please rebase on main so that I could test search in the events table? Search is supposed to be working in this table, correct?

ui/components/AlertsTable.tsx
ozamosi: I think the comma could be confusing - looks a bit like it's doing some kind of `{kind}: {objs.filter(o => o.kind == kind).map(o => o.name).join(', ')}`.

I think I would do `{obj.kind}: {obj.namespace}/{obj.name}`

I almost said "because that's how it's usually displayed" and then I realised that it's not (if you run `flux get all -A` you get the `{obj.kind}/{obj.name}` in one column, and namespace in another one, and if you run `kubectl port-forward --help` then you'll see that you can also use `{obj.kind}/{obj.name}`) - but I can't think of anything better that's still somewhat faithful to the original design.

ozamosi (approved):

PR #2727, opened 2022-09-13
remove display name annotation from docs

ahussein3:
-Resolve [#1508](https://github.com/weaveworks/weave-gitops-enterprise/issues/1508)


foot (approved):
💯

PR #2733, opened 2022-09-13
Automatically raise PRs against EE when PRs are merged, released

ozamosi:
This adds 2 places to raise 2 different PRs: one for releases, one for merges.

In each case, this will
* Check out the latest main of EE
* Merge with the latest OSS tracking branch, if there is one. The branch name is always the same, so it knows which one it is.
  * This means I can push other changes to the tracking branch to make it green, and have this thing collaborate with me. 
  * This will fail the pipeline and stop if there's merge conflicts. This is on purpose - this is supposed to be an assistant, so if you've taken control of the process this will stop. You can choose to rebase the PR manually, or you can close the PR - either way, this'll start back up next time.

For complicated permission reasons, this uses two different bot accounts - one for accessing EE, and a different one for building EE. Neither one account has access to all the things I need, so two it is.

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -244,7 +244,8 @@ jobs:
     if: github.event_name == 'push'
     permissions:
       packages: write
-
+    outputs:
+      js-version: ${{ steps.package-version.outputs.js-version }}
     steps:
       - name: Checkout
         uses: actions/checkout@v2
@@ -258,8 +259,10 @@ jobs:
       - run: npm install
       - run: make ui-lib
       - name: Update package version
+        id: package-version
         run: |
           GITOPS_VERSION=$(git describe)
+          echo "::set-output name=js-version::$GITOPS_VERSION"
           jq '.version = "'$GITOPS_VERSION'" | .name = "@weaveworks/weave-gitops-main"' < dist/package.json > dist/package-new.json
           mv dist/package-new.json dist/package.json
           cp .npmrc dist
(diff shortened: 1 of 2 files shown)


.github/workflows/pr.yaml
```diff
@@ -291,6 +294,64 @@ jobs:
...
+          git config user.email wks-ci-test-bot@weave.works
+          git fetch origin
+          if git rev-parse -q --verify origin/track-latest-oss; then
+              git merge origin/track-latest-oss
+          fi
+      - name: Upgrade main version
+        run: |
+          go get -u github.com/weaveworks/weave-gitops@${{ github.event.after }}
+          go mod tidy
+          cd ui-cra && yarn add @weaveworks/weave-gitops@npm:@weaveworks/weave-gitops-main@${{ needs.ci-publish-js-lib.outputs.js-version }}
```
foot, line 333: 🔥 !

Wicked :)

Great start, we might wanna do a `yarn test -u` to update snapshots and things too. (follow up)

foot (approved):
⭐ ⭐ ⭐

ozamosi:
Added:
* `yarn test -u`
* Link to compare view with all commits
* A simple changelog with all PRs.

These are all for the merge flow, not necessarily for releases. I hope the merge flow will be the one we interact with the most, but it shouldn't be hard to move those changes over as well.

foot:
Super cool!

- the test PR is now on a funny base 🤔 , it pulled in commits from `main` into the update branch. After all these years I don't know exactly how that occurs.

ozamosi:
Yes, merge isn't rebase, is it :see_no_evil: - flipped the merge direction.

PR #2735, opened 2022-09-13
Link to filtered tables from alert event sources

### LINKED ISSUE #2734: Alert event sources should be represented by links
The `eventSources` array of an `Alert` has enough info to be represented by a link to an automations or sources table with corresponding filters. The asterisk wildcard character should mean no filters are applied for that field (can only be `name`...MAYBE `namespace` so I'm gonna implement it in case it happens), while falsy values in any field should (probably) lead to no link being rendered.

joshri:
Closes #2734 

**What changed?**
- Event sources are now links to either the automations or sources table depending on the `kind`
- If any field values are falsy, a `li` is shown rather than a `Link`
- The asterisk wildcard character means no filters are added for that key (name and namespace are implemented, although it's unclear to me whether namespace should be in there or not based on https://github.com/fluxcd/notification-controller/issues/71 - it seems like it's not implemented but I figured I'd future proof things.)

### DIFF
--- a/ui/components/AlertsTable.tsx
+++ b/ui/components/AlertsTable.tsx
@@ -66,22 +66,19 @@ function AlertsTable({ className, rows = [] }: Props) {
         return (
           <ul className="event-sources">
             {a?.eventSources?.map((obj: CrossNamespaceObjectRef, index) => {
-              if (obj.name && obj.namespace && obj.kind)
-                return (
-                  <Link
-                    className="event-sources"
-                    key={index}
-                    to={makeEventSourceLink(obj)}
-                  >
-                    {obj.kind}: {obj.namespace}/{obj.name}
-                  </Link>
-                );
-              else
-                return (
-                  <li className="event-sources" key={index}>
-                    {obj.kind}: {obj.namespace}/{obj.name}
-                  </li>
-                );
+              obj.name && obj.namespace && obj.kind ? (
+                <Link
+                  className="event-sources"
+                  key={index}
+                  to={makeEventSourceLink(obj)}
+                >
+                  {obj.kind}: {obj.namespace}/{obj.name}
+                </Link>
+              ) : (
+                <li className="event-sources" key={index}>
+                  {obj.kind}: {obj.namespace}/{obj.name}
+                </li>
+              );
             })}
           </ul>
         );
--- a/ui/components/AlertsTable.tsx
+++ b/ui/components/AlertsTable.tsx
@@ -1,17 +1,35 @@
+import qs from "query-string";
 import * as React from "react";
 import styled from "styled-components";
 import { useFeatureFlags } from "../hooks/featureflags";
-import { Alert, CrossNamespaceObjectRef } from "../lib/objects";
+import { Alert, CrossNamespaceObjectRef, Kind } from "../lib/objects";
+import { V2Routes } from "../lib/types";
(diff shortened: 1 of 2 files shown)


ozamosi:
Sweet!

Just, I think this is missing filtering on kind.

joshri:
> Sweet!
> 
> Just, I think this is missing filtering on kind.

Oops...in taking out the `*` conditional I just...deleted the whole thing. Fixing.

ozamosi (approved):
Limitations: if you click on an event source that has no matches, then you get an empty list of matches (as expected), but you also don't get the chips you should (which isn't expected). This seems unrelated to this change, though.

joshri:
> Limitations: if you click on an event source that has no matches, then you get an empty list of matches (as expected), but you also don't get the chips you should (which isn't expected). This seems unrelated to this change, though.

Maybe we need a something other than "No data" in the data table...a small message like the new notifications and dependencies have?

ui/components/AlertsTable.tsx
```diff
@@ -1,19 +1,35 @@
...
 } from "./DataTable";
+import { filterSeparator } from "./FilterDialog";
 import KubeStatusIndicator from "./KubeStatusIndicator";
+import Link from "./Link";
 type Props = {
   className?: string;
   rows?: Alert[];
 };
 
+export const makeEventSourceLink = (obj: CrossNamespaceObjectRef) => {
```
opudrovs, line 21: Can you please add tests for this function? It's just asking for it. And I think the function return type should be specified explicitly.

joshri: I started writing some tests, and then I was essentially saying to expect the output to be exactly what the return value of the function is, just with the arguments subbed in. Is there a better way to test it than just plugging in a fake obj and expecting the return value?

opudrovs: Normally, yes, you just provide fake objects with different fields, write expected values (the returned string values that you know should be correct for this object) first and then check if what your function produces matched the expected values (of course you can first run the function and copy/paste the actual values to expected but the former is considered "the right way" ™). Tests also serve as documentation and for catching regressions so they should be useful even if they look simple.

joshri: Tests added under `AlertsTable.test.tsx`!!

ui/components/__tests__/AlertsTable.test.tsx
```diff
@@ -0,0 +1,35 @@
...
+        namespace: "space",
+        kind: "GitRepository",
+        matchLabels: [],
+      };
+      const sourceLink = makeEventSourceLink(allNames);
+      expect(sourceLink.includes("/sources")).toEqual(true);
+      expect(
+        sourceLink.includes("type") && sourceLink.includes("GitRepository")
+      ).toEqual(true);
+      console.log(sourceLink);
```
opudrovs, line 18: Great tests!

The only question: does this call to `console.log` have any special purpose?

joshri: UGH. I'm fired. Thanks for finding that

opudrovs: *side-whispering* Great that I cover my tracks well. Since _some_ totally random time I always search in files for `console.log` and `fmt.Println` before submitting my PRs for review. 🤣

PR #2746, opened 2022-09-15
Add an optional choice of NodePort

kazukousen:
add an option to explicitly specify the nodePort value when service's type is nodePort.


ozamosi (approved):
Thank you!

PR #2748, opened 2022-09-15
Add pop up for Alert yaml

### LINKED ISSUE #2741: Alert Yaml
Display a modal containing the alert yaml when you click on an alert name.

joshri:
Closes #2741 

Adds a Mui Dialog component to the AlertsTable that displays an Alert's yaml when clicked. The pseudo scrollbar element styling added a lot of unnecessary scrollbars to the YamlView, so a couple changes were made in a class and one Mui override to get things looking good.

https://user-images.githubusercontent.com/65822698/190470818-8f79c348-f38f-4570-bca4-204cfb065136.mov

### DIFF
--- a/ui/components/AlertsTable.tsx
+++ b/ui/components/AlertsTable.tsx
@@ -79,7 +79,7 @@ function AlertsTable({ className, rows = [] }: Props) {
         return (
           <ul className="event-sources">
             {a?.eventSources?.map((obj: CrossNamespaceObjectRef, index) => {
-              obj.name && obj.namespace && obj.kind ? (
+              return obj.name && obj.namespace && obj.kind ? (
                 <Link
                   className="event-sources"
                   key={index}
--- a/ui/components/AlertsTable.tsx
+++ b/ui/components/AlertsTable.tsx
@@ -15,7 +15,7 @@ import { filterSeparator } from "./FilterDialog";
 import KubeStatusIndicator from "./KubeStatusIndicator";
 import Link from "./Link";
 import Text from "./Text";
-import YamlView from "./YamlView";
+import { DialogYamlView } from "./YamlView";
 type Props = {
   className?: string;
   rows?: Alert[];
@@ -55,11 +55,7 @@ function AlertsTable({ className, rows = [] }: Props) {
     {
       label: "Name",
       value: (a) => (
-        <Text
-          onClick={() => setYamlView(a)}
-          color="primary10"
-          className="pointer"
-        >
+        <Text onClick={() => setYamlView(a)} color="primary10" pointer>
           {a.name}
         </Text>
       ),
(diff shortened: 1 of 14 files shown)


opudrovs:
The modal is working for me as expected. If I don't see links under Event Sources now with the same YAMLs — is it normal?

Screenshot 2022-09-15 at 21 13 24

Screenshot 2022-09-15 at 21 13 56

opudrovs:
Or do I now need some special YAMLs to test event sources links?

opudrovs:
The `matrix` notifications provider never gets ready for some reason, I think it was ready and had links before.

`Service/flux-system/notification-controller created` 

Maybe I need to upgrade smth.? Will try to reinstall Flux now.

ui/components/AlertsTable.tsx
```diff
@@ -46,10 +49,20 @@ function AlertsTable({ className, rows = [] }: Props) {
...
+
   const alertFields: Field[] = [
     {
       label: "Name",
-      value: "name",
+      value: (a) => (
+        <Text
+          onClick={() => setYamlView(a)}
+          color="primary10"
+          className="pointer"
```
opudrovs, line 61: Passing from outside the class defined within a component is not done usually. The className is reserved for passing external styling/classes. The external components are not supposed to know anything about internals of the Text component. It's better to add a prop for this. Either an optional useHandPointer or smth. like that or an optional string prop to pass a cursor type and apply it within the component like:

`${(props) => props.cursor && `cursor:${props.cursor};`};`

ui/components/Text.tsx
```diff
@@ -42,6 +42,9 @@ const Text = styled.span<TextProps>`
 
   ${(props) => props.noWrap && "white-space: nowrap;"};
```
opudrovs, line 43: We are not adding semicolons in styled components styling we can just remove it too for consistency.

ui/components/AlertsTable.tsx
```diff
@@ -103,12 +116,23 @@ function AlertsTable({ className, rows = [] }: Props) {
...
+        fields={alertFields}
+        rows={rows}
+        filters={initialFilterState}
+      />
+      <Dialog open={yamlView !== null} onClose={() => setYamlView(null)}>
+        {yamlView && (
+          <YamlView
+            object={yamlView}
+            yaml={yamlView?.yaml}
+            className="dialog"
```
opudrovs, line 131: Same here: passing internal className from the outside. It would be better to use a boolean prop (`isDialog`?) for this.

opudrovs:
@joshri if classnames feel more intuitive to you, keep them. Using classnames per se is not an issue even if they sometimes can cause issues with specificity (= lead to hard to debug bugs):
https://styled-components.com/docs/advanced#issues-with-specificity

I feel here it breaks one of the main principles of React — unidirectional data flow:
https://www.educative.io/answers/what-is-unidirectional-data-flow-in-react

The parent component somehow "knows" inner classes of child component and gets this class name, declared within the child component, and passes it back to the child component like a "magic string" (a hardcoded values for which it's not obvious how it was obtained (a developer reading the code will first start looking for this value in the parent component instead of the child component) and which is easy to misspell).

Maybe if keeping to use classnames we could at least add exported string constants in the child component for classnames, import them in the parent component, to show where the data (class name string) used comes from more explicitly.

joshri:
Added a `DialogYamlView` styled component, and added pointer as a prop in Text. I think it's fair to say that styled components aren't meant to be used with class names like that and that's it's worth changing! As we come across it in other components we can change it.

opudrovs:
That's fine with me. Tested it, works as expected. 🎉

PR #2752, opened 2022-09-19
Move `MessageFlex` to a separate `MessageBox` component.

opudrovs:
- As discussed with @joshri, renamed `MessageFlex` to `MessageBox` and moved it to a separate file (to avoid using the ugly `styled(Styled)` wrapper when it was placed in the `Flex` component file module). Also added `MessageBox` to exports in `index.ts`.

- While on it, I repalced the margins in the No Dependencies message with the `Spacer` component for consistency with the project in general and the No Notifications message in particular.


ui/components/NotificationsTable.tsx
```diff
@@ -100,7 +101,7 @@ function NotificationsTable({ className, rows }: Props) {
     return (
       <Flex wide tall column align>
         <Spacer padding="xxl" />
-        <MessageFlex column align={false}>
+        <MessageBox column align={false}>
```
joshri, line 104: align={false} can be deleted I think since I was overriding the default props. If no value is provided align-items will be `start`

joshri (approved):

PR #2760, opened 2022-09-19
Fix console warning about styled components

joshri:
I noticed a console warning about declaring styled components inside the render function in DependenciesView - moving the Text withComponent lines outside of the main function fixed the issue!!

Screenshot of the warning:
image


opudrovs:
@joshri good catch! I cannot confirm this error because it seems it happens only for `npm start` and I see the following error when trying to open the app at the 4567 port after npm start now, in both the main branch and the fix branch:

Screenshot 2022-09-20 at 02 15 36

But at least the app with the fix works for me as before with tilt up.

opudrovs (approved):

PR #2765, opened 2022-09-20
List automations objects

### LINKED ISSUE #2763: Remove listKustomizations/listHelmReleases

joshri:
Closes #2763 

This is a branch off of #2751. It extends the useListObjects hook to Automations (Kustomizations and HelmReleases) - this then allowed for a BUNCH of things to be deleted, which caused a bunch of problems, and so on and so on. The main things being deleted are the full kustomization and helmrelease.go files, and the change of the FluxObjectKind type to just Kind, which then removes a bunch of conversion utilities such as addKind and removeKind. 

There are probably other things that can be deleted - extra types like Syncable were on my radar. But this is certainly a start.


api/core/core.proto
```diff
@@ -227,7 +186,7 @@ message GetObjectResponse {
 
 message ListObjectsRequest {
     string namespace   = 1;
-    string kind        = 2;
+    Kind   kind        = 2;
```
ozamosi, line 189: Could we please turn this (both listObjects and getObject) back to string? I really would like to preserve the feature that these APIs can be embedded into other products that their own add additional kinds to list objects, without requiring updating the core APIs.

api/core/types.proto
```diff
@@ -10,38 +10,32 @@ message Interval {
...
 
-message FluxObjectRef {
-    FluxObjectKind kind      = 1;
-    string         name      = 2;
-    string         namespace = 3;
-}
-
 message ObjectRef {
-    string         kind      = 1;
+    Kind           kind      = 1;
```
ozamosi, line 32: :x: I missed this before - this changes the same "can't dynamically add more kinds and have the endpoint work" thing for listEvents that I was mentioning for listObjects and getObject

core/server/fluxruntime.go
```diff
@@ -164,18 +164,18 @@ func (cs *coreServer) GetReconciledObjects(ctx context.Context, msg *pb.GetRecon
 	var opts client.MatchingLabels
 
 	switch msg.AutomationKind {
-	case pb.FluxObjectKind_KindKustomization:
+	case "Kustomization":
```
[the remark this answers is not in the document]
joshri: This is why it felt weird to me to delete `FluxObjectKind` all together, since if I do a typed thing here it has to match on the frontend request. If the type field on objects is `string` and not `Kind` in the frontend this will not work.......I think? At least that's what I thought was going on while I worked on this. I guess I can convert the string going in into this type and add a slot for "other" or something?

api/core/types.proto
joshri: Same thing here - every time we reference ObjectRef / ClusteredObjectRef in the backend this will have to be a string and not typed?

joshri:
@ozamosi All strings of flux primitives have been changed to their flux controller types - tests are going but things look good from my end. Check it out <|:^)

opudrovs (approved):

ozamosi (approved):

joshri:
Hi @amulyakish!!

The only code remaining in `helm_release.go` is the `getHelmReleaseInventory` function, which is tested in `objects_test.go`. I can see how this might be confusing and I probably should have moved the function into `objects.go`. Was there any other code you noticed that is remaining from the test files that were deleted? Or any bugs that you've come across since this was merged?

amulyakish:
> Hi @amulyakish!!
> 
> The only code remaining in `helm_release.go` is the `getHelmReleaseInventory` function, which is tested in `objects_test.go`. I can see how this might be confusing and I probably should have moved the function into `objects.go`. Was there any other code you noticed that is remaining from the test files that were deleted? Or any bugs that you've come across since this was merged?

No I haven't noticed any errors as such. I'm yet to test this in fact. My comment was based only on the list of changes that I saw in this PR. I missed the fact that the `getHelmReleaseInventory` function is tested in `objects_test.go`. It is not very obvious unless I search the whole codebase. I think we are good for now. I will report back once I've had a chance to test this. It will probably be after the next release though.

Thank you for all the work you've done here!

PR #2766, opened 2022-09-20
increase kubeClientTimeout and add environment variable option

rparmer:
Fixes weaveworks/weave-gitops-enterprise#1549

WGE has been seeing increased timeout issues when connecting to leaf clusters.  This increased the timeout from 8 sec to 30 sec and adds an env var config that can be used to adjust the timeout again if needed.

### DIFF
--- a/core/clustersmngr/factory.go
+++ b/core/clustersmngr/factory.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"net"
+	"os"
 	"sync"
 	"time"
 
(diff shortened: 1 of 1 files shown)


luizbafilho (approved):

yiannistri:
Looks good to me, cc @jmickey as he's introduced the 8 sec timeout for more input.

jmickey (approved):
LGTM. To be completely honest the justification for the original 8 second timeout wasn't exactly scientific, it mostly came down to what felt the most reasonable given the variables in play. 

Keep in mind that if just 1 leaf cluster is slow to respond that all results will be affected. With a timeout of 30 seconds it will take just 1 bad cluster to lead to all results taking (up to) 30 seconds to be displayed. In my discussions with the rest of the Pesto team we felt leaning towards failing quickly in this case is preferable from a user experience point of view.

Having said that, by making it configurable (which was outside the scope of the original ticket that added this timeout) with a 30 second default seems like a perfectly reasonable compromise.

PR #2776, opened 2022-09-23
Add useListProviders to exports for EE

AlinaGoaga:
Just need to add the hook to the exports to be able to show the new Notifications sections in EE. Thanks!


ozamosi:
I think you need `useListAlerts` as well.

PR #2781, opened 2022-09-23
helm: Add network policy

### LINKED ISSUE #2600: Default installation interferes with flux default network policies
**Describe the bug**

When installing weave gitops in a flux2 cluster into the flux-system namespace the application is not accessible using an ingress because flux by default installs network policies that only allow access to port 8080 and weave gitops runs on port 9001 and is not configurable and therefore incompatible with the default flux configuration.

**Environment**

- Weave-Gitops Version v0.9.1
- Flux Version v0.32.0
- Kubernetes version v1.24.3+rke2r1

**To Reproduce**

1. Bootstrap a cluster with flux2
2. Create a helmrelease according to the weave-gitops docs
3. Try to access the pod on port 9001 from somewhere in the cluster

**Expected behavior**

The application can be accessed

**Actual Behavior**

only kubectl port-forward works, no other application in the cluster can access the pod

Screenshot from 2022-08-15 19-25-39

ozamosi:
Our default installation steps tells you to install weave-gitops in flux-system. Flux-system already contains network policies, and weave-gitops doesn't extend them to work, so if you enable network policies and follow our manual, gitops wouldn't work.

This creates enough network policies for weave gitops to work, even if your system default is to deny everything.

This creates network policies for
* Ingress access to the regular dashboard port, and optionally the metrics port - no other ingress should be required.
* Unrestricted egress access - when installed into flux-system this is redundant (flux already provides all pods in the namespace unrestricted egress), but in case it's installed into other namespaces this would make gitops work.

We could look into shrinking the egress permissions, but some of those can be configured (e.g. by configuring the cluster pool), so I'm not going to get into that today.

This resolves #2600.

### DIFF
--- a/charts/gitops-server/templates/network-policy.yaml
+++ b/charts/gitops-server/templates/network-policy.yaml
@@ -1,0 +1,34 @@
+{{- if .Values.networkPolicy.create -}}
+---
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  name: allow-dashboard-ingress
+spec:
+  podSelector:
+    matchLabels:
+      {{- include "chart.selectorLabels" . | nindent 6 }}
+  ingress:
+  - ports:
+    - port: 9001
+      protocol: TCP
+  {{- if .Values.metrics.enabled }}
+    - port: {{ .Values.metrics.service.port }}
+      protocol: TCP
+  {{- end }}
+  policyTypes:
+  - Ingress
+---
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  name: allow-dashboard-egress
+spec:
+  podSelector:
+    matchLabels:
+      {{- include "chart.selectorLabels" . | nindent 6 }}
+  egress:
+    - {}
+  policyTypes:
+  - Egress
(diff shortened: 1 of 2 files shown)


ozamosi:
To test, I was stubborn and stuck to kind, which requires a bit of a complicated setup:

Create a new cluster using this config:
```
kind: Cluster
apiVersion: kind.x-k8s.io/v1alpha4
networking:
  disableDefaultCNI: true # disable kindnet
  podSubnet: 192.168.0.0/16
```

Then run `kubectl apply -f https://raw.githubusercontent.com/projectcalico/calico/v3.24.1/manifests/calico.yaml` and you end up with a network overlay that supports network policies.

I then created a namespace to use and created this reject-all policy:
```
---
apiVersion: networking.k8s.io/v1
kind: NetworkPolicy
metadata:
  name: default-deny-all
  namespace: <the-namespace>
spec:
  podSelector: {}
  policyTypes:
  - Ingress
  - Egress
```
and then made sure I could install and access weave-gitops in that namespace (easiest way: add namespace parameter to the main `helm()` function call in the `Tiltfile`).

opudrovs (approved):

PR #2782, opened 2022-09-23
Caching users clients

luizbafilho:
Closes

**What changed?**

**Why was this change made?**

**Release notes**

**Documentation Changes**

### DIFF
--- a/core/clustersmngr/factory.go
+++ b/core/clustersmngr/factory.go
@@ -95,12 +95,12 @@ type clustersManager struct {
 	clustersNamespaces *ClustersNamespaces
 	// lists of namespaces accessible by the user on every cluster
 	usersNamespaces *UsersNamespaces
+	usersClients    *UsersClients
 
 	initialClustersLoad chan bool
 	scheme              *apiruntime.Scheme
 	newClustersPool     ClusterPoolFactoryFn
 	kubeConfigOptions   []KubeConfigOption
-
 	// list of watchers to notify of clusters updates
 	watchers []*ClustersWatcher
 }
@@ -135,6 +135,7 @@ func NewClustersManager(fetcher ClusterFetcher, nsChecker nsaccess.Checker, logg
 		clusters:            &Clusters{},
 		clustersNamespaces:  &ClustersNamespaces{},
 		usersNamespaces:     &UsersNamespaces{Cache: ttlcache.New(userNamespaceResolution)},
+		usersClients:        &UsersClients{Cache: ttlcache.New(userNamespaceResolution)},
 		log:                 logger,
 		initialClustersLoad: make(chan bool),
 		scheme:              scheme,
@@ -285,6 +286,11 @@ func (cf *clustersManager) GetImpersonatedClient(ctx context.Context, user *auth
 		return nil, errors.New("no user supplied")
 	}
 
+	if client, found := cf.usersClients.Get(user); found {
+		cf.log.Info("returning cached client for", "user1", user.ID)
+		return client, nil
+	}
+
 	pool := cf.newClustersPool(cf.scheme)
 	errChan := make(chan error, len(cf.clusters.Get()))
 
@@ -311,7 +317,10 @@ func (cf *clustersManager) GetImpersonatedClient(ctx context.Context, user *auth
 		result = multierror.Append(result, err)
 	}
 
-	return NewClient(pool, cf.userNsList(ctx, user)), result.ErrorOrNil()
+	client := NewClient(pool, cf.userNsList(ctx, user))
+	cf.usersClients.Set(user, client)
+
+	return client, result.ErrorOrNil()
 }
 
 func (cf *clustersManager) GetImpersonatedClientForCluster(ctx context.Context, user *auth.UserPrincipal, clusterName string) (Client, error) {
--- a/core/clustersmngr/factory_caches.go
+++ b/core/clustersmngr/factory_caches.go
@@ -5,6 +5,7 @@ import (
 	"sort"
 	"strings"
 	"sync"
+	"time"
 
 	"github.com/cheshir/ttlcache"
 	"github.com/weaveworks/weave-gitops/pkg/server/auth"
@@ -147,4 +148,28 @@ func (un *UsersNamespaces) Clear() {
 
 func (un UsersNamespaces) cacheKey(user *auth.UserPrincipal, cluster string) uint64 {
 	return ttlcache.StringKey(fmt.Sprintf("%s:%s", user.ID, cluster))
+}
+
+type UsersClients struct {
+	Cache *ttlcache.Cache
+}
+
+func (uc *UsersClients) cacheKey(user *auth.UserPrincipal) uint64 {
+	return ttlcache.StringKey(fmt.Sprintf("%s:%s", user.ID, strings.Join(user.Groups, "/")))
+}
+
+func (uc *UsersClients) Set(user *auth.UserPrincipal, client Client) {
+	uc.Cache.Set(uc.cacheKey(user), client, 5*time.Minute)
+}
+
+func (uc *UsersClients) Get(user *auth.UserPrincipal) (Client, bool) {
+	if val, found := uc.Cache.Get(uc.cacheKey(user)); found {
+		return val.(Client), true
+	}
+
+	return nil, false
+}
+
+func (uc *UsersClients) Clear() {
+	uc.Cache.Clear()
 }
--- a/core/server/fluxruntime.go
+++ b/core/server/fluxruntime.go
@@ -156,7 +156,7 @@ func filterFluxNamespace(nss []v1.Namespace) *v1.Namespace {
 }
 
 func (cs *coreServer) GetReconciledObjects(ctx context.Context, msg *pb.GetReconciledObjectsRequest) (*pb.GetReconciledObjectsResponse, error) {
-	clustersClient, err := cs.clustersManager.GetImpersonatedClientForCluster(ctx, auth.Principal(ctx), msg.ClusterName)
+	clustersClient, err := cs.clustersManager.GetImpersonatedClient(ctx, auth.Principal(ctx))
 	if err != nil {
 		return nil, fmt.Errorf("error getting impersonating client: %w", err)
 	}
@@ -247,7 +247,7 @@ func (cs *coreServer) GetReconciledObjects(ctx context.Context, msg *pb.GetRecon
 }
 
 func (cs *coreServer) GetChildObjects(ctx context.Context, msg *pb.GetChildObjectsRequest) (*pb.GetChildObjectsResponse, error) {
-	clustersClient, err := cs.clustersManager.GetImpersonatedClientForCluster(ctx, auth.Principal(ctx), msg.ClusterName)
+	clustersClient, err := cs.clustersManager.GetImpersonatedClient(ctx, auth.Principal(ctx))
 	if err != nil {
 		return nil, fmt.Errorf("error getting impersonating client: %w", err)
 	}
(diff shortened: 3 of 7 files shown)


core/server/fluxruntime.go
```diff
@@ -247,7 +247,7 @@ func (cs *coreServer) GetReconciledObjects(ctx context.Context, msg *pb.GetRecon
 }
 
 func (cs *coreServer) GetChildObjects(ctx context.Context, msg *pb.GetChildObjectsRequest) (*pb.GetChildObjectsResponse, error) {
-	clustersClient, err := cs.clustersManager.GetImpersonatedClientForCluster(ctx, auth.Principal(ctx), msg.ClusterName)
+	clustersClient, err := cs.clustersManager.GetImpersonatedClient(ctx, auth.Principal(ctx))
```
jpellizzari, line 250: This seems like a logic change that I wouldn't expect. We don't need to get a client scoped for a cluster?

jpellizzari: After reading more, I think I understand. Should we remove the `GetImpersonatedClientForCluster` method? Is it used anywhere else?

core/clustersmngr/factory_caches.go
```diff
@@ -148,3 +149,27 @@ func (un *UsersNamespaces) Clear() {
...
+type UsersClients struct {
+	Cache *ttlcache.Cache
+}
+
+func (uc *UsersClients) cacheKey(user *auth.UserPrincipal) uint64 {
+	return ttlcache.StringKey(fmt.Sprintf("%s:%s", user.ID, strings.Join(user.Groups, "/")))
+}
+
+func (uc *UsersClients) Set(user *auth.UserPrincipal, client Client) {
+	uc.Cache.Set(uc.cacheKey(user), client, 5*time.Minute)
```
jpellizzari, line 162: Does the 5 minute timeout mean that if I come back 5 minutes later, things take 7 seconds to load again?. I guess we will find out.

jpellizzari (approved):
LGTM, only questions.

Are we guessing that recreating the clients is causing the slowness, or have we been able to gather some data on the operation that is taking the longest?

core/server/fluxruntime.go
luizbafilho: yep, I'll remove it.

core/clustersmngr/factory_caches.go
luizbafilho: very likely, this need to be paired with using pagination, which is implemented in the backend but needs to be used on ui.

jpellizzari: Is a `cursor` tied to a client?

luizbafilho: good question, with this caching stuff we'll need to investigate how it plays with our current pagination solution.

jpellizzari (commented):
Another question @luizbafilho: Why are we adding "watch" RBAC roles here? Do we need them for cacheing, or was it just a convenience thing because you knew you would need in in the future?

luizbafilho:
> Another question @luizbafilho: Why are we adding "watch" RBAC roles here? Do we need them for cacheing, or was it just a convenience thing because you knew you would need in in the future?

That's required by the `DelegatingClient`. That client is the one that is actually used but the kubebuilder controllers. Here is a very good explanation of how it works: https://medium.com/@timebertt/kubernetes-controllers-at-scale-clients-caches-conflicts-patches-explained-aa0f7a8b4332#95a2

luizbafilho:
> LGTM, only questions.
> 
> Are we guessing that recreating the clients is causing the slowness, or have we been able to gather some data on the operation that is taking the longest?

every time we do `client.New` it triggers an authentication call against the cluster, besides that, there is the actual call we need to do, and without a cached client, the call hits Kube API every time.

This caches the client itself, and also the client is of a different kind that caches the called data.

core/server/fluxruntime.go
foot: `GetImpersonatedClientForCluster` is used a little bit in EE for querying details pages.

It will be slightly slower to build up the client set for these cases, mostly if another cluster is unreachable in a funny way (timeout's ~8s).

jpellizzari: @foot Any idea if we can control that timeout?

luizbafilho: The goal of the PR is not to have to build clients every time, I wonder what happens if we keep the clients around for longer than the 5 min we have now, and as Jordan suggested, decrease to a more sensible number.

ozamosi: Why do we want to remove `GetImpersonatedClientForCluster` at all?

I would not use a ttl cache at all, because the resource requirements aren't possible to control - unless there's issues with connections breaking at a certain time, I'd use an LRU.

foot: > @foot Any idea if we can control that timeout?

- We first try to connect to a leaf in `ClientConfigWithUser`, it will apply the `DefaultKubeConfigOptions` one of which is `WithFlowControl`. This does a `HEAD` request to the leaf which can take 5s to hit `kubeClientDialTimeout`. So most unreachable leafs will stop here.
- Then when we `client.New` the client will start up and fill in the restMapper by querying CRDs, this has has 8s to hit the `kubeClientTimeout` I think...

So... yeah, if everything goes terribly it can at least take 4.9s + 7.9s. Both those timeout values are in factory.go to tweak.

foot: > I wonder what happens if we keep the clients around for longer than the 5 min we have now

- We'd need to wipe out the cache if clusters change. There is a `chan` we can subscribe to now though.
- If 1 cluster is briefly unreachable then it wouldn't be in the client set until we try again in 5mins.

luizbafilho: > So... yeah, if everything goes terribly it can at least take 4.9s + 7.9s. Both those timeout values are in factory.go to tweak.

Because it's better to have a single cached client for each cluster, and `GetImpersonatedClientForCluster` was kind of a hack to overcome the slow client creation problem. If we can fix that we should remove it. And when querying a specific cluster we should use the `ClusterName` param on the calls.

ozamosi: > Because it's better to have a single cached client for each cluster

That method should grab the cached client, yeah.

> If we can fix that we should remove it.

But, as @foot explained, clusters come and go, so this change won't affect any of the problems this method works around.

(Which in my opinion suggests `GetImpersonatedClient` doesn't work and `GetImpersonatedClientForCluster` is the one we should keep, and the decision which cluster to query/poll should live together with the decision of when to poll. That would let us easily exclude broken clusters, back off slow clusters, render partial results as they come in until we have responses from all clusters, etc. But I digress)

charts/gitops-server/templates/admin-user-roles.yaml
```diff
@@ -8,40 +8,40 @@ metadata:
 rules:
   - apiGroups: [""]
     resources: ["configmaps", "secrets", "pods", "services", "namespaces", "persistentvolumes", "persistentvolumeclaims"]
-    verbs: [ "get", "list" ]
+    verbs: [ "get", "list", "watch" ]
```
enekofb, line 11: do we need to update any user documentation at the back of this change?

core/clustersmngr/factory.go
```diff
@@ -469,6 +494,102 @@ func (cf *clustersManager) userNsList(ctx context.Context, user *auth.UserPrinci
...
+		Mapper: mapper,
+	})
+	if err != nil {
+		return nil, fmt.Errorf("failed to create leaf client: %w", err)
+	}
+
+	return client, nil
+}
+
+func CachedClientFactory(user *auth.UserPrincipal, cfgFunc ClusterClientConfigFunc, cluster Cluster, scheme *apiruntime.Scheme) (client.Client, error) {
```
enekofb, line 540: is there any test coverage for the code added here?

core/clustersmngr/factory.go
```diff
@@ -469,6 +494,102 @@ func (cf *clustersManager) userNsList(ctx context.Context, user *auth.UserPrinci
...
+
+	client, err := client.New(config, client.Options{
+		Scheme: scheme,
+		Mapper: mapper,
+	})
+	if err != nil {
+		return nil, fmt.Errorf("failed to create leaf client: %w", err)
+	}
+
+	return client, nil
```
enekofb, line 537: this code is common between Client and  CachedClient, do you think might be useful to refactor it to a shared function?

core/clustersmngr/factory.go
```diff
@@ -469,6 +494,102 @@ func (cf *clustersManager) userNsList(ctx context.Context, user *auth.UserPrinci
 	return cf.GetUserNamespaces(user)
 }
 
+func (cf *clustersManager) getOrCreateClient(ctx context.Context, user *auth.UserPrincipal, cfgFunc ClusterClientConfigFunc, cluster Cluster) (client.Client, error) {
```
enekofb, line 497: it could be good to add some test for `getOrCreateClient`

core/server/fluxruntime.go
```diff
@@ -180,42 +179,68 @@ func (cs *coreServer) GetReconciledObjects(ctx context.Context, msg *pb.GetRecon
...
-				// Our service account (or impersonated user) may not have the ability to see the resource in question,
-				// in the given namespace. We pretend it doesn't exist and keep looping.
-				// We need logging to make this error more visible.
-				continue
-			}
+		wg.Add(1)
 
-			return nil, fmt.Errorf("listing unstructured object: %w", err)
-		}
+		go func(clusterName string, gvk *pb.GroupVersionKind) {
```
enekofb, line 196: is this change related to caching client or just enhancement to parallelize?

enekofb (commented):
it looks good, just added a few comments with some suggestions and doubts. suggestions mostly around testing. 

where is this available?

core/clustersmngr/client.go
```diff
@@ -103,6 +111,13 @@ func (c *clustersClient) List(ctx context.Context, cluster string, list client.O
 		return err
 	}
 
+	// Due to how DelegatingClients work, calls that fails never returns,
+	// because it waits the cache to sync before returning it https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/cache/internal/informers_map.go#L206
+	// so we are forced to use a timeout so it doesn't keep the informer up.
+	// TODO: What should be the timeout here?
+	ctx, cancel := context.WithTimeout(ctx, clientTimeout)
```
enekofb, line 118: can you expand a bit more on the issue to understand it in order to input the timeout value?

enekofb: I see now, thanks for the explanation  ... it now makes sense to me

follow up question - is the cache always filled when empty whenever is a get or a list operation?

core/clustersmngr/factory.go
```diff
@@ -469,6 +494,102 @@ func (cf *clustersManager) userNsList(ctx context.Context, user *auth.UserPrinci
...
+	if user == nil {
+		user = &auth.UserPrincipal{
+			ID: "weave-gitops-server",
```
yiannistri, line 500: Why do we need to set this? Can we not return an error instead? I'm thinking this _may_ be lead to privilege escalation if there is RBAC in place that refers to this "user".

yiannistri (commented):
Nicely done 👍 I agree with @ozamosi that we can't control the resources and an LRU cache may be more appropriate but we can tackle that in another PR.

enekofb (approved):

core/clustersmngr/client.go
enekofb: so we have three scenarios 

1. access allowed and cache hot  -> quick answer
2. access allowed and cache cold -> up to the size of the collection 
3. not allowed -> not answer -> timeout

what would be the result for the user when in scenario 2) and timeout = 2sec, to fill up the cache takes more than 2 sec? there would be a timeout and then is there any retry or the client see an error?

I could think that a strategy of failing fast + retry could be a good balance

enekofb: to me sounds sensible, we could enhance this once tested with Protocol Labs if we see that is not a great experience

PR #2783, opened 2022-09-23
Add version and revision info to Helm detail pages

### LINKED ISSUE #2706: Show chart version in a HelmRelease and HelmChart details pages
This information is available in the Applications table under revision, and similar information is available for kustomizations under "applied revision". Whereas for Helm resources the user needs to jump into the YAML view.

I believe it is useful summary detail to surface:
**HelmRelease**
- `spec.chart.spec.Version`
- `status.lastAppliedRevision`
- `status.lastAttemptedRevision`

**HelmChart**
- `spec.version`
- `status.artifact.revision`

joshri:
Closes #2706 

The HelmChart class has two new getters which are displayed in the details view - Version, and Revision (should this be artifact revision?)
The HelmRelease class has three - Version, Last Applied Revision, and Last Attempted Revision

Screenshots below - yes, there's a couple unfortunate bugs I found here with the overflow styling of chip group...and a chip will be created if you just press enter on the search input with no content...........issue incoming :(

image

image

### DIFF
--- a/ui/components/HelmChartDetail.tsx
+++ b/ui/components/HelmChartDetail.tsx
@@ -34,6 +34,8 @@ function HelmChartDetail({ className, helmChart }: Props) {
       info={[
         ["Type", Kind.HelmChart],
         ["Chart", helmChart.chart],
+        ["Version", helmChart.version],
+        ["Revision", helmChart.revision],
         ["Ref", helmChart.sourceRef?.name],
         ["Last Updated", <Timestamp time={helmChart.lastUpdatedAt} />],
         ["Interval", <Interval interval={helmChart.interval} />],
--- a/ui/components/HelmChartDetail.tsx
+++ b/ui/components/HelmChartDetail.tsx
@@ -35,7 +35,7 @@ function HelmChartDetail({ className, helmChart }: Props) {
         ["Type", Kind.HelmChart],
         ["Chart", helmChart.chart],
         ["Version", helmChart.version],
-        ["Revision", helmChart.revision],
+        ["Current Revision", helmChart.revision],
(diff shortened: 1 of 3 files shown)


opudrovs (approved):

ui/components/HelmChartDetail.tsx
```diff
@@ -34,6 +34,8 @@ function HelmChartDetail({ className, helmChart }: Props) {
       info={[
         ["Type", Kind.HelmChart],
         ["Chart", helmChart.chart],
+        ["Version", helmChart.version],
+        ["Revision", helmChart.revision],
```
ozamosi, line 38: :bulb: The "version" vs "revision" distinction is weird to me. At least to me, "version" and "revision" can be used interchangeably, and I'd use revision for "automatic" versioning (i.e. git commits) and version for "active" versioning (i.e. git tags) - and that's not at all what's going on here! Version is a "version mask", "version selector" or "version specification", and revision is "actual version" or "highest version matching the version field". I looked at the flux documentation to see if they spell it out more, but the best they're doing is "artifact revision", which is perfectly accurate but also requires you to have a deep understanding of the source controller for it to be helpful.

This isn't too bad on the helm release page (because "last applied" and "last attempted" helps explain that they're "discovered" and thus status, so that helps suggest "chart version" isn't discovered and thus specified), but here I can't help go "wait, what?".

**tl;dr**: My concrete suggestion is to change "Revision" to "Latest Revision" or "Current Revision". That way it would also "look like" it comes from status, to help distinguish between the two.

ozamosi (approved):
Works as expected, though I also put one suggestion for UI clarification.

PR #2785, opened 2022-09-25
implement flux install using the fluxexec package

### LINKED ISSUE #2740: Use flux-exec to do the flux base install
Replace the custom flux installation with using flux's CLI through the `flux-exec` package.

The end result should be that the same flags we have now still work, and the same resources should be created, but they should be passed on to the flux binary through flux-exec instead of the custom installer.

chanwit:
Fixes #2740


pkg/fluxexec/bootstrap.go
```diff
@@ -180,31 +194,31 @@ func (flux *Flux) bootstrapArgs(opts ...BootstrapOption) []string {
 
 	args := []string{}
 
-	if c.authorEmail != "" {
+	if c.authorEmail != "" && !reflect.DeepEqual(c.authorEmail, defaultBootstrapOptions.authorEmail) {
```
ozamosi, line 197: :grey_question:  Could you explain why you did this (e.g. "all the DeepEquals", I just picked one for the comment) change?

cmd/gitops/beta/run/cmd.go
```diff
@@ -224,40 +226,50 @@ func betaRunCommandRunE(opts *config.Options) func(*cobra.Command, []string) err
...
+			}
 
-			man, err := run.NewManager(log, ctx, kubeClient, kubeConfigArgs)
+			wd, err := os.Getwd()
 			if err != nil {
-				log.Failuref("Error creating resource manager")
 				return err
 			}
 
-			if err := run.InstallFlux(log, ctx, installOpts, man); err != nil {
```
ozamosi, line 247: :warning: I believe there's at least some code that's now unused - could you please have a go at cleaning it up as well?

ozamosi (approved):

pkg/fluxexec/bootstrap.go
ozamosi: The risk is that the default we think flux has diverges from the actual flux default, and then if you try to set "our default" we'd just remove it and flux would then set "the new flux default".

Though I suppose that's entirely hypothetical.

chanwit: yes, the consequence is that we have to revisit this package every time Flux releasing a new version. 
It is the case for HC's tfexec as well.

I'll document how to do so, systematically.

PR #2786, opened 2022-09-26
Take the user through a bootstrap wizard when they close gitops run

### LINKED ISSUE #2699: Take the user through a bootstrap wizard when they close gitops run
Flux bootstrap's per-git forge support allows a user to handle flux authentication in a secure way - we need to support these.

However, that means we need to support things that are difficult to auto-detect - e.g. whether `git.mycompany.com` is an enterprise installation of github, gitlab or something else, or whether `github.com/weaveworks` is a personal github account or an organisation account.

There are probably libraries to help us do this.

**Acceptance Criteria**

* Allow defining a tree of questions. E.g.:
  * What type of git forge is this? [github, gitlab, bitbucket, generic] github
    * Is this a personal or an organizational account? [personal, organizational]
    * Is this a private or public repository? [private, public]
* Specify a way to try to auto-detect an answer. E.g.:
 * Parse the git remote - if the host is `github.com` then use that answer as the git forge type, and don't ask the question
 * If the environment variable `GITOPS_GITHUB_ACCOUNT_TYPE=personal` is set, then we don't ask question about account type.

The outcome of this wizard should be an unambiguous `flux bootstrap` command.

Note that flux already has some of this - if you run `flux bootstrap github` without a `GITHUB_TOKEN` environment variable, it will ask you for a PAT. We are not looking to replace this specific UI - we are looking to extend this wizard thinking so that we can construct the command in the first place.

opudrovs:
Closes #2699

WIP, needs code check and description.


opudrovs:
I will add UI screenshots or video now.

opudrovs:
UI screenshots. Different flag lists are used for different git providers. If remote URL contains `github.com` or `gitlab.com`, some values will be prefilled.

The select git provider table will be displayed only if the git provider vannot be parsed from the remote URL. It is run as a separate bubbletea program. Both programs exit to the main command flow when done. The UI uses default styling from bubbletea examples for now.

Screenshot 2022-09-26 at 13 07 10
Screenshot 2022-09-26 at 13 07 26
Screenshot 2022-09-26 at 13 07 46
Screenshot 2022-09-26 at 13 08 00
Screenshot 2022-09-26 at 13 08 16
Screenshot 2022-09-26 at 13 09 37
Screenshot 2022-09-26 at 13 10 08

opudrovs:
For testing the wizard UI easily I uncomment out previous gitops run code and uncomment checks around calls to`SelectGitProvider`.

chanwit:
It's so beautiful! Thank you so much @opudrovs

chanwit:
`path: ./clusters/<...>` is still missing from the generated command, AFAICT

opudrovs:
@chanwit 

> path: ./clusters/<...> is still missing from the generated command, AFAICT

Was the `path` flag value empty? If yes, it will be removed from flags for now, I don't handle required flags yet.

Or do you mean that it should it be autofilled from the gitops run commandline arguments?

ozamosi:
:tada: Beautiful!

:bulb: I would change `defaultFlagValue` from `string` to something like `type DefaultValueGetter func(*gogit.Repository) string` - that way, the auto-detection bits would sit together with their definitions, instead of the auto-detection being in `NewBootstrapWizard` separate from the definitions.

We could still use something like
```
func constantDefault(value string) DefaultValueGetter {
	return func(_ *gogit.Repository) string {
		return value
	}
}
```
so you only need to change `defaultFlagValue: "true"` to `defaultFlagValue: constantDefault("true")` - but we'd also be able to write more complex getters for e.g. path, branch, hit the github API to check if repo exists, use the OS username as default for owner, etc.

opudrovs:
Thanks! OK, will do it.

cmd/gitops/beta/run/cmd.go
```diff
@@ -525,6 +527,72 @@ func betaRunCommandRunE(opts *config.Options) func(*cobra.Command, []string) err
 			return err
 		}
 
+		// run bootstrap wizard only if Flux was not installed
+		if fluxVersion != "" {
```
ozamosi, line 531: :no_entry: Final thought: it's Release Week, and the follow up tickets won't be done before we want to ship. Could we just add a quick check here that, say, the `GITOPS_RUN_BOOTSTRAP` env var is set, and if not also return nil?

opudrovs: Sure, not a problem.

Should I just check for `os.Getenv("GITOPS_RUN_BOOTSTRAP")` or do I also need a call to `osSetenv` somewhere?

ozamosi: No setenv, just check.

opudrovs:
@ozamosi @chanwit 

 - Added check for os.Getenv("GITOPS_RUN_BOOTSTRAP") in addition to flux version before running the wizard.
- Replace default flag values with callbacks (except for path, did not create a getter for it because it's computed not from repo).
- Improve parsing of remote URLs.
- Added parsing branch as repo.Head().

ozamosi (approved):

opudrovs:
Mwahaha! I am merging it! 😈

PR #2787, opened 2022-09-26
Fix events view to take clusters into account

ozamosi:
We should only get events for the current cluster - we used to search all clusters, and if another cluster had the same object then we'd show incorrect events.

This makes the new, correct object ref kind introduced in #2765 the only object ref used in our APIs, and then copy-pastes the client creation code from c3a407a.


joshri (commented):
A couple lingering ClusteredObjRefs in suspend_test.go - I'm guessing most other places just used ObjectRef?

joshri (approved):
It's a BIG approve. Tested locally and events are coming through.

PR #2794, opened 2022-09-27
add empty table row prop so it can be configured from parent component

ahussein3:
-Resolve [Issue-1566](https://github.com/weaveworks/weave-gitops-enterprise/issues/1566)

What are the changes ?
- add new prop to DataTable "emptyMessagePlaceholder"

Why these changes are done ? 
- To be able to manipulate the "No Data" message with custom message being sent from parent component.

### DIFF
--- a/ui/components/DataTable.tsx
+++ b/ui/components/DataTable.tsx
@@ -59,6 +59,7 @@ export interface Props {
   dialogOpen?: boolean;
   hasCheckboxes?: boolean;
   hideSearchAndFilters?: boolean;
+  emptyMessagePlaceholder?: string | React.ReactNode;
 }
 //styled components
 const EmptyRow = styled(TableRow)<{ colSpan: number }>`
@@ -331,6 +332,7 @@ function UnstyledDataTable({
   hasCheckboxes: checkboxes,
   dialogOpen,
   hideSearchAndFilters,
+  emptyMessagePlaceholder,
 }: Props) {
   //URL info
   const history = useHistory();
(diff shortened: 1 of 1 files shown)


ui/components/DataTable.tsx
```diff
@@ -59,6 +59,7 @@ export interface Props {
   dialogOpen?: boolean;
   hasCheckboxes?: boolean;
   hideSearchAndFilters?: boolean;
+  emptyMessagePlaceholder?: string | React.ReactNode;
```
foot, line 62:
```suggestion
  emptyMessagePlaceholder?: React.ReactNode;
```

`string` seems to already be in there down in `ReactText`

```ts
    type ReactText = string | number;
    type ReactChild = ReactElement | ReactText;

    /**
     * @deprecated Use either `ReactNode[]` if you need an array or `Iterable<ReactNode>` if its passed to a host component.
     */
    interface ReactNodeArray extends ReadonlyArray<ReactNode> {}
    type ReactFragment = {} | Iterable<ReactNode>;
    type ReactNode = ReactChild | ReactFragment | ReactPortal | boolean | null | undefined;
```

ui/components/DataTable.tsx
```diff
@@ -561,7 +563,9 @@ function UnstyledDataTable({
                         size="base"
                       />
                       <Spacer padding="xxs" />
-                      <Text color="neutral30">No data</Text>
+                      <Text color="neutral30">
+                        {emptyMessagePlaceholder || "No data"}
+                      </Text>
```
foot, line 568: Maybe just worth double checking we can add links that look good (preserve their link color etc) as a child element to `<Text />`.. otherwise we could wrap the whole thing?
```suggestion
                      {emptyMessagePlaceholder || (<Text color="neutral30">
                        "No data"
                      </Text>)}
```

Kind of thing?

foot: And maybe add `Text` as an export to `index.ts`

foot (approved):
🙌

ui/components/DataTable.tsx
ahussein3: sound a good idea will commit that

ozamosi (approved):

joshri (approved):

Issue #2796, opened 2022-09-27
Incorrect message from computeMessage function


joshri:
<img width="1176" alt="image" src="https://user-images.githubusercontent.com/65822698/192570666-3abbb8ea-fd2a-49a4-8177-82d8f117ed15.png">

This is displaying the message from the first conditions object (type `AbleToScale`) which has `status: "True"` instead of the second (type: `ScalingActive`) which has `status: "False"`. 

The `computeMessage` function is only looking for "Ready" and "Available" which can lead to an incorrect message. It should look for the message based on computeReady's ReadyType, and handle message for Ready, NotReady, and Reconciling.

joshri:
<img width="2047" alt="image" src="https://user-images.githubusercontent.com/65822698/192886016-1712b044-a205-4f1e-bc0a-c7cc417a1a6d.png">

It gets weirder - CRDs in the reconciled objects table are showing the wrong status. Screenshot shows interesting conditions array with no failures and status is coming out as not ready

joshri:
UPDATE sending out an 🆘 to @ozamosi @JamWils

Starting looking into this - the `computeMessage` function actually IS working as intended. It searches for the condition of type `Ready` and only uses that one. The problem with this is that both CRDs and HorizontalPodAutoscaler here don't have a condition of type `Ready`, so the function just returns false, which obviously does not tell the full story. 

Looking into a better way to process `conditions` didn't lead to great results - you can't rely on the order of the different conditions to signify importance, and although there is apparently a `phase` field on objects you're not supposed to use it? Maybe. Could be outdated. 

If ANY condition in the array has a status of "False" should that override everything else? I'm not entirely sure, although it seems like you'd rather know that one thing is failing than if three are succeeding.  Could this flip the other way, where you could have the Ready condition failing but things being fine overall? Users can even make their own conditions right? 

Maybe instead of `Not Ready` we display something like `<number> failing conditions` with the message column being able to display multiple messages.

ozamosi:
`phase` is interesting - `phase` tells you what we're currently doing, but not whether we were successful at it or not. So `phase: running` doesn't mean that we _are_running, just that we've completed the previous phases - it could still mean that we're failing to run. So, it kinda tells you which condition to look at. But, not all object types have a phase.

> Users can even make their own conditions right?

Yes - each Kind can have their own states.

> If ANY condition in the array has a status of "False" should that override everything else?

[...]

PR #2800, opened 2022-09-27
Add yaml pop-ups for objects that aren't flux primitives

### LINKED ISSUE #2755: Improve navigation and YAML for all objects
As a user, on the detail page I want to be able to see and understand all of the different objects and not just the Flux primitives.  At a minimum, if I click on any k8s object it would be nice to see a modal pop-up with the yaml.

### Acceptance Criteria

The design of the yaml modal follows the design used on the Providers "alerts" tab in the Flux Notification work.

When I go to the details tab of a `kustomization` or `helm release`.   I can click on any object in the name column.
- [ ] When I click on the object that is not a Flux primitive then a modal appears with the yaml for that object.  I can click outside of the modal to dismiss it.
- [ ] When I click on the object that is a Flux primitive then I am taken to the detail page for that particular object.  This list includes `Kustomizations`, `HelmReleases`, `Sources`, `Providers`, `Alerts`, etc.

When I go to the graph tab of a `kustomization` or `helm release`. I can click on the "name" in each node.
- [ ] When I click on the object that is not a Flux primitive then a modal appears with the yaml for that object.  I can click outside of the modal to dismiss it.
- [ ] When I click on the object that is a Flux primitive then I am taken to the detail page for that particular object.  This list includes `Kustomizations`, `HelmReleases`, `Sources`, `Providers`, `Alerts`, etc.

When I go to the dependency tab of a `kustomization` or `helm release`. I can click on the "name" in each node.
- [ ] When I click on the object that is a Flux primitive then I am taken to the detail page for that particular object.  This list includes `Kustomizations`, `HelmReleases`, `Sources`, `Providers`, `Alerts`, etc.

`Secret` is excluded from this functionality.  When I see a "secret" it is not clickable and I cannot view its details.

joshri:
Closes #2755

BACKEND:
- In order to add non-primitive objects to the GetObject formula, they had to be freed from the shackles of the UnstructuredObject type. This mostly involved deleting, but there are results I'm still not sure about - such as certain objects not having a conditions value (we used to make that ourselves and assign it to UnstructuredObject), and the new way of retrieving images in the frontend (I'm not sure how to verify if this is working correctly)
- The good news is that this led to me discovering that `computeMessage` just doesn't really do anything? So I made an issue to fix that: https://github.com/weaveworks/weave-gitops/issues/2796
- I also added tenant in so they'd work with K8sToProto....I don't know what a tenant is so this could easily have been the wrong move. 

getImage on FluxObject:

``` 
get images(): string[] {
    const spec = this.obj.spec;
    if (!spec) return [];
    if (spec.template) return spec.template.spec.containers.map((x) => x.image);
    if (spec.containers) return spec.containers.map((x) => x.image);
    return [];
  }
```
image of types with no conditions:

image

FRONTEND:
- The largest change that doesn't revolve around removing the UnstructuredObject type is adding the GraphNode yaml state to AppContext. I can't put the dialog in GraphNode because then every single node renders it's own dialog. I don't want to put it in any of the graph components because it's reused, and I don't want to add any extra re-renders in there if I can avoid it. Putting it in AutomationDetail seems like the right move, as it's the parent of both graphs. Unfortunately the component that triggers the appearance of the dialog is in GraphNode. This is why I added the state to Context, to avoid passing it down all the way through to DirectedGraph AND DagGraph. 
- For the `Secret` stuff. I put some code in there but I haven't tested it. I can barely make a Kustomization let alone a Secret. It's a miracle I have a HorizontalPodAutoscaler. Or is it? I don't know. 

Graph yaml video:

https://user-images.githubusercontent.com/65822698/192653706-43539d93-e2a1-4842-b3b2-bd434045290e.mov


joshri:
Changes have been made -
@ozamosi - CRD showing incorrect status has been added to the computeMessage issue - both conditions show a status of True so I'm not sure why it's showing Not Ready

@opudrovs - I thought there used to be padding on the footer - I looked and it's not there anymore. There must have been a reason but I added it back in for now until I hear otherwise or see something ugly as a result of the change. I also added your great recommendations for get image on FluxObject - I needed to return `[]` at the end because some objects can have a spec but no template or container and thus return undefined, which crashes the app. Thought it would be better to handle in the class than in ImageLink

JamWils:
Two questions:

1.  How is the width of the modal calculated?  Can we make it about 20% wider?
2. Why does the parent node "git repository" show no status?

opudrovs:
@joshri

>I thought there used to be padding on the footer - I looked and it's not there anymore. There must have been a reason but I added it back in for now until I hear otherwise or see something ugly as a result of the change. 

The missing footer padding/spacing looks like an old issue which is not limited to the current branch. I have not been able to find where we lost it, maybe we didn't. Maybe it just seemed that the footer had a padding/spacing thanks to styling in some other components. On staging, I can see the footer without spacing on the Sources page, but not on Applications (probably because the table on the Sources page has more items). So, it's great you added the fix let's hope it solves the issue.

>I also added your great recommendations for get image on FluxObject - I needed to return [] at the end because some objects can have a spec but no template or container and thus return undefined, which crashes the app. Thought it would be better to handle in the class than in ImageLink

Sounds good! My suggestions were mostly cosmetic, to improve readability. The most important thing was returning [] if the spec does not exist but adding a question mark as you suggested would work too, just with longer lines.

opudrovs:
@JamWils 

>Why does the parent node "git repository" show no status?

@JamWils  @joshri has the parent node (`rootNode`) ever shown status at all? I checked out some older commits, don't see displaying the status for this node. If it's a new feature we need a separate ticket for this.

joshri:
@JamWils @opudrovs The parent node (in this case on the `ReconciliationGraph` it's the `source` of the automation) has no status because we're pulling it's data from the `sourceRef` field of the automation, which only has `apiversion`, `kind`, `name`, and `namespace`. At least we can link the user to the source detail page now? Running a getObject request on the source seems like a lot - if you feel like it looks wrong maybe we better distinguish between the automation as the current node (I feel like the bold text doesn't read as well with the blue links) and the source as.....the source?

JamWils:
It's fine, I was just curious.  Thanks for the context.

opudrovs:
@joshri 

>I feel like the bold text doesn't read as well with the blue links

💡 If after testing it we decide that it's not clear which object is the current object (after adding the links), we could try increasing font weight or adding an asterisk.

opudrovs:
❔ Is it expected behavior that size (especially the popup width) of the YAML popup is different for different objects (I don't know if it's related to object type or content line length)?

Screenshot 2022-09-29 at 20 51 23

Screenshot 2022-09-29 at 20 51 34

Screenshot 2022-09-29 at 20 51 49

opudrovs:
If the PR is still in progress, sorry, just saw that review was requested and didn't check its state.

joshri:
> ❔ Is it expected behavior that size (especially the popup width) of the YAML popup is different for different objects (I don't know if it's related to the object type or content line lengths)?

NOT EXPECTED - this is weird because I'm not getting this on my local. Could be another npm start vs. tilt situation. I added the `maxWidth='md'` and `fullWidth` props to Mui's Dialog component (https://mui.com/material-ui/react-dialog/#optional-sizes) which seemed to work for me.

opudrovs:
⚠️ A couple of minor issues:

- When dragging graph with the pointer over a link in a graph node which has no linked details page and just a YAML popup and releasing the graph, the YAML popup for this node is displayed because the pointer stays over the link.

- When attempting to drag the graph when the pointer is over a link in a node which has a linked details page, the mouse attempts to drag the link and the graph is not dragged.

joshri:
> ⚠️ A couple of minor issues:
> 
> * When dragging graph with the pointer over a link in a graph node which has no linked details page and just a YAML popup and releasing the graph, the YAML popup for this node is displayed because the pointer stays over the link.
> * When attempting to drag the graph when the pointer is over a link in a node which has a linked details page, the mouse attempts to drag the link and the graph is not dragged.

Even if I change the onClick to a onMouseDown with a setTimeout or something, won't there still be some default browser behavior with dragging links? I get that it's annoying for dragging the graph but I'm not sure that overriding that behavior is a good idea (although I'm sure it's possible somehow). Is the expected behavior that a click and hold on a link doesn't do anything and you can drag the graph and release with no action?

opudrovs:
In addition to the behavior noted above, when dragging a graph node with a details page by the link, the node is not dragged, but when releasing the mouse button somewhere it jumps to the pointer location and the graph stays attached to the pointer (it's hard to get the rid of it afterwards 😄  ).

opudrovs:
>Even if I change the onClick to a onMouseDown with a setTimeout or something, won't there still be some default browser behavior with dragging links? I get that it's annoying for dragging the graph but I'm not sure that overriding that behavior is a good idea (although I'm sure it's possible somehow). Is the expected behavior that a click and hold on a link doesn't do anything and you can drag the graph and release with no action?

It depends on which behavior of graph nodes with links is desired. Graph is not a standard browser element so I think we have some freedom in how it should behave.

opudrovs:
In general the PR LGTM except for the minor issues above. 🎉

opudrovs:
I've created a secret with `kubectl` and wanted to test it but have not found where to view it in the dashboard. @ozamosi how can I test a secret in the context of this PR?

ozamosi:
> @ozamosi how can I test a secret in the context of this PR?

Install weave-gitops through it's chart, and let it create an admin user secret.

The secret data still shows up right now - something like this would make all the secret data vanish, even though it'd still show a popup:
```diff
-func K8sObjectToProto(object client.Object, clusterName string, tenant string, inventory []*pb.GroupVersionKind) (*pb.Object, error) {
+func K8sObjectToProto(object *unstructured.Unstructured, clusterName string, tenant string, inventory []*pb.GroupVersionKind) (*pb.Object, error) {
        var buf bytes.Buffer
 
+       if object.GetKind() == "Secret" {
+               object.Object["data"] = "** redacted **"
+       }
```

opudrovs:
@ozamosi thanks! I've installed it and can confirm that the secret displays the YAML popup.

opudrovs (approved):

joshri:
I had secret un-clickable in the graph, but not in the ReconciledObjectsTable - fixed it. 

Table:
image

Graph:
image

joshri:
Secret data is now changed to `{redacted: null}` inside of GetReconciledObjects, so that it's not even being sent to the frontend. This could technically make Secret yaml viewable if we're interested in that @JamWils 

@opudrovs if you could confirm for me that data from the secret in the test-gitops helm release has been changed it would be much appreciated!! Thanks!

PR #2802, opened 2022-09-28
update filter options chip color

### LINKED ISSUE #2801: [UI] Update Table FilterOptions Color

ahussein3:
Resolve #2801 

what are the changes? 
- add new color to the color palette
- add class to options chip to be able to update it's color
- remove opacity from page background as the opacity should be on the footer only


ui/lib/theme.ts
```diff
@@ -26,6 +26,7 @@ export const theme: DefaultTheme = {
     primaryLight: "#98E0F7",
     primary10: "#009CCC",
     primary20: "#006B8E",
+    primary05: "#E5F7FD",
```
foot, line 29: Maybe pop it up the top to keep the ordering

```suggestion
    primary05: "#E5F7FD",
    primary10: "#009CCC",
    primary20: "#006B8E",
```

ui/typedefs/styled.d.ts
```diff
@@ -7,6 +7,7 @@ export namespace colors {
   const primaryLight: string;
   const primary10: string;
   const primary20: string;
+  const primary05: string;
```
foot, line 10: ordering too

ui/components/DataTable.tsx
```diff
@@ -616,6 +616,9 @@ export const DataTable = styled(UnstyledDataTable)`
     overflow: hidden;
     text-overflow: ellipsis;
   }
+  .filterOptionsChip {
```
foot, line 619: css conventions is usually to use kebab-case. We don't have this documented anywhere though. Is it mixed up around the place?

```suggestion
  .filter-options-chip {
```

ui/lib/theme.ts
foot: Nice, is the color straight from figma?

foot (commented):
Nice! Will take it for a spin..

ui/components/ChipGroup.tsx
```diff
@@ -28,11 +28,18 @@ function ChipGroup({ className, chips = [], onChipRemove, onClearAll }: Props) {
...
               onDelete={() => onChipRemove([chip])}
+              className="filter-options-chip"
             />
           </Flex>
         );
       })}
-      {chips.length > 0 && <Chip label="Clear All" onDelete={onClearAll} />}
+      {chips.length > 0 && (
+        <Chip
+          className="filter-clearAll-Chip"
```
foot, line 38: Maybe rm this if not using. Just to keep things tidy, I think if we declare a className I'd assume we're using it.

foot (approved):
👍

ui/components/DataTable.tsx
```diff
@@ -616,6 +616,9 @@ export const DataTable = styled(UnstyledDataTable)`
     overflow: hidden;
     text-overflow: ellipsis;
   }
+  .filter-options-chip {
```
joshri, line 619: 💡 Adding the class name on DataTable will change the color of all the chips - at least in oss. Do we need the class name to have the og color in enterprise somewhere else or can we just change it inside ChipGroup?

ui/components/Page.tsx
```diff
@@ -17,7 +17,7 @@ export type PageProps = {
 };
 
 export const Content = styled(Flex)`
-  background-color: rgba(255, 255, 255, 0.75);
+  background-color: rgba(255, 255, 255, 1);
```
joshri, line 20: 💡 If we're ditching the opacity we can use `props.theme.colors.white` since I know there were plans to do a dark mode eventually. Definitely not a necessary change

joshri (approved):
It works - see my comment about the className for the color change - if it's possible for it to be the default for all chips I'd like that better but it's ultimately fine either way

PR #2804, opened 2022-09-28
Split pkg/run into 3 packages

ozamosi:
There's 3 distinct phases:
 * First, installation of flux and/or the dashboard
 * Then, a watch phase, which is the main phase
 * Finally, a bootstrap phase.

These are now in 3 different sub-packages of run. The run package still provide shared functionality - mainly functionality shared between applying the dashboard and applying the manifests in the bucket.


opudrovs (approved):
Tested it, works fine. And the code structure is easier to grasp now. 🎉

PR #2809, opened 2022-09-28
ui: Don't flatten children for the graph

### LINKED ISSUE #2808: [UI] v0.9.6.rc1 displays incorrect graph elements as children of Helm chart
After upgrading to the new v0.9.6.rc1 the graph started to show direct child-elements of helm charts that shouldn't be considered child elements. For instance, it shows the deployment and its subsequent child elements correctly - however it also shows the replicaset as a direct child, even though it's spawned from the deployment. See attached pictures.

**Environment**

- GitOps version: v0.9.6.rc1 and v0.9.5 - Helm chart 4.0.3
- Flux Version: 0.34.0
- Kubernetes version: 1.22.8

**To Reproduce**

1. Update image to v0.9.6.rc1
2. Check any Helm Chart that has a deployment which spawns pods.
3. Rollback to v0.9.5 and see the difference.

**Expected behavior**
Show hierarchy correctly. 

**Actual Behavior**
Showing child objects as direct child elements of the Helm Chart which are misplaced in the hierarchy multiple times. See below for screenshots which clearly shows the difference:

v0.9.5:
image

v0.9.6.rc1:
image

ozamosi:
I added the flattenChildren in da33a4b, where I assumed that ReconciliationGraph used getChildren directly. Which it doesn't.

flattenChildren thus did it's thing of copying the children out to the top layer so they show up in the top level of the array so they show up in the Details view. But that also created copies at the wrong depth in the graph.

This fixes it by introducing two different ways to get reconciled objects - one that is flat and one that isn't.

I'm hoping we'll change from the tree structure to the d3-dag structure one of these days, at which point it won't need to be nested and this code can be removed.

This fixes #2808

### DIFF
--- a/ui/components/ReconciliationGraph.tsx
+++ b/ui/components/ReconciliationGraph.tsx
@@ -2,7 +2,7 @@ import { Slider } from "@material-ui/core";
 import * as d3 from "d3";
 import * as React from "react";
 import styled from "styled-components";
-import { useGetReconciledObjects } from "../hooks/flux";
+import { useGetReconciledTree } from "../hooks/flux";
 import { ObjectRef } from "../lib/api/core/types.pb";
 import { Automation } from "../lib/objects";
 import DirectedGraph from "./DirectedGraph";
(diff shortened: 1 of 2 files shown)


opudrovs (approved):
Tested it locally, looks like it fixed the issue (if I got what was wrong right). I was able to reproduce it in the main branch and don't see it in the current branch.

PR #2832, opened 2022-10-04
update ListFluxCrds to support querying multiple clusters

rparmer:
adds functionality needed to fix weaveworks/weave-gitops-enterprise#1621


core/server/fluxruntime.go
```diff
@@ -109,37 +109,63 @@ func (cs *coreServer) ListFluxCrds(ctx context.Context, msg *pb.ListFluxCrdsRequ
...
-				Plural: d.Spec.Names.Plural,
-				Group:  d.Spec.Group},
-			Version:     version,
-			Kind:        d.Spec.Names.Kind,
-			ClusterName: msg.ClusterName,
-			Uid:         string(d.GetUID()),
+				r := &pb.Crd{
+					Name: &pb.Crd_Name{
+						Plural: d.Spec.Names.Plural,
+						Group:  d.Spec.Group},
```
jpellizzari, line 160: Formatting looks a little weird here. I would expect `gofmt` to put this end `}` on a new line? Maybe it is not that opinionated...

core/server/fluxruntime.go
```diff
@@ -109,37 +109,63 @@ func (cs *coreServer) ListFluxCrds(ctx context.Context, msg *pb.ListFluxCrdsRequ
...
 	opts := client.MatchingLabels{
 		coretypes.PartOfLabel: FluxNamespacePartOf,
 	}
 
-	list := &apiextensions.CustomResourceDefinitionList{}
+	for clusterName, lists := range clist.Lists() {
+		for _, l := range lists {
+			list, ok := l.(*apiextensions.CustomResourceDefinitionList)
+			if !ok {
+				continue
```
jpellizzari, line 143: Are we sure we want to just `continue` here? Has something gone horribly wrong if we fail this type assertion? (not rhetorical, I really don't know).

jpellizzari (approved):
No blockers for me.

core/server/fluxruntime.go
rparmer: `go fmt` didn't care, but yeah I agree it looks better on a new line

core/server/fluxruntime.go
rparmer: Good question.  I saw that is what was done for the tf stuff so i just mirrored it.  IMO continue seems ok.  We could add some logging so something but we probably don't want to throw an error back to the user

rparmer:
> Looks ok, my only questions are about existing code rather than code being added as part of this PR.
> 
> I noticed that the request object includes `Namespace` which is not being used ? Can we remove it from the definition if it's not required?
> 
> Also noticed that in other parts of the system i.e. `ListObjects` we do the following
> 
> ```
> if msg.ClusterName != "" {
>     clustersClient, err = cs.clustersManager.GetImpersonatedClientForCluster(ctx, auth.Principal(ctx), msg.ClusterName)
> } else {
>     clustersClient, err = cs.clustersManager.GetImpersonatedClient(ctx, auth.Principal(ctx))
> }
> ```
> 
> Why don't we need to call `GetImpersonatedClientForCluster` here too?

I think it is just a difference of implementation.  I mirrored the `ListFluxRuntimeObjects` which does not use the `GetImpersonatedClientForCluster` function.  However, currently the request options for `ListFluxRuntimeObjects` and `ListFluxCrds` are not used.  We should probably look into if they are still needed or if they should be refactored to use the `GetImpersonatedClientForCluster` call.  It sounded like the original issue was high priority and it made more sense to fix the issue now and look into refactoring later.

core/server/fluxruntime.go
```diff
@@ -109,37 +109,64 @@ func (cs *coreServer) ListFluxCrds(ctx context.Context, msg *pb.ListFluxCrdsRequ
...
+			}
 
-	if err := clustersClient.List(ctx, msg.ClusterName, list, opts); err != nil {
-		respErrors = append(respErrors, &pb.ListError{ClusterName: msg.ClusterName, Message: fmt.Sprintf("%s, %s", errors.New("could not list CRDs"), err)})
-	}
+			if err := clustersClient.List(ctx, clusterName, list, opts); err != nil {
```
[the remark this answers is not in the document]
rparmer: Ah ok.  That makes sense.  I should have realized that 🤦‍♂️.  I moved the opts to the `ClusteredList` call and it is working as expected.

PR #2835, opened 2022-10-05
Fix styling bug on filter chips

### LINKED ISSUE #2784: ChipGroup styling on overflow
image

This marvelous screenshot contains a couple unfortunate bugs:
- a user can submit an empty input in the search input resulting in an empty chip (you can see this in the first chip)
- chip.search(filterSeparator) needs to be true before making the comparison between lengths. Lengths of 1 character return -1, which is equal to a false chip.search return (see cnull chip)
- Most egregious - ChipGroup is somehow allowed to push CheckboxActions and the search and filter buttons out of view. This is actually breaking some functionality of the app, although what you gain from having that many text filters active I do not know.

joshri:
Closes #2784

- ChipGroup properly overflows without pushing CheckboxActions or the icon buttons off the screen
- A bug where a one character text search could meet the requirements for an undefined filter has been fixed
- An empty text search is not able to be submitted as a valid filter

https://user-images.githubusercontent.com/65822698/194111349-33b058ba-0dbb-40e1-9748-973ac9aae02c.mov

### DIFF
--- a/ui/components/ChipGroup.tsx
+++ b/ui/components/ChipGroup.tsx
@@ -18,11 +18,13 @@ function ChipGroup({ className, chips = [], onChipRemove, onClearAll }: Props) {
   return (
     <Flex className={className} wide align start>
       {_.map(chips, (chip, index) => {
+        const hasSeparator = chip.search(filterSeparator);
         const isUndefined =
+          hasSeparator !== -1 &&
           //javascript search finds first occurence of substring, returning index
-          chip.search(filterSeparator) ===
-          //if first occurence of filterSeparator is the end of the chip string, it's an undefined value
-          chip.length - filterSeparator.length;
+          hasSeparator ===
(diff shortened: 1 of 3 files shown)


opudrovs:
Tried submitting text strings consisting of space characters only. Submitted the following strings individually ` `, `  `, `   `, `    `, `     `, `      ` (from 1 space to 6 spaces) and it resulted in six small chips.

Is it expected behavior or is it beyond the scope of this issue?

Screenshot 2022-10-05 at 18 57 43

opudrovs:
Sounds good. Can we have a search term ending with a space or spaces? What columns are we searching within now? Can their values have spaces?

opudrovs (approved):

opudrovs:
Just noticed, a test is failing. Otherwise it LGTM!

PR #2837, opened 2022-10-05
Handle conditions without a type of "Ready"

### LINKED ISSUE #2796: Incorrect message from computeMessage function

joshri:
Closes #2796 

In adding display of a lot more objects in #2765, we noticed that `computeReady` and `computeMessage` weren't working as expected in some of the new object kinds (the most visible example was in CRDs in the `ReconciledObjectsTable`). 

What was happening here is that we were specifically looking for conditions with a `type` of `Ready` or `Available`, and assigning `NotReady` to the status when that type didn't exist. It turns out that CRDs don't have a conditions object with a type of Ready, but that doesn't mean that they are failing.

What made the most sense to me for now was to add a little extra handling for when Ready isn't there - if ANY conditions object has a status of "False", we assign `NotReady` and display that conditions message, otherwise, we assign "Ready" and display the message of the first conditions object in the list. 

The problems with this are:
- Some conditions use negative polarity, where the `normal` state is actually `False`. There's no standard for indicating this. 
- When all conditions are True, the message of the first condition in the list might not be the most relevant. There is no standard on how to indicate which conditions are more important than others.

I found this which is a.....semi-recent discussion on setting some new standards for conditions
https://github.com/kubernetes/community/pull/4521

Here's what the guidelines say right now:
https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#spec-and-status

Here's a screenshot of the ReconciledObjectsTable after my changes:
image


opudrovs (approved):

opudrovs:
LGTM! 🎉 A very interesting issue.

PR #2838, opened 2022-10-06
implement vcluster-based session

chanwit:


opudrovs:
@chanwit 

> The outer process starts in a session by default (with --session=<session name> and --disable-session=false).

As far as I understood from code, if these flags are not provided when running the command, the command will run as before without a vcluster, correct?

>The outer process starts in a session by default (with --session=<session name> and --disable-session=false).
>Then the outer process forks the inner process (with --disable-session=true).
>The inner process installs GitOps Run related components as it runs outside a session.

And how does the watcher run now? I see it in `runCommandWithoutSession`, does it mean it runs in the inner process without a session?

chanwit:
@opudrovs please try running it on macOS, as it needs a test for process enumeration and signaling.

opudrovs:
@chanwit on it!

opudrovs:
Your instructions in the PR description list `--session` flag. So, I had to replace it with `--session-name` otherwise the output complains

```
Error: unknown flag: --session
```

Output for two attempts to run the command against a clean cluster:

```
macbook-pro:podinfo olga$ /Volumes/Macintosh_HD_Data/_job_ww/_src/weave-gitops/bin/gitops beta run ./deploy/overlays/dev --timeout=3m --port-forward namespace=dev,resource=svc/backend,port=9898:9898 --session-name=session1 --disable-session=false
Error: couldn't find vcluster session1
```

and without the `session-name`

```
macbook-pro:podinfo olga$ /Volumes/Macintosh_HD_Data/_job_ww/_src/weave-gitops/bin/gitops beta run ./deploy/overlays/dev --timeout=3m --port-forward namespace=dev,resource=svc/backend,port=9898:9898 --disable-session=false
Error: couldn't find vcluster master-79f81383-dirty
```

So, what should the correct command to test the new functionality in this branch look like?

opudrovs:
The third step of the instructions (I guess it was implied) also requires specifying the `--session-name` or `--disable-session` flag , otherwise I get the following error:
```
Error: couldn't find vcluster master-79f81383-dirty
```

I've tested the code locally according to the instructions, everything works fine for me, both with the vcluster and when disabling the session. 🎉 I can access both dashboard and podinfo with port forwarding in both cases.

opudrovs:
>it needs a test for process enumeration and signaling.

It seems to work fine but if there is another command to get more detailed data on processes on macOS that you need please let me know.

pkg/run/git_exec.go
```diff
@@ -0,0 +1,44 @@
+package run
+
+import (
+	"os/exec"
+	"strings"
+)
+
+func GetBranchName() (string, error) {
+	cmd := exec.Command("git", "rev-parse", "--abbrev-ref", "HEAD")
```
opudrovs, line 9: @ozamosi does it do the same thing as getting the branch name with go-git for the bootstrap wizard? What is the preferred way to get the branch name: this one or what we used in the wizard?

pkg/run/git_exec.go
```diff
@@ -0,0 +1,44 @@
...
+	out, err := cmd.Output()
+
+	if err != nil {
+		return "", err
+	}
+
+	return strings.TrimSpace(string(out)), nil
```
opudrovs, line 16: ❔ Why is trimming the space required here? Can this command return a string with trailing or leading spaces as the branch name?

pkg/run/session.go
```diff
@@ -0,0 +1,68 @@
...
+		allProcesses, err := ps.Processes()
+
+		if err != nil {
+			return
+		}
+
+		for _, proc := range allProcesses {
+			if proc.PPid() == thisProc {
+				// ok it's a child process, notify it
+				pp, err := os.FindProcess(proc.Pid())
```
opudrovs, line 49: ❔ Here I just didn't quite get what we are doing, could you please clarify it a bit?

`Pid`s are unique for each process, correct? So, if we already have a process `proc` and just checked that its parent process's ID is `thisProc`, why do we also need to search for `pp` here `pp, err := os.FindProcess(proc.Pid())`? Because if process IDs are unique, wouldn't be the `pp` process returned by `os.FindProcess(proc.Pid())`  the same process as `proc`?

❔ BTW, how should the `pp` abbreviation/variable name be deciphered here?

opudrovs (approved):
Tested it, works fine. Totally cool functionality! 🎉

pkg/run/git_exec.go
chanwit: It's for eliminating "\n" "\r" printed out by the command.

pkg/run/git_exec.go
chanwit: Let's discuss!

pkg/run/session.go
chanwit: The loop is to find it (an `int`), and `os.FindProcess()` is to obtain the process object (a `struct` representing the process).

opudrovs: Ah, ` proc` is of type `ps.Process` and `pp`, returned by `os.FindProcess(` is of type `*os.Process`, so they are different types and we actually need to get the object of type `*os.Process` in the end?

chanwit: Oh. Good catch. Let me revisit it.

chanwit: No. `proc` is of `ps.Process`. What we need to use is `*os.Process`.
`ps.Process` belongs to the `go-ps` package which is for process enumeration.

opudrovs: Yes, that's what I wrote. Thanks for confirming it.

PR #2840, opened 2022-10-06
Add `gitops set config` and `gitops get config` GitOps CLI commands.

### LINKED ISSUE #2772: GitOps set config command

### LINKED ISSUE #2811: GitOps get config command
As a user, I want to see a print out of my CLI configuration for Weave GitOps.

### Acceptance Criteria
GIVEN I run the command `gitops get config`

1.  WHEN it is successful I THEN see a simple JSON printout of the "key" and "value" such as 

```
{
    "analytics": true,
    "userId": "12345"
}
```

2. WHEN it is unsuccessful THEN I let the user know there was an error when trying to process the file and that there is an issue with the formatting.  As well as simple instructions that it should represent a JSON object format such as:

```
{
    "analytics": true,
    "userId": ""
}
```

3. WHEN I run the command on the Enterprise CLI THEN I get the same behavior.

opudrovs:
Part of #2772
Part of #2811

- Added the `gitops set`, `gitops set config`, and `gitops get config` GitOps CLI commands.

- Added the `pkg/config` package with functions to read and write the CLI configuration for Weave GitOps and generate a user ID for analytics.

- Added tests for selected (where it makes sense) functions from the `pkg/config` package.

- Tested that new commands work with WG Enterprise. Will raise a PR in the enterprise repo with the `add-gitops-set-config-and-get-config-commands` branch when I get the write access.

- Removed enterprise examples from the `gitops get` command.

Notes and questions:

- The existing options type at `github.com/weaveworks/weave-gitops/cmd/gitops/config` can be mixed up with the new `github.com/weaveworks/weave-gitops/pkg/config` package containing config utility functions.

Maybe the name of the `github.com/weaveworks/weave-gitops/pkg/config` package should be changed?

- Do we need to provide example user ID here?

```
{
    "analytics": true,
    "userId": ""
}
```
In acceptance criteria it's 12345, but we changed it to a base64 encrypted string.

### DIFF
--- a/cmd/gitops/get/cmd.go
+++ b/cmd/gitops/get/cmd.go
@@ -5,6 +5,7 @@ import (
 
 	"github.com/weaveworks/weave-gitops/cmd/gitops/config"
 	"github.com/weaveworks/weave-gitops/cmd/gitops/get/bcrypt"
+	configCmd "github.com/weaveworks/weave-gitops/cmd/gitops/get/config"
 )
 
 func GetCommand(opts *config.Options) *cobra.Command {
@@ -12,17 +13,16 @@ func GetCommand(opts *config.Options) *cobra.Command {
 		Use:   "get",
 		Short: "Display one or many Weave GitOps resources",
 		Example: `
-# Get all CAPI templates
-gitops get templates
+# Get the CLI configuration for Weave GitOps
+gitops get config
 
-# Get all CAPI credentials
-gitops get credentials
-
-# Get all CAPI clusters
-gitops get clusters`,
+# Generate a hashed secret
+PASSWORD="<your password>"
+echo -n $PASSWORD | gitops get bcrypt-hash`,
 	}
 
 	cmd.AddCommand(bcrypt.HashCommand(opts))
+	cmd.AddCommand(configCmd.ConfigCommand(opts))
 
 	return cmd
 }
--- a/cmd/gitops/get/config/cmd.go
+++ b/cmd/gitops/get/config/cmd.go
@@ -1,0 +1,53 @@
+package config
+
+import (
+	"fmt"
+	"os"
+
+	"github.com/spf13/cobra"
+	"github.com/weaveworks/weave-gitops/cmd/gitops/config"
+	clilogger "github.com/weaveworks/weave-gitops/cmd/gitops/logger"
+
+	gitopsConfig "github.com/weaveworks/weave-gitops/pkg/config"
+)
+
+func ConfigCommand(opts *config.Options) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "config",
+		Short: "Prints out the CLI configuration for Weave GitOps",
+		Example: `
+# Prints out the CLI configuration for Weave GitOps
+gitops get config`,
+		SilenceUsage:      true,
+		SilenceErrors:     true,
+		RunE:              getConfigCommandRunE(opts),
+		DisableAutoGenTag: true,
+	}
+
+	return cmd
+}
+
+func getConfigCommandRunE(opts *config.Options) func(*cobra.Command, []string) error {
+	return func(cmd *cobra.Command, args []string) error {
+		var err error
+
+		log := clilogger.NewCLILogger(os.Stdout)
+
+		cfg, err := gitopsConfig.GetConfig(log, false)
+		if err != nil {
+			return err
+		}
+
+		log.Successf("Your CLI configuration for Weave GitOps:")
+
+		cfgStr, err := cfg.String()
+		if err != nil {
+			log.Failuref("Error printing config")
+			return err
+		}
+
+		fmt.Println(cfgStr)
+
+		return nil
+	}
+}
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -1,0 +1,208 @@
+package config
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io"
+	"io/fs"
+	"math/rand"
+	"os"
+	"path/filepath"
+
+	"github.com/weaveworks/weave-gitops/pkg/logger"
+)
+
+const (
+	ConfigFileName       = "weave-gitops-config.json"
+	WrongConfigFormatMsg = `Your CLI configuration for Weave GitOps should represent a JSON object format:
+{
+	"analytics": true,
+	"userId": ""
+}
+Please set you configuration with: gitops set config`
+)
+
+var letters = []byte("abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789+/")
+
+type GitopsCLIConfig struct {
+	Analytics bool   `json:"analytics"`
+	UserID    string `json:"userId"`
+}
+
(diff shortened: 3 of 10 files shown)


pkg/config/config.go
```diff
@@ -0,0 +1,123 @@
...
+	configPath := filepath.Join(userConfigDir, filename)
+
+	return configPath, nil
+}
+
+// OpenConfigFile opens existing config file or creates one if it does not exist
+func OpenConfigFile(configPath string, shouldCreate bool) (*os.File, error) {
+	_, err := os.Stat(configPath)
+	if err != nil {
+		if !errors.Is(err, os.ErrNotExist) || !shouldCreate {
```
opudrovs, line 47: @ozamosi should the outer and inner conditions be combined on the same line? I tried it, it looks too complex.:)

pkg/config/config.go
```diff
@@ -0,0 +1,123 @@
...
+		return fmt.Errorf("error writing to config file: %w", err)
+	}
+
+	return nil
+}
+
+var letters = []byte("abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ")
+
+// GenerateUserID generates a string of specified length made of random characters and encodes it in base64 format
+func GenerateUserID(numChars int, seed int64) string {
```
ozamosi, line 114: Given that you've chosen to make the original ID printable, there's no real need to base64 encode it again - you could add `0123456789+/` to the letters array, and it would already be a base64 representation of a random value.

cmd/gitops/get/config/cmd.go
```diff
@@ -0,0 +1,70 @@
...
+		log.Actionf("Reading GitOps CLI config ...")
+
+		configPath, err := gitOpsCLIConfig.GetConfigPath(gitOpsCLIConfig.ConfigFileName)
+		if err != nil {
+			return err
+		}
+
+		configFile, err := gitOpsCLIConfig.OpenConfigFile(configPath, false)
+		if err != nil {
+			log.Failuref("Error opening config file: %w at path: %s", err, configPath)
```
ozamosi, line 45: This needs to be `%v` to wrap the error, instead of `%w` to debug-print the error. (I think the same thing happens in set config)

Right now it prints `✗ Error opening config file: %!w(*fmt.wrapError=&{failed to open config file: open /home/robin.sonefors/.config/weave-gitops-config.json: no such file or directory 0xc0000376e0})` whereas with `%v` it prints properly.

cmd/gitops/get/cmd.go
```diff
@@ -5,24 +5,24 @@ import (
...
+gitops get config
 
-# Get all CAPI credentials
-gitops get credentials
-
-# Get all CAPI clusters
-gitops get clusters`,
+# Generate a hashed secret
+# PASSWORD="<your password>"
+# echo -n $PASSWORD | gitops get bcrypt-hash`,
```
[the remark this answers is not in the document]
opudrovs: Don't know, copied them from bcrypt directly, but probably not.  😄  Will remove the`#`s.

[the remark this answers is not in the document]
opudrovs: And in `get bcrypt` too then.

pkg/config/config.go
ozamosi: To me, the bit that makes it look complex comes from the number of negations. I'd probably find this more readable if written as:

```
if err != nil {
  if errors.Is(err, os.ErrNotExist) && shouldCreate {
    // we'll just create a new file
    err = nil
  } else {
    return nil, err
  }
}
```

cmd/gitops/get/config/cmd.go
opudrovs: Just replaced `%w` with `%v` in config-related files everywhere, I hope it works.

pkg/config/config.go
opudrovs: Reworked it as suggested, just had to remove the `err = nil` assignment because I was getting `::error file=pkg/config/config.go,line=49,col=4::ineffectual assignment to err (ineffassign)` from the linter.

cmd/gitops/get/cmd.go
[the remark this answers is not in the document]
opudrovs: Removed, from both `gitops get config` and `gitops get bcrypt` commands.

pkg/config/config.go
opudrovs: Added the characters and removed the additional base24 encoding.

> Given that you've chosen to make the original ID printable

What would making the original ID non-printable look like? Some binary operations on numbers or what?

cmd/gitops/get/config/cmd.go
opudrovs: OK, re-replaced `%v` with `%w` where appropriate, I hope NOW it works.

pkg/config/config.go
ozamosi:
```
b := make([]byte, 10)
_, err := rand.Read(b)
```

opudrovs: Cool! Would you prefer to use it instead?

pkg/config/config.go
```diff
@@ -0,0 +1,123 @@
...
+		return "", fmt.Errorf("user config directory could not be determined: %w", err)
+	}
+
+	configPath := filepath.Join(userConfigDir, filename)
+
+	return configPath, nil
+}
+
+// OpenConfigFile opens existing config file or creates one if it does not exist
+func OpenConfigFile(configPath string, shouldCreate bool) (*os.File, error) {
```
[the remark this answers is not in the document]
opudrovs: Reworked the functions, exported only suggested functions, fixed some logic/flow errors. The code now seems to work as expected (including autoreplacing an invalid config for the `set config` command). But some if/else's look rather complex, not sure how to simplify them more with keeping the logic.

The only way I can think of is splitting `openConfigFile` to two functions without the current `shouldCreate` param: one of them will work as the previous `openConfigFile` with `shouldCreate`, the other one as one without.

pkg/config/config.go
```diff
@@ -0,0 +1,208 @@
...
+	}
+
+	configFile, err := openConfigFile(configPath, true)
+	if err != nil {
+		return fmt.Errorf("error opening config file: %w", err)
+	}
+
+	defer configFile.Close()
+
+	err = configFile.Truncate(0)
```
ozamosi, line 112: :warning: You know you just opened the file, so you don't need to truncate and seek - you would need to explicitly open it in append mode for "open(); write()" to not do what you are doing here.

:bulb: It's common for data encoding and decoding APIs to have a way to use a IO reader or writer (which is to say, an open file) to stream data to/from - in the json module, there's an `Encoder` and a `Decoder` that you could use directly with the file descriptor. This might have been important if the data could be "big" (since it saves copying the data), but for this it doesn't really matter.

ozamosi (approved):

pkg/config/config.go
opudrovs: Thanks, good to know!

opudrovs: Removed unneeded calls to `Truncate` and `Seek`.

PR #2845, opened 2022-10-07
Add LinkResolver for customized routing when running as "embedded"

jpellizzari:
Part of https://github.com/weaveworks/weave-gitops-enterprise/issues/1661

Allows us to override links in specific instances from "outside" the component tree:

```jsx
    const MyComponent = () => {
      const linkResolver = useLinkResolver();

      return (
        <a data-testid="link" href={linkResolver("")}>
          My Link
        </a>
      );
    };

//...

      <LinkResolverProvider
        resolver={() => {
          return "/some/path";
        }}
      >
        <MyComponent />
      </LinkResolverProvider>

// ^^ This will result in `a href="/some/path" />` allowing us to add custom routing in Enterprise
```

### DIFF
--- a/ui/components/ReconciledObjectsTable.tsx
+++ b/ui/components/ReconciledObjectsTable.tsx
@@ -2,11 +2,12 @@ import _ from "lodash";
 import * as React from "react";
 import styled from "styled-components";
 import { AppContext } from "../contexts/AppContext";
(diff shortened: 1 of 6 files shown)


joshri (approved):
Tested all routes locally, everything looks good 🌞

PR #2847, opened 2022-10-10
create vcluster instance in session

### LINKED ISSUE #2839: GitOps Run - create an vcluster instance based on the session name
This ticket is a follow-up one for the session control.
It is for creating a vcluster instance by applying its Helm chart (similar to how we install the dashboard).

We would allow optional service mapping too. But we don't know how to let the users specify a list of services they want to map.

chanwit:
This PR creates a vcluster instance in a session.
Fixes #2839 


ozamosi (approved):

opudrovs:
❔ Just a question, why is `OpenRepository ` and `GetRemoteURL` preferred naming in Go over `ParseGitRemote` and `ParseRemoteURL`? To hide implementation details, like we are not really interested in how we get them, we just need the result?

chanwit:
@opudrovs reverted during the rebase

PR #2849, opened 2022-10-11
Actually default-yes for installing dashboard

ozamosi:
The instructions and the implementation didn't line up - we claimed yes was default, but just hitting enter caused the value to be "" which meant we didn't actually install it.

This is caused by using `IsConfirm`, where it doesn't set the default as the default value, but rather uses the error value as `nil` if the user agreed, and promptui.ErrAbort (which is `errors.New("")`) if the user didn't.

### DIFF
--- a/cmd/gitops/beta/run/cmd.go
+++ b/cmd/gitops/beta/run/cmd.go
@@ -6,12 +6,9 @@ import (
 	"os"
 	"os/signal"
 	"path/filepath"
-	"strings"
 	"sync/atomic"
 	"syscall"
 	"time"
-
-	"github.com/weaveworks/weave-gitops/pkg/kube"
 
 	"github.com/fsnotify/fsnotify"
 	"github.com/manifoldco/promptui"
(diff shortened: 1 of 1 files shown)


cmd/gitops/beta/run/cmd.go
```diff
@@ -339,8 +339,10 @@ func runCommandWithoutSession(cmd *cobra.Command, args []string) error {
 			Default:   "Y",
 		}
 
-		result, err := prompt.Run()
-		if err == nil && strings.ToUpper(result) == "Y" {
+		// Answering "n" causes err to not be nil. Hitting enter without typping
```
opudrovs, line 342: a typo "typping"

ozamosi:
The code says `Default: "Y"`, so it prints `[Y/n]` - I don't particularly care whether yes or no is the default, but the code reads misunderstands the return value.

opudrovs:
Ah, I see. Good catch will test it later.

opudrovs:
Tested and confirmed the issue. isConfirm sets input to "" in the promptui code:

```
var inputErr error
input := p.Default
if p.IsConfirm {
    input = ""
}
```

opudrovs (approved):

PR #2852, opened 2022-10-11
gitops run: Add silly little struct for paths

ozamosi:
I'm not sure how much of a good idea this actually is - it's definitely under-developed. However, I had a small breakdown while trying to get paths right and had to try something to sort some paths out.

Basically, we expect the git repo to be "somewhere". Inside that git repository, there's a manifests directory, which is the same as the current working directory. Inside that directory, there's a clusters directory. Somewhere, probably but not necessarily inside the manifests directory but definitely inside the git directory, there's a workload directory. All of these paths can be refered to as relative paths, absolute paths, and some of them (in flux) should be prefixed with `./`.

This bans all somethingPath variables, and puts a paths variable instead. That variable is a struct that has documented members so we can remember which path is which. There's no variable called `relativePath`.

Inside this struct, we should check those invariants described above, but for the moment I mainly just copy-pasted the code we already had.

### DIFF
--- a/cmd/gitops/beta/run/cmd.go
+++ b/cmd/gitops/beta/run/cmd.go
@@ -186,43 +186,12 @@ func runCommandWithoutSession(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
-	gitRepoRoot, err := install.FindGitRepoDir()
-	if err != nil {
-		return err
-	}
-
-	rootDir := flags.RootDir
-	if rootDir == "" {
-		rootDir = gitRepoRoot
-	}
-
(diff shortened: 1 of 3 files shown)


opudrovs:
Great job! 🎉 

⚠️ The only minor point: could you please move this new type + methods + related functions out of the main GitOps Run cmd file? It looks like quite a standalone thing. As we discussed today, the cmd file has grown too much. I see that you are removing some lines, but it would be nice to reduce its size even more.

opudrovs (approved):

PR #2862, opened 2022-10-12
gitops run: Stop the wizard accepting missing values

ozamosi:
Right now every field in the wizard is mandatory. Therefore, I've made it refuse to accept an empty field.

At some point, we'll probably have to re-introduce optional fields and isRequired, and add some kind of type validation. That day is not today.

Also change the return value to be a map instead of a list.

### DIFF
--- a/pkg/run/bootstrap/bootstrap_wizard_ui.go
+++ b/pkg/run/bootstrap/bootstrap_wizard_ui.go
@@ -159,7 +159,7 @@ func (m preWizardModel) View() string {
 type wizardModel struct {
 	textInputs []textinput.Model
 	prompts    []string
-	msgChan    chan []*BootstrapCmdOption
+	msgChan    chan BootstrapCmdOptions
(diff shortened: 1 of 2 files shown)


pkg/run/bootstrap/bootstrap_wizard_ui.go
```diff
@@ -242,23 +244,19 @@ func (m wizardModel) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
...
 
 				for i, input := range m.textInputs {
 					prompt := m.prompts[i]
 
 					value := strings.TrimSpace(input.Value())
 
 					if value == "" {
-						continue
+						m.errorMsg = "Missing value in " + input.Placeholder
+						return m, tea.Batch()
```
ozamosi, line 256: @opudrovs I'm pretty sure I'm violating all the rules here - do you know how I'm supposed to write this?

PR #2869, opened 2022-10-12
update status icons used in tables update user setting style

ahussein3:
- Partially Close [1666](https://github.com/weaveworks/weave-gitops-enterprise/issues/1666)
what are the Changes ? 
- update account setting style to match the design.
- update fail-icon to be the SVG imported from MUI instead of the image to be able to accept size prop.
- updating the DataTable header to match the design


joshri (approved):
Only thing I noticed is that the header looks slightly different with checkboxes - The padding on the checkbox button adjusts the height a little bit:

image

image

ahussein3:
@joshri  you are right will update the padding on the checkbox so that header match with/without checkbox. Thanks

PR #2871, opened 2022-10-13
fix bootstrapping process in the session mode

chanwit:
This PR fixes the bootstrap process after closing a session by:
  - Making GitOps Run aware of bootstrapping after closing a session
  - Adding `--dashboard-hashed-password`. The outer process is now responsible for asking the dashboard question and it passes the hashed password into the inner process.

Fixes #2863

PR #2877, opened 2022-10-13
Improve gitops run bootstrap questions

ozamosi:
This contains 2 fixes:
 * Remove question mark from questions - promptui already adds them.
 * Don't forward errors from promptui - error is how it signals a negative answer, and that's OK.

This addresses comments from #2871


opudrovs (approved):

ozamosi:
:facepalm: This is what happens when you try to do a quick fix.

4 question marks removed.

opudrovs:
Well, it was a quick 50% fix. Still not bad. 😸

PR #2881, opened 2022-10-13
Add analytics to helm chart on dashboard installation

### LINKED ISSUE #2771: Update dashboard commands for Pendo
As a user, when I run `gitops run` or `gitops create dashboard` I should have the analytics of the helm chart set to the appropriate value based on my response to providing "feedback" when I first run a CLI command.

### Acceptance Criteria

GIVEN I run `gitops get config` and I see `analytics: true`

1. WHEN I run `gitops run` and I am asked whether to install the GitOps Dashboard THEN I should have the value WEAVE_GITOPS_FEATURE_TELEMETRY set to true.
2. WHEN I run `gitops create dashboard` THEN I should have the value WEAVE_GITOPS_FEATURE_TELEMETRY set to true. 
3. WHEN I run either of the above commands and I am unable to parse the config file THEN I should just set WEAVE_GITOPS_FEATURE_TELEMETRY to false.  The commands should not be interrupted.

The place to put this is under `values`, so it shows up as 
```
envVars:
  - name: WEAVE_GITOPS_FEATURE_TENANCY
    value: "true"
```

joshri:
Closes #2771 

When creating the Helm Chart for the dashboard, we will now check the user's config file and set analytics accordingly

### DIFF
--- a/pkg/run/install/install_dashboard.go
+++ b/pkg/run/install/install_dashboard.go
@@ -297,8 +297,8 @@ func makeValues(username string, passwordHash string) ([]byte, error) {
 			}
 	}
 
-	gitopsConfig, _ := config.GetConfig(false)
-	if gitopsConfig.Analytics {
+	gitopsConfig, err := config.GetConfig(false)
+	if err == nil && gitopsConfig.Analytics {
 		valuesMap["WEAVE_GITOPS_FEATURE_TELEMETRY"] = "true"
 	}
 
--- a/pkg/run/install/install_dashboard.go
+++ b/pkg/run/install/install_dashboard.go
@@ -9,6 +9,7 @@ import (
 	helmv2 "github.com/fluxcd/helm-controller/api/v2beta1"
 	sourcev1 "github.com/fluxcd/source-controller/api/v1beta2"
 	loglevels "github.com/weaveworks/weave-gitops/core/logger"
+	"github.com/weaveworks/weave-gitops/pkg/config"
 	"github.com/weaveworks/weave-gitops/pkg/logger"
 	"github.com/weaveworks/weave-gitops/pkg/run"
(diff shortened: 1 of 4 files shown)


pkg/run/install/install_dashboard.go
```diff
@@ -284,20 +285,31 @@ func makeHelmRelease(log logger.Logger, name string, namespace string, username
...
+				"create":       true,
+				"username":     username,
+				"passwordHash": passwordHash,
+			}
 	}
 
-	jsonRaw, err := json.Marshal(valuesMap)
-	if err != nil {
-		return nil, fmt.Errorf("encoding values failed: %w", err)
+	analytics, _ := config.GetConfig(nil, false)
```
opudrovs, line 298: 💡 Would it be better to call this var `cfg` or smth. like that instead of `analytics`? We  are getting the whole CLI config here, not just the analytics field from it (even though what it contains now is related to analytics).

❔ Why are we ignoring the error here? Does it mean that the value of `valuesMap["WEAVE_GITOPS_FEATURE_TELEMETRY"]` should be always unset if we get an error when getting the config and we are not interested why exactly the error occured?

opudrovs (approved):

opudrovs:
@ozamosi I've
- added the global `sessionConfig` var to be able to set it for unit testing.
- fixed the helm release and values unit tests.
- fixed text messages and prompt (which was worse because it just hang the terminal) being forwarded to generated dashboard manifests if the `create dashboard` with `--export` flag on was the first command (before the analytics config was created) when running the GitOps CLI.

opudrovs:
Tested it, seems to work fine.

opudrovs:
This should be merged before the enterprise PRs.

ozamosi (approved):

PR #2900, opened 2022-10-17
Put logr between cli logger and stdout

### LINKED ISSUE #2857: Abstract the logger
In the `gitops run` codebase we want to be able to send logs to more than just the console.  We should update the code so we are no longer using the os packages `stdout` and `stdin` commands.  We should use a logger that can take levels and payloads.

ozamosi:
Before, we had a very simple log interface that wrote messages with added emoji to a writer, in practice stdout.

This deletes the interface altogether - instead, the log functions are now simply calls logr with an emoji prefix.

Then, it changes the fluxexec interface and the port forwarding function to steal the stdout and stderr messages, and log them as logr messages - info level for stdout, error level for stderr.

This means you can easily get out of the custom interface and switch to using it as logr, with verbosity and structured output right there
- but you'd have to worry about any emoji yourself.

Still sent to stdout without going through a logger are the dots during the upload progress (which, they're not log messages), and interactive questions related to the dashboard or the bootstrap.

The one thing that doesn't work is when the CLI returns an error out of cobra without logging it - we need to stop doing that.

To hook this up to something else, you need another zapcore Core thingy that writes somewhere - to test this, I just wrote json to stderr. I've left the code in there, commented out.

This fixes #2857.


opudrovs (approved):
Cool! 🎉

pkg/logger/logger.go
```diff
@@ -1,33 +1,77 @@
...
+}
 
-package logger
+// NewCLILogger returns a wrapped logr that logs to the specified writer
+// Note: unless you're doing CLI work, you should use core/logger.New instead
+func NewCLILogger(writer io.Writer) Logger {
+	return Logger{defaultLogr()}
+}
+
+// From wrapps a logr instance with the extra emoji generating helpers
```
opudrovs, line 24: A typo.

ozamosi: Fixed, thank you.

PR #2903, opened 2022-10-17
Add labels metadata display

### LINKED ISSUE #2889: Add labels section to flux automation details pages
Display a label section on all objects' details pages.

This section should live just under the metadata section, and work like the label view on clusters in enterprise:

joshri:
Closes #2889 

Adds label display as part of the Metadata component and a label getter to the FluxObject class

https://user-images.githubusercontent.com/65822698/196262483-4f8032c8-94eb-495d-ad14-b82f5cd9d808.mov


opudrovs (approved):
LGTM, great job! 🎉

The only minor issue is that horizontal scrolling in case of several tens of labels on an object is annoying. I wonder how many labels are usually present on real-world objects. But if it's expected behavior, fine. Maybe we could consider adding horizontal scrolling on mouse wheel when the pointer is over the labels or chips (or where else we have horizontal scrolling) view later?

ozamosi (approved):
For scrolling - I kind-of would have assumed it would overflow onto a second row if it didn't fit. But then, I didn't say that. Let's see how this works, we can always change it later.

> I'm getting the labels from `metadata.labels` - I'm assuming this is what we want and that `matchLabels` in selectors are something else 

Correct - `matchLabels` specifies a way to search for `metadata.labels`.

Will this change mean any code can be removed from enterprise now so it's consistent from now on?

opudrovs:
@ozamosi for the future improvements, yes, I agree that the labels should wrap and maybe introduce vertical scrolling within the area the labels are wrapped in if the area's height is exceeded (for example, if the max height of the area means that only 5 or 6 rows of wrapped labels can fit and we have 7 rows).

JamWils:
Yeah, the wrapping would be better for sure; horizontal scrolling should always be kept to a minimum 😂

PR #2914, opened 2022-10-19
Copy to clipboard button on Git auth flow

TheGostKasper:
Closes https://github.com/weaveworks/weave-gitops-enterprise/issues/1713
Currently the user is asked to copy across a code for auth with GitHub. Triple clicking to select all results in an additional space (as shown in screenshot) and then doesn't paste into the field on GitHub.

If we could provide a copy to clipboard button next to the code, that would make things much simpler than having to highlight.

image
image

### DIFF
--- a/ui/components/GithubDeviceAuthModal.tsx
+++ b/ui/components/GithubDeviceAuthModal.tsx
@@ -26,11 +26,23 @@ type Props = {
 const Pad = styled(Flex)`
   padding: 8px 0;
 `;
-
+const PointerIcon = styled(Icon)`
+  cursor: pointer;
+`;
 const ModalContent = styled(({ codeRes, onSuccess, onError, className }) => {
   // Move this to a component so that we get the cancel logic when the modal closes.
   const { getGithubAuthStatus } = useAuth();
   const [loading, setLoading] = React.useState(true);
+  const [copied, setCopied] = React.useState(false);
+
+  const handleCopy = React.useCallback(() => {
+    navigator.clipboard.writeText(codeRes.userCode).then(() => {
+      setCopied(true);
+      setTimeout(() => {
+        setCopied(false);
+      }, 3000);
+    });
+  }, []);
 
   React.useEffect(() => {
     if (!codeRes) {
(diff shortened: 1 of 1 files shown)


ui/components/GithubDeviceAuthModal.tsx
```diff
@@ -26,11 +26,23 @@ type Props = {
...
 const ModalContent = styled(({ codeRes, onSuccess, onError, className }) => {
   // Move this to a component so that we get the cancel logic when the modal closes.
   const { getGithubAuthStatus } = useAuth();
   const [loading, setLoading] = React.useState(true);
+  const [copied, setCopied] = React.useState(false);
+
+  const handleCopy = React.useCallback(() => {
+    navigator.clipboard.writeText(codeRes.userCode).then(() => {
+      setCopied(true);
+      setTimeout(() => {
```
joshri, line 41: ❔ Would it better for this to have a `.then` and a `.catch` instead of the `setTimeout`? We could use `notifyError` that way. This definitely returns a promise and I didn't treat it that way in `YamlView` - you'd be doing me a huge favor if you could move this function to `.utils` and export it to both this component and YamlView!!

ui/components/GithubDeviceAuthModal.tsx
```diff
@@ -54,7 +66,20 @@ const ModalContent = styled(({ codeRes, onSuccess, onError, className }) => {
...
-        <Text size="extraLarge">{codeRes.userCode}</Text>
+        <Text
+          size="extraLarge"
+          onClick={handleCopy}
+          style={{
+            display: "inline-flex",
+          }}
+        >
+          {codeRes.userCode}
+          <PointerIcon
```
joshri, line 77: 💡 I did this as a button in `YamlView` - if that code would work here you could get rid of the styled component with `cursor: pointer`!

joshri (approved):
Looks good! If this had a specific design from Figma let me know so I can unify it with the one in the YamlView component

ui/components/GithubDeviceAuthModal.tsx
TheGostKasper: `setTimeout` here is meant to change the icon back to copy instead of notifying the user with a copied message

ui/components/GithubDeviceAuthModal.tsx
TheGostKasper: good point :sparkles:

TheGostKasper:
Closing it as we're moving it to WGE

PR #2931, opened 2022-10-26
Replace text inputs with checkboxes in the bootstrap wizard UI

### LINKED ISSUE #2870: Replace text inputs with checkboxes in the bootstrap wizard UI
Replace text inputs with text-based checkboxes (`[x]`) for boolean values in the bootstrap wizard UI.

It will improve user experience and will just look cool.

opudrovs:
Closes #2870 

- Added type `bootstrapWizardInput` and related types.

- Replaced text fields with checkboxes for Boolean values in the bootstrap wizard. Checkboxes can be toggles with the Space key when focused.

- Minor refactoring:
-- added local variables for input lookup (`input := inputs[i]`) for readability
-- removed unnecessary copying of styles. I previously copied it from Bubble Tea examples, so I'm not sure why `style.Copy()` was used not only when creating new styles based on old styles but also when just rendering elements with in this style. So, removed it and tested it locally, it seems to work fine without the redundant copying. If anything goes wrong, we can always put it back, but it should work fine.

Questions:

- Maybe some type names could be shorter? But they were clashing with some other names or reserved words or didn't informative enough.

Checkbox toggled with space:

Screenshot 2022-10-26 at 13 39 14

Greyed out underscore marks false value. The color of the underscore can be changed or or the underscore can be removed easily if needed.

Focuses checkbox with true value + unfocused checkbox:

Screenshot 2022-10-26 at 13 39 21

Screenshot 2022-10-26 at 13 39 33

### DIFF
--- a/pkg/run/bootstrap/bootstrap_wizard_ui.go
+++ b/pkg/run/bootstrap/bootstrap_wizard_ui.go
@@ -43,6 +43,7 @@ const (
 
 type bootstrapWizardInput struct {
 	inputType     bootstrapWizardInputType
+	flagName      string
 	prompt        string
 	textInput     textinput.Model
 	checkboxInput *checkbox
@@ -203,6 +204,8 @@ func makeInput(task *BootstrapWizardTask, isFocused bool) *bootstrapWizardInput 
 		inputType = bootstrapWizardInputTypeTextInput
 	}
 
+	flagName := task.flagName
+
 	prompt := task.flagName + flagSeparator + task.flagDescription
 
 	ti := textinput.Model{}
@@ -234,6 +237,7 @@ func makeInput(task *BootstrapWizardTask, isFocused bool) *bootstrapWizardInput 
 
 	return &bootstrapWizardInput{
 		inputType:     inputType,
+		flagName:      flagName,
 		prompt:        prompt,
 		textInput:     ti,
 		checkboxInput: cb,
@@ -265,7 +269,7 @@ func (input *bootstrapWizardInput) getView(isFocused bool) string {
 		close = focusedStyle.Render(close)
 	}
 
-	return fmt.Sprintf("%s%s%s", open, checkmark, close)
+	return fmt.Sprintf("%s%s%s %s", open, checkmark, close, input.flagName)
 }
 
 func initialWizardModel(tasks []*BootstrapWizardTask, msgChan chan BootstrapCmdOptions) wizardModel {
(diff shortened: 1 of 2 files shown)


opudrovs:
Tested that command params are built properly with checkboxes.

```
params:
{1 map[branch:master hostname:github.com owner:stefanprodan pat:123 personal:true private:false repository:podinfo]}
```

pkg/run/bootstrap/bootstrap_wizard_ui.go
```diff
@@ -354,21 +446,24 @@ func (m wizardModel) getContent() string {
...
 	b.WriteString("Please enter the following values" + "\n" +
 		"(Tab and Shift+Tab to move input selection," + "\n" +
+		"(Space to toggle the currently focused checkbox," + "\n" +
 		"Enter to move to the next input or submit the form, " + "\n" +
 		"up and down arrows to scroll the view, Ctrl+C twice to quit):" + "\n\n\n")
 
-	for i := range m.textInputs {
-		b.WriteString(m.prompts[i])
+	for i := range m.inputs {
+		input := m.inputs[i]
```
ozamosi, line 454: :bulb: 

```suggestion
	for i, input := range m.inputs {
```
There's a few other for loops that can be simplified by just grabbing the second argument from range.

pkg/run/bootstrap/bootstrap_wizard_ui.go
```diff
@@ -162,62 +194,93 @@ func (m preWizardModel) getContent() string {
...
 	if isFocused {
-		ti.Focus()
-		ti.PromptStyle = focusedStyle
-		ti.TextStyle = focusedStyle
+		open = focusedStyle.Render(open)
+		close = focusedStyle.Render(close)
 	}
 
-	return ti
+	return fmt.Sprintf("%s%s%s", open, checkmark, close)
```
ozamosi, line 268: :warning: Could we maybe add the placeholder value here? When I see a line just saying `[x]` I panic that I don't know what it's asking me to do and refuse to read the help text above until I've figured it out, whereas `[x] private` would calm me down enough to pay attention.

ozamosi (approved):

pkg/run/bootstrap/bootstrap_wizard_ui.go
opudrovs: Good catch! JS habits die hard. 🤣

opudrovs: Reworked the loops.

pkg/run/bootstrap/bootstrap_wizard_ui.go
opudrovs: Added placeholders.

PR #2945, opened 2022-10-28
can support azure devops

enekofb:
required for https://github.com/weaveworks/weave-gitops-enterprise/issues/1704


foot:
I think we have moved this forward in another PR?

enekofb:
> I think we have moved this forward in another PR?

yes, closing thins and the related ones as it was part of the initial discovery work

PR #2951, opened 2022-10-31
update templates

sympatheticmoose:
- Makes it clearer which templates are for community users vs project contributors and weaveworkers
- adds ✨


makkes (approved):
🥇

Issue #2946, opened 2022-10-28
Refactor and dedup getKubeClient


chanwit:
Follow-up #2921

The getKubeClient codes are related to many parts, including the global flags, and are difficult to refactor. So they should be addressed separately in this ticket.

katya-makarevich:
@ozamosi please review as a first priority

PR #2956, opened 2022-11-01
re-organize gitops cmds to dedup codes

chanwit:
This PR re-organizes `gitops` sub-commands to 

- remove duplicated codes #2946
- fix persistent flags problems


cmd/gitops/config/options.go
```diff
@@ -1,11 +1,9 @@
...
-	GitHostTypes          map[string]string
-	InsecureSkipTLSVerify bool
-	Username              string
-	Password              string
-	Kubeconfig            string
+	Endpoint          string
+	OverrideInCluster bool
+	GitHostTypes      map[string]string
+	Username          string
+	Password          string
```
ozamosi, line 8: I think these all _should_ really be in enterprise - except username and password is reused in the dashboard install command.

cmd/gitops/root/root.go
```diff
@@ -0,0 +1,147 @@
...
+	"k8s.io/client-go/rest"
+)
+
+const defaultNamespace = "flux-system"
+
+var (
+	rootCmdFlags      = &config.Options{}
+	kubeconfigArgs    = genericclioptions.NewConfigFlags(false)
+	kubeclientOptions = new(runclient.Options)
+)
```
ozamosi, line 26: While this is very sleek, I think this won't work easily with enterprise right now, as it's just a copy of this file. I don't think it's insurmountable, but they _will_ be in another package.

ozamosi (commented):
:tada: I really like the general direction this is taking.

bigkevmcd:
@chanwit I'm going to close this just now, feel free to reopen if you want to pick it back up

PR #2960, opened 2022-11-02
Add error logging when validating the userinfo.

bigkevmcd:
This adds some additional logging to the server logs when we fail to parse or query the userinfo endpoint.

It also simplifies the calls to the UserInfo endpoint.

**Why was this change made?**
Currently, we only log the errors to the user's browser, which isn't particularly easy to get to for admins.

The UserInfo function was literally just function returning a function which wasn't necessary.

It's a logging change, tests are passing.

**Release notes**
Additional logging for authentication failures.


foot (approved):
👌

PR #2962, opened 2022-11-02
Update filterByTypeCallback

### LINKED ISSUE #2959: Unable to filter on Type in Applications details page
**Describe the bug**
We appear to have lost the filter on type capability as shown below, looks like it was introduced in `0.10.0` as a test on `0.9.6` shows it working properly (final screenshot)

**Error present:**
Screenshot 2022-11-02 at 08 28 02

It _may_ just be on this page, as I can filter on type in the source details view - however, I guess that could also be linked to the fact that the latter will show only Flux objects.

**Sources working:**
Screenshot 2022-11-02 at 08 33 47

**Working normally - 0.9.6:**
Screenshot 2022-11-02 at 08 42 19

**Environment**

- Weave-Gitops Version 0.10.1
- Flux Version 0.34.0
- Kubernetes version

**To Reproduce**

**Expected behavior**

**Actual Behavior**

joshri:
Closes #2959 

We had a special method for filtering type in the ReconciledObjectsTable that isn't needed anymore - especially since it breaks things now. Removed. I removed the specific test for it as well (we're already testing filtering type as a string from key 'type'). Also removed from exports - I checked the enterprise repo and it's not being used there so we should be good. 

image

### DIFF
--- a/ui/components/DataTable.tsx
+++ b/ui/components/DataTable.tsx
@@ -103,7 +103,7 @@ export const filterByStatusCallback = (v) => {
   else return ReadyType.NotReady;
 };
 
-export const filterByTypeCallback = (v) => _.get(v, "groupVersionKind.kind");
+export const filterByTypeCallback = (v) => v.type;
 
 export function filterConfig(
   rows,
--- a/ui/components/DataTable.tsx
+++ b/ui/components/DataTable.tsx
@@ -103,8 +103,6 @@ export const filterByStatusCallback = (v) => {
   else return ReadyType.NotReady;
 };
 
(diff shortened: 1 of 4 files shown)


jpellizzari (approved):
Code LGTM. @joshri Why don't we need the method for reconciled objects?

ozamosi (approved):

joshri:
> Code LGTM. @joshri Why don't we need the method for reconciled objects?

The method was looking at a specific field - but all reconciled objects have been slotted into the FluxObject class, so we can just look at the type field and be good to go!

PR #2970, opened 2022-11-07
add yaml validation

### LINKED ISSUE #2887: Gitops run client-side yaml validation
Client-side YAML validation would be useful to stop gitops run trying to reconcile invalid yaml.

How would it be implemented - would it be annoying if it's constantly complaining? 

Workflow:

* changes detected
* run client-side validation with [1] (policies would also be used here)
  * passed -> upload to the bucket
  * failed -> say error messages
[1] - https://github.com/yannh/kubeconform

Pulled out of #2882

chanwit:
This PR adds yaml validation at an early stage before uploading YAML files to the bucket.
It validates yaml files under the `targetDir` only.

Fixes #2887 

### DIFF
--- a/cmd/gitops/beta/run/cmd.go
+++ b/cmd/gitops/beta/run/cmd.go
@@ -32,6 +32,7 @@ import (
 	"github.com/weaveworks/weave-gitops/pkg/run/bootstrap"
 	"github.com/weaveworks/weave-gitops/pkg/run/install"
 	"github.com/weaveworks/weave-gitops/pkg/run/watch"
+	"github.com/weaveworks/weave-gitops/pkg/validate"
 	"github.com/weaveworks/weave-gitops/pkg/version"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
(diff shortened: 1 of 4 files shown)


ozamosi (approved):
* Tested "standard podinfo stuff" (create namespace, create flux resources): works :heavy_check_mark: 
* Tested incorrect kustomization schema: rejects my change :heavy_check_mark: 
* Tested other CRD I'm not installing here: document is ignored :heavy_check_mark: 
  * This opened my eyes to why my idea "why don't we just ask the cluster for the CRD" is nonsense - I might be installing the CRD together with the resource :facepalm:  
* Tested a document that was just `this: "is not a resource"`: rejected :heavy_check_mark: 
*  kustomization.yaml includes are documented as being out of scope, not tested those.

So overall: :heavy_check_mark: 

:bulb: The output is e.g.
```
► Validating files under test/ ...
test/podinfo.yaml - Kustomization podinfo is invalid: For field spec: prune is required
✗ Error validating: validation failed
```
I had a second where I read the last line saying "validation failed" and assumed that was the start of the error output so I should be looking for the error after it - there's nothing after it, so I thought "well that's unhelpful". A possible enhancement would be that if the error is "validation failed" specifically, the error line could be changed to imply it's a summary - "Validation failed - review any errors and try again".

:question: This feature is not optional - is there any risk of false positive where we'd get in the user's way instead of helping them? For example, if we're using different schema versions from the cluster.

chanwit:
Next ticket: #2986

PR #3001, opened 2022-11-14
rename gitops run objects and add metadata

### LINKED ISSUE #2813: Metadata for bucket and kustomization
As a user when I launch gitops run and install the dashboard I have no idea what is going on.  At a minimum, it would be nice to have metadata on the two objects we create to explain their purpose to the user.

### Acceptance Criteria

When I run `gitops run` and `dev-bucket` and `dev-ks` are created I see the following metadata.  The names should be `run-dev-bucket` and `run-dev-ks` respectively.

**dev-bucket**
* Description: This is a temporary bucket created by gitops run.  This will be cleaned up when this instance of gitops run is ended.
* Run Id: The unique ID of this particular context of gitops run.
* Username: Your OS username.

**dev-ks**
* Description: This is a temporary kustomization created by gitops run.  This will be cleaned up when this instance of gitops run is ended.
* Run Id: The unique ID of this particular context of gitops run.
* Username: Your OS username.

chanwit:
Fixes #2813 

- rename objects to run-dev-bucket, and run-dev-ks
- add metadata to run-dev-bucket and run-dev-ks
- rename dev-bucket server namespace to gitops-run
- add a hidden flag --x-session-name to pass down the session name to the sub-process

### DIFF
--- a/cmd/gitops/beta/run/cmd.go
+++ b/cmd/gitops/beta/run/cmd.go
@@ -7,6 +7,7 @@ import (
 	"io/fs"
 	"os"
 	"os/signal"
+	"os/user"
 	"path/filepath"
 	"strconv"
 	"strings"
@@ -75,6 +76,9 @@ type RunCommandFlags struct {
 
 	// Flags, created by genericclioptions.
 	Context string
+
+	// Hidden session name for the sub-process
+	HiddenSessionName string
 }
 
 var flags RunCommandFlags
@@ -127,6 +131,9 @@ gitops beta run ./deploy/overlays/dev --timeout 3m --port-forward namespace=dev,
 	cmdFlags.BoolVar(&flags.NoSession, "no-session", false, "Disable session management. If not specified, the session will be enabled by default.")
 	cmdFlags.BoolVar(&flags.NoBootstrap, "no-bootstrap", false, "Disable bootstrapping at shutdown.")
 	cmdFlags.BoolVar(&flags.SkipResourceCleanup, "skip-resource-cleanup", false, "Skip resource cleanup. If not specified, the GitOps Run resources will be deleted by default.")
+
+	cmdFlags.StringVar(&flags.HiddenSessionName, "x-session-name", "", "The session name acknowledged by the sub-process. This is a hidden flag and should not be used.")
+	_ = cmdFlags.MarkHidden("x-session-name")
 
 	kubeConfigArgs = run.GetKubeConfigArgs()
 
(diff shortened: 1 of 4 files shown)


chanwit:
Selection_051

pkg/run/watch/install_bucket_source_and_ks.go
```diff
@@ -47,33 +56,43 @@ func SetupBucketSourceAndKS(ctx context.Context, log logger.Logger, kubeClient c
 	}
 	source := sourcev1.Bucket{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      devBucket,
-			Namespace: namespace,
+			Name:      RunDevBucketName,
+			Namespace: params.Namespace,
+			Annotations: map[string]string{
+				"metadata.weave.works/description": "This is a temporary Bucket created by GitOps Run. This will be cleaned up when this instance of GitOps Run is ended.",
```
opudrovs, line 62: 💡 A suggestion: maybe `this instance of GitOps Run is done.` or `this instance of GitOps Run is finished.` would work better (here and in the other place) instead of `this instance of GitOps Run is ended`? It looks like `process instance is done` or `finished` is used more often.

cmd/gitops/beta/run/cmd.go
```diff
@@ -128,6 +132,12 @@ gitops beta run ./deploy/overlays/dev --timeout 3m --port-forward namespace=dev,
 	cmdFlags.BoolVar(&flags.NoBootstrap, "no-bootstrap", false, "Disable bootstrapping at shutdown.")
 	cmdFlags.BoolVar(&flags.SkipResourceCleanup, "skip-resource-cleanup", false, "Skip resource cleanup. If not specified, the GitOps Run resources will be deleted by default.")
 
+	cmdFlags.StringVar(&flags.HiddenSessionName, "x-session-name", "", "The session name acknowledged by the sub-process. This is a hidden flag and should not be used.")
+
+	if err := cmdFlags.MarkHidden("x-session-name"); err != nil {
+		return nil
```
opudrovs, line 138: ❔ What is the reason for returning `nil` if marking this flag as hidden results in an error?

opudrovs (approved):
LGTM! 🎉 Tested that objects are renamed and the metadata is added.

cmd/gitops/beta/run/cmd.go
chanwit: good question actually.

what do you think would be used instead of returning nil here?

pkg/run/watch/install_bucket_source_and_ks.go
chanwit: They are just copied from @JamWils' texts here:

https://github.com/weaveworks/weave-gitops/issues/2813#issue-1390112050

opudrovs: Ah, OK, thanks for the info.

PR #3006, opened 2022-11-15
docs to install weave gitops enterprise  airgap / offline mode 

enekofb:
Result of https://github.com/weaveworks/weave-gitops-enterprise/issues/1809

**What changed?**
Added guide 

**Why was this change made?**
To guide wge admins while installing wge in a restricted netwrok

docs

locally running the doc

**Release notes**
Documentation added to guide platform admin while installing weave gitops enterprise in a restricted network

**Documentation Changes**
No

### DIFF
--- a/website/docs/enterprise/installation/airgap.mdx
+++ b/website/docs/enterprise/installation/airgap.mdx
@@ -1,0 +1,542 @@
+---
+title: Airgap environments
+sidebar_position: 1
+hide_title: true
+toc_max_heading_level: 4
+---
+
+# Installing Weave Gitops Enterprise in Airgap environments
+
+From [wikipedia](https://en.wikipedia.org/wiki/Air_gap_(networking))
+
+>An air gap, air wall, air gapping or disconnected network is a network security measure employed on one or more computers
+to ensure that a secure computer network is physically isolated from unsecured networks, such as the public Internet or an unsecured local area network...
+
+This document guides on how to install Weave Gitops Enterprise in a restricted environment.
+
+# Before Start
+
+There are multiple restrictions that could happen within an airgap environment. This guide assumes that you have egress network
+restrictions. In order to install Weave Gitops Enterprise (WGE), the required artifacts are required to be loaded
+from a private registry. This guide helps you with the task to identity the helm charts
+and container images required to install WGE and to load them into your private registry.
+
+It also assumes that you could prepare the installation from a proxy host. A proxy host is defined here
+as a computer that is able to access to both the public and private network. It could take different shapes,
+for example, it could be a bastion host, a corp laptop, etc.
+
+Access to both public and private network is required during the airgap installation but not simultaneously.
+It is expected to have an online stage to gather the artifacts first, and an offline stage later,
+to load the artifacts in the private network.
+
+Finally, we aim to provide an end to end example to use it as a guidance more than a recipe. Feel free to adapt the details
+that do not fit within your context.
+
+# Install
+
+There are different variations of the following stages and conditions. We consider that installing 
+Weave Gitops Enterprise in an airgap environment could follow the following stages.
+
+1. Setup a WGE install environment.
+2. Collect artifacts and publish to a private registry.
+3. Install Weave Gitops Enterprise in the airgap environment.
+
+## Setup a WGE install environment
+
+The main goal of this stage is to recreate a local Weave Gitops Enterprise within your context, to collect
+the container images and helm charts, that will be required in your private registry for the offline installation.
+
+A three-step setup is followed.
+
+1. Setup a proxy host
+2. Setup a private registry
+3. Install Weave Gitops Enterprise
+
+### Setup a proxy host
+
+There are many possible configurations for this host. This guide will assume that the host has installed the following:
+
+- [docker](https://www.docker.com/) as container runtime.
+- [kubectl and kind](https://kubernetes.io/docs/tasks/tools)
+- [helm](https://helm.sh/docs/intro/install/)
+- [skopeo](https://github.com/containers/skopeo) to manage container images
+- [flux](https://fluxcd.io/flux/cmd/) to boostrap flux in the environment.
+- [clusterctl](https://cluster-api.sigs.k8s.io/user/quick-start.html#install-clusterctl) to replicate the cluster management
+capabilities.
+
+#### Create Kind Cluster
+
+Create a kind cluster with registry following [this guide](https://kind.sigs.k8s.io/docs/user/local-registry/)
+
+#### Install flux
+
+You could just use `flux install` to install flux into your kind cluster
+
+#### Setup a helm repo
+
+We are going to install [ChartMuseum](https://chartmuseum.com/) via flux.
+
+Remember to also install helm plugin [cm-push](https://github.com/chartmuseum/helm-push)
+
+```yaml
+apiVersion: source.toolkit.fluxcd.io/v1beta2
+kind: HelmRepository
+metadata:
+  name: chartmuseum
+  namespace: flux-system
+spec:
+  interval: 10m
+  url: https://chartmuseum.github.io/charts
+---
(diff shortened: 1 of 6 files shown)


enekofb:
updated guide 

>What are your thoughts @darrylweaver l

website/docs/enterprise/installation/airgap.mdx
```diff
@@ -0,0 +1,560 @@
...
+3. Install Weave Gitops Enterprise in the airgap environment.
+
+## Setup a WGE install environment
+
+The main goal of this stage is to recreate a local Weave Gitops Enterprise within your context, to collect
+the container images and helm charts, that will be required in your private registry for the offline installation.
+
+A three-step setup is followed.
+
+1. Setup a proxy host
```
darrylweaver, line 43: This is a little redundant as it was stated already above the steps.

website/docs/enterprise/installation/airgap.mdx
```diff
@@ -0,0 +1,560 @@
...
+2. Setup a private registry
+3. Install Weave Gitops Enterprise
+
+### Setup a proxy host
```
darrylweaver, line 47: I think a proxy host assumption for some customers like banks will be fine, would be good to check with our most secure customers if they can also use a proxy host.

darrylweaver (commented):
Overall the method is good provided the proxy host assumption is correct.
I do not see how clusterctl is deployed in these instructions.
You initialise the cluster API provider for capmvm, but you do not seem to have a way to install that in the target environment.

website/docs/enterprise/installation/airgap.mdx
darrylweaver: I checked this and we would have to be compatible with a laptop to download the artefacts then disconnect from the internet and connect to the secure network and then be able to upload the artefacts to the private registry.
Therefore an intermediate step of writing out a tarball and then uploading in a separate step.

enekofb: thanks for checking, yes proxy host here should be understood as a host that have access to both networks, could be of multiple forms. adding that.

enekofb: updated

website/docs/enterprise/installation/airgap.mdx
```diff
@@ -0,0 +1,560 @@
...
+
+pull-charts:
+	curl -L https://github.com/weaveworks/weave-gitops-profile-examples/releases/download/$(CERT_MANAGER)/$(CERT_MANAGER_CHART) --output $(CERT_MANAGER_CHART)
+	curl -L https://s3.us-east-1.amazonaws.com/weaveworks-wkp/releases/charts-v3/$(WGE_CHART) --output  $(WGE_CHART)
+
+push-charts:
+	helm cm-push -f $(CERT_MANAGER_CHART) $(PRIVATE_HELM_REPO_NAME)
+	helm cm-push -f $(WGE_CHART) $(PRIVATE_HELM_REPO_NAME)
+```
+where `image-pull-push.sh` is similar to
```
bigkevmcd, line 425: Another option is [skopeo](https://www.redhat.com/en/blog/skopeo-copy-rescue) which doesn't require Docker to be running locally which might be an issue for enterprises who don't allow users to install things?

bigkevmcd: I'm not sure we can rely on folks being able to install Kind either :-)

enekofb: agree, the doc itself is more to provide guidance than a recipe 

i tried to capture [it here](https://github.com/weaveworks/weave-gitops/pull/3006/files#diff-c807d0873ce72f2d1cd51982bddcd77f162faed3a34f2ada80c03b8b1e9968dfR49) but let me expand on that

bigkevmcd: One reason not to use docker pull/push is that it loses the manifest.

Most of the tools that can copy images will retain the manifests, which means things like different architectures are retained, e.g. regctl https://github.com/regclient/regclient

To be clear, I have no particular attachment to skopeo (I use regctl personally) but it's at least supported by Red Hat :-)

enekofb: Expanded here https://github.com/weaveworks/weave-gitops/pull/3006/commits/fc2d3297f29d6d233f1c609a9a51edf258cabea3

darrylweaver:
What I think is missing the installation of clusterctl into the destination cluster, as I said.
You collect the artefacts, which would include some containers, it should save the capi containers too.
But then for the installation in the air-gapped cluster, you only say install the WGE helm chart with some images specified from the local container repo.
You do not say to run `clusterctl init` and the problem with running that would be it would connect to the repo for the provider to init the cluster.
So, once you have the containers in your local repo, how do you provide kubernetes manifests to install them for CAPI?

darrylweaver:
Another comment:
Why are we making the customer run the build of the artefact to cross over the airgap, when we could easily run all of that in our own CI and just publish the tarball they need to download?
We just need download location and pushing to the local repo rather than installing Kind, clusterctl, WGE etc and having to do that for each release version themselves.
If we just make it part of our CI we can automatically publish it whenever a new version is released.  They can then follow the install instructions from the tarball whenever they want to upgrade.

website/docs/enterprise/installation/airgap.mdx
```diff
@@ -147,45 +147,86 @@ At this stage you have already a private registry for container images and helm
 
 This step is to gather the artifacts and images in your local environment to push to the private registry.
 
-#### Cert Manager
+####  Cluster API
```
enekofb, line 150: @darrylweaver this added to address your comment on CAPI 

https://github.com/weaveworks/weave-gitops/pull/3006#issuecomment-1327415216

let me know whether that covers it

darrylweaver: Yes, that would cover it for now.

darrylweaver (approved):

website/docs/enterprise/installation/airgap.mdx
```diff
@@ -423,24 +415,27 @@ charts: pull-charts push-charts
...
 	helm cm-push -f $(WGE_CHART) $(PRIVATE_HELM_REPO_NAME)
 ```
-where `image-pull-push.sh` is similar to
+where `image-sync.sh` is similar to
 
 ```shell
-docker pull $1
-docker tag $1 $2/$1
-docker push $2/$1
+#--dest-tls-verify=false to work over http with the registry in this guide, if you have an https registry, you could remove this setting
```
enekofb, line 431: @bigkevmcd let me know whether makes more sense to you now

website/docs/enterprise/installation/airgap.mdx
```diff
@@ -423,24 +415,27 @@ charts: pull-charts push-charts
...
+skopeo copy docker://$1 docker://$2/$1 --dest-tls-verify=false --preserve-digests --multi-arch=all
 ```
```
bigkevmcd, line 433: Let's remove the --dest-tls-verify=false, if they need to not verify, they can add that...but it should never be the default.

enekofb: done, yes it is a bit weird to have it

website/docs/enterprise/installation/airgap.mdx
```diff
@@ -423,24 +415,27 @@ charts: pull-charts push-charts
...
 
+>[Skopeo](https://github.com/containers/skopeo) allows you to configure a range a security features to meet your requirements.
+For example, configuring trust policies before pulling or singing containers before making them available in your private network.
```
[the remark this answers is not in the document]
enekofb: changed

website/docs/enterprise/installation/airgap.mdx
enekofb: updated

PR #3017, opened 2022-11-16
validate with k8s server and flux versions

chanwit:


cmd/gitops/beta/run/cmd.go
```diff
@@ -539,6 +540,19 @@ func runCommandWithoutSession(cmd *cobra.Command, args []string) error {
...
+		return err
+	}
+
+	serverVersion, err := discoveryClient.ServerVersion()
+	if err != nil {
+		return err
+	}
+
+	// We need the server version to pass to the validation package
+	kubernetesVersion := strings.TrimPrefix(strings.TrimRight(serverVersion.GitVersion, "+"), "v")
```
ozamosi, line 554: Is this supposed to turn e.g. "1.25.1+gke1234" into "1.25"? Because I don't think it does.

ozamosi (commented):
Still validates :+1: 

Checked that "most" k8s releases are supported (the docs are misleading).

Checked that if I have a non-supported k8s version then that just turns off validation.

Added a question where I think the code is incorrect.

cmd/gitops/beta/run/cmd.go
chanwit: Thanks for spotting this!

test cases added.

ozamosi (approved):

Issue #2898, opened 2022-10-17
GitOps run table view


josefaworks:
As a user with GitOps Run already running, I would like to see a list of clusters where it is running so that I can see more details and dive deeper if I want. 
In the list of clusters I would also like to see
- the name of the cluster
- how long it has been running (duration) (no filter)
- automation
- source
- CLI version
- port forward (no filter)
- commands used (no filter)

Note: this is enterprise only

Figma link: https://www.figma.com/file/IVHnM9iyeFWpd11evtY8ux/Weave-GitOps?node-id=11227%3A53921

![Screen Shot 2022-10-17 at 10.13.56 AM.png](https://images.zenhubusercontent.com/6138b1b3d7038f238a0ab7fb/fcddb173-a014-43ba-9bf1-9d7c78cca71a)
![Screen Shot 2022-10-17 at 10.13.50 AM.png](https://images.zenhubusercontent.com/6138b1b3d7038f238a0ab7fb/4ffdc73f-1db2-47b0-9753-acb50732a3d4)

ozamosi:
The implementation of sessioned and unsessioned mode would have to be separate. Let's start with sessioned mode (because otherwise you can't have more than one, and then it's not much of a list).

We should define a "session" as "HelmRelease with the labels `app.kubernetes.io/managed-by=GitopsRun` and `app.kubernetes.io/component=session`". These labels aren't currently set, we need to set them automatically.

We should store the other bits of information given to the CLI we want to display - cli version, port-forward, command, bootstrap, dashboard - as annotations. These annotations can be prefixed by `run.weave.works' so e.g. the annotation `run.weave.works/cli-version=0.9.2-rc.2` would print that CLI version field.

We can't link to automation or source in session mode right now, so skip that. This needs further work to connect to inside the vcluster.

ozamosi:
From #2907 session metadata, we get:
* the name of the cluster
* CLI version
* port forward (no filter)
* commands used (no filter)

To add:
* how long it has been running (duration) (no filter) - this is the age of the statefulset

Moved to weaveworks/weave-gitops-enterprise#1772, skip in this ticket:
* automation
* source

ozamosi:
We don't have anywhere to link the name to yet - don't link it.

ozamosi:
Add the ability to filter by label to listObjects, and make the call to the backend include the labels added in #2907

PR #3021, opened 2022-11-16
Add transformFunc argument to filterConfig function

joshri:
Part of #2898 

This PR allows you to supply your own custom filter-value-getter to `filterConfig` without ALSO hard-coding it into the `filterRows` function (you can see this on line 137 of DataTable in the changes).

Removing the hard-coded version from `filterRows` just means we have to pass it through everything else, hence the changes to the `FilterConfig` type.

Doing this lets me provide a custom function to filterConfig inside enterprise without hard-coding it inside core.

### DIFF
--- a/ui/components/DataTable.tsx
+++ b/ui/components/DataTable.tsx
@@ -141,6 +141,7 @@ export function filterRows<T>(rows: T[], filters: FilterConfig) {
 
       if (!_.includes(vals.options, value)) {
         ok = false;
+        return ok;
       }
     });
 
--- a/ui/components/DataTable.tsx
+++ b/ui/components/DataTable.tsx
@@ -121,7 +121,7 @@ export function filterConfig(
     []
   );
 
-  return { [key]: config };
+  return { [key]: { options: config, transformFunc: computeValue } };
 }
 
 export function filterRows<T>(rows: T[], filters: FilterConfig) {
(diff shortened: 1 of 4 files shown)


jpellizzari (approved):
:+1: . Just for my own info, the transform function allows you to change the filter label text?

ui/components/DataTable.tsx
```diff
@@ -134,14 +134,12 @@ export function filterRows<T>(rows: T[], filters: FilterConfig) {
...
-        value = filterByStatusCallback(row);
-      }
+
+      if (vals.transformFunc) value = vals.transformFunc(row);
       // strings
       else value = row[category];
 
-      if (!_.includes(vals, value)) {
+      if (!_.includes(vals.options, value)) {
         ok = false;
```
opudrovs, line 143: ❓ Do we have to go through remaining elements if this condition is true `!_.includes(vals.options, value)` and ok becomes false? Can we call `return false;` after `ok = false` to exit the Lodash `each` loop or if it does not work can we use a regular JS loop with break instead?

ui/components/DataTable.tsx
```diff
@@ -134,14 +134,12 @@ export function filterRows<T>(rows: T[], filters: FilterConfig) {
 
     _.each(filters, (vals, category) => {
       let value;
-      // status
-      if (category === "status") {
-        value = filterByStatusCallback(row);
-      }
+
+      if (vals.transformFunc) value = vals.transformFunc(row);
```
opudrovs, line 138: 💡 Adding curly braces to if and else might make the code more readable.

joshri:
@jpellizzari transform function supplies a display value for the filter dialog that will override obj[key].

opudrovs (approved):

ui/components/DataTable.tsx
joshri: that return makes sense - added it and nothing broke

PR #3026, opened 2022-11-17
Add passing a `Writer` to the CLI logger.

opudrovs:
Test branch.


chanwit:
@opudrovs I happened to fix this in #3027 
would you mind reviewing it instead?

We are fixing it with the same logic.

opudrovs:
@chanwit approved. Have not tested it yet but it looks like it should work. 🙂

opudrovs:
Closed as a test PR.

PR #3027, opened 2022-11-18
implement s3 log writer

### LINKED ISSUE #2848: GitOps Run create a CLI log forwarding mechanism
GitOps Run log receiver is an important server-side component to help store logs on the server side.

A GitOps Run instance would printouts logs to the os.Stdout, os.Stderr, while sending a copy of logs to the log receiver simultaneously.

The receiver receives the logs, and stores them in RAM.

Out of scope:
The GitOps Run UI retrieves logs stored by the log receiver, re-styling each line of the logs, then displaying them on the UI - in real time.

chanwit:
Fixes #2848 

### DIFF
--- a/cmd/gitops/beta/run/cmd.go
+++ b/cmd/gitops/beta/run/cmd.go
@@ -274,7 +274,8 @@ func fluxStep(log logger.Logger, kubeClient *kube.KubeHTTP) (fluxVersion string,
 			return "", false, err
 		}
 
-		flux.SetLogger(log.Logger)
+		// This means that Flux logs will be printed to the console, but not be sent to S3
+		flux.SetLogger(log.L())
 
 		var components []fluxexec.Component
 		for _, component := range flags.Components {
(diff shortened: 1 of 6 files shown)


chanwit:
Requested @opudrovs for writer and addsyncer.
@ozamosi for general s3 forwarder.

opudrovs (approved):

ozamosi:
So this is implemented as a wrapper around logr, instead of as a zap writer? I think we'll regret that - it means that every time we call L(), we won't get any logs to the server, right?

chanwit:
@ozamosi yes.
It works at the level of our interface.

If we chose to not send it to S3, we would be able to call .L() to bypass the outer layer.

ozamosi:
Yeah, but that means it doesn't actually forwards e.g. any flux logs? Which is bad?

chanwit:
It depends on the context.

For example, Flux installation logs happen before we start the S3 bucket. In any mean, we cannot grab them.

Also, Flux bootstrapping logs happen after we stop the S3 bucket. We don't have any chance to grab them either.

PR #3029, opened 2022-11-18
Fix status icon colors

ozamosi:
The colors were renamed in #3019, but the colors returned from this function wasn't updated.

### DIFF
--- a/ui/components/KubeStatusIndicator.tsx
+++ b/ui/components/KubeStatusIndicator.tsx
@@ -2,6 +2,7 @@ import _ from "lodash";
 import * as React from "react";
 import styled from "styled-components";
 import { Condition } from "../lib/api/core/types.pb";
+import { colors } from "../typedefs/styled.d";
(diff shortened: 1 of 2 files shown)


ui/components/KubeStatusIndicator.tsx
```diff
@@ -146,15 +146,15 @@ function KubeStatusIndicator({
     if (ready === ReadyType.Reconciling) {
       readyText = ReadyType.Reconciling;
       icon = IconType.ReconcileIcon;
-      iconColor = "primary";
+      iconColor = "primaryOriginal";
```
foot, line 149: This one should stay as `primary` I think.

We could type it as we do in https://github.com/weaveworks/weave-gitops/blob/de7ed81057ee4dc497038062c4aa186d3a53b2ab/ui/components/PageStatus.tsx#L21

foot (approved):
💯

PR #3039, opened 2022-11-23
Break up installation docs

dholbach:
@makkes talked me through the installation docs yesterday and we both thought that breaking them up into separate docs would be helpful in order to have a better flow and to make it clearer what's expected from the user.

The aim was to reduce duplication and offer a clearer flow.

1. installation page
2. `<choose your install method>`
3. get started with the dashboard

How does everyone else feel about this?

Signed-off-by: Daniel Holbach <daniel@weave.works>


dholbach:
→ https://staging.docs.gitops.weave.works/installation-straw-doc/docs/next/installation/

makkes:
I like this new structure a lot. It reduces size of the individual installation pages and makes them much simpler to digest.

If I'm not mistaken, you stripped the Helm chart installation method from the OSS page, was that on purpose? Do we not want to mention that going forward and rather focus on the CLI installation method?

dholbach:
> If I'm not mistaken, you stripped the Helm chart installation method from the OSS page, was that on purpose? Do we not want to mention that going forward and rather focus on the CLI installation method?

That's the part I was unsure about. It looked like there were two different ways to install Weave GitOps. I'm happy to go with whatever. Or offer two separate ways to do it..? :man_shrugging:

sympatheticmoose:
I agree this is better, however, the reason I've not done this yet is that the header is not versioned and needs to link to a given page as you can see in the recording below. 
https://user-images.githubusercontent.com/19860021/203531211-c81737bd-d0d1-40a1-b5b8-7cb207d63dd6.mov

This is part of the reason why I recommend we abandon the idea of versioning the docs as it would free us a lot more.

makkes:
> I agree this is better, however, the reason I've not done this yet is that the header is not versioned and needs to link to a given page as you can see in the recording below.

Did we try to make the header versioned the same way as the other pages?

sympatheticmoose:
@makkes I don't believe this is possible as its part of the theme configuration... happy to be wrong here

dholbach:
I tried using the redirect plugin, but that didn't work either. Here is another attempt using a local redirect file. LMK if that works for you.

It could indeed be a good idea to move to a `stable` vs `next` interface to avoid carrying too much weight of URL stability with you.

dholbach:
If anyone would like to take a crack at breaking out the CLI installations, sure that WFM. I just wasn't sure about it as it'd be a separate flow for WG vs WGE, right?

website/docs/installation/weave-gitops.mdx
```diff
@@ -0,0 +1,177 @@
...
+* 0.35
+* 0.36
+
+## Before you begin
+
+We will provide a complete walk-through of getting Flux installed and Weave GitOps configured. However, if you have:
+- an existing cluster bootstrapped Flux version >= 0.32.0 🎉
+- followed our [installation](./overview.mdx) doc to configure access to the Weave GitOps dashboard then install Weave GitOps 👏
+
+Then you can skip ahead to [Part 1](#part-1---weave-gitops-overview) 🏃
```
makkes, line 27:
```suggestion
Then you can skip ahead to [the Weave GitOps overview](../getting-started.mdx#part-1---weave-gitops-overview) 🏃
```

website/docs/installation/weave-gitops.mdx
```diff
@@ -0,0 +1,177 @@
...
+     --owner=$GITHUB_USER \
+     --repository=fleet-infra \
+     --branch=main \
+     --path=./clusters/my-cluster \
+     --personal
+   ```
+
+Full installation documentation including how to work with other Git providers is available [here](https://fluxcd.io/docs/installation/).
+
+The bootstrap command above does following:
```
makkes, line 78:
```suggestion
The bootstrap command above does the following:
```

website/docs/installation/weave-gitops.mdx
```diff
@@ -0,0 +1,177 @@
...
+
+- Creates a git repository fleet-infra on your GitHub account
+- Adds Flux component manifests to the repository
+- Deploys Flux Components to your Kubernetes Cluster
+- Configures Flux components to track the path /clusters/my-cluster/ in the repository
```
makkes, line 83:
```suggestion
- Create a git repository fleet-infra on your GitHub account
- Add Flux component manifests to the repository
- Deploy Flux Components to your Kubernetes Cluster
- Configure Flux components to track the path /clusters/my-cluster/ in the repository
```

dholbach:
I filed #3051 and #3052 as follow-up items.

website/docs/installation/weave-gitops.mdx
```diff
@@ -0,0 +1,177 @@
+---
+title: Weave GitOps
+sidebar_position: 2
+hide_title: true
+pagination_next: "getting-started"
+---
+
+## Installing Weave GitOps
+
+#### Install Flux
```
makkes, line 10: There is already a section with the exact same name under "Before you begin" so I think what we should do is move the content of this section to the other one.

website/docs/installation/weave-gitops.mdx
```diff
@@ -0,0 +1,177 @@
...
+
+Weave GitOps is an extension to Flux and therefore requires that Flux 0.32 or later has already been installed on your Kubernetes cluster. Full documentation is available at: [https://fluxcd.io/docs/installation/](https://fluxcd.io/docs/installation/).
```
makkes, line 12:
```suggestion
```
We have a section for installing Flux on this very page so telling people they need to have Flux already installed just to tell them how to install Flux right after is a little strange.

website/docs/installation/weave-gitops.mdx
```diff
@@ -0,0 +1,177 @@
...
+
+This version of Weave GitOps is tested against the following Flux releases:
+* 0.32
+* 0.33
+* 0.34
+* 0.35
+* 0.36
```
makkes, line 19: This should go in the "Install Flux" section below.

website/docs/installation/weave-gitops.mdx
```diff
@@ -0,0 +1,177 @@
...
+
+Then you can skip ahead to [Part 1](#part-1---weave-gitops-overview) 🏃
+but note ⚠️ you may need to alter commands where we are committing files to GitHub ⚠️.
+
+To follow along, you will need the following:
+- A Kubernetes cluster - such as [Kind](https://kind.sigs.k8s.io/docs/user/quick-start/).
+- A [GitHub](https://github.com) account and [personal access token with repo permissions](https://help.github.com/github/authenticating-to-github/creating-a-personal-access-token-for-the-command-line).
+- [kubectl](https://kubernetes.io/docs/tasks/tools/#kubectl).
+
+### Install Flux
```
makkes, line 35:
```suggestion
## Install Flux
```

makkes (commented):
Great job!

makkes (approved):
Very very great change!

scottrigby (commented):
Merged before I finished my review – which is great. This change reads SO much better IMO 👏 👏

PR #3043, opened 2022-11-23
add helm chart reloading capability

### LINKED ISSUE #2886: Minimal helm chart support in gitops run
When I run gitops run against a target that contains a file called `Chart.yaml`, gitops run should create a helmrelease instead of a kustomization.

chanwit:
Fixed #2886 

How to test this PR:
```
git clone https://github.com/stefanprodan/podinfo
cd podinfo
gitops beta run ./chart/podinfo \
  --timeout 3m \
  --port-forward namespace=flux-system,resource=svc/run-dev-helm-podinfo,port=9898:9898
```


ozamosi:
I tried running this against the weave-gitops chart, and I get a validation error with
```
charts/gitops-server/templates/admin-user-creds.yaml - failed validation: error unmarshalling resource: error converting YAML to JSON: yaml: did not find expected node content
```
It doesn't really seem to matter what I set in the values file, it still errors out with the same thing.

I think that might mean we need a different validation approach with helm charts? That should probably be its own ticket.

chanwit:
@ozamosi we skipped the validation entirely during the Helm mode.
I'll take a look at what hit you.

chanwit:
@ozamosi you should see something like this:
image

with the Helm label conflict messages as the dashboard was already install inside the cluster by GitOps Run.

image

Maybe you ran an older version.

chanwit:
Found another bug:
I should fix it for this PR.
```
✗ Error requesting reconciliation: <nil>
```

ozamosi:
Refetched, recompiled, restarted - works now!

PR #3047, opened 2022-11-23
Allow dynamic claims

### LINKED ISSUE #2993: Support for custom groups claim
**Problem**
Using OIDC with group impersonate is currently not working when custom group claims are used. 
Only the  default groups claim is hardcoded right now and thus providers like Auth0 are not working. 

**Solution**
Make the claim name which is used to obtain groups configurable

bigkevmcd:
Closes #2993 

**What changed?**
This adds support for configuring the claims that we use to extract the Principal ID and Groups.

**Why was this change made?**
To allow users to customise their authorisation config.

There are two new configuration elements, the username claim, and the groups claim, these default to "email" and "groups" respectively which is what they were before.

When parsing the token claims, instead of a hard-coded struct, this uses a map and then looks for the configured keys and parses them appropriately.

The naming of the options is the same as the Kubernetes kube-apiserver version, this has also been tested by one 

Additionally, you can configure the custom claims via the same secret as used for the other OIDC configuration by adding two additional keys:

```yaml
apiVersion: v1
kind: Secret
metadata:
  name: oidc-auth
  namespace: default
type: Opaque
stringData:
  clientID: <insert your OIDC client ID here>
  clientSecret: <insert your OIDC client secret here>
  issuerURL: https://example.com/issue
  redirectURL: https://server.example.com/callback/oauth
  claimUsername: sub
  claimGroups: test_groups
```

Logged in via an Okta instance with custom claims for the groups.

**Release notes**
Support for custom claims in the OIDC configuration

**Documentation Changes**
We need to document the new command-line options _and_ the option for providing them in the OIDC Secret.


pkg/server/auth/jwt.go
```diff
@@ -139,6 +119,7 @@ func (pg *JWTAdminCookiePrincipalGetter) Principal(r *http.Request) (*UserPrinci
 		return nil, nil
 	}
 
+	// TODO: Fix This
 	return parseJWTAdminToken(pg.verifier, cookie.Value)
```
foot, line 123: So this one seems to get in the way. If you don't specify

```yaml
additionalArgs:
  - --auth-methods=oidc
```

In the values then we always end up here, even when logging in w/ OIDC, and so we end up w/ the `aud` claim. Which is a bit funny.

foot: Sorry my bad `sub`, not `aud`.

foot: This was `devMode` short-circuiting things

foot (approved):
Tested with 

```bash
  --namespace default \
  --from-literal=issuerURL=https://REDACTED/oauth2/default \
  --from-literal=clientID=REDACTED \
  --from-literal=clientSecret=REDACTED\
  --from-literal=redirectURL=http://localhost:9001/oauth2/callback \
  --from-literal=claimUsername=name \
  --from-literal=claimGroups=groupz
```

and creating / removing rolebindings for user and then group independently..

and it worked great after disabling tilt+devmode.

The values for `name` and `groupz` were not returned in userinfo but thats it.

Great stuff!

eloo:
I have also tested it at least for the groupsClaim and it seems to work as expected :) 
Thanks

foot:
Works great! Now get the correct userClaim value in the UI too ✨

chanwit (approved):
I'm worrying a bit about some possibilities of `s.OIDCConfig` being a zero struct.
But in general it looks good to me.

Issue #2985, opened 2022-11-09
docs/intro: Show video of Weave GitOps in action


dholbach:
It'd be nice to offer a video or two of Weave GitOps in action on the landing page. As many users prefer to see things in action before they dive into the docs, this could be helpful.

There's probably more, but I already found these:

- https://www.youtube.com/watch?v=5b91fhbLcE0
- https://www.youtube.com/watch?v=nke29laH8Xk

Or re-use the one form the Weaveworks Product page:

```html
<video
   src="https://assets.contentstack.io/v3/assets/blt300387d93dabf50e/blt351ecb1bd5aba155/629a9719df98c90faab8980f/Core_Redux_-_Sonja_Cut.mp4"
   controls="" class="core-video"
   poster="https://images.contentstack.io/v3/assets/blt300387d93dabf50e/blt79ece502817b2187/6191e8f59dccbf6fa4178c8a/Screen_Shot_2021-11-14_at_8.57.55_PM.png?format=webp&amp;width=600">
</video>
```

PR #3048, opened 2022-11-23
show video on get-started page

dholbach:
Fixes: #2985

Signed-off-by: Daniel Holbach <daniel@weave.works>


dholbach:
This uses the video from the `weave.works` website. It'd be good if that video was available on a more permanent place than current contentstack.

Maybe it'd be worth adding e.g.

-    https://www.youtube.com/watch?v=5b91fhbLcE0
-    https://www.youtube.com/watch?v=nke29laH8Xk

to a dedicated resources page?

dholbach:
→ https://staging.docs.gitops.weave.works/add-wg-video/docs/next/intro/

scottrigby:
This looks like a good placement for the video. And it's a GOOD video (Steve W's voice is so easy to listen to!)
Only concern for this embedded video is we should add a thumbnail. Right now before playing it's not easy to see it's there at all

dholbach:
> This looks like a good placement for the video. And it's a GOOD video (Steve W's voice is so easy to listen to!) Only concern for this embedded video is we should add a thumbnail. Right now before playing it's not easy to see it's there at all

I added a preview image.

PR #3058, opened 2022-11-24
docs: remove duplicate "installation" entry from left nav

makkes:
As they were two "installation" pages in the final site, the left navigation had two entries for it. The new structure in this change fixes this by adding an index page to the "installation" folder that is rendered for the '/installation' URL path.


yiannistri (approved):
Tested updated links and verified that the incorrect section link has been removed 👍

PR #3070, opened 2022-11-28
dashboard: Use fallback namespace if no part-of=flux label

### LINKED ISSUE #2382: Flux Runtime view is empty when Flux is installed on OpenShift via OperatorHub
If flux wasn't installed with the flux cli, we usually end up missing labels which breaks the flux runtime view.

If we can't find a namespace with the correct labels when we're looking for the flux namespace, we should fall back to looking for `flux-system`. 

**Original content**

WORKAROUND: `kubectl label ns flux-system app.kubernetes.io/part-of='flux'`

As a user of Weave GitOps I should be able to install Flux via my preferred method and benefit from the observability capabilities in Weave GitOps.

When running OpenShift, many users will install software via the Operator Catalog/Operator Hub. 

After installing Flux via the community operator  https://operatorhub.io/operator/flux and installing Weave GitOps as a HelmRepository and Helm Release, the Flux Runtime view is empty:
Screenshot 2022-06-30 at 13 37 42

checking the network tab shows: `{"deployments":[],"errors":[{"clusterName":"Default","namespace":"","message":"could not find flux namespace in cluster"}]}`

Other views appear to work ok and show all resources:
Screenshot 2022-06-30 at 13 37 51

**Environment**

- Weave-Gitops Version 0.9.0
- Flux Version 0.30.2
- Kubernetes version (CRC version: 2.5.1+5b02b826 / OpenShift version: 4.10.18 / K8s - v1.23.5+3afdacb)

**To Reproduce**

1. Go to https://console.redhat.com/openshift/create/local
2. Download CRC
3. Run `crc setup`, `crc config set memory 16384`, `crc start`, `crc console` 
4. login to openshift console using kubeadmin (password provided after `crc start`
5. Install the Flux Operator
6. Create a HelmRepository and HelmRelease to install Weave GitOps
7. Expose the ww-gitops service as a route (can do via `oc expose service ww-gitops-weave-gitops -n flux-system`)
8. Login to dashboard
9. Navigate to Flux Runtime view
10. See error
-->

**Expected behaviour**

See the Flux controllers

**Actual Behavior**

Blank page

ozamosi:
If flux was installed through mechanisms other than the flux CLI (operatorhub, aws marketplace, microsoft.flux, etc), then the namespace usually doesn't have the part-of=flux label.

As a workaround, this still prefers the part-of label if it's set, but if we can't find one then we fall back to the namespace "flux-system". I hope this will be harmless - it's only used when the "proper" way has already failed, so I expect it'll either "fix" the feature, or do nothing.

However, the default in those third party installers might be to install flux somewhere else - e.g. operatorhub defaults to "operators". For that case I made a random undocumented environment variable that a user could set so that they get a different default than flux-system. I don't expect end users to do anything with that, but it could be used to ship this software together with flux in one of these stores - so e.g. if we throw in weave-gitops in the operatorhub flux installation, then we'd set `WEAVE_GITOPS_FALLBACK_NAMESPACE` in that distribution to make it match where flux goes.

This fixes #2382.

### DIFF
--- a/core/server/fluxruntime.go
+++ b/core/server/fluxruntime.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"os"
 	"sync"
 
 	"github.com/hashicorp/go-multierror"
@@ -39,7 +40,17 @@ var (
 	ErrFluxNamespaceNotFound = errors.New("could not find flux namespace in cluster")
 	// ErrListingDeployments no deployments found
 	ErrListingDeployments = errors.New("could not list deployments in namespace")
+
+	DefaultFluxNamespace = lookupEnv("WEAVE_GITOPS_FALLBACK_NAMESPACE", "flux-system")
 )
+
+func lookupEnv(envVar string, fallback string) string {
+	if val, ok := os.LookupEnv(envVar); ok {
+		return val
+	}
+
+	return fallback
+}
 
 func (cs *coreServer) ListFluxRuntimeObjects(ctx context.Context, msg *pb.ListFluxRuntimeObjectsRequest) (*pb.ListFluxRuntimeObjectsResponse, error) {
 	respErrors := []*pb.ListError{}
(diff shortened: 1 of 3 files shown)


core/server/fluxruntime.go
```diff
@@ -39,8 +40,18 @@ var (
 	ErrFluxNamespaceNotFound = errors.New("could not find flux namespace in cluster")
 	// ErrListingDeployments no deployments found
 	ErrListingDeployments = errors.New("could not list deployments in namespace")
+
+	DefaultFluxNamespace = getDefaultString("WEAVE_GITOPS_FALLBACK_NAMESPACE", "flux-system")
 )
 
+func getDefaultString(envVar string, fallback string) string {
```
makkes, line 47:
```suggestion
func lookupEnv(envVar string, fallback string) string {
```
This makes code using this function a little more human-parseable. Alternatively can call it `getEnvOrDefault` or similar.

core/server/fluxruntime.go
```diff
@@ -171,13 +182,19 @@ func (cs *coreServer) ListFluxCrds(ctx context.Context, msg *pb.ListFluxCrdsRequ
 }
 
 func filterFluxNamespace(nss []v1.Namespace) *v1.Namespace {
-	for _, ns := range nss {
+	var fluxSystem *v1.Namespace = nil
```
makkes, line 185:
```suggestion
	var fluxSystem *v1.Namespace
```
nil is the default value, anyway.

makkes (commented):
Thanks, this should help a lot with 3rd-party integrators.

makkes (approved):

PR #3075, opened 2022-11-28
Allow the wego-admin user read access to all objects

### LINKED ISSUE #2966: Ensure default cluster user role includes permissions for all core kubernetes resources
**Problem**

Users sometimes hit issues where the Weave GitOps UI is not showing all reconciled objects for a given kustomization or helmrelease due to missing permissions, such as #2897.

**Solution**

Updated: 
We should wildcard "get", "list", "watch" to all resource, adding "patch" to the flux resources.

This change should occur for both OSS and Enterprise admin users.

jpellizzari:
Closes #2966 

After this change, the default service account for the WeGO UI will have read access to all `apiGroups` and all `resources`. This allows objects that we have not enumerated to show up in the list of reconciled objects. Note the `ValidatingWebhookConfiguration` and `Task` objects (`Task` is a Tekton object, which WeGO UI doesn't "understand"):

Screenshot from 2022-11-28 12-09-39

### DIFF
--- a/charts/gitops-server/templates/admin-user-roles.yaml
+++ b/charts/gitops-server/templates/admin-user-roles.yaml
@@ -43,24 +43,35 @@ rules:
   - apiGroups: ["*"]
     resources: ["*"]
     verbs: [ "get", "list", "watch" ]
+
   - apiGroups: ["source.toolkit.fluxcd.io"]
     resources: [ "buckets", "helmcharts", "gitrepositories", "helmrepositories", "ocirepositories" ]
     verbs: [ "get", "list", "watch", "patch" ]
+
   - apiGroups: ["kustomize.toolkit.fluxcd.io"]
     resources: [ "kustomizations" ]
     verbs: [ "get", "list", "watch", "patch" ]
+
   - apiGroups: ["helm.toolkit.fluxcd.io"]
     resources: [ "helmreleases" ]
     verbs: [ "get", "list", "watch", "patch" ]
+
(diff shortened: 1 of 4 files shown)


charts/gitops-server/templates/admin-user-roles.yaml
```diff
@@ -36,9 +18,6 @@ rules:
   - apiGroups: ["helm.toolkit.fluxcd.io"]
     resources: [ "helmreleases" ]
     verbs: [ "get", "list", "watch", "patch" ]
-  - apiGroups: [ "notification.toolkit.fluxcd.io" ]
```
sympatheticmoose, line 39: Not blocking - can't validate what the page offers since its broken currently, but given you can suspend and resume alerts/providers it may make sense to have patch on these resources. We can always add back later.

sympatheticmoose: Nit: the notification resources are listed separately below in the ClusterRole, should likely be consistent

sympatheticmoose (approved):

charts/gitops-server/templates/admin-user-roles.yaml
jpellizzari: Fixed

sympatheticmoose: @jpellizzari notification resources missing patch in the ClusterRole

ozamosi:
Please also update the documentation.

ozamosi:
I think that's the main one :+1:

PR #3076, opened 2022-11-28
Support multiple instances of Flux in Flux Runtime view

### LINKED ISSUE #2910: Support multiple instances of Flux in Flux Runtime view
**Short description**
As a platform operator I want to see all Flux controllers on my cluster, when running multiple instances of Flux for multi-tenancy resource isolation, so that I can monitor their health and check their versions.

**Acceptance criteria**
- [ ] Flux Runtime Controller view shows controllers from multiple namespaces with the `part-of: flux` label
- [ ] New namespace column is added to the Controller tab

Reference doc: https://fluxcd.io/flux/security/best-practices/#resource-isolation

To set up a sample environment:
- run `flux install --watch-all-namespaces=false`
- then run `flux install -n <different-namespace> --watch-all-namespaces=false`

TheGostKasper:
Closes #2910 

**What changed?**
Dynamically support multiple instance of Flux in the UI to cover both OSS and WGE as in 
- Show *Title* of single FLux version if there's only one Flux instance is used 
- Display *Table* of Flux versions if multiple versions applied 
- Namespace column added to controller table

*Test scenario* :
- multiple clusters => will display a table 
- single cluster & singe flux version => will display a single line above tables
- single cluster & multiple flux versions (if happens) => will display a table

### DIFF
--- a/website/docusaurus.config.js
+++ b/website/docusaurus.config.js
@@ -43,7 +43,7 @@ module.exports = {
       items: [
         {
           type: "doc",
-          docId: "installation",
+          docId: "installation/index",
           position: "left",
           label: "Installation",
         },
--- a/website/versioned_docs/version-0.11.0/cluster-management/getting-started.mdx
+++ b/website/versioned_docs/version-0.11.0/cluster-management/getting-started.mdx
@@ -12,7 +12,7 @@ import BrowserOnly from "@docusaurus/BrowserOnly";
 
 ## Creating your first CAPD Cluster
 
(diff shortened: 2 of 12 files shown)


joshri (approved):
LGTM! I think the `supportMultipleFlux` in the `Props` type on line 28 can be removed, since it's no longer used

TheGostKasper:
@joshri i'll include the removal of `supportMultipleFlux` from `prop` in https://github.com/weaveworks/weave-gitops/pull/3080

PR #3080, opened 2022-11-29
[fix] Chip filters are lost on refresh

### LINKED ISSUE #3074: Chip filters are lost on refresh
**Describe the bug**

Chips for filters on tables are not visible if user is following an exact URL to a filtered list, or the user refreshes the page. This can wrongly give the impression that they are viewing an unfiltered list.

**Environment**

- Weave-Gitops Version 0.11.0
- Flux Version 0.36.0
- Kubernetes version

**To Reproduce**

1. go to applications page `/applications`
2. apply a filter, i.e. to only show the flux-system namespace `/applications?filters=namespace%3A%20flux-system_`
3. Here the UI will show a chip that a filter is applied
Screenshot 2022-11-28 at 17 58 43
4. Refresh the page / or copy paste the full URL into a different browser - the chip has disappeared, but the filter applies

To remove the filter, you have to first apply the filter again, then remove it

**Expected behavior**

UI should help the user understand that a filter is applied, chip should be visible.

**Actual Behavior**

Chip is invisible

TheGostKasper:
Closes #3074 

**What changed?**
```
FilterConfig = {
    [key: string]: {
        options: string[];
        transformFunc?: (option: any) => string;
    };
}
```
filterConfig now has an `options` instead of setting the value directly to the `key` and that was the issue to not display chips and selected filters.

changing the `initialFormState` to bind to the new field solves it

### DIFF
--- a/ui/components/DataTable.tsx
+++ b/ui/components/DataTable.tsx
@@ -186,7 +186,7 @@ function filterText(
   });
 }
 
-export function initialFormState(cfg: FilterConfig, initialSelections?) { 
+export function initialFormState(cfg: FilterConfig, initialSelections?) {
   if (!initialSelections) {
     return {};
   }
--- a/ui/components/DataTable.tsx
+++ b/ui/components/DataTable.tsx
@@ -186,14 +186,14 @@ function filterText(
   });
 }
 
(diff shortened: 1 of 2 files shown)


joshri (approved):
Thank you!! Good fix

PR #3092, opened 2022-11-30
Pendo Agent for Enterprise

opudrovs:
Part of https://github.com/weaveworks/weave-gitops-enterprise/issues/1941

- Added the `Pendo` component to exports.

- Added the `defaultTelemetryFlag` prop to the `Pendo` component.

- Added the `js-sha3` package to the dependencies.

- Turned telemetry off in Tilt by default.

- Reworked the logic for detecting if Pendo should be initialized (in order to initialize it only after the feature flags are available in enterprise).

- Moved providing the default value for feature flags from `CoreClientContext` to `useFeatureFlags`.

- Removed unneeded default values and optional chaining for data returned by `useFeatureFlags` (it looks like they were not needed even before the latest change).

- Moved initializing telemetry to the gitops server cmd.

- Moved the check if should initialize telemetry to outside the `InitTelemetry` function.

### DIFF
--- a/cmd/gitops-server/cmd/cmd.go
+++ b/cmd/gitops-server/cmd/cmd.go
@@ -36,6 +36,7 @@ import (
 	"github.com/weaveworks/weave-gitops/pkg/server"
 	"github.com/weaveworks/weave-gitops/pkg/server/auth"
 	"github.com/weaveworks/weave-gitops/pkg/server/middleware"
+	"github.com/weaveworks/weave-gitops/pkg/telemetry"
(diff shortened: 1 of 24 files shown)


opudrovs:
@ozamosi this PR is ready for review. Please see the notes, update:

- Moved initializing telemetry to the gitops server cmd.

- Moved the check if should initialize telemetry to outside the `InitTelemetry` function.

I tested it with enterprise, will update my enterprise PR with the latest main after the current PR is merged.

PR #3098, opened 2022-12-01
Add HTTPS support to bucket server

makkes:
The bucket server used by `gitops beta run` now serves over HTTP as well as HTTPS. HTTP is still necessary as Flux's Bucket resource doesn't have a field for providing a custom CA certificate.

This is a backwards-incompatible change as the ports the server is listening on have to provided through flags. Also, providing TLS cert and key is mandatory.


chanwit:
@makkes `make lint` would help you get rid of those errors.

pkg/http/server_test.go
```diff
@@ -0,0 +1,108 @@
+package http_test
```
chanwit, line 1: I think we don't need a separate package for test cases, right?

cmd/gitops-bucket-server/main.go
```diff
@@ -2,60 +2,59 @@ package main
...
+		certFile, keyFile   string
+	)
 
-	// create a listener with the desired port.
-	listener, err := net.Listen("tcp", ":"+port)
-	if err != nil {
-		log.Fatal(err)
-	}
+	flag.IntVar(&httpPort, "http-port", 9000, "TCP port to listen on for HTTP connections")
+	flag.IntVar(&httpsPort, "https-port", 9001, "TCP port to listen on for HTTPS connections")
```
chanwit, line 36: The default HTTPS port would be another number like 9443, as we are using `9001` for our default dashboard port.

chanwit (changes requested):

pkg/http/server_test.go
makkes: I usually do that to make sure to only test the public surface of a package.

chanwit (approved):

PR #3102, opened 2022-12-02
implement auth middleware for s3

chanwit:

### DIFF
--- a/cmd/gitops-bucket-server/main.go
+++ b/cmd/gitops-bucket-server/main.go
@@ -11,21 +11,49 @@ import (
 	"github.com/johannesboyne/gofakes3"
 	"github.com/johannesboyne/gofakes3/backend/s3mem"
 	"github.com/weaveworks/weave-gitops/pkg/http"
+	"github.com/weaveworks/weave-gitops/pkg/s3"
 )
 
 func main() {
+	logger := log.New(os.Stdout, "", 0)
+
+	awsAccessKeyID := os.Getenv("AWS_ACCESS_KEY_ID")
+	if awsAccessKeyID == "" {
+		minioRootUser := os.Getenv("MINIO_ROOT_USER")
+		if minioRootUser == "" {
+			logger.Fatal("AWS_ACCESS_KEY_ID or MINIO_ROOT_USER must be set")
+			return
+		}
+
+		awsAccessKeyID = minioRootUser
+	}
+
+	awsSecretAccessKey := os.Getenv("AWS_SECRET_ACCESS_KEY")
+	if awsSecretAccessKey == "" {
+		minioRootPassword := os.Getenv("MINIO_ROOT_PASSWORD")
+		if minioRootPassword == "" {
+			logger.Fatal("AWS_SECRET_ACCESS_KEY or MINIO_ROOT_PASSWORD must be set")
+			return
+		}
+
+		awsSecretAccessKey = minioRootPassword
+	}
+
 	ctx, cancel := signal.NotifyContext(
 		context.Background(),
 		syscall.SIGINT,
 		syscall.SIGTERM)
 	defer cancel()
 
-	logger := log.New(os.Stdout, "", 0)
-	backend := s3mem.New()
-	s3 := gofakes3.New(backend,
+	s3Server := gofakes3.New(s3mem.New(),
 		gofakes3.WithAutoBucket(true),
-		gofakes3.WithLogger(gofakes3.StdLog(logger, gofakes3.LogErr, gofakes3.LogWarn, gofakes3.LogInfo)))
-	s3Server := s3.Server()
+		gofakes3.WithLogger(
+			gofakes3.StdLog(
+				logger,
+				gofakes3.LogErr,
+				gofakes3.LogWarn,
+				gofakes3.LogInfo,
+			))).Server()
 
 	var (
 		httpPort, httpsPort int
@@ -54,7 +82,7 @@ func main() {
 		Logger:    logger,
 	}
 
-	if err := srv.Start(ctx, s3Server); err != nil {
+	if err := srv.Start(ctx, s3.AuthMiddleware(awsAccessKeyID, awsSecretAccessKey, s3Server)); err != nil {
 		logger.Fatalf("server exited unexpectedly: %s", err)
 	}
 }
--- a/pkg/s3/auth_middleware.go
+++ b/pkg/s3/auth_middleware.go
@@ -1,0 +1,301 @@
+package s3
+
+import (
+	"bytes"
+	"crypto/hmac"
+	"crypto/sha256"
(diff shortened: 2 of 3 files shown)


pkg/http/auth_middleware.go
```diff
@@ -0,0 +1,352 @@
...
+
+			buf.WriteByte('\n')
+		}
+	}
+
+	return buf.String()
+}
+
+func encodePath(pathName string) string {
+	var reservedObjectNames = regexp.MustCompile("^[a-zA-Z0-9-_.~/]+$")
```
pjbgf, line 189: Could be initialised at `init()` or as a global `var`, as it only needs to happen once.

cmd/gitops-bucket-server/main.go
```diff
@@ -14,13 +14,24 @@ import (
 )
 
 func main() {
+	logger := log.New(os.Stdout, "", 0)
+
+	if os.Getenv("AWS_ACCESS_KEY_ID") == "" {
```
makkes, line 19: I wonder if we should make these command-line flags just as the other parameters. This would bring all parameters into alignment.

pkg/http/auth_middleware.go
```diff
@@ -0,0 +1,352 @@
...
+	return credential{
+		AccessKeyID: parts[0],
+		Time:        parts[1],
+		Location:    parts[2],
+		Service:     parts[3],
+		Request:     parts[4],
+	}, nil
+}
+
+func authMiddleware(handler http.Handler) http.Handler {
```
makkes, line 51: How about this instead so the passing of the AWS credentials isn't hidden in the func implementation but rather is part of its signature:
```suggestion
func authMiddleware(awsKeyID, awsSecretKey string, handler http.Handler) http.Handler {
```

cmd/gitops-bucket-server/main.go
chanwit: I would rather want to bind them via envs with Secret. It would be hard to hide those values if they were on the command line. wdyt?

makkes: Yes, agreed.

pkg/http/auth_middleware.go
```diff
@@ -0,0 +1,339 @@
...
+}
+
+func getCanonicalHeaders(req http.Request, signedHeaders string) string {
+	headers := []string{}
+	vals := make(map[string][]string)
+
+	for _, k := range strings.Split(signedHeaders, ";") {
+		vv := req.Header.Values(k)
+		headers = append(headers, strings.ToLower(k))
+		vals[strings.ToLower(k)] = vv
```
pjbgf, line 141: Avoid double op:
```suggestion
		lk := strings.ToLower(k)
		headers = append(headers, lk)
		vals[lk] = vv
```

pkg/http/auth_middleware.go
```diff
@@ -0,0 +1,339 @@
...
+			utf8.EncodeRune(u, s)
+
+			for _, r := range u {
+				encodedPathname.WriteString("%" + strings.ToUpper(hex.EncodeToString([]byte{r})))
+			}
+		}
+	}
+
+	return encodedPathname.String()
+}
```
pjbgf, line 211: Can we use [url.PathEscape](https://github.com/golang/go/blob/master/src/net/url/url.go#L281) instead?

pjbgf:
After the review I noticed that upstream exposes some of this functionality:
https://github.com/aws/aws-sdk-go-v2/blob/main/aws/signer/v4/v4.go

We may benefit from using their implementation for the signing and verification part. They also have a middleware implementation, but I am not sure how useful that would be for us.

pkg/http/auth_middleware.go
chanwit: I'm not so sure about this as it is taken from what the MinIO client used for encoding the path. I used the same code for the compatibility.

chanwit:
> After the review I noticed that upstream exposes some of this functionality: https://github.com/aws/aws-sdk-go-v2/blob/main/aws/signer/v4/v4.go
> 
> We may benefit from using their implementation for the signing and verification part. They also have a middleware implementation, but I am not sure how useful that would be for us.

Here's some background.
The codes I'm using was taken from the v4 signing of the MinIO lib (Apache license, also used by Flux).
Instead of signing, I reverse its process to implement the verification method.

pkg/http/auth_middleware.go
pjbgf: As this is [publicly exposed upstream](https://github.com/minio/minio-go/blob/master/pkg/s3utils/utils.go#L327), I think we could just add it as a dependency. The upstream implementation provides context, and given the larger use base would help with long term maintainability/improvements.

pjbgf:
> Here's some background.
The codes I'm using was taken from the v4 signing of the MinIO lib (Apache license, also used by Flux).
Instead of signing, I reverse its process to implement the verification method.

@chanwit thank you for the context. Unfortunately neither MiniIO nor AWS seems to expose the entire logic, so it does make sense the approach.

pkg/http/auth_middleware.go
chanwit: ha. I should have done that.
v7 of the cli had been imported as a dep already.

thank you for spotting that.

pjbgf (approved):
LGTM

thanks @chanwit! :bow:

pkg/http/auth_middleware.go
```diff
@@ -0,0 +1,301 @@
+package http
```
makkes, line 1: Since this is very much S3-specific code, a better package name would actually be `s3` IMHO.

makkes (commented):
Tested it locally with:
```
AWS_ACCESS_KEY_ID=foo AWS_SECRET_ACCESS_KEY=some_secret_key ./bin/gitops-bucket-server --cert-file ~/dev/local-ca/ca.pem --key-file ~/dev/local-ca/ca.key
```
and the minio CLI:
```
$ mc alias set test https://localhost:9443 foo some_secret_key --insecure
Added `test` successfully.
$ $ /tmp/mc ls test --insecure
[2022-12-05 10:48:01 CET]     0B probe-bucket-sign-gza44hmacqgz/
[2022-12-05 10:48:29 CET]     0B minio/
[2022-12-05 10:55:29 CET]     0B probe-bucket-sign-wt0spypubjre/
```
Works fine, good job!

Just a remark on picking a different package name for the middleware code.

pkg/http/auth_middleware.go
chanwit: ack

pkg/s3/server.go
```diff
@@ -1,4 +1,4 @@
-package http
+package s3
```
makkes, line 1: This needs to stay in `http` as it is a generic server.

makkes (approved):

PR #3106, opened 2022-12-02
use HTTPS in `gitops run` S3 bucket server

makkes:
When installing the dev bucket server on the cluster, it is now configured with a self-signed certificate generated on demand. The certificate is valid for 3 days which should be enough for a typical `gitops run` use case.

Instead of using HTTP in the forwarding from the host machine into the bucket server pod, the CLI now uses HTTPS so that traffic can't be intercepted on its way between the developer's machine and the cluster.

The certificate, after being generated, is put into a Kubernetes Secret which is in turn mounted into the bucket server pod.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,7 +9,7 @@ GIT_COMMIT?=$(shell which git > /dev/null && git log -n1 --pretty='%h')
 VERSION?=$(shell which git > /dev/null && git describe --always --match "v*")
 FLUX_VERSION=0.37.0
 CHART_VERSION=$(shell which yq > /dev/null && yq e '.version' charts/gitops-server/Chart.yaml)
-DEV_BUCKET_CONTAINER_IMAGE=ghcr.io/weaveworks/gitops-bucket-server@sha256:8fbb7534e772e14ea598d287a4b54a3f556416cac6621095ce45f78346fda78a
+DEV_BUCKET_CONTAINER_IMAGE?=ghcr.io/weaveworks/gitops-bucket-server@sha256:157fa617e893e3ab0239547d8f1e820664b10c849fbd652c7f8738920b842f13
 TIER=oss
 
 # Go build args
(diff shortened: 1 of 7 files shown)


Makefile
```diff
@@ -9,7 +9,7 @@ GIT_COMMIT?=$(shell which git > /dev/null && git log -n1 --pretty='%h')
 VERSION?=$(shell which git > /dev/null && git describe --always --match "v*")
 FLUX_VERSION=0.37.0
 CHART_VERSION=$(shell which yq > /dev/null && yq e '.version' charts/gitops-server/Chart.yaml)
-DEV_BUCKET_CONTAINER_IMAGE=ghcr.io/weaveworks/gitops-bucket-server@sha256:8fbb7534e772e14ea598d287a4b54a3f556416cac6621095ce45f78346fda78a
+DEV_BUCKET_CONTAINER_IMAGE?=ghcr.io/weaveworks/gitops-bucket-server@sha256:157fa617e893e3ab0239547d8f1e820664b10c849fbd652c7f8738920b842f13
```
makkes, line 12: This lets devs override the image by issuing
```
DEV_BUCKET_CONTAINER_IMAGE=something make gitops
```

cmd/gitops/beta/run/cmd.go
```diff
@@ -615,21 +615,46 @@ func runCommandWithoutSession(cmd *cobra.Command, args []string) error {
...
-	unusedPorts, err := run.GetUnusedPorts(1)
+	unusedPorts, err := run.GetUnusedPorts(2)
 	if err != nil {
 		cancel()
 		return err
 	}
 
-	devBucketPort := unusedPorts[0]
-	cancelDevBucketPortForwarding, err := watch.InstallDevBucketServer(ctx, log0, kubeClient, cfg, devBucketPort)
+	devBucketHTTPPort := unusedPorts[0]
```
pjbgf, line 624: Can we get away with only having a single port, and supporting HTTPS only?

makkes: Unfortunately not because Flux doesn't (yet) support providing a custom CA certificate to a Bucket.

makkes:
@pjbgf as discussed, I changed the code to use a custom transport instead of relying on the env var consumed by the minio lib.

chanwit (approved):
LGTM

PR #3110, opened 2022-12-05
Fix incorrect message display for reconciled objects

### LINKED ISSUE #3109: Improve Status Message Display
The wrong message is displaying here in `ReconciledObjectsTable`
image

If there is a condition of type `Ready` or `Available` we should be prioritizing that condition in both `computeReady` and `computeMessage`. Here it's working in computeReady but not for computeMessage.

You can see the correct message in the yaml here:
image

joshri:
Closes #3109 

If we always display the message of the first conditions object, we sometimes don't prioritize conditions with types of `Ready` or `Available`. It looks like if conditions are not undefined, don't have types of ready or available, and don't have a condition with a status of false, `computeMessage` will return the first condition's message anyway, even without the code I've removed here in `ReconciledObjectsTable`.

@jpellizzari Am I missing something related to the `DaemonSets` here that require the message of the first condition?

before:
image

after:
image

### DIFF
--- a/ui/components/ReconciledObjectsTable.tsx
+++ b/ui/components/ReconciledObjectsTable.tsx
@@ -1,4 +1,3 @@
-import _ from "lodash";
 import * as React from "react";
 import styled from "styled-components";
 import { AppContext } from "../contexts/AppContext";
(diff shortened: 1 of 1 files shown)


jpellizzari:
> Am I missing something related to the DaemonSets here that require the message of the first condition?

@joshri DaemonSets don't use conditions like other objects. We have to calculate that manually ourselves.

jpellizzari (approved):

PR #3114, opened 2022-12-06
Generate secret for bucket server New

opudrovs:
WIP

Code re-added because of rebase issues.


pkg/run/watch/secret.go
```diff
@@ -0,0 +1,39 @@
+package watch
+
+import (
+	"math/rand"
```
makkes, line 4: We shouldn't use this package for security-sensitive applications.
```suggestion
	"crypto/rand"
```

> Package rand implements pseudo-random number generators unsuitable for security-sensitive work. ([source](https://pkg.go.dev/math/rand#pkg-overview))

pkg/run/watch/secret.go
```diff
@@ -0,0 +1,39 @@
...
+	return []byte(string(secretKey))
+}
+
+func generateRandomKey(numChars int, seed int64, letters []rune) []byte {
+	rand.Seed(seed)
+
+	key := make([]rune, numChars)
+
+	for i := range key {
+		key[i] = letters[rand.Intn(numChars)]
```
makkes, line 35:
```suggestion
		num, err := rand.Int(rand.Reader, big.NewInt(int64(len(letters))))
		if err != nil {
			return nil, fmt.Errorf("failed to get random number: %w", err)
		}
		key[i] = letters[num.Int64()]
```

makkes (changes requested):
apart from the math/rand => crypto/rand transition this looks fine to me. I'd like @pjbgf to have a look, too.

pkg/s3/auth_middleware_test.go
```diff
@@ -14,6 +14,11 @@ import (
 	"github.com/minio/minio-go/v7/pkg/signer"
 )
 
+const (
+	secretKey = "AKIAJIFCAFBEBI"
```
chanwit, line 18: I would expect these values to be random from the function?

chanwit (dismissed):

pkg/s3/auth_middleware_test.go
chanwit: probably changing the package name to be `s3_test` would fix the cycling dependency?

opudrovs: Please see the PR Notes:
>Tests in `auth_middleware_test.go` use hardcoded secret key and access key values because if trying to generate them in that file I am getting the `import cycle not allowed in testgo list` error (probably because the tests are in the `watch` package and import `gomega` and `ginkgo` too).

If you know how to fix this warning or think that I should move some tests for key generation to a separate package please let me know.

opudrovs: >probably changing the package name to be s3_test would fix the cycling dependency?

Yes, it should work, but we usually try to keep the tests in the same package. But if it's required let's do it.

Are random values good for unit tests in general? Maybe it's better to use a list of pre-generated key values for consistency? Ir is it some sort of an integration test?

opudrovs: And if we change the package name can I export the `verifySignature` function use in the tests too? Because it's private now.

makkes: IMHO it's a much more idiomatic way to have tests in a dedicated *_test package so that you're forced to only test the public API.

makkes (approved):
Let's make sure to squash the commits before merging

PR #3115, opened 2022-12-06
Pipeline docs for using GitOpsTemplates

sympatheticmoose:


website/docs/enterprise/pipelines/pipeline-templates.mdx
```diff
@@ -0,0 +1,278 @@
...
+      description: Environment the cluster is a part of within a pipeline.
+  resourcetemplates:
+  - apiVersion: notification.toolkit.fluxcd.io/v1beta1
+    kind: Provider
+    metadata:
+      name: ${RESOURCE_NAME}
+      namespace: ${RESOURCE_NAMESPACE}
+    spec:
+      address: http://${PROMOTION_HOST}/promotion/${APP_NAME}/${ENV_NAME}
+      type: generic
```
yiannistri, line 249:
```suggestion
      type: generic-hmac
```
if using `generic-hmac` then we should include the `secretRef` field, currently on appRef

sympatheticmoose: added to the provider

PR #3130, opened 2022-12-09
implement the get session logs endpoint for backend

### LINKED ISSUE #3033: GitOps Run - log retrieval on server-side for UI
As a UI developer,
I want to be able to call an API to obtain logs for a specific session,
so that these logs can be displayed on the GitOps Run UI.

Acceptance criteria:

- a Backend endpoint for the UI to call.
- It accepts a session name, the string to filter file query by suffix.
- It accepts an offset, ideally a value used for log pagination.
- it returns a set of strings, contents of each file, sorted by date.

chanwit:
Fixes #3033
This PR implements the get session logs endpoint for the backend.


chanwit:
Tested with `curl`

image

chanwit:
Payload example:
```
  --data-raw '{"sessionId": "run-master-8615cb75-dirty", "token":"", "clusterName": "Default", "namespace":"flux-system"}' 
```
Token must be "" to start retrieving from the first log entry.
Then you keep the nextToken value and pass it to the next call to retrieve the rest.

`namespace` is the namespace of the dev-bucket source from which we obtain the bucket information.

opudrovs (approved):

PR #3131, opened 2022-12-09
update guidance

### LINKED ISSUE #3097: adminUser passwordSecret
The install documentation recommends using something like SOPS or similar rather than using the default PASSWORD hash.
The issue is the chart doesn't provide an option to pass along the secret name to use.

So for example create a admin-dashboard-secret this would need to be able to be passed to the HelmRelease in place of adminUser.passwordHash ... so something like adminUser.passwordSecretName ?

sympatheticmoose:
Closes #3097

**What changed?**

**Why was this change made?**

**Release notes**

**Documentation Changes**

### DIFF
--- a/website/docs/installation/weave-gitops.mdx
+++ b/website/docs/installation/weave-gitops.mdx
@@ -139,6 +139,8 @@ brew install weaveworks/tap/gitops
 
 :::warning
 This command stores a hash of a password.  While this is relatively safe for demo and testing purposes it is recommend that you look at more secure methods of storing secrets (such as [Flux's SOPS integration](https://fluxcd.io/docs/guides/mozilla-sops/)) for production systems.
+
+More guidance and alternative login methods can be found in [Securing access to the dashboard](../../configuration/securing-access-to-the-dashboard).
 :::
 
 1. Commit and push the `weave-gitops-dashboard.yaml` to the `fleet-infra` repository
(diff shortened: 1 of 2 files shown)


website/docs/installation/weave-gitops.mdx
```diff
@@ -139,6 +139,8 @@ brew install weaveworks/tap/gitops
 
 :::warning
 This command stores a hash of a password.  While this is relatively safe for demo and testing purposes it is recommend that you look at more secure methods of storing secrets (such as [Flux's SOPS integration](https://fluxcd.io/docs/guides/mozilla-sops/)) for production systems.
+
+More guidance and alternative login methods can be found in [Securing access to the dashboard](../../configuration/securing-access-to-the-dashboard).
```
bigkevmcd, line 143: Should this be a single `..` ?

This is `website/docs/installation/weave-gitops.mdx` and the referenced file is in `website/docs/configration/securing-access-to-the-dashboard.mdx` ?

bigkevmcd: I could be misunderstanding how we generate the links when it's deployed tho'

yiannistri: Seems to work https://staging.docs.gitops.weave.works/install-update/docs/installation/weave-gitops/#deploying-weave-gitops

yiannistri (approved):

website/docs/installation/weave-gitops.mdx
sympatheticmoose: tried single `..` just to check and got `Page Not Found`

PR #3134, opened 2022-12-10
Keyboard shortcuts for "port forwards" on GitOps Run

### LINKED ISSUE #2738: Keyboard shortcuts for "port forwards" on GitOps Run

opudrovs:
Closes #2738 

- Added building and printing a list of set up port forwards.
- Added opening the selected port forward URL on key press. Listening for keypress is performed with the `github.com/mattn/go-tty` package (other options required pressing Enter after a keypress, this catches just a single numeric keypress) and opening URLs with the `github.com/pkg/browser` package.

Notes:
- As stated in the ticket description, we display the port forward passed in the `--port-forward` flag even though we are unable to say yet if it was set up successfully or not, right after "Reconciliation is done" is printed.
- I use the `Name` field from `PortForwardSpec` to get the service name for the portforward from the port forward flag, not the whole repository name as in the example. Because the spec says `[service-name]`.

Questions:
- If anyone knows of a better name for the `PortForwardShortcut` struct, I am all ears.


opudrovs:
@chanwit it is OK to print the list of set up port forwards with `fmt.Println`, correct?

opudrovs:
Added a screenshot:

opudrovs:
@chanwit great, thank you!

opudrovs:
The command to test it:

```
gitops beta run ./deploy/overlays/dev --timeout=3m --port-forward namespace=dev,resource=svc/backend,port=9898:9898
```

chanwit (approved):

cmd/gitops/beta/run/cmd.go
```diff
@@ -830,12 +830,39 @@ func runCommandWithoutSession(cmd *cobra.Command, args []string) error {
 						log.Successf("Reconciliation is done.")
 					}
 
+					portForwards := map[string]*watch.PortForwardShortcut{}
```
makkes, line 833: It feels a little odd that the key type here is a string while the map only ever holds numbers. We can just as well make it a slice:
```suggestion
					portForwards := make([]watch.PortForwardShortcut, 0)
```

cmd/gitops/beta/run/cmd.go
```diff
@@ -830,12 +830,39 @@ func runCommandWithoutSession(cmd *cobra.Command, args []string) error {
...
+
+					if dashboardInstalled {
+						portForwards[fmt.Sprint(len(portForwards)+1)] = &watch.PortForwardShortcut{
+							Name:     dashboardName,
+							HostPort: flags.DashboardPort,
+						}
```
makkes, line 839:
```suggestion
						portForwards = append(portForwards, watch.PortForwardShortcut{
							Name:     dashboardName,
							HostPort: flags.DashboardPort,
						})
```

pkg/run/watch/forward_port.go
```diff
@@ -126,3 +134,40 @@ func ForwardPort(log logr.Logger, pod *corev1.Pod, cfg *rest.Config, specMap *Po
 
 	return fw.ForwardPorts()
 }
+
+func ShowPortForwards(log clilogger.Logger, portForwards map[string]*PortForwardShortcut) {
```
makkes, line 138:
```suggestion
func ShowPortForwards(log clilogger.Logger, portForwards []PortForwardShortcut) {
```

pkg/run/watch/forward_port.go
```diff
@@ -126,3 +134,40 @@ func ForwardPort(log logr.Logger, pod *corev1.Pod, cfg *rest.Config, specMap *Po
...
+	fmt.Printf("\n\033[1m%s\033[0m\n\n", "We set up port forwards for you, use the number below to open it in the browser")
+
+	for key, portForward := range portForwards {
+		fmt.Printf("(%s) %s: http://localhost:%s\n", key, portForward.Name, portForward.HostPort)
+	}
+
+	fmt.Println()
+
+	// listen for keypresses
+	go func() {
```
makkes, line 150: It is idiomatic Go to never start a goroutine without a way to shut it down. I would recommend passing in a context.Context and if that's closed, exit from this goroutine.

opudrovs:
@makkes when exactly did you press Ctrl + C? After the port forward list was displayed? We might have some other issues with Ctrl+C I suppose.

cmd/gitops/beta/run/cmd.go
makkes: good point, I changed the suggestion.

pkg/run/watch/forward_port.go
opudrovs: And here should it be a slice or a map[rune]portforward?

pkg/run/watch/forward_port.go
opudrovs: I would appreciate an example.

pkg/run/watch/forward_port.go
```diff
@@ -126,3 +134,40 @@ func ForwardPort(log logr.Logger, pod *corev1.Pod, cfg *rest.Config, specMap *Po
...
+		}
+		defer tty.Close()
+
+		for {
+			r, err := tty.ReadRune()
+			if err != nil {
+				log.Failuref("Error reading keypress: %v", err)
+			}
+
+			portForward := portForwards[string(r)]
```
makkes, line 163:
```suggestion
			portForward := portForwards[r]
```

pkg/run/watch/forward_port.go
```diff
@@ -126,3 +134,40 @@ func ForwardPort(log logr.Logger, pod *corev1.Pod, cfg *rest.Config, specMap *Po
...
+
+	for key, portForward := range portForwards {
+		fmt.Printf("(%s) %s: http://localhost:%s\n", key, portForward.Name, portForward.HostPort)
+	}
+
+	fmt.Println()
+
+	// listen for keypresses
+	go func() {
+		tty, err := tty.Open()
```
makkes, line 151: This shadows the tty package name.
```suggestion
		tt, err := tty.Open()
```

makkes:
> @makkes when exactly did you press Ctrl + C? After the port forward list was displayed? We might have some other issues with Ctrl+C I suppose.

I believe the reason is that the TTY is never closed so you'd have to call `tty.Close()` before exiting.

pkg/run/watch/forward_port.go
opudrovs: Took it from their example, will change it.

opudrovs:
@makkes

>I believe the reason is that the TTY is never closed so you'd have to call tty.Close() before exiting.

You mean `defer tty.Close()` from their example is not working (mayeb when called from a goroutine)? I need to check it.

opudrovs:
@makkes 

- Addressed the changes:

-- Added passing context to the `ShowPortForwards` function.

-- Added two goroutines to this function for listening to keypresses and closing tty. I hope they look right.

-- Changed the type of the `portForwards` map key from `string` to `rune`.

-- Renamed the `tty` variable to `tt` to prevent shadowing of the `tty` package name.

-- Added closing tty properly in a separate goroutine.

For now keeping the changes in a distinct commit.

- I added returning (without an error, just return) if opening a tty returns an error. I am not sure if it's right, maybe it's we should actually return the error from the `ShowPortForwards` function. But I am not sure what is the right way to do it with those goroutines.

- I have not been able to reproduce the original issue in my terminal, but at least the temp debug message I added is printed before closing tty. So, it should be working.

- I saw a weird issue once, but before replacing the strings as map keys with runes, so maybe it was smth. intermediary:

Screenshot 2022-12-15 at 20 56 26

Can map keys iterated over with range be printed in the wrong order? I have not been able to reproduce the issue.

- Found another minor issue while testing it (and I think I saw it before but didn't pay attention): if pressing one of the keys which are present in the shortcuts map after a "requesting reconciliation..." message is displayed but before "Reconciliation is done." is displayed, the tty somehow reads these  keypresses and opens these shortcuts. is ot normal behavior or what should be done about it?

opudrovs:
@makkes so, now it seems to work, but do I still need to pass context to the second goroutine too (I think I was getting `panic: close of closed channel` message on Ctrl+C when I tried it)? And do I need to add more `for` or `select`s (I was getting some linter warnings but maybe I was adding them in the wrong way)?

cmd/gitops/beta/run/cmd.go
```diff
@@ -830,12 +830,39 @@ func runCommandWithoutSession(cmd *cobra.Command, args []string) error {
 						log.Successf("Reconciliation is done.")
 					}
 
+					portForwards := map[rune]watch.PortForwardShortcut{}
+
+					if dashboardInstalled {
+						portForwards[watch.PortForwardShortcutRunes[len(portForwards)+1]] = watch.PortForwardShortcut{
```
makkes, line 836: this has the potential to cause an "index out of range" error if the size of the `portForwards` slice is bigger than `watch.PortForwardShortcutRunes`. I think we should constrain the number of shortcuts available to the user to be no higher than the number of runes available in `watch.PortForwardShortcutRunes`.

This is kind of an edge case but it would backfire sonner or later I suppose.

makkes (commented):
lgtm, just need to find a way to not run into the out of range error.

pkg/run/watch/forward_port.go
```diff
@@ -126,3 +137,50 @@ func ForwardPort(log logr.Logger, pod *corev1.Pod, cfg *rest.Config, specMap *Po
 
 	return fw.ForwardPorts()
 }
+
+func ShowPortForwards(log clilogger.Logger, ctx context.Context, portForwards map[rune]PortForwardShortcut) {
```
makkes, line 141:
```suggestion
func ShowPortForwards(ctx context.Context, log clilogger.Logger, portForwards map[rune]PortForwardShortcut) {
```
Having the context as first param is more idiomatic.

makkes:
> do I still need to pass context to the second goroutine too (I think I was getting `panic: close of closed channel` message on Ctrl+C when I tried it)?

Theoretically you'd need to close the TTY in the case where the context isn't cancelled and thus the first goroutine not closing it. Since the 2nd goroutine is supposed to be running until the context is closed, anyway, we shouldn't have to worry about that.

makkes:
> Can map keys iterated over with range be printed in the wrong order?

The [order of iteration over maps is non-deterministic](https://go.dev/ref/spec#For_statements) ("The iteration order over maps is not specified and is not guaranteed to be the same from one iteration to the next") so the answer is "yes, we can't be sure".

cmd/gitops/beta/run/cmd.go
opudrovs: Added a guard against it for now.

We'll need to add getting and sorting map keys when building the port forward shortcut list in a followup PR later. So, we can add a better guard with letter shortcuts (in addition to numeric shortcuts) supported as well.

opudrovs:
@makkes if there are no other issues with the new `GetNextPortForwardKey` function, I am merging it.

And if passing logger before context as a function param is not idiomatic in Go, then we'll need to reorder them in other functions to, we have several tens of such functions, I suppose.

PR #3136, opened 2022-12-12
guidance on oss release notes

sympatheticmoose:
Follow up to https://github.com/weaveworks/weave-gitops/pull/3133

### DIFF
--- a/website/docs/releases.mdx
+++ b/website/docs/releases.mdx
@@ -7,6 +7,10 @@ import TierLabel from "./_components/TierLabel";
 
 # Releases <TierLabel tiers="enterprise" />
 
+:::note
+This page details the changes for Weave GitOps Enterprise and its associated components. For Weave GitOps OSS - please see the release notes on [GitHub](https://github.com/weaveworks/weave-gitops/releases).
+:::
+
 ## v0.11.0
 2022-11-25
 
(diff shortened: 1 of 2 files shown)


website/docs/releases.mdx
```diff
@@ -7,6 +7,10 @@ import TierLabel from "./_components/TierLabel";
 
 # Releases <TierLabel tiers="enterprise" />
 
+:::note
+This page details the changes for Weave GitOps Enterprise and its associated components. For Weave GitOps OSS - please see the release notes on [GitHub](https://github.com/weaveworks/weave-gitops/releases).
```
makkes, line 11: Why can't we incorporate the OSS release notes into this section as well?

sympatheticmoose: I _think_ when we break this out from the OSS repo, we would want - instead of a page - a section like:

- Releases
-- Weave GitOps Enterprise (Includes summarised content from all of the below with links to individual pages under the component list)
-- Weave GitOps OSS
-- Policy Agent
-- Terraform Controller
-- Pipelines Controller
-- Cluster Controller
-- Cluster Bootstrap Controller

Most we should be able to just pull from the releases on GitHub, the WGE one we could curate more.

But, that will take a little time 🙂 so I hoped to get a little plaster over this for now.

makkes (approved):

website/docs/releases.mdx
sympatheticmoose: I'll turn this suggestion into an issue before its lost to the ages

sympatheticmoose: raised https://github.com/weaveworks/weave-gitops/issues/3137

PR #3139, opened 2022-12-12
add namespace annotation to gitops run sessions

### LINKED ISSUE #3126: Make GitOps Run namespace flag value available in UI server session
As a developer,
I'd like to obtain additional information for example the "namespace" from the session resource.

chanwit:
This PR adds the annotation `run.weave.works/namespace` to session objects.
The value of this annotation is taken from `--namespace` (flags.Namespace) of GitOps Run (default to `flux-system`).

Fixes #3126

### DIFF
--- a/pkg/run/install/install_vcluster.go
+++ b/pkg/run/install/install_vcluster.go
@@ -35,10 +35,7 @@ func makeVClusterHelmRepository(namespace string) (*sourcev1.HelmRepository, err
 	return helmRepository, nil
 }
 
-func makeVClusterHelmRelease(name string, namespace string, portForwards []string, automationKind string) (*helmv2.HelmRelease, error) {
-	args := append([]string{filepath.Base(os.Args[0])}, os.Args[1:]...)
-	command := strings.Join(args, " ")
-
+func makeVClusterHelmRelease(name string, namespace string, command string, portForwards []string, automationKind string) (*helmv2.HelmRelease, error) {
 	helmRelease := &helmv2.HelmRelease{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      name,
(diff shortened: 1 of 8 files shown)


chanwit:
cc @joshri

pkg/run/install/install_vcluster.go
```diff
@@ -76,9 +76,16 @@ func makeVClusterHelmRelease(name string, namespace string, portForwards []strin
     "run.weave.works/cli-version": "%s",
     "run.weave.works/port-forward": "%s",
     "run.weave.works/command": "%s",
-    "run.weave.works/automation-kind": "%s"
+    "run.weave.works/automation-kind": "%s",
+    "run.weave.works/namespace": "%s"
```
makkes, line 80: can we add some unit tests for these annotations?

pkg/run/session/list.go
```diff
@@ -61,11 +63,12 @@ func List(kubeClient client.Client, namespace string) ([]*InternalSession, error
...
-			Namespace:   s.Namespace,
-			Command:     annotations["run.weave.works/command"],
-			CliVersion:  annotations["run.weave.works/cli-version"],
-			PortForward: strings.Split(annotations["run.weave.works/port-forward"], ","),
+			SessionName:      s.Name,
+			SessionNamespace: s.Namespace,
+			Command:          annotations["run.weave.works/command"],
+			CliVersion:       annotations["run.weave.works/cli-version"],
+			PortForward:      strings.Split(annotations["run.weave.works/port-forward"], ","),
+			Namespace:        annotations["run.weave.works/namespace"],
```
makkes, line 71: it'd be great to have this backed up by a test as well.

pkg/run/session/remove.go
```diff
@@ -31,8 +32,8 @@ func Remove(kubeClient client.Client, session *InternalSession) error {
 
 	if err := kubeClient.Get(context.Background(),
 		types.NamespacedName{
-			Namespace: session.Namespace,
-			Name:      session.Name,
+			Namespace: session.SessionNamespace,
```
makkes, line 35: I suppose this fixes the issue with `gitops remove run` not removing the session that a user would expect it would, right? We should definitely call this fix out explicitly, at best by creating an issue and referring to it in this PR.

chanwit: No it's not related to any other issue. The field was renamed as the part of this PR only.
We have "session namespace" and the "namespace" which is from the --namespace flag.

This PR changed "namespace" field to "session namespace" field
and used the "namespace" field to hold the value from --namespace flag.

PR #3145, opened 2022-12-12
fix formatting

davidstauffer:
Closes

**What changed?**

**Why was this change made?**

**Release notes**

**Documentation Changes**

### DIFF
--- a/website/docs/releases.mdx
+++ b/website/docs/releases.mdx
@@ -18,7 +18,7 @@ This page details the changes for Weave GitOps Enterprise and its associated com
 **We highly recommend users of v0.11.0 upgrade to this version as it includes fixes for a number of UI issues.**
 
 #### GitOps Templates
-- Support to specify Helm charts inside the CRD, instead of annotations. We’ve provided [documentation](.gitops.weave.works/docs/pipelines/pipeline-templates/) which has a example. 
+- Support to specify Helm charts inside the CRD, instead of annotations. We’ve provided [documentation](.gitops.weave.works/docs/pipelines/pipeline-templates.mdx) which has a example. 
 
 ```yaml
 spec:
@@ -64,7 +64,7 @@ config
 - From the Pipelines view you can add new Pipelines in a way which leverages GitOpsTemplates, additionally - to help users configure these, we’ve provided [documentation](.gitops.weave.works/docs/pipelines/pipeline-templates/) which has some samples.
 
 #### Support for multiple Flux instances on a single cluster
-- Support for running multiple flux instances in different namespaces on a single cluster for better resource isolation.
+- Support for running multiple flux instances in different namespaces on a single cluster for resource isolation.
 
 #### Minor fixes
 
--- a/website/docs/releases.mdx
+++ b/website/docs/releases.mdx
@@ -72,7 +72,7 @@ config
 Users of the Terraform Controller will be pleased to know we’ve addressed the issue where an error would be displayed if it had not been installed on all connected clusters.
 
 **Management cluster renaming**
-If a customer chose to rename the local cluster as it is known to Weave GitOps Enterprise from the default of management there were certain workflows where errors would occur, such as fetching profiles, this has now been resolved.
+If the name of the cluster where Weave GitOps Enterprise is installed, was changed from the default of management through the config.cluster.name parameter, certain workflows could fail such as fetching profiles, this has now been resolved.
 
 ### Dependency versions​
 weave-gitops v0.12.0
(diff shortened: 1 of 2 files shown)


foot (approved):

davidstauffer:
Will add tomorrow some examples for the new templating features @sympatheticmoose and @foot

website/docs/releases.mdx
```diff
@@ -44,13 +44,13 @@ config
 
 #### Minor fixes
 
-- Clicking on objects resulting in crashed UI
+**Clicking on objects resulting in crashed UI**
 We highly recommend users upgrade to v0.12.0 from 0.11.0, as unfortunately a bug slipped in where clicking on objects in the details tab for reconciled resources often resulted in the UI displaying a raw error message.
```
sympatheticmoose, line 48: We can't have a "Minor fix" with the suggestion "we highly recommend users upgrade" due to it. 

I would either remove the recommendation for now or move to the Highlights with something like "We highly recommend users of v0.11.0 upgrade to this version as it includes fixes for a number of UI issues, depending on when 0.12.1 is expected to land as this will contain similar guidance for the provider issue. My preference would be to move to highlights as the provider issue is less significant.

Then the error description could be more simple as "Clicking on certain objects in the details tab..."

Additionally suggest we rename Minor Fixes to Bug Fixes.

website/docs/releases.mdx
```diff
@@ -44,13 +44,13 @@ config
...
 
-- Terraform CRD Error
+**Terraform CRD Error**
 Users of the Terraform Controller will be pleased to know we’ve addressed the issue where an error would be displayed if it had not been installed on all connected clusters.
 
-- Management cluster renaming
+**Management cluster renaming**
 If a customer chose to rename the local cluster as it is known to Weave GitOps Enterprise from the default of management there were certain workflows where errors would occur, such as fetching profiles, this has now been resolved.
```
sympatheticmoose, line 54: Would rephrase, "If a customer chose" doesn't read well in the release notes.

"If the name of the cluster where Weave GitOps Enterprise is installed, was changed from the default of `management` through the `config.cluster.name` parameter, certain workflows could fail such as fetching profiles, this has now been resolved."

website/docs/releases.mdx
sympatheticmoose: Multi-Flux is too internally facing.

Would rename as "Support for multiple Flux instances on a single cluster" with a simple description. Note you will also either need to add the referenced screenshot or remove the reference

sympatheticmoose:
@davidstauffer I'd also suggest adding a known issues section which details the issue with the Provider and that a patch release will be available shortly.

website/docs/releases.mdx
```diff
@@ -15,8 +15,32 @@ This page details the changes for Weave GitOps Enterprise and its associated com
 
 ### Highlights
 
+**We highly recommend users of v0.11.0 upgrade to this version as it includes fixes for a number of UI issues.**
+
 #### GitOps Templates
-- Support to specify Helm charts inside the CRD, instead of annotations.
+- Support to specify Helm charts inside the CRD, instead of annotations. We’ve provided [documentation](.gitops.weave.works/docs/pipelines/pipeline-templates/) which has a example.
```
davidstauffer, line 21: @foot Link to be replaced once the documentation has landed.

website/docs/releases.mdx
davidstauffer: Done. Thanks.

website/docs/releases.mdx
davidstauffer: Done. Thanks.

website/docs/releases.mdx
```diff
@@ -37,21 +61,18 @@ config
...
 
 #### Pipelines - adding Pipelines through Templates
-- From the Pipelines view you can add new Pipelines in a way which leverages GitOpsTemplates, additionally - to help users configure these, we’ve provided documentation which has some samples.
+- From the Pipelines view you can add new Pipelines in a way which leverages GitOpsTemplates, additionally - to help users configure these, we’ve provided [documentation](.gitops.weave.works/docs/pipelines/pipeline-templates/) which has some samples.
 
-#### Multi-Flux
-- We can now support users running multiple instances of Flux on a single cluster for resource isolation, note the two versions within different namespaces on the cluster flux-system/dev
+#### Support for multiple Flux instances on a single cluster
+- Support for running multiple flux instances in different namespaces on a single cluster for better resource isolation.
```
sympatheticmoose, line 67: nit: remove the word "better", it's not a clear recommendation to do this, its just an option.

davidstauffer: done.

PR #3148, opened 2022-12-13
Add a prop for hiding seconds in the formatted text of the `Timestamp…

opudrovs:
Closes 

- Added the `hideSeconds` prop to the formatted text of the `Timestamp` component. If it's on, we display a hardcoded string "less than a minute ago" if the formatted time contains `seconds` (it only uses largest units, so there will be no text like `5 minutes 3 seconds`, only `5 minutes`) instead of the default formatted time string.

- Add the `minWidth` prop to the `DataTable` component.

Notes:

- I decided not to deal with time conversions and checks, because they would be harder to read and would also use hardcoded strings, but as keys, not as regular text:

```
if (hideSeconds && Math.abs(t.diffNow("seconds").seconds) < 60) {
  formattedTime = "less than a minute ago";
} else {
  formattedTime = t.toRelative();
}
```


joshri (approved):
Poked around in core and everything seems to be working as expected. LGTM!!!!

PR #3149, opened 2022-12-13
use dynamic Secret name for admin credentials

### LINKED ISSUE #2237: The admin credentials secret name is hardcoded
In https://github.com/weaveworks/weave-gitops/blob/main/charts/gitops-server/templates/admin-user-creds.yaml we hardcode the name of the secret, given that Helm charts should generate unique names for every release, we should change the secret metadata to conform with Helm.

Instead of:
```
apiVersion: v1
kind: Secret
metadata:
  name: cluster-user-auth
  namespace: flux-system
```

Should be:
```
apiVersion: v1
kind: Secret
metadata:
  name: {{ include "chart.fullname" . }}-user-auth
```

Then we need to pass the secret name to the gitops-server process so it knows which secret to look for in the same namespace where gitops-server runs .

makkes:
The gitops-server now has a flag `--admin-secret` for users to provide
the name of the Secret that stores the admin credentials. All
hard-coded occurrences of this Secret's name are removed.

The Helm chart now has the value `.adminUser.secretName` that lets
users specify the name of that secret in case they pre-created it or
want to otherwise maintain it outside of the Helm release.

When the Secret is generated through the chart by setting
`.adminUser.createSecret` the Secret gets a generated name derived
from the release name and a static suffix.

closes #2237


makkes:
Including @foot here mainly as a fyi. As soon as that's merged and enterprise bumps its dependencies, there are some things that need to be changed in enterprise, too.

website/docs/configuration/securing-access-to-the-dashboard.mdx
```diff
@@ -49,6 +49,12 @@ Once the HTTP server starts unauthenticated users will have to click the 'login
 
 ## Login via a cluster user account
 
+### Weave GitOps OSS
+
+If you followed the [installation instructions](../installation/weave-gitops.mdx) you don't need to perform any additional steps to enable this login method. You will be able to login to the dashboard using the credentials generated in the installation step.
```
sympatheticmoose, line 54: IMO this oversimplifies matters and doesn't point the user towards how they can configure manually should they want/need to - for example, if they have a requirement that all secrets are generated / managed via an external service.

Configuration docs should explain the mechanics and options available to the user. 

My suggestion would be to remove the OSS / WGE split and instead have a note like:

\`\`\`Weave GitOps OSS If you followed the [installation instructions](../installation/weave-gitops.mdx) you don't need to perform any additional steps to enable this login method. You will be able to login to the dashboard using the credentials generated in the installation step
\`\`\`

opudrovs (approved):

website/docs/configuration/securing-access-to-the-dashboard.mdx
makkes: Just discussed with @sympatheticmoose in-person: I will change this section to be a complete guide on the several ways you can configure a local admin user because there's so many different ways to install the dashboard that it's hard to provide step-by-step guidance.

opudrovs:
Closing as outdated.

PR #3151, opened 2022-12-14
Fixes the PR commit hash we publish npm modules under

foot:
- We derive the name of the npm artifact from the git ref.
- on.pull_request creates an intermediate merge commit and messes this up

**What changed?**

Updates the actions/checkout step to avoid a quirk of `on.pull_request` checking out the code to a "non-existant" merge commit. 

**Why was this change made?**

We use the ref from the commit to name the npm package. This should reflect the actual commit and not the temporary merge commit actions/checkout creates.

Straight up yaml engineering.

Checked the _publish js library_ CI output which now shows:
> `npm notice 📦  @weaveworks/weave-gitops-main@0.12.0-20-g116fd4bc`

For the head of this PR as of writing which matches the last commit.

**Release notes**

**Documentation Changes**

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -247,8 +247,13 @@ jobs:
       js-version: ${{ steps.package-version.outputs.js-version }}
     steps:
       - name: Checkout
-        uses: actions/checkout@v3
+        uses: actions/checkout@v2
         with:
+          # avoid the merge commit that on.pull_request creates
+          # fallback to github.sha if not present (e.g. on.push(main))
+          # https://github.com/actions/checkout#checkout-pull-request-head-commit-instead-of-merge-commit
+          # We want the correct sha so we can tag the npm package correctly
+          ref: ${{ github.event.pull_request.head.sha || github.sha }}
           fetch-depth: 0
       - uses: actions/setup-node@v2
         with:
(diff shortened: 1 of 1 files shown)


foot:
@opudrovs you should be able to run `make update-weave-gitops BRANCH=fixes-publishing-npm-module-on-pr` on this branch!

.github/workflows/pr.yaml
```diff
@@ -249,6 +249,11 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v2
         with:
+          # avoid the merge commit that on.pull_request creates
+          # fallback to github.sha if not present (e.g. on.push(main))
+          # https://github.com/actions/checkout#checkout-pull-request-head-commit-instead-of-merge-commit
+          # We want the correct sha so we can tag the npm package correctly
+          ref: ${{ github.event.pull_request.head.sha || github.sha }}
```
foot, line 256: I don't know if this will work for main. I _think_ so. Other solutions?

opudrovs:
@foot thanks, will have a look at it right after the sync.

makkes (approved):
I think it's worth a try. Thanks for addressing the issue, @foot.

opudrovs:
Done with my PR, will check it now.

opudrovs (approved):
LGTM.

The tag appeared for me after running the command you provided (with enterprise main checked out). I hope it is working as expected:

PR #3158, opened 2022-12-15
fix flaky multi-server test

### LINKED ISSUE #3144: TestMultiServerServesOverBothProtocols is flaky
The test is failing from time to time which can easily be reproduced:

```sh
$ (pass=0; fail=0; for i in $(seq 1 100) ; do go test -count=1 -v -run '^TestMultiServerServesOverBothProtocols$' ./pkg/http/ && pass=$(($pass+1)) || fail=$(($fail+1)) ; done; echo -e "Pass: $pass\nFail: $fail")
[...]
Pass: 84
Fail: 16
```

So in that run the test had a 16% failure rate.

makkes:
The HTTP server might take some time to boot up so the `http.Get` call
failed sometimes. This race condition is now fixed by using
`g.Eventually` to try the request until it succeeds.

```
$ (pass=0; fail=0; for i in $(seq 1 1000) ; do go test -count=1 -v -run '^TestMultiServerServesOverBothProtocols$' ./pkg/http/ && pass=$(($pass+1)) || fail=$(($fail+1)) ; done; echo -e "Pass: $pass\nFail: $fail")

[...]

Pass: 1000
Fail: 0
```
closes #3144

### DIFF
--- a/pkg/http/server_test.go
+++ b/pkg/http/server_test.go
@@ -70,8 +70,14 @@ func TestMultiServerServesOverBothProtocols(t *testing.T) {
 
 	// test HTTP
 
-	resp, err := http.Get(fmt.Sprintf("http://localhost:%d/", httpPort))
-	g.Expect(err).NotTo(HaveOccurred())
+	var resp *http.Response
+
+	g.Eventually(func() error {
+		var err error
+		resp, err = http.Get(fmt.Sprintf("http://localhost:%d/", httpPort))
+		return err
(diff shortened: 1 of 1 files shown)


opudrovs:
Neat fix!

The CI Static Checks step is failing, please run make lint. Otherwise  LGTM!

chanwit (approved):
LGTM

opudrovs (approved):

Issue #3154, opened 2022-12-14
Pendo: Tier key and values


josefaworks:

opudrovs:
@JamWils 

>For both OSS and Enterprise, latest version should be another key, with value as the latest version that was used.

What does the latest version mean in this context? Is it `most recent browser version and name` as in the doc you linked (which, I supposed, should be already logged by Pendo with the default visitor metadata by default) or is it the current dashboard version?

JamWils:
By `latestVersion` I am referring to the dashboard version they are using.  So we can target users based on "new features" we are trying to promote.

opudrovs:
OK, thanks! Then I will submit whichever dashboard version I can get (I remember that before it was displayed with a commit hash if running from main, but now I see `v0.12.0` without a commit, maybe smth. was changed).

PR #3172, opened 2022-12-19
Pendo: Track if users are coming from OSS or Ent and latest version

opudrovs:
Part of #3154

- Added tier and version to the `Pendo` component props.

- Added analytics as an optional `Signin` component prop.

- Added tracking tier and version as Pendo visitor metadata.

- Switched to initializing and re-identifying Pendo with objects instead of strings.

- Added the `PendoContainer` component to provide the version to Pendo (it should be used where using `useVersion` hook/calling the API in general is possible, within `AuthCheck`, because hooks can be called only at the top level in a function component and not conditionally).

- Switched to rendering the `Pendo` component twice: at the Signin page without the version (= without the `PendoContainer` component) and at other pages with the version.

- Reworked props of the `Footer`, and `Version` components.

- Moved getting the app version to a separate function `getAppVersion` and added unit tests for it.

- Updated some UI test snapshots (which were not updated before probably).

Notes:
- The name of the `shouldDisplayApiVersion` local var within the `getAppVersion` function is a convention from when we discussed adding the get version endpoint. It's like a full version built dynamically from values set at build time vs. the hardcoded release version from package.json.

- There is a known followup issue to this (in OSS only):
https://github.com/weaveworks/weave-gitops/issues/3191

Testing:
- To test if Pendo has been initialized and with which visitor metadata, run the following command in the browser console
```
console.log(window.pendo.getSerializedMetadata().visitor)
```

If Pendo is not initialized, this will throw an error.

In OSS, Pendo should be always initialized (with the version) on pages on which the user is logged in and should only be initialized on the Sign In page (but without the version) if there is a user data stored in the window's localStorage.

To clear localStorage for testing, run:

```
window.localStorage.clear()
```
Tested it locally, works as expected.

### DIFF
--- a/ui/App.tsx
+++ b/ui/App.tsx
@@ -14,6 +14,7 @@ import { ThemeProvider } from "styled-components";
 import ErrorBoundary from "./components/ErrorBoundary";
 import Layout from "./components/Layout";
 import Pendo from "./components/Pendo";
+import PendoContainer, { tier } from "./components/PendoContainer";
 import AppContextProvider from "./contexts/AppContext";
 import AuthContextProvider, { AuthCheck } from "./contexts/AuthContext";
 import CoreClientContextProvider from "./contexts/CoreClientContext";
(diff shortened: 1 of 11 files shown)


chanwit (approved):
code LGTM!

PR #3173, opened 2022-12-19
Jp OIDC scratch

jpellizzari:
Closes

**What changed?**

**Why was this change made?**

**Release notes**

**Documentation Changes**

### DIFF
--- a/core/clustersmngr/altclient.go
+++ b/core/clustersmngr/altclient.go
@@ -1,0 +1,35 @@
+package clustersmngr
+
+import (
+	"context"
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func NewAltClient() (Client, error) {
+
+	return &altClient{}, nil
+}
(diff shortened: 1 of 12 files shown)


core/clustersmngr/altclient.go
```diff
@@ -0,0 +1,180 @@
...
+	// build a client from values extracted from ctx or read from cache
+	fmt.Println("doing other client")
+	cfg, err := rest.InClusterConfig()
+	if err != nil {
+		return fmt.Errorf("building in-cluster config: %w", err)
+	}
+
+	user := auth.Principal(ctx)
+
+	cfg.BearerToken = user.Token()
```
foot, line 48: We might also have to `cfg.BearerTokenFile = ""` (which I guess is the ServiceAccount), as it seems to take precedence (https://github.com/kubernetes/client-go/blob/master/transport/round_trippers.go#L289)

PR #3177, opened 2022-12-20
Adds user-guide on how to use new template paths

foot:
**What changed?**

Docs for new paths feature in templates

**Why was this change made?**

User to understand and use this cool new feature.

Markdown

`GA_KEY=fakekey yarn start` and had a read through

### DIFF
--- a/website/docs/assets/templates/capd-template.yaml
+++ b/website/docs/assets/templates/capd-template.yaml
@@ -1,4 +1,4 @@
-apiVersion: capi.weave.works/v1alpha1
+apiVersion: capi.weave.works/v1alpha2
 kind: CAPITemplate
 metadata:
   name: cluster-template-development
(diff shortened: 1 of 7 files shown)


website/docs/assets/templates/capd-template.yaml
```diff
@@ -1,4 +1,4 @@
-apiVersion: capi.weave.works/v1alpha1
+apiVersion: capi.weave.works/v1alpha2
 kind: CAPITemplate
```
davidstauffer, line 2: Are we not deprecating now with this change slowly CAPI templates? and merge them all under the umbrella of GitOpsTemplates?

website/docs/cluster-management/profiles.mdx
```diff
@@ -92,7 +92,7 @@ As shown above, some profiles will be optional whereas some profiles will be req
 To allow editing of the yaml values for required profiles, the `editable` flag can be added in the annotation describing the required profile in the template. For example: 
 
 ```
-apiVersion: capi.weave.works/v1alpha1
+apiVersion: capi.weave.works/v1alpha2
 kind: CAPITemplate
```
davidstauffer, line 96: CAPITemplates vs GitOpsTemplate...

website/docs/gitops-templates/templates.mdx
```diff
@@ -136,6 +136,74 @@ Where
...
+          metadata:
+            name: ${CLUSTER_NAME}-cert-manager
+          ...
+```
+
+### Default paths
+
+If the `spec.resourcetemplates[].path` is omitted a default path for the rendered template is calculated.
+In this case some some of the submitted params are used. You must provide one of the following parameters:
+- `CLUSTER_NAME`
```
davidstauffer, line 185: Only are required of you are not specifying the path correct?

website/docs/assets/templates/capd-template.yaml
foot: Yeah we should good shout

website/docs/gitops-templates/templates.mdx
foot: Yes, _or_ if you are using profiles/kustomizations.

davidstauffer (approved):

Issue #3119, opened 2022-12-07
Improve golangci-lint configuration


makkes:
The current golangci-lint configuration isn't ideal for the following reasons:

- `wsl` eats developer time. It should be up to developers where to put blank lines and where not to. [whitespace](https://github.com/ultraware/whitespace) is much more lenient and actually helps improve readability.
- important linters such as [gosec](https://github.com/securego/gosec), [unconvert](https://github.com/mdempsky/unconvert) or [gocritic](https://github.com/go-critic/go-critic) are disabled.

Todos:

- [x] #3187 
- [x] #3282
- [x] #3284

makkes:
We are going to convert this into a task.

PR #3187, opened 2022-12-21
add explicit linter configuration; disable wsl

makkes:
This configuration enables all the same linters that have been enabled
so far except for wsl which is more a distraction than anything else
and does not add to the code's readability.

The new configuration lists all of those linters that we should enable
gradually going forward.

This allows for us to enable the linters one-by-one any time we have
some leeway to do so.

refs #3119

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -1,10 +1,84 @@
 run:
-  skip-dirs:
-      - pkg/vendorfakes
+  timeout: 5m
+  go: '1.18'
 
 linters:
+  disable-all: true
   enable:
-    - wsl
+    #- containedctx
+    #- contextcheck
+    - deadcode
+    #- depguard
+    - errcheck
+    #- errchkjson
+    #- errname
+    #- gochecknoinits
+    #- gci
+    #- goconst
+    #- gocritic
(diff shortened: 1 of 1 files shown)


opudrovs:
Does this PR close the linked issue completely or is just a part of closing it?
https://github.com/weaveworks/weave-gitops/issues/3119

makkes:
> grey_question Does this PR close the linked issue completely or is just a part of closing it? #3119

It doesn't close it completely, that's why I used `refs` and not `closes`.

.golangci.yaml
```diff
@@ -1,10 +1,84 @@
 run:
-  skip-dirs:
-      - pkg/vendorfakes
```
joshri, line 3: I'm guessing the new rules make it okay to lint everything in the vendorfakes directory?

joshri (approved):
The linter demanding whitespace was probably my least favorite thing about writing backend code so this is a very large thumbs up 👍

.golangci.yaml
makkes: Yes, it passed with this `skip-dirs` directive so I didn't bother to keep it there.

PR #3196, opened 2022-12-23
Make context the first param of Go functions

### LINKED ISSUE #3165: Make context the first param of Go functions
We have many functions in our Go code to which we pass a `context.Context` object, but it's not always passed as the first param.

According to the official recommendation @makkes provided, it should be passed as the first param:
https://pkg.go.dev/context#pkg-overview

Acceptance Criteria
- The current functions in Go code which accept a `context.Context` structure should accept the context as the first param. Refactoring should not introduce any regressions.

opudrovs:
Closes #3165 

- Made any params of type `context.Context` the first param.

Notes:
- Did not touch the `contextName` params, because they are related to Kubernetes contexts not the context form the standard package.
- Tested imported WG OSS in enterprise, have not found anything to fix there, related to this refactoring.


chanwit (approved):
LGTM

PR #3212, opened 2023-01-04
Fix brew hanging

### LINKED ISSUE #3193: brew upgrade seems to hang
The `weaveworks/tap/gitops` tap seems to hang on upgrade. This isn't something that is super critical, however it's persistent enough for me to mention as an issue, and a quick look at the brew files didn't reveal anything super obvious to jump out at me what might be causing it, so cause still unknown.

I've seen this with every gitops CLI upgrade recently through `brew upgrade` – my platform is Darwin/OSX Ventura 13.0.1

Brew makes it somewhat difficult for me to provide instructions to reproduce, I'd have to start with an outdated CLI and run `brew upgrade` – there has just been a release so likely someone out there can attempt to reproduce this now, but I'd have to make some checkout configuration or version pinning with git in order to get the tap back on an older version, and I'm not sure how to undo such a pin in a way that would produce the same environment we originally had when we did the upgrade normally 🤷 

Expected behavior is, `brew upgrade` does not hang – actual behavior, it hangs in the terminal for at least minutes, and the installation seems to complete normally after pressing `^C` – but also does not return right away, so having been impatient, I likely pressed `^C` multiple times, and interrupted some `brew cleanup` task the second time by mistake. Whoops...

opudrovs:
Closes #3193

- Replaced `popen_read` with `safe_popen_read` (the way it's recommended in the HomeBrew docs and issues). The actual thing that fixes brew upgrade or install hanging is adding `{ :err => :err }`.
- Added a hidden `no-analytics` flag to the root command in gitops to disable displaying analytics prompt (which runs in PersistentPreRun) when running the auto-generated `completion` command, in the brew formula. After I fixed install hanging, the analytics prompt started displaying while running the formula (maybe it's related to the hanging, like the prompt could be blocking smth. without an error message).

Testing:

I tested it by
- downloading the generated formula (the URL to which is visible when adding the `--verbose` and `--debug` options) from (in my case, `/usr/local/Homebrew/Library/Taps/weaveworks/homebrew-tap/Formula/gitops.rb`)
- adding changes from the current PR to the formula
-  building the gitops executable manually, archiving and serving it to brew from a local server (Chrome WebServer extension) by replacing the archive URL and checksum in the local formula
- uninstalling the gitops formula and installing it with `brew install gitops.rb --verbose --debug`

Maybe there is a better way.

When tested locally, the installation does not hang anymore, the analytics prompt is not displayed, and autocompletion works for the installed binary.

### DIFF
--- a/cmd/gitops/root/cmd.go
+++ b/cmd/gitops/root/cmd.go
@@ -2,14 +2,15 @@ package root
 
 import (
 	"fmt"
-	"github.com/weaveworks/weave-gitops/cmd/gitops/logs"
-	"github.com/weaveworks/weave-gitops/cmd/gitops/replan"
-	"github.com/weaveworks/weave-gitops/cmd/gitops/resume"
-	"github.com/weaveworks/weave-gitops/cmd/gitops/suspend"
 	"log"
 	"os"
 	"strings"
 	"time"
+
+	"github.com/weaveworks/weave-gitops/cmd/gitops/logs"
+	"github.com/weaveworks/weave-gitops/cmd/gitops/replan"
+	"github.com/weaveworks/weave-gitops/cmd/gitops/resume"
+	"github.com/weaveworks/weave-gitops/cmd/gitops/suspend"
 
 	"github.com/weaveworks/weave-gitops/cmd/gitops/remove"
 
@@ -97,6 +98,10 @@ func RootCmd() *cobra.Command {
 				os.Exit(1)
 			}
 
+			if options.NoAnalytics {
+				return
+			}
+
 			var gitopsConfig *config.GitopsCLIConfig
 
 			gitopsConfig, err = config.GetConfig(false)
@@ -143,8 +148,10 @@ func RootCmd() *cobra.Command {
 	rootCmd.PersistentFlags().StringToStringVar(&options.GitHostTypes, "git-host-types", map[string]string{}, "Specify which custom domains are running what (github or gitlab)")
 	rootCmd.PersistentFlags().BoolVar(&options.InsecureSkipTLSVerify, "insecure-skip-tls-verify", false, "If true, the server's certificate will not be checked for validity. This will make your HTTPS connections insecure")
 	rootCmd.PersistentFlags().StringVar(&options.Kubeconfig, "kubeconfig", "", "Paths to a kubeconfig. Only required if out-of-cluster.")
+	rootCmd.PersistentFlags().BoolVar(&options.NoAnalytics, "no-analytics", false, "Don't ask to enable/disable analytics.")
 	cobra.CheckErr(rootCmd.PersistentFlags().MarkHidden("override-in-cluster"))
 	cobra.CheckErr(rootCmd.PersistentFlags().MarkHidden("git-host-types"))
+	cobra.CheckErr(rootCmd.PersistentFlags().MarkHidden("no-analytics"))
 
 	rootCmd.AddCommand(version.Cmd)
 	rootCmd.AddCommand(get.GetCommand(options))
--- a/.goreleaser.brew.yml
+++ b/.goreleaser.brew.yml
@@ -11,9 +11,9 @@ brews:
     install: |
         bin.install "gitops"
         # Install bash completion
-        output = Utils.popen_read("#{bin}/gitops completion bash")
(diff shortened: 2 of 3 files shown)


chanwit:
@opudrovs could you give me quick instructions how to test it locally?

opudrovs:
@chanwit please see the last part of the PR description. I don't know of a better way right now. This is how I tested.

opudrovs:
@makkes I think the brew package is only built during the release? Do you know of a more convenient way to test it locally on PR without so many manual changes to the formula and archive?

makkes:
I've only used brew as an end user before but I'll see what I can find out.

makkes:
What confuses me more is that the "PR CI Workflow / PR CI Pipeline (pull_request)" job turns up twice in the CI checks. 🤔

opudrovs:
Same here. I read the Brew formula cookbook first time yesterday, but have not found anything useful for local testing so far except for running `brew install`. But maybe it's just hidden somewhere in it.

makkes:
> Added a hidden no-analytics flag to the root command in gitops to disable displaying analytics prompt (which runs in PersistentPreRun) when running the auto-generated completion command, in the brew formula. After I fixed install hanging, the analytics prompt started displaying while running the formula (maybe it's related to the hanging, like the prompt could be blocking smth. without an error message).

Instead of changing the command wouldn't it be more reasonable to pre-create the ~/.config/weave-gitops-config.json file?

opudrovs:
@makkes we need to ask the user during the first normal command run if they would like to turn on analytics or not. With which value should we pre-create the file during brew install and doesn't brew install run in sandbox in general? It is supposed to clean up after itself probably? Or what do you mean?

opudrovs:
@makkes 
> What confuses me more is that the "PR CI Workflow / PR CI Pipeline (pull_request)" job turns up twice in the CI checks.

I see the step repeated twice in other PRs, for example here:
https://github.com/weaveworks/weave-gitops/pull/3218

It looks like it runs once and it is skipped once.

makkes:
> I see the step repeated twice in other PRs, for example here: #3218
> 
> It looks like it runs once and it is skipped once.

This is definitely not right and it prevents the merge button from being green. We need to investigate this.

opudrovs:
>prevents the merge button from being green

There are no reviews yet. I think it's not supposed to be green before there is a review?

makkes:
> @makkes we need to ask the user during the first normal command run if they would like to turn on analytics or not. With which value should we pre-create the file during brew install and doesn't brew install run in sandbox in general? It is supposed to clean up after itself probably? Or what do you mean?

Oh I see. Sorry I should should've looked close what it is used for. So the reason why brew hangs is that gitops is prompting the user? Why doesn't this happen all the time someone tries to install gitops? Can we clarify this in the PR description, please?

opudrovs:
From the current PR description:

```
After I fixed install hanging, the analytics prompt started displaying while running the formula (maybe it's related to the hanging, like the prompt could be blocking smth. without an error message).
```

opudrovs:
So, I am not 100% sure it's the cause of the hanging, I just suppose it may be true because we only added the analytics prompt a few months ago. And I think now brew install hangs every time until Ctrl+C is pressed.

cmd/gitops/root/cmd.go
```diff
@@ -143,8 +148,10 @@ func RootCmd() *cobra.Command {
 	rootCmd.PersistentFlags().StringToStringVar(&options.GitHostTypes, "git-host-types", map[string]string{}, "Specify which custom domains are running what (github or gitlab)")
 	rootCmd.PersistentFlags().BoolVar(&options.InsecureSkipTLSVerify, "insecure-skip-tls-verify", false, "If true, the server's certificate will not be checked for validity. This will make your HTTPS connections insecure")
 	rootCmd.PersistentFlags().StringVar(&options.Kubeconfig, "kubeconfig", "", "Paths to a kubeconfig. Only required if out-of-cluster.")
+	rootCmd.PersistentFlags().BoolVar(&options.NoAnalytics, "no-analytics", false, "Disable analytics.")
```
makkes, line 151:
```suggestion
	rootCmd.PersistentFlags().BoolVar(&options.NoAnalytics, "no-analytics", false, "Don't ask to enable/disable analytics.")
```

makkes (approved):

PR #3218, opened 2023-01-04
Lookup namespaces with user client

foot:
Iterates on the work done in #3173. This is a more workaroundy approach that just makes It work.

**Performance implications**

- All clusters including the management cluster itself are now queried using the users OIDC token (you can strip the gitops-server's SA permissions right down now).
- We cache the namespaces a user has access to for 30s. Then we re-query them all using self subjectaccessreviews.
- There is now an additional 1 query per cluster to get the whole list of namespaces first.

**What changed?**

Lookup queryable namespaces using user clients instead of the
SA/kubeconfig-SA.

The user must have permissions to list namespaces for this to work at all.

**Why was this change made?**

- Supports use cases where a kubeconfig-secret for a cluster does not
  provide credentials (usually a service-account) that we can use to
  list all available namespaces. Instead we do this lookup in-request
  using the user's OIDC credentials
- The kubeconfig SA usually provides impersonation permissions, so
  instead of impersonation we again use the OIDC token provided by the
  user to do the usual subsequent requests

Tweaking the existing namespace lookup code so that it does the query at request time rather that in "in the background".

- `tilt`
- no tests yet.

Enabling:

Update values with:
```yaml

additionalArgs:
  - --auth-methods=oidc,token-passthrough

envVars:
  - name: WEAVE_GITOPS_FEATURE_OIDC_AUTH_PASSTHROUGH
    value: "true"
  - name: WEAVE_GITOPS_FEATURE_USE_USER_CLIENT_FOR_NAMESPACES
    value: "true"
```

**Release notes**

**Documentation Changes**


core/clustersmngr/cluster/single.go
```diff
@@ -75,6 +75,8 @@ func getImpersonatedConfig(config *rest.Config, user *auth.UserPrincipal) (*rest
 		return nil, fmt.Errorf("no user ID or Token found in UserPrincipal")
 	} else if tok := user.Token(); tok != "" {
```
bigkevmcd, line 76: We can simplify this, remove the else, because it's returning.

core/clustersmngr/cluster/single.go
```diff
@@ -75,6 +75,8 @@ func getImpersonatedConfig(config *rest.Config, user *auth.UserPrincipal) (*rest
...
 		cfg.BearerToken = tok
+		// Clear the token file as it seems to take precedence over the token
+		cfg.BearerTokenFile = ""
```
bigkevmcd, line 79: It feels like there's an opportunity for some refactoring to remove the duplication?

https://github.com/weaveworks/weave-gitops/blob/8d316717a22809bca21fe1ebc0b4671c27ff75e2/pkg/kube/config_getter.go#L40

core/clustersmngr/factory.go
```diff
@@ -314,10 +327,27 @@ func (cf *clustersManager) watchNamespaces(ctx context.Context) {
...
+		return cf.GetServerClient(ctx)
+	})
+}
+
+// Used when clusters have kubeconfig secrets that DO NOT contain credentials to connect to the cluster.
+// We need to use the user's token from the incoming request to connect to the cluster and list its namespaces
+// instead.
+func (cf *clustersManager) updateNamespacesFromUserClient(ctx context.Context, user *auth.UserPrincipal) error {
+	return cf.updateNamespacesWithClient(ctx, func() (Client, error) {
+		return cf.getImpersonatedClientWithNamespaces(ctx, user, false)
```
bigkevmcd, line 343: maybe this is a bit confusing, it's explicitly _not_ using impersonation?

core/clustersmngr/factory.go
```diff
@@ -542,6 +584,13 @@ func (cf *clustersManager) userNsList(ctx context.Context, user *auth.UserPrinci
 		return userNamespaces
 	}
 
+	if cf.useUserClientForNamespaces {
+		err := cf.updateNamespacesFromUserClient(ctx, user)
+		if err != nil {
+			cf.log.Error(err, "failed updating namespaces from user client")
+		}
```
bigkevmcd, line 591: We could call out why we don't return an error here?

core/clustersmngr/cluster/single.go
foot: Good catch!

core/clustersmngr/factory.go
foot: :thumbsup:

core/clustersmngr/factory.go
foot: True!

pkg/kube/config_getter.go
```diff
@@ -33,25 +33,36 @@ func NewImpersonatingConfigGetter(cfg *rest.Config, insecure bool) *Impersonatin
...
+
+// ConfigWithPrincipal returns a new config with the principal set as the
+// impersonated user or bearer token.
+func ConfigWithPrincipal(user *auth.UserPrincipal, config *rest.Config) *rest.Config {
+	cfg := rest.CopyConfig(config)
+
+	if tok := user.Token(); tok != "" {
+		cfg.BearerToken = tok
+		// Clear the token file as it takes precedence over the token.
+		cfg.BearerTokenFile = ""
```
jpellizzari, line 59: Came to the same conclusion in my branch, but it hadn't been pushed yet :+1:

bigkevmcd (approved):

foot:
> LGTM. Just so I understand, it uses the existing mechanism, just generates the client a different way?

Yeah exactly, it generates and calls the "list all namespaces" client:
1. using the token instead of SA
2. in-request instead of in a go-routine in the background

Issue #3203, opened 2022-12-27
Exiting "gitops run" leaves terminal in indeterminate state


pfcurtis:
**Describe the bug**

Once "gitops run" has exited, the terminal (curses) state is not returned to the default.

**Environment**

- OS: macOS X
- Weave-Gitops Version 0.13
- Flux Version 0.37
- Kubernetes version 1.25

**To Reproduce**

1. Open a terminal window
2. Execute "gitops run"
3. Exit that using `<Ctrl-C>`

**Expected behavior**

Terminal curses state returned to default state.

**Actual Behavior**

Terminal requires a "tset" in order to have the terminal work properly. The common example is newlines are not returned to the terminal when the <Enter> (return) key is pressed. This normally due to the libcurses state not being reset prior to exiting the application.

**Additional Context (screenshots, logs, etc)**

makkes:
Thanks for filing this bug @pfcurtis. I have two requests:

1. Would you mind expaning on step three of the "To Reproduce" section, please? It currrently says "Exit that using".
2. At which point do you exit the `gitops` program? Would you mind pasting the terminal output from around that phase, please?

pfcurtis:
There was a "Ctrl-C" at the end of step 3.

I exit "gitops run" when I'm done with the session. Typically after I've committed and pushed the code.

makkes:
@pfcurtis Would you mind pasting the terminal output from around that phase, please?

makkes:
First thing to do here would be to try to reproduce it and gather some more information on the exact point in time where this happens.

makkes:
I'm  not able to reproduce this one. @chanwit is there a chance you can give it a try as you said you run into it from time to time?

makkes:
I tried all different ways to reproduce it but to not avail. Closing.

PR #3224, opened 2023-01-04
Fix irrelevant error messages about log upload during shutdown for gitops run

### LINKED ISSUE #3064: gitops run: Irrelevant error messages about log upload during shutdown
Sometimes, after hitting ctrl+c to turn off gitops run, after every line in the terminal output there's another line printing
```
✗ failed to put log to s3: Put "http://localhost:45423/gitops-run-logs/no-session/20221125-124034.74331.txt": dial tcp [::1]:45423: connect: connection refused
```

I expect the log sender to stop sending logs at the same time as the S3 bucket is shut down.

chanwit:
Fixes #3064 

### DIFF
--- a/cmd/gitops/beta/run/cmd.go
+++ b/cmd/gitops/beta/run/cmd.go
@@ -15,8 +15,6 @@ import (
 	"syscall"
 	"time"
 
-	"k8s.io/client-go/discovery"
-
 	"github.com/fluxcd/go-git-providers/gitprovider"
 	kustomizev1 "github.com/fluxcd/kustomize-controller/api/v1beta2"
 	sourcev1 "github.com/fluxcd/source-controller/api/v1beta2"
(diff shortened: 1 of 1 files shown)


cmd/gitops/beta/run/cmd.go
```diff
@@ -548,6 +548,18 @@ func runCommandWithoutSession(cmd *cobra.Command, args []string) error {
 	// 2. log is the S3 logger that also delegates its outputs to "log0".
 	log0 := logger.NewCLILogger(os.Stdout)
 
+	state, err := term.GetState(int(os.Stdin.Fd()))
```
makkes, line 551: Do you know exactly why this is necessary? I've never ran across terminal issues after exiting gitops run and I'd like us to be aligned on why the terminal state isn't as expected when exiting.

chanwit: I pressed Ctrl +C many times and was hit by the bug. It's not so hard to reproduce (but still a random behaviour).

Also, Paul Curtis reported it here https://github.com/weaveworks/weave-gitops/issues/3203

makkes: Yes, I've seen it being reported but I'd like to understand the circumstances under which it occurs.

makkes:
as discussed in standup, let's please split the PR up into two so that we get proper release notes entries.

makkes (approved):

Issue #3217, opened 2023-01-04
BE: API to query for ImageRepositories


yiannistri:
API endpoint for #3223

Issue #3216, opened 2023-01-04
BE: API to query for ImagePolicies


yiannistri:
API endpoint for #3222

PR #3232, opened 2023-01-05
Allowing image automations listing

### LINKED ISSUE #3215: BE: API to query for ImageUpdateAutomations

luizbafilho:
Closes #3215 #3216 #3217 

**What changed?**
Allows Image Automations objects to be listed using `ListObjects` handler

**Why was this change made?**

**Release notes**

**Documentation Changes**


yiannistri (approved):
Small but elegant 🥇

yitsushi (approved):

tools/reboot.sh
```diff
@@ -23,6 +23,7 @@ do_bootstrap(){
 
         flux bootstrap "$provider" \
             --owner="$owner" \
+            --components-extra=image-reflector-controller,image-automation-controller \
```
makkes, line 26: Does this mean that image automation will need to be installed for Weave GitOps to work going forward?

makkes: That'd be very nice. Progressive enhancement ftw.

yiannistri: Good point, if it's causing UI errors perhaps we can put this behind a flag that the user can enable.

yiannistri: Yes we should handle this gracefully and show a message that explains why the image automation list is empty (when the CRD is not installed). Strictly speaking this is not an error because the controllers are not expected to be installed (in the majority of cases).

pkg/kube/kubehttp.go
```diff
@@ -34,6 +36,8 @@ func CreateScheme() (*apiruntime.Scheme, error) {
 		rbacv1.AddToScheme,
 		authv1.AddToScheme,
 		notificationv2.AddToScheme,
+		reflectorv1.AddToScheme,
```
enekofb, line 39: @luizbafilho it might be useful to just raise a quick ADR to record that we are not creating its own endpoints because XWZ

PR #3234, opened 2023-01-05
Allow customisation of the scopes.

bigkevmcd:
**What changed?**
This provides a mechanism for customising the requested scopes from an upstream OIDC server.

**Why was this change made?**
Some customers need to be able to customise the set of scopes that are requested, perhaps because they need specialised ones for a claim.

It's a pretty trivial change, defaults to the existing defaults, can be parsed from the OIDC Secret (as the other related values can) and it drops the "must have email and groups", retaining "must have openid" because we must have that.

Tests and got James to test it against Okta

**Release notes**
Support for custom OIDC scopes when authenticating with an upstream server.

**Documentation Changes**
We need documentation on how to configure the OIDC Secret.

### DIFF
--- a/cmd/gitops-server/cmd/cmd.go
+++ b/cmd/gitops-server/cmd/cmd.go
@@ -113,6 +113,7 @@ func NewCommand() *cobra.Command {
 	cmd.Flags().DurationVar(&options.OIDC.TokenDuration, "oidc-token-duration", time.Hour, "The duration of the ID token. It should be set in the format: number + time unit (s,m,h) e.g., 20m")
 	cmd.Flags().StringVar(&options.OIDC.ClaimsConfig.Username, "oidc-username-claim", auth.ClaimUsername, "JWT claim to use as the user name. By default email, which is expected to be a unique identifier of the end user. Admins can choose other claims, such as sub or name, depending on their provider")
 	cmd.Flags().StringVar(&options.OIDC.ClaimsConfig.Groups, "oidc-groups-claim", auth.ClaimGroups, "JWT claim to use as the user's group. If the claim is present it must be an array of strings")
+	cmd.Flags().StringSliceVar(&options.OIDC.Scopes, "custom-oidc-scopes", auth.DefaultScopes, "Customise the requested scopes for then OIDC authentication flow - openid will always be requested")
 	// Metrics
 	cmd.Flags().BoolVar(&options.EnableMetrics, "enable-metrics", false, "Starts the metrics listener")
 	cmd.Flags().StringVar(&options.MetricsAddress, "metrics-address", ":2112", "If the metrics listener is enabled, bind to this address")
--- a/pkg/server/auth/server.go
+++ b/pkg/server/auth/server.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"net/http"
 	"net/url"
+	"strings"
 	"time"
 
 	"github.com/coreos/go-oidc/v3/oidc"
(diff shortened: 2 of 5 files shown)


cmd/gitops-server/cmd/cmd.go
```diff
@@ -113,6 +113,7 @@ func NewCommand() *cobra.Command {
 	cmd.Flags().DurationVar(&options.OIDC.TokenDuration, "oidc-token-duration", time.Hour, "The duration of the ID token. It should be set in the format: number + time unit (s,m,h) e.g., 20m")
 	cmd.Flags().StringVar(&options.OIDC.ClaimsConfig.Username, "oidc-username-claim", auth.ClaimUsername, "JWT claim to use as the user name. By default email, which is expected to be a unique identifier of the end user. Admins can choose other claims, such as sub or name, depending on their provider")
 	cmd.Flags().StringVar(&options.OIDC.ClaimsConfig.Groups, "oidc-groups-claim", auth.ClaimGroups, "JWT claim to use as the user's group. If the claim is present it must be an array of strings")
+	cmd.Flags().StringSliceVar(&options.OIDC.Scopes, "custom-oidc-scopes", auth.DefaultScopes, "Customise the requested scopes for then OIDC authentication flow - openid will always be requested")
```
foot, line 116: This flag doesn't seem to be connected right now, maybe we remove it for now or?

pkg/server/auth/server.go
```diff
@@ -110,9 +120,29 @@ func NewOIDCConfigFromSecret(secret corev1.Secret) OIDCConfig {
 
 	cfg.TokenDuration = tokenDuration
 
+	scopes := splitAndTrim(string(secret.Data["customScopes"]))
+	if len(scopes) == 0 {
```
foot, line 124: I think I may have been doing the odd `if slice == nil` in other places, will keep the distinction in mind..

pkg/server/auth/server.go
```diff
@@ -311,7 +334,7 @@ func (s *AuthServer) Callback() http.HandlerFunc {
 			return
 		}
 
-		token, err = s.oauth2Config(nil).Exchange(ctx, code)
+		token, err = s.oauth2Config(s.OIDCConfig.Scopes).Exchange(ctx, code)
```
foot, line 337: Is it important to include the scopes when doing the exchange?

foot (approved):
Tested with okta, got a nice little popup asking if I wanted to allow the extra scopes. 👍

pkg/server/auth/server.go
bigkevmcd: In this case, `result := []string{}` will return a slice of zero length.

`var result []string` would mean that result _could_ be nil.

Taking the length of a nil slice yields `0` any which way.

cmd/gitops-server/cmd/cmd.go
bigkevmcd: I think it is connected, through a long line of bits.

It is passed here https://github.com/weaveworks/weave-gitops/blob/2deffd819e9d9482e033bc3cb002b62becb21aaa/cmd/gitops-server/cmd/cmd.go#L175

pkg/server/auth/server.go
bigkevmcd: No, you're right, they're not, I think I did a quick `s/scopes/s.OIDCConfig.Scopes/` good catch :-)

cmd/gitops-server/cmd/cmd.go
foot: Oh you're right. I was testing with mixed secret/cli config and then missed this

https://github.com/weaveworks/weave-gitops/blob/2deffd819e9d9482e033bc3cb002b62becb21aaa/pkg/server/auth/init.go#L40-L41

LGTM

bigkevmcd: I find it a bit confusing that the secret overrides the cmd-line, but that's the existing behaviour for other options, consistency wins out :-)

PR #3239, opened 2023-01-07
wip: add automated age decryption to GitOps Run

makkes:
This way users can specify the path to an age key that is used to
decrypt Secrets created by the GitOps Run Kustomization.

Background: Often a workload contains Secrets that need to be created
and in a GitOps workflow those Secrets are often stored encrypted.
Flux supports that use case by offering a `.spec.decryption` field to
tell Flux which decryption provider to use (currently only supports
SOPS) and where to find the Kubernetes Secret with the decryption key.

Usage example:
```
gitops beta run --decryption-sops-key-file=../sops.agekey .
```


makkes:
I will wait for #3250 to get merged so I can add tests for this particular PR.

chanwit (approved):

PR #3251, opened 2023-01-10
Loop over namespaces when listing reconciled objects

### LINKED ISSUE #3231: Unable to see resources created by HelmReleases/Kustomizations
**Describe the bug**

I'm logged in to the dashboard through OIDC and I can see the Kustomizations and HelmReleases but as I click on one of them the list of resources in the lower table is always empty. With the admin user the table lists all the resources as expected.

**Environment**

- Weave-Gitops Version: 0.14.1
- Flux Version: 0.37.0
- Kubernetes version: 1.26

**To Reproduce**

1. Login to the dashboard through OIDC with a user that only has access to read resources in a certain namespace
1. Click on one of the HelmReleases or Kustomizations on the "Applications" page

**Expected behavior**

I can see all the resources created by the Kustomization/HelmRelease in the lower table.

**Actual Behavior**

Table is empty.

Admin view:

OIDC user view:

timricese:
Fixes #3231 

**What changed?**
When getting reconciled objects, instead of just looping over kinds and retrieving all objects cluster-wide, loop over only namespaces which the current user has access to, as well as kinds as before.

**Why was this change made?**
When an OIDC impersonated user does not have cluster-wide privileges, the weave-gitops UI will not show any resources that are reconciled as part of kustomize/helmrelease etc objects. This change means that only objects in namespaces which the user has permissions in will be listed.

Simply providing another option to the client when calling List(), which specifies the namespace to list from, and looping over the namespaces which the user has access to. The list of user-accessible namespaces was already defined a few lines later in the code.

Tested locally/in our production clusters. 

We are expecting that existing tests should already cover this and dont need updating.

**Release notes**
Bugfix, nothing notable

**Documentation Changes**
None

### DIFF
--- a/core/server/events_test.go
+++ b/core/server/events_test.go
@@ -11,6 +11,7 @@ import (
 	. "github.com/onsi/gomega"
 	pb "github.com/weaveworks/weave-gitops/pkg/api/core"
 	"github.com/weaveworks/weave-gitops/pkg/kube"
+	"google.golang.org/grpc/metadata"
 	corev1 "k8s.io/api/core/v1"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/rand"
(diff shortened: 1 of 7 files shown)


opudrovs:
@timricese thank you for submitting this PR! 🎉 We'll review it as soon as we have some bandwidth.

Some checks are failing. Could you please fix whatever can be fixed? (All that is not caused by some unpatched JS vulnerabilities or similar issues.)

timricese:
Thanks!

The failing test (just linting whitespace) should work now.

makkes:
Great stuff @timricese. I added a commit providing integration tests for the intended behaviour. Let me know what you think.

makkes (approved):

PR #3254, opened 2023-01-10
poll to wait for a pod before starting port forward

### LINKED ISSUE #2883: gitops run: race condition in setting up port forwarding
At first run, it's possible for the flux resources to be healthy and available, but for the pods not to be.

To reproduce, grab a cluster without podinfo running. Create a new directory with a file with this in it:
```
---
apiVersion: v1
kind: Namespace
metadata:
  name: dev
---
apiVersion: source.toolkit.fluxcd.io/v1beta2
kind: HelmRepository
metadata:
  name: podinfo
  namespace: dev
spec:
  interval: 1m0s
  url: https://stefanprodan.github.io/podinfo
---
kind: HelmRelease
metadata:
  name: podinfo
  namespace: dev
spec:
  chart:
    spec:
      chart: podinfo
      reconcileStrategy: ChartVersion
      sourceRef:
        kind: HelmRepository
        name: podinfo
  interval: 1m0s
  targetNamespace: dev
```
and run gitops run as

`gitops beta run ./podinfo --no-session --port-forward namespace=dev,resource=svc/dev-podinfo,port=9898:9898`

The first run you should see `✗ Error getting pod from specMap: no running pods found for service`. Re-save the file, and it'll update and the port forwarding will work.

chanwit:
Fixes #2883

### DIFF
--- a/cmd/gitops/beta/run/cmd.go
+++ b/cmd/gitops/beta/run/cmd.go
@@ -34,8 +34,10 @@ import (
 	"github.com/weaveworks/weave-gitops/pkg/s3"
 	"github.com/weaveworks/weave-gitops/pkg/validate"
 	"github.com/weaveworks/weave-gitops/pkg/version"
+	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/cli-runtime/pkg/genericclioptions"
 	"k8s.io/client-go/discovery"
 	"k8s.io/client-go/rest"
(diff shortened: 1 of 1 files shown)


cmd/gitops/beta/run/cmd.go
```diff
@@ -897,12 +899,26 @@ func runCommandWithoutSession(cmd *cobra.Command, args []string) error {
...
+
+							log.Waitingf("Waiting for a pod from specMap: %v", podErr)
+							return false, nil
+						}); pollErr != nil {
+							log.Failuref("Waiting for a pod from specMap: %v", pollErr)
 						}
 
-						if pod != nil {
+						if pod == nil {
+							log.Failuref("Error getting pod from specMap")
```
makkes, line 920: shouldn't we return here?

cmd/gitops/beta/run/cmd.go
```diff
@@ -897,12 +899,26 @@ func runCommandWithoutSession(cmd *cobra.Command, args []string) error {
...
+						if pollErr := wait.PollImmediate(2*time.Second, flags.Timeout, func() (bool, error) {
+							pod, podErr = run.GetPodFromResourceDescription(thisCtx, namespacedName, specMap.Kind, kubeClient)
+							if pod != nil && podErr == nil {
+								return true, nil
+							}
+
+							log.Waitingf("Waiting for a pod from specMap: %v", podErr)
+							return false, nil
+						}); pollErr != nil {
+							log.Failuref("Waiting for a pod from specMap: %v", pollErr)
```
makkes, line 916: shouldn't we return here?

makkes (commented):
lgtm, just 2 questions.

cmd/gitops/beta/run/cmd.go
chanwit: we cannot.

returning means stopping the go routine.
we have to log and wait for the next trigger.

opudrovs (approved):
LGTM! 🎉

makkes (approved):

PR #3273, opened 2023-01-16
fix infinite loop when finding the root of a git repo

chanwit:
Fixes #

The fixed version first gets the absolute path of the current directory using `filepath.Abs(".")` and assigns it to the `gitDir` variable.
Then it enters a loop that checks if the current directory is a git repository by calling `os.Stat(filepath.Join(gitDir, ".git"))`.
If the current directory is not a git repository, it goes to the parent directory by calling `filepath.Clean(filepath.Join(gitDir, ".."))` and assigns the result to `gitDir`.

Then it gets the absolute path of the parent directory using `filepath.Abs(gitDir)` and assigns it to `absGitDir`.
It compares the parent directory of `absGitDir` and `gitDir` using `filepath.Dir(absGitDir) == gitDir`, and if it evaluates to true, it returns an error message indicating that it is not in a git repository.
Finally, it assigns `absGitDir` to `gitDir` and the loop continues.

This implementation also checks for permission issues by calling `os.IsPermission(err)` and returns an error message if there is a permission issue.  This implementation should be able to handle the cases where the parent directory is not reachable or not accessible due to permission issues.


opudrovs (approved):
LGTM! 🎉  Tested it locally, works as expected.

PR #3276, opened 2023-01-17
feat: endpoint to query CRDs

### LINKED ISSUE #3271: Add endpoint to indicate whether image reflector/automation CRDs are present in the cluster
To properly support image automation objects in OSS, we should add an endpoint that the UI can use to query whether the relevant CRDs are installed. This is so that end users do not get an error when they navigate to that section in the UI and are prompted instead with a friendly message to add the image automation controllers to their cluster. 

There are 2 additional controllers that need to be installed `image-reflector-controller` and `image-automation-controller` and 3 related CRDs `imagepolicies`, `imagerepositories` and `imageupdateautomations` . I'm not sure whether we need to get too sophisticated about this but I would suggest to add a single endpoint that can indicate to the UI whether all 3 are installed or not. After all the UI will error if any of the 3 are queried but are not present.

yitsushi:
Closes #3271

**What changed?**
A single endpoint to check if given CRD is available on clusters or not, so we can show a quick-start guide or simply do not even try to query resources if the CRD is not installed.

**Why was this change made?**

A background go routine updates an internal cache a list of custom resources available. An endpoint to query this information.

Return structure from the endpoint:

```json
{
  "Default": true,
  "Cluster1": true,
  "Cluster2": false
}
```

**Considered**

I considered using a string list as an input to check multiple resources in one query, but settled on one per query because I think it will be called mostly when we navigate to a page where we want to work with a given resource. Allowing multiple resources at once would give the idea to query everything and store the result in a global scope on the UI and refresh periodically, which is not a bad idea, but harder to extend the logic as we add more features, potentially behind feature flags and we may end up in a situation where we have 2-3 places to query CRDs with 2-3 different logic.

If we want to extend this to be able to query multiple, on the back-end it's a simple additional loop, resources are cached, and it has no additional cost.

The return structure would look something like this:

```json
{
  "resource1": {
    "Default": true,
    "Cluster1": true,
    "Cluster2": false
  },
  "resource2": {
    "Default": false,
    "Cluster1": false,
    "Cluster2": false
  }
}
```

Tests

**Release notes**
I don't know.

**Documentation Changes**
No.


pkg/services/crd/fetcher.go
```diff
@@ -0,0 +1,111 @@
...
+	s.Lock()
+	defer s.Unlock()
+
+	for clusterName, crds := range s.crds {
+		// Set this to be sure the key is there with false value if the following
+		// look-up does not say it's there.
+		result[clusterName] = false
+
+		for _, crd := range crds {
+			if crd.Name == name {
```
makkes, line 103: Usually in Kubernetes you would check the GVK or GK and not just the name. In this particular case you'd additionally query (and check) `.spec.group` and optionally `.spec.versions.name`.

makkes (commented):
How would this work for users that don't have permission to query for cluster-scoped resources, i.e. users that only have access to a certain Namespace?

pkg/services/crd/fetcher.go
makkes: This might be a helpful resource: https://helm.sh/docs/chart_template_guide/function_list/#capabilitiesapiversionshas

yitsushi: The only purpose of this endpoint is to check if given resource is available or not for example `canaries.flagger.app` to determine so we can show a getting started guide to the user if it's not installed.

yitsushi:
> users that only have access to a certain Namespace?

It uses the cluster role in a background process.

pkg/services/crd/fetcher.go
yitsushi: Technically this is just a port (copy + update) of the same functionality from progressive delivery, but as we go forward we want more resources to check.

makkes:
> > users that only have access to a certain Namespace?
> 
> It uses the cluster role in a background process.

Doesn't this leak information to a user who doesn't have access to that information from an RBAC perspective?

yitsushi:
> Doesn't this leak information to a user who doesn't have access to that information from an RBAC perspective?

Yes it does, we had this discussion back then (last spring or summer), but I don't remember what we said "not an issue". Do you remember @yiannistri ?

I think it was something like (but maybe that was just one reason, really don't really remember):
yes it does, but the existence of a crd shouldn't be an issue as they can't create, view, update, or delete those resources even if a user knows they are available on the cluster.

yiannistri:
Yes this is a valid concern, however the purpose of this is to present a helpful message to the user to prompt them to install image automation controllers that are currently optional. As long as the admins can control the permissions (and by consequence whether the UI errors or not) I think it's a reasonable compromise between usability/progressive user engagement.

yiannistri:
Tested with curl
```
curl 'http://localhost:9001/v1/crd/is_available?name=gitrepository.source.toolkit.fluxcd.io' -X GET -H 'Cookie: ...'
{"clusters":{"Default":false}}

curl 'http://localhost:9001/v1/crd/is_available?name=gitrepositories.source.toolkit.fluxcd.io' -X GET -H 'Cookie: ...'
{"clusters":{"Default":true}}
```
Note that this uses the plural version of the CRD name.

pkg/services/crd/suit_test.go
```diff
@@ -0,0 +1,136 @@
+package crd_test
```
yiannistri, line 1: s/suit_test.go/suite_test.go

yiannistri (approved):
Well-written code and good test coverage 👏 💯 

It would be good to add doc comments to the new files being added, `fetcher.go` and `nocachefetcher.go`.

core/server/crd.go
```diff
@@ -0,0 +1,13 @@
+package server
+
+import (
+	"context"
+
+	pb "github.com/weaveworks/weave-gitops/pkg/api/core"
+)
+
+func (cs *coreServer) IsCRDAvailable(ctx context.Context, msg *pb.IsCRDAvailableRequest) (*pb.IsCRDAvailableResponse, error) {
```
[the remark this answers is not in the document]
yitsushi: We don't really have comments on endpoint functions as we have comment already in the proto file, but i'll add one here.

PR #3294, opened 2023-01-23
Add exports and updates for gitopsset details in EE

AlinaGoaga:
Part of https://app.zenhub.com/workspaces/team-pesto-6127680175876a0014fc5657/issues/gh/weaveworks/weave-gitops-enterprise/2142

Added relevant exports and a type `ReconciledObjectsAutomation` instead of sending either a HelmRelease or a Kustomization to the graph / reconciled objects table. This allows us to reuse these components for GitOpsSets also.


ui/components/AutomationDetail.tsx
```diff
@@ -66,6 +80,35 @@ function AutomationDetail({
...
+  } = automation
+    ? useGetReconciledTree(
+        automation.name,
+        automation.namespace,
+        Kind[automation.type],
+        automation.inventory,
+        automation.clusterName
+      )
+    : { data: [], error: null, isLoading: false };
+  //add extra nodes
```
opudrovs, line 97: ⚠️ Redundant comment (probably copy-pasted with a code snippet?).

opudrovs (approved):
LGTM! 🎉 Tested it locally, the graph and reconciled objects table work fine. The only minor fix is the duplicate comment, but it does not affect the app's behavior.

ui/components/AutomationDetail.tsx
AlinaGoaga: Ah good catch thanks Olga, will remove :)

PR #3295, opened 2023-01-23
Set up a new bucket for gitops run logs

### LINKED ISSUE #3268: Set up a new bucket for gitops run logs
Set up a new bucket to store the collected logs (in addition to the current two buckets)

# Implementation Details

In order to set up a new bucket to store the collected logs in addition to the current two buckets, the following steps can be taken:

1. Create A new S3-compatible bucket named `pod-logs` in the dev-bucket server.

# Acceptance Criteria

- A bucket named `pod-logs` is created in the dev-bucket server.

opudrovs:
Closes #3268 

- Added passing a MinIO client to the `NewS3LogWriter` function and removed some other params. The difference between the secure and insecure S3 logger is determined by the MinIO client which is passed to `NewS3LogWriter`. So, I removed the `NewInsecureS3LogWriter` function.

-  Added a new `CreateBucket` function which also accepts a MinIO client (generally, the same client which is used for creating the S3 logger.).

- Add constants for log bucket names.

- Added creating a bucket for pod logs in addition to the existing bucket for session logs.

### DIFF
--- a/cmd/gitops/beta/run/cmd.go
+++ b/cmd/gitops/beta/run/cmd.go
@@ -9,7 +9,6 @@ import (
 	"os/signal"
 	"os/user"
 	"path/filepath"
-	"strconv"
 	"strings"
 	"sync/atomic"
 	"syscall"
@@ -658,7 +657,23 @@ func runCommandWithoutSession(cmd *cobra.Command, args []string) error {
 		return fmt.Errorf("unable to install S3 bucket server: %w", err)
 	}
 
-	log, err := logger.NewS3LogWriter(sessionName, fmt.Sprintf("localhost:%d", devBucketHTTPSPort), accessKey, secretKey, cert, log0)
+	minioClient, err := s3.NewMinioClient(fmt.Sprintf("localhost:%d", devBucketHTTPSPort), accessKey, secretKey, cert)
+	if err != nil {
+		cancel()
+		return err
+	}
+
+	if err := logger.CreateBucket(minioClient, logger.SessionLogBucketName); err != nil {
+		cancel()
+		return err
+	}
+
+	if err := logger.CreateBucket(minioClient, logger.PodLogBucketName); err != nil {
+		cancel()
+		return err
+	}
+
+	log, err := logger.NewS3LogWriter(minioClient, sessionName, log0)
 	if err != nil {
 		cancel()
 		return fmt.Errorf("failed creating S3 log writer: %w", err)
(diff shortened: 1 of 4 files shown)


cmd/gitops/beta/run/cmd.go
```diff
@@ -658,7 +658,23 @@ func runCommandWithoutSession(cmd *cobra.Command, args []string) error {
...
+		cancel()
+		return err
+	}
+
+	if err := logger.CreateBucket(minioClient, logger.PodLogBucketName); err != nil {
+		cancel()
+		return err
+	}
+
+	log, err := logger.NewS3LogWriter(minioClient, sessionName, log0)
```
makkes, line 677: ⚠️ Given this function's enormous size already we should move this code into the a separate function.

pkg/logger/s3_log_writer.go
```diff
@@ -18,60 +16,31 @@ type S3LogWriter struct {
...
-
-	if err != nil {
-		return nil, err
-	}
-
-	if err := minioClient.MakeBucket(context.Background(), logBucketName, minio.MakeBucketOptions{}); err != nil {
-		return nil, err
-	}
-
+func NewS3LogWriter(minioClient *minio.Client, id string, log0 Logger) (Logger, error) {
```
makkes, line 26: ✔️ 
```suggestion
func NewS3LogWriter(minioClient *minio.Client, id string, log Logger) (Logger, error) {
```

cmd/gitops/beta/run/cmd.go
```diff
@@ -719,7 +735,7 @@ func runCommandWithoutSession(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
-	minioClient, err := s3.NewMinioClient("localhost:"+strconv.Itoa(int(devBucketHTTPSPort)), accessKey, secretKey, cert)
+	minioClient, err = s3.NewMinioClient("localhost:"+strconv.Itoa(int(devBucketHTTPSPort)), accessKey, secretKey, cert)
```
makkes, line 738: ✔️ to be consistent with line 661 above this should be:
```suggestion
	minioClient, err = s3.NewMinioClient(fmt.Sprintf("localhost:%d", devBucketHTTPSPort), accessKey, secretKey, cert)
```

makkes (commented):
✨ This looks great. Just a couple of minor remarks.

chanwit (approved):

cmd/gitops/beta/run/cmd.go
opudrovs: Yes, the code of GitOps Run can be split further but it is beyond the scope of this PR. We will do it later.

pkg/logger/s3_log_writer.go
opudrovs: It's @chanwit 's naming preference, there are multiple options for this name, so, you could discuss it later.

PR #3308, opened 2023-01-25
Adds more debug logging to help figure out OIDC/RBAC issues

foot:
**What changed?**

- We log out the namespaces we've guessed a user has access
- Include user a bit more in the logs too
- Remove "attempt to read token from auth header" as its not that useful

**Why was this change made?**

Make it easier to debug why things may be missing on the apps/sources etc pages

Adding more logging

Manually

**Release notes**

**Documentation Changes**

### DIFF
--- a/core/clustersmngr/factory.go
+++ b/core/clustersmngr/factory.go
@@ -13,6 +13,7 @@ import (
 	"github.com/hashicorp/go-multierror"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/weaveworks/weave-gitops/core/clustersmngr/cluster"
+	"github.com/weaveworks/weave-gitops/core/logger"
 	"github.com/weaveworks/weave-gitops/core/nsaccess"
 	"github.com/weaveworks/weave-gitops/pkg/featureflags"
 	"github.com/weaveworks/weave-gitops/pkg/server/auth"
@@ -382,6 +383,8 @@ func (cf *clustersManager) updateNamespacesWithClient(ctx context.Context, creat
 		}
 	}
 
+	cf.log.V(logger.LogLevelDebug).Info("Updated namespaces cache", "namespaces", clusterNamespacesLogValue(cf.clustersNamespaces.namespaces))
+
 	opsUpdateNamespaces.Inc()
 
 	return result.ErrorOrNil()
@@ -589,13 +592,17 @@ func (cf *clustersManager) userNsList(ctx context.Context, user *auth.UserPrinci
 		if err != nil {
 			// This may not completely fail the request, e.g. if some of the clusters
 			// are able to respond with their namespaces. So log the error and continue.
-			cf.log.Error(err, "failed updating namespaces from user client")
+			cf.log.Error(err, "error updating namespaces from user client", "user", user)
 		}
 	}
 
 	cf.UpdateUserNamespaces(ctx, user)
 
-	return cf.GetUserNamespaces(user)
+	userNamespaces = cf.GetUserNamespaces(user)
+
+	cf.log.V(logger.LogLevelDebug).Info("Updated namespace access cache for user", "userNamespaces", clusterNamespacesLogValue(userNamespaces), "user", user, "ttl", userNamespaceTTL.String())
+
+	return userNamespaces
 }
 
 func (cf *clustersManager) getOrCreateClient(ctx context.Context, user *auth.UserPrincipal, cluster cluster.Cluster) (client.Client, error) {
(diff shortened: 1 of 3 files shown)


pkg/server/auth/jwt.go
```diff
@@ -74,8 +74,6 @@ func NewJWTAuthorizationHeaderPrincipalGetter(log logr.Logger, verifier tokenVer
 }
 
 func (pg *JWTAuthorizationHeaderPrincipalGetter) Principal(r *http.Request) (*UserPrincipal, error) {
-	pg.log.Info("attempt to read token from auth header")
```
bigkevmcd, line 77: This could be a debug level one, and we should ensure that all the `Principal` implementations have very similar calls so we can see what we're trying to do to authenticate.

core/clustersmngr/factory.go
```diff
@@ -589,13 +592,17 @@ func (cf *clustersManager) userNsList(ctx context.Context, user *auth.UserPrinci
...
+			cf.log.Error(err, "Error updating namespaces from user client", "user", user)
 		}
 	}
 
 	cf.UpdateUserNamespaces(ctx, user)
 
-	return cf.GetUserNamespaces(user)
+	userNamespaces = cf.GetUserNamespaces(user)
+
+	cf.log.V(logger.LogLevelDebug).Info("Updated namespace access cache for user", "userNamespaces", clusterNamespacesLogValue(userNamespaces), "user", user, "ttl", userNamespaceTTL.String())
```
makkes, line 603: This reveals sensitive information (user names) in the logs. I don't know if this is even a problem but wanted to make sure this is a deliberate decision. 🙂

makkes (commented):
Love the additional styling guidelines! ✨ Just a small question.

core/clustersmngr/factory.go
davidstauffer: Interesting points!
Likely the group is the more relevant information for debugging. We should start with the group instead of the user. Besides that, its probably a great idea to have somewhere in the docs a note about what information related to the user we log.

foot: https://github.com/weaveworks/weave-gitops/issues/3318

core/clustersmngr/factory.go
```diff
@@ -631,5 +641,22 @@ func (cf *clustersManager) getOrCreateClient(ctx context.Context, user *auth.Use
 
 	cf.usersClients.Set(user, cluster.GetName(), client)
 
+	log.V(logger.LogLevelDebug).Info("Client created and added to cache")
+
 	return client, nil
 }
+
+// Format the namespaces as a map[clusterName][]namespacesNames
```
bigkevmcd, line 649: Will this scale?

What if you have 100+ namespaces? 1000+ ?

For clarity, 1000+ is not uncommon

pkg/server/auth/auth_test.go
```diff
@@ -385,7 +385,7 @@ func TestUserPrincipal_String(t *testing.T) {
 	// principal is logged out.
 	p := auth.NewUserPrincipal(auth.ID("testing"), auth.Groups([]string{"group1", "group2"}), auth.Token("test-token"))
 
-	if s := p.String(); s != `id="testing" groups=[group1 group2]` {
+	if s := p.String(); s != `id="testing" groups=[group1 group2] tokenLength=10` {
 		t.Fatalf("principal.String() got %s, want %s", s, `id="testing" groups=[group1 group2]`)
```
bigkevmcd, line 389: This hasn't been updated...time to extract to a string and compare :-)

pkg/server/middleware/middleware.go
```diff
@@ -87,7 +87,7 @@ func WithProviderToken(jwtClient auth.JWTClient, h http.Handler, log logr.Logger
 		tokenSlice := strings.Split(tokenStr, "token ")
 
 		if len(tokenSlice) < 2 {
-			log.V(logger.LogLevelDebug).Info("missing or invalid token.")
+			log.V(logger.LogLevelDebug).Info("missing or invalid git-provider-token.")
```
bigkevmcd, line 90: Could we improve this with the guidance above?

https://github.com/weaveworks/weave-gitops/pull/3308/files#diff-1ea5dccd5193cc193367354f068acde964eb10df67d3201f7cd63b0757aaecf5R73

core/clustersmngr/factory.go
foot: Good question. For the scenarios we've been exploring ~10-15 namespaces its been a real help. But for 1000 I guess some part of your logging system might get grumpy.

```
{ "myCluster": { "namespaces": ["foo", "bar"], "totalCount": "3" } # show a sensible N namespaces and report the total count
```

I don't know if the structured logging has nested structures in mind..

The count alone would still be useful.

pkg/server/auth/auth_test.go
foot: Yes! I support we still have `.String()` as a "fallback" (?), to protect the token in case someone accidentally mis-uses and dumps it out, however its a pointer now so I'm not sure if thats a problem anymore. Will test.

We could implement `MarshalLog` to return a struct for logging though.

bigkevmcd: We could implement `MarshalLog` the pointer on the token is to guard against accidental leakage, something printing out a Principal, perhaps bad error handling for example?

foot: > time to extract to a string and compare

Ah I see what you meant, good catch.

LappleApple:
@foot Is this still in progress?

Issue #3313, opened 2023-01-27
Get Session Logs endpoint returns an error instead of log data


opudrovs:
The following error is returned by the Get Session Logs OSS endpoint when testing it from both enterprise (recommended) and directly in OSS, even thought a GitOps Run session is running:

<img width="1527" alt="Screenshot 2023-01-27 at 12 56 32" src="https://user-images.githubusercontent.com/8824708/215085761-85c08143-ec84-4fc1-99db-1df37f0667af.png">

Testing:

The enterprise PR to test the issue:
https://github.com/weaveworks/weave-gitops-enterprise/pull/2283

- To run the session, I use the following command in the podinfo directory:
```
gitops beta run ./deploy/overlays/dev --timeout=3m --port-forward namespace=dev,resource=svc/backend,port=9898:9898
```

The session is displayed in the list of GitOps Run sessions but a vcluster is not displayed in the list of vclusters.

- Please replace the following hardcoded values in the `GitOpsRunDetail` component with appropriate values:

```
const { isLoading, data, error } = useGetLogs({
    namespace,
    sessionId: 'run-master-fdd0a0b7',
    clusterName: 'default/run-master-fdd0a0b7',
  });

  console.log('logs:');
  console.log(data);
```

- To show the `GitOpsRunDetail` component, please run a GitOpsRun session and click on the session's name in the list of sessions.

opudrovs:
I mean except for getting their logs. Links to object detail pages work for the sessions.

opudrovs:
Also added commented out previous args next to current hardcoded args passed to the GetSessionLogsRequest.

chanwit:
do we still need this ticket?
It's not blocking us any more.

chanwit:
^- cc @opudrovs

opudrovs:
But the GetSessionLogs endpoint is not working on the frontend. We are not able to use it.

chanwit:
ok let's keep it.

chanwit:
To set up the dev env for testing the session endpoint:

```
make gitops
make install
gitops beta run --no-session ./deploy/overlays/dev/

# create kind with local registry
make local-kind-cluster-with-registry

# tilt up
tilt up
```

Then go to https://localhost:9001

To test the session_logs endpoint, use:

```
curl ... -d '{"namespace":"flux-system","clusterName":"Default"}'
```

PR #3314, opened 2023-01-27
Return timestamp, source, level and message in GitOps Run log endpoint

### LINKED ISSUE #3280: Return timestamp, source, level and message in GitOps Run log endpoint
Currently, the get-session-logs endpoint returns logs as an array of strings, which is not enough to be used by the front end.

We should return logs as an array of tuples, each of which contains
- timestamp (standard logging format)
- source (the source of the log - it would be "gitops-run-client" to indicate that the log is from the GitOps Run console, or "$namespace/$resource" to indicate that it's from pods of the resource (e.g. deployment, statefuleset, daemonset, etc)
- level (the logs level) they are 3 levels, "info", "warning" and "error".
- message (the log itself)

opudrovs:
Closes #3280

- Added the `LogEntry` structure to the `GetSessionLogsResponse` type.

- Ran `make proto`.

- Added collecting data for `Log` structure fields.

- Added marshaling session log data to JSON.

- Added unmarshaling log entries from JSON in the `GetSessionLogs` endpoint.

- Updated integration tests for `GetSessionLogs`.

Testing:
I just updated integration tests for the `GetSessionLogs` endpoint for now. Unable to test it in the UI because of https://github.com/weaveworks/weave-gitops/issues/3313 But for the scope of this ticket automated tests should work fine.


pkg/logger/s3_log_writer.go
```diff
@@ -36,13 +40,38 @@ func CreateBucket(minioClient *minio.Client, bucketName string) error {
...
+	if len(msg) > 0 {
+		if strings.HasPrefix(msg, "✗ ") {
+			level = "error"
+		} else if strings.HasPrefix(msg, "⚠️ ") {
+			level = "warning"
+		}
+	}
+
+	result := &pb.LogEntry{
+		Timestamp: now.Format(time.RFC3339),
```
chanwit, line 56: :+1:

pkg/logger/s3_log_writer.go
```diff
@@ -36,13 +40,38 @@ func CreateBucket(minioClient *minio.Client, bucketName string) error {
 }
 
 func (l *S3LogWriter) putLog(msg string) {
+	now := time.Now()
+
+	level := "info"
+
+	if len(msg) > 0 {
+		if strings.HasPrefix(msg, "✗ ") {
```
chanwit, line 48:
```suggestion
		if strings.HasPrefix(msg, "✗") {
```
without the space would be just fine :-)

pkg/logger/s3_log_writer.go
```diff
@@ -18,6 +21,7 @@ type S3LogWriter struct {
 
 const SessionLogBucketName = "gitops-run-logs"
 const PodLogBucketName = "pod-logs"
+const SessionLogSource = "gitops-run-client"
```
chanwit, line 24: :+1:

chanwit (commented):
LGTM in general!

Will do the second round when it's ready.

pkg/logger/s3_log_writer.go
opudrovs: Removed the space :)

opudrovs: And in the other place too.

pkg/logger/s3_log_writer.go
opudrovs: I wasn't sure which timestamp format you need for this endpoint, so copied it from the timestamp field in another endpoint. Glad it is OK.

chanwit (approved):

PR #3315, opened 2023-01-27
Add initial (naive) BitBucket Server support

jpellizzari:
Part of https://github.com/weaveworks/weave-gitops-enterprise/issues/2271

Adds repo URL parsing support for BitBucket using a simple path match.

### DIFF
--- a/pkg/gitproviders/factory.go
+++ b/pkg/gitproviders/factory.go
@@ -12,9 +12,10 @@ import (
 type GitProviderName string
 
 const (
-	GitProviderGitHub GitProviderName = "github"
-	GitProviderGitLab GitProviderName = "gitlab"
-	tokenTypeOauth    string          = "oauth2"
+	GitProviderGitHub          GitProviderName = "github"
+	GitProviderGitLab          GitProviderName = "gitlab"
+	GitProviderBitBucketServer GitProviderName = "bitbucket-server"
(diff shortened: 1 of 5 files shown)


jpellizzari:
@yiannistri It wasn't immediately obvious how we can detect a bitbucket provider URL using Viper or env vars. Any ideas there?

yiannistri:
@jpellizzari  we don't need to detect the provider from the URL, there's a map[string]string that gets populated from a Kubernetes secret that maps the URL to the provider type, for example in the `detectGitProviderFromURL` function the `gitHostTypes` parameter will be
```
gitHostTypes[bitbucket.acme.com: bitbucket-server
```
So we can use that value to determine the provider.

yiannistri:
There's a reference to this in the docs but it's not super clear https://github.com/weaveworks/weave-gitops/blob/main/website/docs/installation/weave-gitops-enterprise/index.mdx?plain=1#L233-L236

jpellizzari:
> @jpellizzari we don't need to detect the provider from the URL, there's a map[string]string that gets populated from a Kubernetes secret that maps the URL to the provider type, for example in the `detectGitProviderFromURL` function the `gitHostTypes` parameter will be

Fixed! PTAL

yiannistri (approved):

PR #3316, opened 2023-01-30
Relax sub-sub-group constraint on getting owner from URL

foot:
**What changed?**

Relaxes the subsubgroup constraint on gitlab, go-git-providers does support subgroup/subsubgroup etc with some small tweaks. https://github.com/fluxcd/go-git-providers/issues/183

**Why was this change made?**

Some users use gitlab with subgroups and they cannot use WGE right now.

Removing the constraint

- Unit tests
- Manual testing

**Release notes**

**Documentation Changes**


pkg/gitproviders/repo_url.go
```diff
@@ -87,23 +87,11 @@ func (n RepoURL) Protocol() RepositoryURLProtocol {
...
-			return "", fmt.Errorf("a subgroup in a subgroup is not currently supported")
-		}
-
-		if len(parts) > 2 {
-			return parts[0] + "/" + parts[1], nil
-		}
-	}
-
-	return parts[0], nil
+	return strings.Join(parts[:len(parts)-1], "/"), nil
```
ranatrk, line 94: could someone provide a url with `len(parts)>4`? in that case it shouldn't be valid right?

foot: Great question, I'm not sure what the valid max depth of subgroups are. I've tested at 3, we should be able to handle however many gitlab can, which doesn't seem to be documented afaict.

If there is some limit then an error will appear when we attempt to create the PR, the user can fix their URL at that point I guess.

ranatrk: apparently up to 20 subgroups can be created according to [gitlab](https://about.gitlab.com/solutions/subgroups/#:~:text=You%20can%20create%20up%20to,an%20incredible%20level%20of%20flexibility.)

ranatrk (approved):

PR #3322, opened 2023-01-31
docs: Configure sidebars from a single place

Callisto13:
This makes it easier to both change/control sidebar ordering, and to use some of the more advanced Docusaurus sidebar features.

There are no content/ordering/docs changes to review here, simply a bit of tidying up before I PR actual changes which would have been very messy if I lumped it all together.


website/versioned_sidebars/version-0.15.0-sidebars.json
```diff
@@ -4,5 +4,11 @@
       "type": "autogenerated",
       "dirName": "."
     }
+  ],
+  "docs": [
+    {
+      "type": "autogenerated",
+      "dirName": "."
```
Callisto13, line 11: I have no idea whether this is strictly necessary, but my local build was unhappy if I clicked on certain pages and this was not included so perhaps the real one will be too? Will I need to backfill all the others?

chanwit (approved):
using the sidebar file makes sense to me.
thank you @Callisto13

PR #3327, opened 2023-01-31
Add gitops run command as an alias to gitops beta run

### LINKED ISSUE #3317: Add `gitops run` command as an alias to `gitops beta run`
**Short description**

As a GitOps Run user I want to use it with `gitops run` so that I don't have to provide the `beta` sub-command.

**Acceptance criteria**

- [ ] `gitops run` and `gitops beta run` work the exact same way with one effectively being an alias of the other.

**Context**

GitOps Run will still be considered beta from a product perspective after this story has been implemented. This is just one step in the direction of moving it out of beta later on.

opudrovs:
Closes #3317 

- Added the `run` command to  the root (`gitops`) command. So, calls to `gitops beta run` and `gitops run` will work exactly the same.

Notes:
The only limitation: usage examples in help for `gitops run` will display `gitops beta run`, because they are hardcoded strings, but for our purposes it should work. We can update help after we move GitOps Run our of beta.

Testing that both commands work the same:

`gitops run`

Screenshot 2023-01-31 at 14 09 31

`gitops beta run`

Screenshot 2023-01-31 at 14 16 27

### DIFF
--- a/cmd/gitops/root/cmd.go
+++ b/cmd/gitops/root/cmd.go
@@ -18,6 +18,7 @@ import (
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
 	"github.com/weaveworks/weave-gitops/cmd/gitops/beta"
+	"github.com/weaveworks/weave-gitops/cmd/gitops/beta/run"
 	"github.com/weaveworks/weave-gitops/cmd/gitops/check"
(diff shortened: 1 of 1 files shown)


chanwit (approved):

makkes (approved):

makkes:
✨ the most concise implementation!

PR #3330, opened 2023-01-31
docs: Add landing page with features highlight

### LINKED ISSUE #2984: docs/intro: visually represent key fatures
It'd be great to show representations of the key features of Weave GitOps visually in the first part of the landing page.

The docusaurus page does this well

grafik

Relevant code:

- https://github.com/facebook/docusaurus/blob/e519858ea7c88cd3d16410dc6184f0bd6f491d07/website/src/pages/index.tsx
- https://github.com/facebook/docusaurus/blob/e519858ea7c88cd3d16410dc6184f0bd6f491d07/website/src/data/features.tsx

A simpler approach could be to use Font Awesome icons like on fluxcd.io

grafik

In any case it'd lighten the landing page up some and make it easier to parse what Weave GitOps is about.

Callisto13:
A flashier landing page gives us a place to hook people in with a demo and features highlights.

The features section links to both OSS and EE respectively.

Closes #2984 

Notes on this:
- Edit: changed to use another colour from our logo now after I found out what it was ~I do not love the "built-in" landing page style which Docusaurus gives you (usually I don't mind it, but I hate the colour orange and we are somewhat tied by our logo scheme). I would be up for restyling if others agree, someone else would need to do that work though as I cannot React for my life.~
- The tagline text under the title is not fixed, suggestions welcome. 
- The text I have added under the banner is optional. I can't decide whether I like it or not. I am pretty sure I do not like the font.
- Edit: changed these, all fine in dark mode ~The feature icons are free ones I got from FontAwesome to demonstrate what the landing would look like. If we have an FA pro account I will change these to ones which work better in dark mode.~
- Could someone check that I have got the features under the correct tiers? I was going based on which docs are tagged, but I have found that some are not-tagged but should be. 

Edit: Changed the banner colour to something less garish:

https://user-images.githubusercontent.com/8898786/220344080-6f04b423-bcc5-4cc5-9df3-a8d4d4908be4.mov


Callisto13:
cc @joe-terrible

dholbach:
Thanks for taking on this work - this already looks much better than what's currently up. :-)

makkes:
Love it. Just some remarks:

- The spacing between the header and the content is too narrow:  
- The link in "Available on all Tiers" leads to the intro page whereas I would expect to be presented with a list of the different available tiers when clicking on a link named "all Tiers".
- The link in "Available only on the Enterprise Tier" leads to a 404.

Callisto13:
> The link in "Available on all Tiers" leads to the intro page whereas I would expect to be presented with a list of the different available tiers when clicking on a link named "all Tiers".

good point 🤔 i will think of how that page can look

> The link in "Available only on the Enterprise Tier" leads to a 404.

Sound about right 😂

Callisto13:
Actually I think half that (now that i have figured out the css shorthand)

Callisto13:
> The link in "Available only on the Enterprise Tier" leads to a 404.

Ah actually I know why. The landing page is not versioned, and when you click off it you will be taken to 0.16.0, which does not have an ee page yet as i am not copying it over until the ee-page pr is merged

makkes:
> image (2)
> 
> Too much spacing or just about right?

I would try to make the lower spacing the same height as the upper one.

makkes:
Not that I have any knowledge of UX design btw.

Callisto13:
@davidstauffer @joe-terrible do we want to assign a designer to this piece?

Callisto13:
@makkes @JamWils @joe-terrible @dholbach 

Updated with:
- better colour scheme
- icons which also look good in dark mode
- better spacing between banner and text

if this is okay to have while we wait for a designer i will merge 👍 

https://user-images.githubusercontent.com/8898786/220345023-5ae4fecd-a004-4f5b-b057-c59730586d32.mov

makkes (approved):
Checked it out and it looks glamorous! Good to merge from my pov.

JamWils:
This is so great.  I'll queue up a task on the design board for icons, but again this is a million times better compared to what we had before.

PR #3332, opened 2023-01-31
docs: Add Helm and CLI reference to nav bar

### LINKED ISSUE #2684: Set non-versioned documents as pages instead of docs in the navigation bar with zero maintainance effort
Pages

At the very least if we do this, I would probably convert Help and Support , Feedback and Telemetry and maybe Introduction into pages rather than docs - and be consistent with Security (edited) 

The trick
- not versioned by weave gitops version
- no maintenance effort

### LINKED ISSUE #3323: Make the helm chart reference easier to find
Some feedback has mentioned that it is buried in the docs

Callisto13:
Feedback has been that these are hard to find. Tilt has something like this in their docs which does make it easy to get to.

I have also moved the Help and Feedback pages out of versioned docs and into the navbar.

The navbar is a little busy now, though, but I am planning to tackle that in my next PR.

Closes #2684 
Closes #3323 

Screenshot 2023-01-31 at 16 40 36


website/docusaurus.config.js
```diff
@@ -53,6 +53,22 @@ module.exports = {
           position: "left",
           label: "Getting Started",
         },
+        {
+          type: 'docSidebar',
+          position: 'left',
+          label: 'Helmchart & CLI',
```
makkes, line 59:
```suggestion
          label: 'Helm Chart & CLI',
```

makkes (commented):
✨ only one nit

makkes (approved):

PR #3328, opened 2023-01-31
docs: Add a dedicated Enterprise page

Callisto13:
Or I guess bring it back since I feel it existed before?

The purpose of this page is to highlight the additional features offered by enterprise. It also gives us a place to slot the confusing `Releases` tab under.

Also:
- Jazz up the intro and install pages, making the differences between tiers very clear and adding a bit of colour to break up the text.
- Add some missing `enterprise` tags.

Part of https://github.com/weaveworks/weave-gitops-enterprise/issues/2274

https://user-images.githubusercontent.com/8898786/215781385-45b37424-fbfd-4e83-9c45-676c5a67c961.mov

Closes #3135 #3324 https://github.com/weaveworks/weave-gitops-enterprise/issues/2302

PR #3339, opened 2023-02-01
docs: Tidy the navbar

Callisto13:
And break up 'Getting Started'. This layout is inspired by Tilt's docs, which are often mentioned as Good. The sidebar here is set to always expanded so they are always visible by default.

image (27)

This follows changes in https://github.com/weaveworks/weave-gitops/pull/3332 which crowded the navbar and can go in after https://github.com/weaveworks/weave-gitops/pull/3328

The navbar is not versionable which is a right pain, so either we never touch the navbar, or we live with it 🤷‍♀️ .


website/sidebars.js
```diff
@@ -11,31 +11,58 @@
...
+          collapsed: false,
+          link: {
+            type: 'doc',
+            id: 'getting-started/intro',
+          },
+          items: [
+            {
+              type: 'doc',
+              id: 'installation/index',
+              label: '0. Install Weave GitOps',
```
makkes, line 58: I know we're all nerds but are our customers fine with starting to count from 0?

website/sidebars.js
```diff
@@ -11,31 +11,58 @@
...
       items: [
-        'installation/weave-gitops',
         {
           type: 'category',
-          label: 'Weave GitOps Enterprise',
+          label: 'Installation',
           link: {
             type: 'doc',
-            id: 'installation/weave-gitops-enterprise/index',
+            id: 'installation/index',
```
makkes, line 28: Putting my comment here as the page in question isn't changed as part of this PR: The page's title should follow a similar pattern as the "part 1" and "part 2" pages and have a title preceded by "Part 0: ".

Callisto13: it does and doesn't make sense to do that.

the installation section is still separate from the getting started section, the getting started just links back to installation in its first.

if i were to pull installation under getting started, then yes having it be part 0 makes sense (i can even 1-index more naturally there). but if we want installation to remain separate (and there is a good case for doing that), then it makes less sense to have titled with a number for people who are coming just to the installation bit

Callisto13: image (5)

Looks like this now

makkes (approved):
Love it! ✨

PR #3345, opened 2023-02-01
Fix aud verification

bigkevmcd:
**What changed?**
This adds support for verifying configured audiences to ensure that a JWT comes from an approved service.

**Why was this change made?**
To improve the security

Allow configuration of the audiences

Tests and manual validation

**Release notes**
Can now restrict the set of allowed audiences for JWTs

**Documentation Changes**

### DIFF
--- a/pkg/server/auth/jwt.go
+++ b/pkg/server/auth/jwt.go
@@ -28,20 +28,20 @@ type tokenVerifier interface {
 // JWTCookiePrincipalGetter inspects a cookie for a JWT token
 // and returns a principal object.
 type JWTCookiePrincipalGetter struct {
-	log          logr.Logger
-	verifier     tokenVerifier
-	cookieName   string
-	claimsConfig *ClaimsConfig
+	log        logr.Logger
+	verifier   tokenVerifier
+	cookieName string
+	oidcConfig OIDCConfig
 }
 
 // NewJWTCookiePrincipalGetter looks for a cookie in the provided name and
 // treats that as a JWT token that can be decoded to a Principal.
-func NewJWTCookiePrincipalGetter(log logr.Logger, verifier tokenVerifier, cookieName string, config *ClaimsConfig) PrincipalGetter {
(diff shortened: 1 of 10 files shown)


pkg/server/auth/jwt.go
```diff
@@ -99,13 +101,22 @@ func extractToken(s string) string {
...
 
-	return cc.PrincipalFromClaims(token)
+	if oc.Audiences != nil {
+		for _, aud := range oc.Audiences {
+			if !stringsContain(token.Audience, aud) {
+				// Do not add the valid audiences to this it would be exposing data
+				return nil, fmt.Errorf("token audience %q is not valid for this request", strings.Join(token.Audience, ", "))
+			}
+		}
+	}
```
foot, line 117: I'm surprised the verfier.Verify doesn't do this, I guess thats more of an OIDC specific concern than JWT?

pkg/server/auth/server.go
```diff
@@ -128,6 +131,11 @@ func NewOIDCConfigFromSecret(secret corev1.Secret) OIDCConfig {
 
 	cfg.Scopes = scopes
 
+	audiences := splitAndTrim(string(secret.Data["audiences"]))
+	if len(audiences) > 0 {
+		cfg.Audiences = audiences
```
foot, line 136: Should it be the clientId by default?

pkg/server/auth/jwt.go
bigkevmcd: Yes, it's at the OIDC/Application layer, rather than at the JWT, it's not whether or not the token is valid, it's where it originates that's the issue.

pkg/server/auth/server.go
foot: (Alright, this piece of code is perhaps not the place to check for default though)

bigkevmcd: actually, yes, that's a good idea!

foot (approved):
✨

PR #3355, opened 2023-02-02
Improve formatting of log message timestamp in the log UI

opudrovs:
Part of https://github.com/weaveworks/weave-gitops-enterprise/issues/2319

- Add the `formatLogTimestamp` function to format log timestamp with a custom format string with a UTC timezone offset.

- Add unit tests for `formatLogTimestamp`.

Notes:
- There is a technical problem with importing the luxon package directly into enterprise (a type-related error message which is unresolvable for now, so I decided to export this function from OSS and import in the other PR to enterprise, because luxon is already imported in OSS).

Design from Figma:
Screenshot 2023-02-02 at 23 54 04

Results of formatting with this function (in enterprise UI), the only difference with Figma is, as @josefaworks  suggested, it should display local UI user's time, not time in +0 UTC timezone (we can change exact UTC format later):

Screenshot 2023-02-02 at 23 37 50

- I also applied the current timestamp formatter to the header timestamp, because in Figma it shows some short time format and this needs to be refined, for now the long timestamp should work.

- The `zone` function param is used for unit tests only, otherwise the string would be formatted to local time which would be different on a developer machine and on server (so the tests would fail). 

- We could in principle remove zero UTC from timestamp but in this case the `UTC` strings will not be roughly at the same level as UTC in other timestamps with non-zero UTC, so we can remove it later if needed.

### DIFF
--- a/ui/lib/utils.ts
+++ b/ui/lib/utils.ts
@@ -1,4 +1,5 @@
 import _ from "lodash";
+import { DateTime } from "luxon";
 import { toast } from "react-toastify";
 import { computeReady, ReadyType } from "../components/KubeStatusIndicator";
 import { AppVersion, repoUrl } from "../components/Version";
@@ -176,4 +177,27 @@ export function getAppVersion(
     versionText,
     versionHref,
   };
+}
+
+// formatLogTimestamp formats a timestamp string in the RFC3339 format
+// to a human-readable format with UTC offset.
+// If the timestamp is undefined or an empty string, it returns "-".
+export function formatLogTimestamp(timestamp?: string, zone?: string): string {
+  if (!timestamp) {
+    return "-";
+  }
+
+  let dt = DateTime.fromISO(timestamp);
+
+  if (zone) {
+    dt = dt.setZone(zone);
+  }
+
+  let formattedTimestamp = `${dt.toFormat("yyyy-LL-dd HH:MM:ss 'UTC'Z")}`;
+
+  if (dt.offset === 0) {
+    formattedTimestamp = formattedTimestamp.replace("UTC+0", "UTC");
+  }
+
+  return formattedTimestamp;
 }
--- a/ui/index.ts
+++ b/ui/index.ts
@@ -67,7 +67,12 @@ import {
 } from "./lib/objects";
 import { muiTheme, theme } from "./lib/theme";
 import { V2Routes } from "./lib/types";
-import { isAllowedLink, poller, statusSortHelper } from "./lib/utils";
+import {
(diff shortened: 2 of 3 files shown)


ui/lib/utils.ts
```diff
@@ -177,3 +178,20 @@ export function getAppVersion(
...
+    return "-";
+  }
+
+  let dt = DateTime.fromISO(timestamp);
+
+  if (zone) {
+    dt = dt.setZone(zone);
+  }
+
+  return `${dt.toFormat("yyyy-LL-dd HH:MM:ss ZZ")} UTC`;
```
makkes, line 196: I want to point out that printing the offset **and** "UTC" might come across as confusing. Usually you either print the offset or the **actual** timezone of the time displayed. [Here](https://pkg.go.dev/time#pkg-constants) is a list of common time/date patterns, none of which suggest printing both, the offset and the timezone, especially not a timezone that's different from the offset.

Suggestion:

```suggestion
  return `${dt.toFormat("yyyy-LL-dd HH:MM:ss ZZ")}`;
```

opudrovs: Those patterns look somewhat unusual and not exhaustive, at least I don't think it's what is usually used on the frontend as human-readable formatted time. (Actually, I would not trust a list in Golang docs as a good example of time formatted for friendly user experience.)

In the formatted string +05:00 UTC was, strictly speaking, supposed to specify an offset, I can rename the param zone to offset or format it as "UTC+5" or "UTC+05:00", what is I think a more standard way of specifying offset.

In Figma the design looked like that:

Screenshot 2023-02-02 at 23 54 04

But we agreed with Josefa, we want local time instead of time in UTC time zone. So, it's either specifying offset (with or without UTC) or specifying a local timezone (like time with PST, for example, which I would need to check how to implement with Luxon).

I can ask Josefa what she would prefer.

opudrovs: Wrote to Josefa about preferred time format. A useful link from Luxon docs on timezones:
https://github.com/moment/luxon/blob/master/docs/zones.md

opudrovs: No, in Luxon API name for passed "UTC+3" (whatever) string is zone, so it's a fixed-offset zone and the param name will have to stay zone.

opudrovs:
After discussing it with @josefaworks , switched to a time format with short offset:
```
2023-01-31 19:01:56 UTC+1
```

And for `UTC+0` (which Luxon uses internally) I am displaying `UTC`.

joshri (approved):

PR #3357, opened 2023-02-02
Handle undefined icon type

jpellizzari:
Kept getting this error:

Screenshot from 2023-02-02 11-28-50


joshri (commented):
What component is this happening in? I thought the whole point of IconType being required is that this would be avoided? Maybe it's in a KubeStatusIndicator with special conditions?

joshri (approved):

PR #3363, opened 2023-02-03
enhance get session logs to support pod logs, filters, and return logging sources

### LINKED ISSUE #3269: Write an API that allows users to retrieve logs by pod name and namespace

### LINKED ISSUE #3279: Add endpoint to fetch GitOps Run sources of logs
The task is to create an endpoint function that will populate a list of sources of logs to be used in the UI. The function will take in a token as an argument and will read a page of logs, extract the values from the log name, remove any duplicates, and cache the results. The cached results will be returned in a list, with "gitops-run-client" always prepended as the first element. Additionally, the function will also return the next token for pagination if one is present.

# Implementation Details

- The function will take in a token as an argument. If the token is empty, it will be assumed that it is the first page of logs.
- The function will then read a page of logs using the token.
- The function will extract the values from the log name and remove any duplicates.
- The function will cache the results.
- The function will prepend "gitops-run-client" as the first element in the list of cached results.
- The function will return the list of cached values, along with the next token for pagination if one is present.

# Acceptance Criteria

- The function must be named "PopulateLogSources."
- The function must take in a token as an argument and return a list of sources of logs.
- The function must extract the values from the log name and remove any duplicates.
- The function must cache the results.
- The function must always prepend "gitops-run-client" as the first element in the list of cached results.
- The function must return the next token for pagination if one is present.

### LINKED ISSUE #3344: Return pod logs from GitOps Run session from GetSessionLogs function

chanwit:
Fixes #3269 
Fixes #3279
Fixes #3344 

**TODO**
- [x] Get pod logs
- [x] Merged tokens
- [x] Merged logs
- [x] Sorted logs by timestamp
- [x] Extract logging source information
- [x] LogSource and LogLevel Filters
- [x] Return Log sources

## Enhance GetSessionLogs

The change of this PR is for a protobuf file in the API core of the project. The file contains the definitions for communication between client and server in the API.

In the `GetSessionLogsRequest` message, two new fields have been added:

- logSourceFilter (string)
- logLevelFilter (string)

In the `GetSessionLogsResponse` message, a new repeated field has been added:

- logSources (repeated string)

The purpose of these changes is to allow filtering of log entries based on log source and log level. The new logSources field in `GetSessionLogsResponse` returns a list of available log sources.

## Implement get pod logs

This PR modifies the function GetSessionLogs in the coreServer struct. The function is used to retrieve logs from either GitOps run logs or pod logs. The changes in this diff do the following:

1. Change the naming of the `virtualClusterName` variable to `clusterName` and handle a special case where the session name is equal to "no-session". If it is, the cluster name is set to the default cluster name.

1. Instead of returning an error on failure, the function returns a `pb.GetSessionLogsResponse` with an error message in the `Error` field.

1. Add the ability to filter logs based on the log source and log level filter. The logs can now be filtered by these values.

1. Add the ability to retrieve logs from the GitOps run logs. If the log source filter is equal to the session log source or empty, the logs will be retrieved from the GitOps run logs.

1. Add the ability to retrieve logs from the pod logs. If the log source filter is not empty, the logs will be retrieved from the pod logs.

1. Sort the logs based on the timestamp before returning the logs in the `pb.GetSessionLogsResponse`.

## Detect log level

This PR is adding a new function called `detectLogLevel` that takes a string as a parameter (representing a log message). The function uses regular expressions to determine if the log message contains keywords indicating the log level, such as "error" or "warning". If it finds a match, it returns a string representation of the log level, either "error", "warn", or "info". If it doesn't find a match, it returns "info" as the default log level.

## Improve getPodLogs

This PR introduces a new struct named `PodLog` that represents the logs data. The struct contains the log data such as `date`, `time`, `log`, `msg`, and `Kubernetes` `metadata` such as pod name, namespace name, pod ID, labels, annotations, host, container name, docker ID, container hash, and container image.

The function `getPodLogs` was modified to read logs data using the new `PodLog` struct. The function uses the Minio client to retrieve log data from an S3 bucket. The logs data is read from the S3 bucket and is then split into individual log entries. The log entries are then transformed into the `PodLog` struct.

If the log entry is empty or has an inner message of an empty string, it is ignored. The log level of each log entry is detected using the `detectLogLevel` function. If a log source filter or log level filter is provided, the logs that do not match the filter are ignored.

The final result of the function is a slice of log entries represented as `pb.LogEntry` objects and a last token representing the last key processed in the S3 bucket.

## Fluent-bit in vcluster

We have encountered a problem with Fluent-Bit in our vcluster environment. The issue is that there is no log directory in vcluster and Fluent-Bit is retrieving information directly from the host's /var/log/containers, which causes a mismatch in container names.

As a result, Fluent-Bit is trying to access information from the vcluster control-plane using the wrong container names (the names at the host level, not the vcluster level). We need to find a solution to this problem to ensure accurate logging in our vcluster environment.

Fortunately, from vcluster v0.13 and later, there is a component called HostPath Mapper to help correct logging information from host so that we can collect them correctly from within vclusters.

For more information.
https://www.vcluster.com/docs/operator/monitoring-logging#logging

## Improve --dashboardImage

This PR also modifies the function `makeValues` in the file `install_dashboard.go`. The main changes in this diff are the addition of a new function `parseImageRepository` and modification of the `makeValues` function to use this new function.

The `parseImageRepository` function takes an input string in the format "repository/image:tag" or "image:tag". The function splits the input string into its component parts (repository, image, and tag) and returns them. If the input string is not in the correct format, the function returns an error.

The `makeValues` function has been updated to call the new `parseImageRepository` function when processing the `dashboardImage` argument. The `dashboardImage` argument is used to specify the Docker image for the GitOps Dashboard. If `dashboardImage` is not an empty string, the function creates a `map[string]interface{}` and sets the image key to a map that contains the repository, image, and tag extracted from the `dashboardImage` argument.

This PR provides a more robust and flexible way of parsing the `dashboardImage` argument and ensures that the input string is in the correct format.

## How to test this PR

- start a WGE instance
- update Weave GitOps dependencies to this branch (`BRANCH=enhance-get-session-logs`)
- go to a local directory of any Git repo you have
- start a GitOps Run session `gitops beta run --skip-dashboard-install ./manifests`
- logs in the GitOps Run UI would appear


opudrovs:
Great PR description @chanwit , thank you! ✨ This makes reviewing a complex PR so much easier! 💖

core/server/session_logs.go
```diff
@@ -114,35 +126,191 @@ func (cs *coreServer) GetSessionLogs(ctx context.Context, msg *pb.GetSessionLogs
...
-		return nil, err
+		return &pb.GetSessionLogsResponse{Error: err.Error()}, err
+	}
+
+	var (
+		logEntries []*pb.LogEntry
+		firstToken string
+	)
+
+	if msg.GetLogSourceFilter() == "" || msg.GetLogSourceFilter() == logger.SessionLogSource {
```
opudrovs, line 137: ❓  Can we assign `msg.GetLogSourceFilter()` to a constant for these checks in order not to make this call twice?

core/server/session_logs.go
```diff
@@ -114,35 +126,191 @@ func (cs *coreServer) GetSessionLogs(ctx context.Context, msg *pb.GetSessionLogs
...
+			}
+
+			podLog := &PodLog{}
+			if err := json.Unmarshal([]byte(message), podLog); err != nil {
+				return nil, "", nil, err
+			}
+
+			innerMessage := podLog.Log
+			if innerMessage == "" {
+				innerMessage = podLog.Msg
```
opudrovs, line 259: ❔ What does `podLog.Msg` contain? Is it some sort of a system/error message?

core/server/session_logs.go
```diff
@@ -114,35 +126,191 @@ func (cs *coreServer) GetSessionLogs(ctx context.Context, msg *pb.GetSessionLogs
...
-		logEntry := &pb.LogEntry{}
-		err = json.Unmarshal([]byte(log), &logEntry)
+	return &pb.GetSessionLogsResponse{
+		Logs:       logEntries,
+		NextToken:  firstToken + "," + secondToken,
+		LogSources: append([]string{logger.SessionLogSource}, logSources...),
+	}, nil
+}
+
+func detectLogLevel(message string) string {
```
opudrovs, line 181: ❔ In regular pod logs, obtained with `kubectl logs`, I see a log `level` field:

```
{"level":"info","ts":"2023-02-08T22:43:02.558Z","caller":"podinfo/main.go:151","msg":"Starting podinfo","version":"6.3.3","revision":"ae3fe3da989f5c0173ad9477975cca185b280455","port":"9898"}
{"level":"info","ts":"2023-02-08T22:43:02.559Z","caller":"api/server.go:223","msg":"Starting HTTP Server.","addr":":9898"}
```

Is something like that present in logs, collected by Fluent-Bit, and can we use it in principle instead of this custom function? Or do we need to use our own log levels which do not match default pod log levels?

opudrovs:
While testing it, I ran into some issues, can re-test tomorrow.

Everything is green, sessions are discovered by the UI, as expected:

Screenshot 2023-02-09 at 01 15 36

Screenshot 2023-02-09 at 01 16 13

When I go to the session details page (or sometimes when just clicking on a session name in the list of session), the UI becomes unresponsive for some time with loading indicator displayed in the tab.

Then I on the details page I see logs appearing for a moment and then disappearing (replaced by the no logs found message again):

Screenshot 2023-02-09 at 01 18 02

Screenshot 2023-02-09 at 01 18 29

Did you by chance change the polling logic (on the backend or the frontend), I mean how the token should be used?

And besides maybe we need to display a loading indicator while logs are loading, but it's for the future.

opudrovs:
Let me re-test it with everything rebuilt from the start.

opudrovs:
Sorry, false alarm. Rebuild everything, it's working fine now.

opudrovs (approved):
Great job! 🎉 Tested it with the following commands

```
gitops beta run ./deploy/overlays/dev --timeout=3m --port-forward namespace=dev,resource=svc/backend,port=9898:9898 --skip-dashboard-install
gitops beta run ./deploy/overlays/dev --timeout=3m --port-forward namespace=dev,resource=svc/backend,port=9898:9898 --session-name=session2 --skip-dashboard-install
```

Everything is working as expected! ✨

core/server/session_logs.go
chanwit: I would have done that. Thank you Olga! Will fix it.

core/server/session_logs.go
chanwit: Different log formats have different ways of storing their inner messages. Some programs store inner messages in "log" field, some store them in the "msg" field. 

> How does the inner message in the logs work?
We do not and will not know before hand what field like this would appear in the future. 

If something is broken, we have to add support more field names here.

core/server/session_logs.go
chanwit: YES!

I would have detected it with this before falling back to the detectLogLevel function.

core/server/session_logs.go
chanwit: OK, `const x = msg.GetLogSourceFilter()` didn't work. 
This kind of getter would get optimized nicely by Go. I think we don't have to worry much about it.

opudrovs: Sorry, wrote const but meant smth. like using a regular var `x := msg.GetLogSourceFilter()` (well, it was late at night while I was reviewing it 😅  ). But it's of very minor concern.

core/server/session_logs.go
opudrovs: Got it, thanks!

PR #3366, opened 2023-02-06
Initial import of the GitOpsSets user-guide

foot:
**What changed?**

Adds user-guide for the EE gitopssets feature.

**Why was this change made?**

To provide some public facing docs for this

- Duplicating the internal GitOpsSets guide here
- Adding a big warning up the top about API / security model changes coming soon

**Release notes**

**Documentation Changes**


website/docs/gitopssets/gitopssets.mdx
```diff
@@ -0,0 +1,377 @@
...
+## Generators
+
+We currently provide these generators:
+
+ * list
+ * pullRequests
+ * gitRepository
+ * matrix
+
+## List generator
```
Callisto13, line 158:
```suggestion
### List generator
```
Guessing this is supposed to have same weight as the next one?

Callisto13 (commented):
Just one note.

Because of the title "Initial...": is there more coming for Gitopssets? I ask because there is nothing here which explains what they are for/the problem they solve. For each generator it is also not clear what the expected outcome is, the effect on the cluster(s).

davidstauffer (approved):

Issue #3337, opened 2023-02-01
Improve detection of OSS dashboard


opudrovs:
Right now we are using a HelmChart to detect if OSS dashboard is installed in the host cluster or vcluster. Switch to using a HelmRelease to detect if the OSS dashboard is installed (via `Spec.Chart.Spec.Chart` string which in the OSS dashboard HelmRelease should be equal to `weave-gitops`).

As discussed, GitOps Run should only detect dashboards installed by following the official docs for Weave Gitops at
https://docs.gitops.weave.works/docs/installation/weave-gitops/

**Acceptance Criteria:**

- When running GitOps Run, the presence or absence of the OSS dashboard, installed via OSS HelpChart (as recommended in the docs), should be detected successfully on both regular Kind and enterprise cluster (created with the `./tools/reboot.sh` script in enterprise). The latter is checked to be on the safe side, normally we are not supposed to install the OSS dashboard to the enterprise cluster.

To install the OSS dashboard via a HelmChart, use the following instructions:
https://docs.gitops.weave.works/docs/installation/weave-gitops/#deploying-weave-gitops

opudrovs:
Detection of the enterprise dashboard will be added in https://github.com/weaveworks/weave-gitops/issues/2891

Issue #2891, opened 2022-10-17
GitOps Run cannot detect Enterprise Dashboard


JamWils:
As an enterprise user when I am using `gitops run` it is always asking me if I want to install the dashboard.  This feature should be able to detect when I have the enterprise dashboard installed and not ask this question.

opudrovs:
Created a prerequisite issue https://github.com/weaveworks/weave-gitops/issues/3337

Will use the same code/principle to detect the enterprise dashboard.

Moved the second part of this ticket

>It should port-forward the enterprise dashboard, so it's consistent with what happens when we find an installed OSS dashboard.

to 

https://github.com/weaveworks/weave-gitops/issues/3338

because it requires modifying the port-forwarding functionality which is separate from dashboard detection.

PR #3371, opened 2023-02-06
Improve detection of OSS dashboard

opudrovs:
Closes https://github.com/weaveworks/weave-gitops/issues/3337

- Switched from using a HelmChart to detect OSS dashboard to using a HelmRelease.

- Added a `helmChartName` parameter to `IsDashboardInstalled` and `getDashboardHelmRelease` functions (with the hope to reuse the dashboard detection code in https://github.com/weaveworks/weave-gitops/issues/2891 ).

- Removed unused function parameters.

- Updated tests for `getDashboardHelmRelease`.

Testing:

Tested it on both regular Kind and enterprise (created with the `./tools/reboot.sh` script) clusters.

To install the OSS dashboard via a HelmChart, please use the following instructions:
https://docs.gitops.weave.works/docs/installation/weave-gitops/#deploying-weave-gitops

Notes:
- I kept the same error-handling logic for getting a list of HelmReleases, as before, when getting a HelmChart. If there is an error when calling the client method, we just return nil for the HelmRelease and do not log the error (= we don't care why exactly we are unable to get the HelmReleases).


chanwit (approved):
LGTM, thank you @opudrovs

PR #3380, opened 2023-02-07
Add text search to table urls

### LINKED ISSUE #3359: dashboard UX: refine permalink support (list: search, sort)
**Problem**

As a flux user
in order to share permalinks to dashboard pages with others
I need the permalinks to include navigation actions such as:
* in list (applications, sources, ...):
   * search criteria
   * sorting criteria

Currently, the browser address bar is not reflecting the search or sorting actions previous performed:
image

**Solution**

Improve the permalink support in the dashboard to capture the navigation actions

**Additional context**

## Acceptance Criteria:

1. The permalinks in the dashboard capture the navigation actions performed by the user.
1. The permalinks include the search criteria used in the list views of applications, sources, etc.
1. The permalinks include the sorting criteria used in the list views of applications, sources, etc.
1. The browser address bar accurately reflects the search and sorting actions performed by the user.
1. The permalinks are easily shareable with others and accurately reflect the user's navigation actions.

joshri:
Closes #3359 

Out of the two proposed ideas in 3359, adding the text search to urls gave the most value - as I was coding this it got a little scary to start slapping these query-string and history.replace actions everywhere. It's fewer clicks to sort by a column/reverse sort and I don't think the hit to the code/performance is worth adding it to the url.

### DIFF
--- a/ui/components/DataTable.tsx
+++ b/ui/components/DataTable.tsx
@@ -217,17 +217,21 @@ function toPairs(state: FilterState): string[] {
   return _.concat(out, state.textFilters);
 }
 
-export function parseFilterStateFromURL(search: string): FilterSelections {
+export function parseFilterStateFromURL(search: string) {
   const query = qs.parse(search) as any;
+  const state = { initialSelections: {}, textFilters: [] };
   if (query.filters) {
     const split = query.filters.split("_");
     const next = {};
     _.each(split, (filterString) => {
       if (filterString) next[filterString] = true;
     });
-    return next;
+    state.initialSelections = next;
   }
-  return null;
+  if (query.search) {
+    state.textFilters = query.search.split("_").filter((item) => item);
+  }
+  return state;
 }
 
 export function filterSelectionsToQueryString(sel: FilterSelections) {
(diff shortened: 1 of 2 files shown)


opudrovs:
Is clicking Clear All supposed to remove search params from the address bar? When I click it, the params stay in the address bar, so if I share this link or reload the page, search will be applied again.

Screenshot 2023-02-08 at 12 38 28

Screenshot 2023-02-08 at 12 38 42

joshri:
Latest push should fix this issue thanks for spotting it - did a lil screen recording showing clear all after the commit:

https://user-images.githubusercontent.com/65822698/217621385-62994188-94b4-4e87-b836-bed9af0f66da.mov

opudrovs (approved):
Works fine for me. Great job! 🎉 Tested it with Applications, Sources, and reconciled objects tables.

A minor inconsistency I noticed is that if I add search terms, there is an underscore appended to the URL in the browser address bar. But if I start removing the terms one by one, the final underscore disappears from the browser address bar.  I am not sure if it's expected behavior. But it does not affect functionality anyway.

After adding search terms:

Screenshot 2023-02-08 at 21 07 25

Screenshot 2023-02-08 at 21 07 36

After removing a search term:

Screenshot 2023-02-08 at 21 07 52

joshri:
wowee that is such a good catch. The join function adds the separator character between each item, but not on the end. I think it's nice to have even though it doesn't break it, just because once you then add another item after you've removed one, they get smushed together as though they're the same search. It helps the readability...even though I hope that no one has to read these urls. I fixed it by always adding a `_` to the end of the string post-join on chip remove.

PR #3381, opened 2023-02-08
Update templates.mdx

darrylweaver:
Changed the path for template content to remove the ./ at the beginning.
The relative path is not required as it is relative to the repo root.
Using the ./ means the PR creation fails with the error from github that there is a malformed path component.

Using the ./ means the PR creation fails with the error from github that there is a malformed path component.


bigkevmcd (approved):
Thank you for fixing this!

PR #3382, opened 2023-02-08
add imagePolicy details

### LINKED ISSUE #3384: Add ImagePolicy details page
Provide a details page for imagePolicy to keep it consistence with UI

TheGostKasper:
closes #3384  

**What changed?**
Add `ImagePolicy` details page to `ImagePolicies` to keep consistency  along UI 

**Why was this change made?**
Based on DX team  feedback, there's a room of info to be shown for ImagePolicy than listed in the table view  

https://user-images.githubusercontent.com/4614360/217514670-c1fc167a-399a-46dc-8439-073f3a9f30b5.mp4

One issue will arise and it's related to Navigation highlighter when navigate to ImageAutomation subroute details as for now it's not available for `Breadcrumb` and `Nav` items as both uses the same function `getParentNavValue` to check the route and display the name for it in breadcrumb, so if we wanted to display the right Breadcrumb path [ImageAutomation] for subroute's details page then we won't be able to highlight the Nav item "ImageAutomation".

##### adding a new function to cover the Nav items and use it in Navigation component
```
export const getParentNavRouteValue = (
  path: string
): PageRoute | V2Routes | boolean => {
  const [, currentPage] = _.split(path, "/");
  switch (`/${currentPage}`) {
   ....

    case V2Routes.FluxRuntime:
      return V2Routes.FluxRuntime;

    case V2Routes.ImageAutomation:
    case V2Routes.ImageAutomationUpdatesDetails:
    case V2Routes.ImageAutomationRepositoryDetails:
    case V2Routes.ImagePolicyDetails:
      return V2Routes.ImageAutomation;

    ....

    default:
      // The "Tabs" component of material-ui wants a bool
      return false;
  }
};
```


jpellizzari (approved):

yiannistri (approved):
Tested with Tilt, works great 👍 
Screenshot 2023-02-09 at 10 07 32

Issue #3152, opened 2022-12-14
Add tooltip for absolute time in GitOps run duration column 


josefaworks:
In the GitOps run UI we are not able to show exact time because it creates an issue with the column width. However some users may still want to access this information instead of seeing a rounded value. 

As a user I want to be able to hover over the duration and see a tooltip that shows me the exact time something has been running.

PR #3385, opened 2023-02-08
Add optional tooltip to Timestamp component

joshri:
Part of #3152 

This tooltip prop is currently only going to be used in the GitOps Run table in Enterprise, hence the specificity on localeString, but it's easy to open it up for different formats if we want it to be more flexible and implemented in other tables/components.

Example:
image

### DIFF
--- a/ui/components/AutomationsTable.tsx
+++ b/ui/components/AutomationsTable.tsx
@@ -147,7 +147,6 @@ function AutomationsTable({ className, automations, hideSource }: Props) {
       label: "Last Updated",
       value: (a: Automation) => (
         <Timestamp
-          tooltip
           time={_.get(_.find(a.conditions, { type: "Ready" }), "timestamp")}
         />
       ),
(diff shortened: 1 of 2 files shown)


opudrovs:
- Tested this PR in enterprise, I only see date without time (what defeats the purpose of the tooltip):

Screenshot 2023-02-09 at 02 28 26

Probably this is how `toLocaleString` formats DateTime based on my locale (`de`) Based on Luxon's examples for the `toLocaleString` method, smth. like `DateTime.now().toLocaleString(DateTime.DATETIME_SHORT)` or another example might work instead to display the desired format.

- Also during testing discovered an issue unrelated to the tooltip: I am not able to scroll just the table itself horizontally to view the Created column on my laptop. I can only scroll the whole content area and cannot drag the scrollbar with the mouse when the window is at full height. I have to resize the window for the scrollbar to work with the mouse without trackpad. With trackpad I also have to scroll the whole content horizontally to see the column, but at least it's easier.

Screenshot 2023-02-09 at 02 29 23

Screenshot 2023-02-09 at 02 29 10

opudrovs (approved):
Re-tested it with enterprise, works as expected! 🎉 

Screenshot 2023-02-09 at 17 26 21

PR #3398, opened 2023-02-13
docs: document how to set up Tekton with Pipeline promotion events

yitsushi:
**What changed?**
An extra page added to the documentation website on how to set up pipeline notifications to trigger a Tekton pipeline.

**Why was this change made?**
Document a feature.

Created a new file

Checked the page with `yarn start`

**Release notes**
NA

**Documentation Changes**


website/docs/pipelines/pipelines-with-tekton.mdx
```diff
@@ -0,0 +1,312 @@
...
+# Setting Up Pipelines to Trigger a Tekton Pipeline <TierLabel tiers="Enterprise" />
+
+Using Flux's [Notification
+Controller](https://fluxcd.io/flux/components/notification/), a [Tekton EventListener](https://tekton.dev/docs/triggers/eventlisteners/) can be triggered on Pipeline promotion events.
+
+## Configuring Tekton Pipeline
+
+### Tekton Tasks
+
+We have only two tasks to demonstrate how to use parameter values from the
```
Callisto13, line 20:
```suggestion
In this tutorial, we have two tasks to demonstrate how to use parameter values from the
```

website/docs/pipelines/pipelines-with-tekton.mdx
```diff
@@ -0,0 +1,312 @@
...
+Pipeline event payload, both tasks print out messages with information about the
```
Callisto13, line 21:
```suggestion
Pipeline event payload. Both tasks print out messages with information about the
```

website/docs/pipelines/pipelines-with-tekton.mdx
```diff
@@ -0,0 +1,312 @@
...
+    value: $(body.involvedObject.namespace)
+  - name: name
+    value: $(body.involvedObject.name)
+  - name: message
+    value: $(body.message)
+```
+
+### Tekton TriggerTemplate
+
+The template has the same parameters as the `Pipeline` resources has:
```
Callisto13, line 173:
```suggestion
The template has the same parameters as the `Pipeline` resources:
```

Callisto13 (commented):
Could you summarise at the start what the outcome is? What do they expect to see on their cluster, their gitops repo, the wego UI, etcetc

website/docs/pipelines/pipelines-with-tekton.mdx
```diff
@@ -0,0 +1,312 @@
...
+  serviceAccountName: tekton-ww-pipeline-robot
+  triggers:
+    - name: ww-pipeline-trigger
+      bindings:
+      - ref: ww-pipeline-binding
+      template:
+        ref: ww-pipeline-template
+```
+
+At this point, we should have a `Service` for our `EventListener`.
```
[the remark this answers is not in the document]
yitsushi: Not necessarily. Yes, but no. It's enough if they add it to the cluster in any other ways if they have a hybrid solution (even with `kubectl apply`), and to be fair, if they add them all together it still works.

It's most like "see it creates a server and you can inspect it" at any point, for example if they want to `port-forward` it for testing.

I do not expect any guides here to state "and commit it to your giptops repo otherwise it will not be there".

yitsushi:
> Could you summarise at the start what the outcome is? What do they expect to see on their cluster, their gitops repo, the wego UI, etcetc

I have no idea, other than:
> At the end a Tekton Pipeline will be started on Weave GitOps Pipeline promotion events.

but that's kind of the title itself.

Callisto13 (approved):

website/docs/pipelines/pipelines-with-tekton.mdx
```diff
@@ -0,0 +1,312 @@
...
+❯ kubectl get service -n ww-pipeline
+NAME                      TYPE        CLUSTER-IP     EXTERNAL-IP   PORT(S)             AGE
+el-ww-pipeline-listener   ClusterIP   10.96.250.23   <none>        8080/TCP,9000/TCP   3d
+```
+
+## Configure Notification Provider
+
+In this case, we are using Tekton in the same cluster, so we can use an internal
+address to access the `EventListener` service. If they are not in the same
+cluster, exposing the service may required through an ingress or a service mesh.
```
yiannistri, line 279:
```suggestion
cluster, exposing the service may be required through an ingress or a service mesh.
```

yiannistri (approved):
Well-written guide 👌

Issue #3392, opened 2023-02-10
docs: Go through GitOps Run


Callisto13:
at `website/docs/gitops-run`.

they seem pretty solid, but just walk through them, fill any gaps etc

PR #3400, opened 2023-02-14
docs: Revamp the GitOps Run docs

Callisto13:
Might also be worth a note that versions pre v0.17.0 come with an unpleasant bug?

Closes https://github.com/weaveworks/weave-gitops/issues/3392


yitsushi (approved):
I really love the added folds.

PR #3410, opened 2023-02-15
docs: Add instructions for BitBucket Server

yiannistri:
**What changed?**
Added docs for BitBucket Server

**Why was this change made?**
To assist users with setting up BitBucket Server OAuth flow in WGE.

N/A

N/A

**Release notes**
N/A

**Documentation Changes**
N/A


website/docs/installation/weave-gitops-enterprise/index.mdx
```diff
@@ -243,6 +244,44 @@ kubectl create secret generic git-provider-credentials --namespace=flux-system \
...
+Save your application and take note of the **Client ID** and **Client Secret** and save
+them into the `git-provider-credentials` secret along with:
+
+- `GIT_HOST_TYPES` to tell WGE that the host is bitbucket-server
+- `BITBUCKET_SERVER_HOSTNAME` where the OAuth app is hosted
+
+**Replace values** in this snippet and run:
+
+```bash
+kubectl create secret generic git-provider-credentials --namespace=flux-system \
```
yitsushi, line 272: [only a note]
I think it's worth to mention to "create **or update existing secret**".
A cluster can have a `git-provider-credentials` secret already and a user may think "it's already there, I'll just delete the old one" and that can be a painful step.

I don't think it's in the scope of this PR, but long term, maybe we should update this part of the docs to split it into a "generic" section and a provider specific one. In this form, if someone reads the whole page, they will get it, but if someone has a cluster already somewhere, they try to add a new provider, they will just scroll down to this part and if they were not experts, that can cause pain.

yitsushi (approved):
lgtm

Issue #2994, opened 2022-11-10
ID Token not verified from cookie causes RateLimits


eloo:
**Describe the bug**

Hi, i'm trying to get the OIDC setup running properly but it seems that there is an issue with the tokens in cookie storage.
Right now the order which cookie is verifies is hardcoded to `access_token`, `id_token`.
Using auth0 the access_token can not be used because it contains a different (non-valid) schema but the id_token could be used but only the first one is taken into account.

Nevertheless also when the correct token would be read from the cookies is could not get verified in the userinfo code because https://github.com/weaveworks/weave-gitops/blob/main/pkg/server/auth/server.go#L379 only the HMAC verifier is present.
This result in a call to the userinfo endpoint of the oidc provider everytime. 

In general this is going to work. But its creating a lot of unnecessary requests against the oidc-provider and in the case of Auth0 the rate limit is kicking in and make the UI unusable.

So i guess the best would be to add the correct verifiers to the userinfo code so the token from the cookies can be used. 

**Environment**

- Weave-Gitops Version: 0.10.1

**To Reproduce**

- Configure oidc with auth0
- Click around in the ui or call .../oauth2/userinfo multiple times
- You will se 429 in logs and you will be logged out of the ui

**Expected behavior**

IdToken from cookie is used

**Actual Behavior**

Everytime userinfo is called a request against the oidc-provider is made

**Additional Context (screenshots, logs, etc)**

chanwit:
Thank you again @eloo for reporting this issue!
We'll take a look at it shortly.

PR #3418, opened 2023-02-16
Drop call to UserInfo.

bigkevmcd:
**What changed?**
This drops the UserInfo call, and extracts the user credentials from the ID token

**Why was this change made?**
Turns out that not all UserInfo calls return the same data as we can claim via the scopes.

In this case, valid scopes from the ID token may not exist in the UserInfo response, which was causing problems returning an error.

We already have the data in the ID token (or not) so this was an additional API call in the request context.

It's a trivial refactoring, removing the call, and replacing it by parsing the token.

Tests are still passing

**Release notes**
Remove UserInfo call, which reduces the number of API calls to an OIDC server, and handles cases where the UserInfo doesn't return the data that is available via scopes.

**Documentation Changes**
No


pkg/server/auth/server.go
```diff
@@ -464,12 +464,10 @@ func (s *AuthServer) UserInfo(rw http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	info, err := s.provider.UserInfo(r.Context(), oauth2.StaticTokenSource(&oauth2.Token{
-		AccessToken: c.Value,
-	}))
+	info, err := s.verifier().Verify(r.Context(), c.Value)
```
foot, line 467: c.Value is the access_token rather than the id_token?

bigkevmcd: Not necessarily...look at findToken.

There's an argument to be made that findToken should not return the access_token at all

And a further argument that if we can't use the ID token, we should fall back to the access token, but...it's pretty bad if we can't use the ID token?

foot: Yeah. I guess accesstoken was just to query user info? We can drop it now maybe?

Findtoken only returns access token if it's present tho? 👀

foot (approved):
💯 tested 
- okta with oidc and user-account enabled and worked fine, pulled out the correct user/group claims

foot:
I think this addresse https://github.com/weaveworks/weave-gitops/issues/2994

PR #3419, opened 2023-02-16
address GitOps Run "ask twice" issue by handling "no" answer correctly

### LINKED ISSUE #3038: `gitops run` asks twice to install the dashboard
**Describe the bug**

When running `gitops run` it asks me twice whether I wanted to install the dashboard:

```
1 If you are running GitOps Run for the first time, it may take a few minutes to download the required images.
2 ✗ Would you like to install the GitOps Dashboard:
3 ► Creating GitOps Run session run-main-be6a3d6f in namespace default ...
4 [...]
5 ✔ Port forwarding for run-dev-bucket is ready.
6 ► Checking if GitOps Dashboard is already installed ...
7 ? Would you like to install the GitOps Dashboard? [Y/n] █
```

see lines 2 and 7.

**Expected behavior**

I should only have to answer this question once within the same execution.

chanwit:
Fixes #3038

The reason for these changes is that the previous implementation of the GitOps Dashboard installation process had a flaw in the dashboard prompt. When the prompt was answered with "no," it was still asked again by the sub-process, resulting in a redundant prompt. 

This PR makes changes in two files. In the `cmd/gitops/beta/run/cmd.go` file, the `dashboardStep` function is modified by adding a constant block with three possible values, `unknown`, `wantToInstall`, and `doNotWantToInstall`. The function handles these values and updates the `wantToInstallTheDashboard` variable accordingly.  If the user chooses not to install the dashboard, it sets the `flags.SkipDashboardInstall` to true.

In the `pkg/run/install/session.go` file, the "NewSession" function, and the "Session" struct have a new parameter `skipDashboardInstall`. The "vcluster.ConnectCmd" command receives this value as a flag. If the "skipDashboardInstall" flag is set to true, then the `--skip-dashboard-install` flag is passed to the sub-process. The "Connect" function checks for the "skipDashboardInstall" flag, and if set, it will not forward the dashboard password to the session.

To test this PR:
- Run GitOps Run in session mode.
- Answer "n" when it asks to install the dashboard, and we should not see the redundant prompt asked by the sub-process.

### DIFF
--- a/cmd/gitops/beta/run/cmd.go
+++ b/cmd/gitops/beta/run/cmd.go
@@ -356,7 +356,6 @@ func dashboardStep(ctx context.Context, log logger.Logger, kubeClient *kube.Kube
 	switch dashboardType {
 	case install.DashboardTypeEnterprise:
 		flags.SkipDashboardInstall = true
-
 		log.Warningf("GitOps Enterprise Dashboard is found. GitOps OSS Dashboard will not be installed")
 	case install.DashboardTypeOSS:
 		log.Successf("GitOps Dashboard is found")
@@ -373,9 +372,12 @@ func dashboardStep(ctx context.Context, log logger.Logger, kubeClient *kube.Kube
 
 			// Answering "n" causes err to not be nil. Hitting enter without typing
 			// does not return the default.
-			_, err := prompt.Run()
+			answer, err := prompt.Run()
 			if err == nil {
 				wantToInstallTheDashboard = true
+			} else if answer == "n" || answer == "N" {
+				wantToInstallTheDashboard = false
+				flags.SkipDashboardInstall = true
 			}
 		}
 
@@ -508,6 +510,7 @@ func runCommandOuterProcess(cmd *cobra.Command, args []string) (retErr error) {
 		flags.SessionNamespace,
 		fluxVersionInfo.FluxNamespace, // flux namespace of the session
 		portForwardsForSession,
+		flags.SkipDashboardInstall,
 		dashboardHashedPassword,
 		kind,
 	)
--- a/pkg/run/install/session.go
+++ b/pkg/run/install/session.go
@@ -21,6 +21,7 @@ type Session struct {
 	kubeClient              client.Client
 	log                     logger.Logger
 	dashboardHashedPassword string
+	skipDashboardInstall    bool
 	portForwards            []string
 	automationKind          string
 }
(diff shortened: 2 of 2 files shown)


cmd/gitops/beta/run/cmd.go
```diff
@@ -361,9 +361,14 @@ func dashboardStep(ctx context.Context, log logger.Logger, kubeClient *kube.Kube
...
+		const (
+			unknown = iota
+			wantToInstall
+			doNotWantToInstall
+		)
+		wantToInstallTheDashboard := unknown
 		if dashboardHashedPassword != "" {
-			wantToInstallTheDashboard = true
+			wantToInstallTheDashboard = wantToInstall
 		} else if !flags.SkipDashboardInstall && dashboardHashedPassword == "" {
```
makkes, line 372:
```suggestion
		} else if flags.SkipDashboardInstall {
			wantToInstallTheDashboard = doNotWantToInstall
		} else {
```
This makes it a little easier to read.

cmd/gitops/beta/run/cmd.go
```diff
@@ -373,13 +378,18 @@ func dashboardStep(ctx context.Context, log logger.Logger, kubeClient *kube.Kube
...
+			answer, err := prompt.Run()
 			if err == nil {
-				wantToInstallTheDashboard = true
+				wantToInstallTheDashboard = wantToInstall
+			} else if answer == "n" || answer == "N" {
+				wantToInstallTheDashboard = doNotWantToInstall
+				flags.SkipDashboardInstall = true
+			} else {
+				wantToInstallTheDashboard = unknown
 			}
```
makkes, line 389: If I'm not mistaken this is the only case where `wantToInstallTheDashboard` would be set to `unknown` (which it already is, anyway), so this assignment is unnecessary.

If we decide to fail for any answer other than "n" or "N" then we could completely get rid of the `unknown` state and make `wantToInstallTheDashboard` a bool.

wdyt?

makkes: Thinking about it, we don't need the `unknown` state, anyway because further down we only evaluate `wantToInstallTheDashboard == wantToInstall`.

opudrovs (approved):
LGTM! 🎉 Tested it, works as expected.

cmd/gitops/beta/run/cmd.go
chanwit: Yep. That's an over design. Converted back to a Boolean and it's still semantically correct.

makkes (approved):

PR #3424, opened 2023-02-20
fix enterprise installation docs

juozasg:
install WGE CLI example use latest version


makkes (approved):

makkes:
Thanks @juozasg, good find! I'm led to believe this should be covered by our automation but wasn't for some reason.

Issue #3379, opened 2023-02-07
Error getting GitOps Run logs before GitOps Run objects are ready


opudrovs:
This issue is similar to:
https://github.com/weaveworks/weave-gitops/issues/3297

If we go to the GitOps Run session details pages too early (before some objects, probably session objects or secrets?) are ready, there is an error displayed in the browser console and no logs are loaded on the page.

<img width="1529" alt="Screenshot 2023-02-13 at 18 02 43" src="https://user-images.githubusercontent.com/8824708/218534044-76497271-c013-4764-ae48-0e86eed07cc7.png">

**Acceptance Criteria**

- Change the error message on getting scoped client in the `GetSessionLogs` endpoint to `"session %s not found: %w", clusterName, err`.

- Add a check for the `run-devbucket-credentials` secret created and do not return an error if the secret is not found, but instead assign a message string to the `Error` field of the `GetSessionLogs` response.

opudrovs:
Added a screenshot to the issue description.

Error messages from the browser console:

```
index.js:9317          POST http://localhost:8000/v1/session_logs 500 (Internal Server Error)

query.js:356 {code: 2, message: 'secrets "run-dev-bucket-credentials" not found', details: Array(0)}
```

Issue #3297, opened 2023-01-23
Errors on GitOps Run object details pages before GitOps Run objects are ready


opudrovs:
When starting a GitOps Run session and looking at it in the GitOps Run UI, there is a moment when a session is already displayed in the sessions table, with links to GitOps Run objects

<img width="1500" alt="Screenshot 2023-01-23 at 13 16 29" src="https://user-images.githubusercontent.com/8824708/214075743-5046b1dc-4bac-4a8c-b917-6156b2af8aa7.png">

but clicking the links displays the following errors:

<img width="1521" alt="Screenshot 2023-02-13 at 17 54 44" src="https://user-images.githubusercontent.com/8824708/218526369-ba1bd3e8-9f5e-4a0a-adee-dfc69f175f6e.png">

<img width="1528" alt="Screenshot 2023-02-13 at 18 08 25" src="https://user-images.githubusercontent.com/8824708/218526412-6d243287-27cc-4f38-996e-7577b53372f7.png">

<img width="1536" alt="Screenshot 2023-02-13 at 18 08 48" src="https://user-images.githubusercontent.com/8824708/218526598-79bbef96-5ba8-431b-b461-30a656df05ec.png">

I guess it happens because GitOps Run objects are not reconciled/ready yet.
After a while, object detail pages are displayed as expected when clicking the links.

joshri:
We're constructing these links from pieces of data in the annotations, and we're definitely making an assumption that the automation and source will be ready. I'm not sure that there's a way to easily avoid this in the data as-is, but at the very least this error should be handled in a way that makes it look like we know it's there

opudrovs:
I thought about returning a flag alongside with data returned from the backend, smth. like "objects are ready". Like first we return the data for a gitops run session (or whatever it is, have not looked at specific data yet) without this flag and display the session but do dnot display working links to object detail pages of now. But after re-polling if the flag is present we can display the links.

Otherwise if we just wait until returning the whole data we will not be able to display the session until the objects are ready, so there would be a considerable delay between starting a gitops run session and showing it in the sessions table.

opudrovs:
Updated screenshots in the issue description. Full error logs from the browser console:

```
http://localhost:8000/v1/object/run-dev-ks?namespace=flux-system&kind=Kustomization&clusterName=default%2Frun-master-fdd0a0b7 500 (Internal Server Error)

[...]

PR #3429, opened 2023-02-21
Error getting GitOps Run logs before GitOps Run objects are ready

opudrovs:
Closes https://github.com/weaveworks/weave-gitops/issues/3379

- Updated the error message for getting a scoped client.

- Added a check for the log secret already existing.

- Added unit tests for `isSecretCreated`.

- Removed unneeded comments.

Testing:

- Pull in the current PR branch, GitOps Run and go to the session details page before the session is completely ready (you should see the "cluster not found" error in the UI, this is expected behavior).
- 
When the session objects get ready and polling logs succeeds, logs are displayed without errors.

### DIFF
--- a/core/server/session_logs.go
+++ b/core/server/session_logs.go
@@ -4,13 +4,14 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"github.com/weaveworks/weave-gitops/pkg/compositehash"
 	"io"
 	"regexp"
 	"sort"
 	"strconv"
 	"strings"
 	"time"
+
+	"github.com/weaveworks/weave-gitops/pkg/compositehash"
 
 	sourcev1 "github.com/fluxcd/source-controller/api/v1beta2"
 	"github.com/minio/minio-go/v7"
@@ -25,6 +26,10 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	secretNotCreated = "secret not created"
 )
 
 type s3Reader interface {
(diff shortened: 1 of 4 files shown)


chanwit (approved):
LGTM :+1:

core/server/session_logs.go
```diff
@@ -27,6 +28,10 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+const (
+	secretNotCreated = "secret not created"
```
makkes, line 32: ⚠️ I think this doesn't need to be a constant as it's only used once, anyway. If we want it to be one let's add a comment explaining what it's used for.

core/server/session_logs.go
```diff
@@ -141,6 +146,13 @@ func (cs *coreServer) GetSessionLogs(ctx context.Context, msg *pb.GetSessionLogs
 
 	logSourceFilter := msg.GetLogSourceFilter()
 	if logSourceFilter == "" || logSourceFilter == logger.SessionLogSource {
+		// check if we can get session logs already
+		// if the secret is not created yet, we should not display an error in the browser console
+		_, err = isSecretCreated(ctx, cli, constants.GitOpsRunNamespace, constants.RunDevBucketCredentials)
```
makkes, line 151: ❔ Why does `isSecretCreated` return a `bool` if it's not used anywhere?

core/server/session_logs.go
opudrovs: @makkes thanks for the comment. We don't know yet if the original issue can be reproduced now or not 100%, I just was not able to reproduce it.

So, this constant is used for consistency with the constant used in a similar issue (which we were able to reproduce): https://github.com/weaveworks/weave-gitops/issues/3297 

If this issue happens again, we'll use a similar constant in the enteprise UI to check specifically for this issue and display a non-error notification to the user.

If we are still unable to reproduce this issue after testing, we can just remove this check completely. But I want to be on the safe side.

opudrovs: As discussed, will add this comment and change the return type of the check function to just bool in a next related PR for developer information.

PR #3431, opened 2023-02-21
add hashed key to session logs response

### LINKED ISSUE #3415: gitops run reloading skews timestamps of its session logs
## User story

As a user of the GitOps Run feature, I want to be able to rely on the accuracy of the timestamps in the session logs, so that I can troubleshoot any issues that may arise during the deployment process.

However, I have noticed that when I make changes to the files and the GitOps Run command triggers file reloading, the timestamps in the session logs become skewed. This makes it difficult for me to determine the exact order of events and to accurately diagnose any errors that may have occurred during the deployment.

I would like for the timestamps in the session logs to remain consistent and accurate, regardless of any file reloading that may occur during the GitOps Run process. This would help me to quickly identify any issues that may have arisen and to take corrective action in a timely manner.

## Acceptance Criteria

- The timestamps in the session logs should remain consistent and accurate, regardless of any file reloading that may occur during the GitOps Run process.
- The order of events in the session logs should be clear and accurate, enabling users to troubleshoot any issues that may have arisen during the deployment process.

chanwit:
Fixes #3415

## New `key` field

This PR adds a new field called "key" in the `LogEntry` protobuf message, which is used as the primary sorting key for the log entries. The corresponding change is made to the schema definition for `GetSessionLogsResponse`.

In the `GetSessionLogs` function in `session_logs.go`, the logEntries slice is sorted based on this new "key" field instead of the "timestamp" field, which was used earlier for sorting the logs. The reason for adding this "key" field is to handle a scenario where multiple log entries have the same timestamp. In such a scenario, the sorting order of the logs would be non-deterministic, which can lead to issues in some cases.

To ensure that the log entries are sorted in a deterministic order, this PR introduces a new Go package called `compositehash`, which provides a function for generating a unique identifier from a string and a timestamp.

## `compositehash` package

The `compositehash` package contains one file called composite_hash.go, which defines the logic for generating the unique identifier. The New() function uses the FNV-1a hash function to compute a hash value of the string, and then combines this hash value with the timestamp, expressed in milliseconds since the epoch, to generate a unique 64-bit integer. This integer is guaranteed to be unique and sortable for the given string and timestamp.

The `compositehash` package is used in two different files:

### `s3_log_writer.go`

In `s3_log_writer.go`, the `compositehash` package is used to generate a unique key for each log entry that is written to an S3 bucket. The key is used to sort the log entries and ensure that they are unique. The `New()` function in the `compositehash` package is called with the log message and the current timestamp as input to generate the key.

### `session_logs.go`

In `session_logs.go`, the `compositehash` package is used to calculate a hash of the log entry message and the timestamp to create a unique key. This key is used to sort the log entries and ensure that they are sorted in a deterministic order.

## Misc

This PR also changes "warn" to "warning" in the function `detectLogLevel` and fixes its unit tests.


api/core/core.proto
```diff
@@ -307,6 +307,8 @@ message LogEntry {
     string source    = 2;
     string level     = 3;
     string message   = 4;
+    // sorting key
```
opudrovs, line 310: 💡 Maybe we could remove this comment and just name the field `sortingKey`?

core/server/session_logs.go
```diff
@@ -190,7 +192,7 @@ func detectLogLevel(message string) string {
 	if errorRegex.MatchString(message) {
 		return "error"
 	} else if warnRegex.MatchString(message) {
-		return "warn"
+		return "warning"
```
opudrovs, line 195: ✨ You updated the warning level string for consistency, great!

opudrovs (approved):

api/core/core.proto
chanwit: done!
Thank you @opudrovs

PR #3435, opened 2023-02-22
skip reconciliation of dotfiles in GitOps Run

### LINKED ISSUE #3202: "gitops run" should ignore dot files
**Describe the bug**

When using "gitops run" on a local kind cluster and local directory, certain editors create backup or recovery files in the same sub-directory as the file being edited. When the editor is started, these "dot" files trigger a reconciliation which can cause delays in the actual fixed code from being deployed. For example, "vim" creates a recovery file ".swp.<filename>". Others have similar capabilities.

**Environment**

- Weave-Gitops Version 0.13
- Flux Version 0.37
- Kubernetes version 1.25

**To Reproduce**

Set up "gitops run" environment. Edit a watched file with "vim". Watch the reconcile occur before you actually save the file.

**Expected behavior**

Editing artifacts shouldn't trigger reconciliation.

**Actual Behavior**

File created in the directory that are editing artifacts trigger reconciliation.

makkes:
closes #3202

### DIFF
--- a/cmd/gitops/beta/run/cmd.go
+++ b/cmd/gitops/beta/run/cmd.go
@@ -826,13 +826,23 @@ func runCommandInnerProcess(cmd *cobra.Command, args []string) error {
 			case <-stopUploadCh:
 				return
 			case event := <-watcher.Events:
+				info, err := os.Stat(event.Name)
+				if err != nil {
+					continue
+				}
 				if event.Op&fsnotify.Create == fsnotify.Create ||
 					event.Op&fsnotify.Remove == fsnotify.Remove ||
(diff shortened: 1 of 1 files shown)


opudrovs (approved):
Tested it with podinfo.

For test dot files everything is working as expected: the files are skipped and no change events are detected + no reconciliation is requested. 🎉 

For test dot directories (with or without files) I observe the following behavior: when moving a dot directory out of the podinfo directory (from which gitops run is started), no change event is detected. But when moving a dot directory into the podinfo directory, a change event is detected and reconciliation requested.

I see that we are not skipping dot dirs in this PR, so this is probably expected behavior and the linked issue does not cover dot directories (only dot files), correct?

PR #3437, opened 2023-02-23
Fixes "e.containers is null" UI page crash

foot:
- We're sometimes seeing an "e.containers" is null error that crashes the UI
- This makes the container lookup a bit more defensive
- Any reconciled object can be viewed in the UI and can have an unexpected shape if its a new CR or something

**What changed?**

Don't crash if `spec.template.spec` is in the resource but `spec.template.spec.containers` is not.

**Why was this change made?**

To try and prevent occasional crashes of the UI

Changing the images lookup system to be a bit more defensive.

unit tests

**Release notes**

**Documentation Changes**


jpellizzari (approved):
`lodash` to the rescue. Digging the tests as well.

Issue #3432, opened 2023-02-21
Weave GitOps Left Menu


JamWils:
As a user, I want to have as much screen real estate as possible for data.  This means we want the menu to be able to expand and collapse so I can focus on the work at hand.  
* When the menu is collapsed I am able to clearly see the icons that are set up.
* When I hover over an icon when the menu is collapsed I will see a tooltip per the design.
* I do not see the category names when the menu is collapsed, instead there is space which clearly defines the groupings, see designs.

As a user, I don't often click my user icon, this means I don't see the notifications section that are hidden in the pop-up menu.  Move the notifications into the left-side menu.

As a user, I find the "nesting" of certain features to be awkward.  Please remove the "child" areas and make each menu item a parent object.

As an enterprise user, the menu is rather noisy, there are all of these different technologies to work with and there doesn't seem to be any order to how things are laid out.  It would be nice if a set of products were grouped together so the experience is easier to understand and is more seamless.  Follow the designs and group all items according to the designs.  There are two menu items missing and they should go in:

1. Sources belongs in "Platform"
2. Image automation belongs in "Delivery"

As a user, I expect the "expanded" option to be the default.  However, if I hit "collapse" it would be nice to have that persisted in my local settings.

As a user I also expect to see the logo of "Weave GitOps" disappear when the menu is collapsed, per the design.

https://www.figma.com/proto/IVHnM9iyeFWpd11evtY8ux/Weave-GitOps?page-id=14519%3A99472&node-id=14946%3A94881&viewport=119%2C-1606%2C0.5&scaling=scale-down&starting-point-node-id=14946%3A94881&show-proto-sidebar=1

PR #3444, opened 2023-02-24
Add collapsible nav redesign

joshri:
Part of #3432 

https://user-images.githubusercontent.com/65822698/221301321-eff75773-ede5-482b-8f2a-3b20c711dc9a.mov

Preview of new nav in OSS - any glaring omissions?

Missing icons I need a design choice on:
Sources
Image Automation

Next steps:
- add missing icons (the notifications bell probably exists in MUI already)
- remove link to notifications inside user settings
- add capability for section headers for enterprise nav
- implement nav in enterprise


ui/components/Icon.tsx
```diff
@@ -167,6 +171,51 @@ function getIcon(i: IconType) {
     case IconType.PauseIcon:
       return PauseIcon;
 
+    case IconType.NotificationsBell:
+      return NotificationsBellIcon;
+
+    case IconType.SourcesIcon:
+      return () => (
+        <svg
```
jpellizzari, line 179: I've done this before, no problems for me here. I tend to throw them in their own component just to clean things up:

```jsx
export default function CoolIcon() {
  <svg >...</svg>
} 
```

If you add a `className` prop + styled-components wrapper they can be styled like everything else.

ui/components/Icon.tsx
```diff
@@ -167,6 +171,51 @@ function getIcon(i: IconType) {
...
+    case IconType.ImageAutomationIcon:
+      return () => (
+        <svg
+          width="24"
+          height="24"
+          viewBox="0 0 24 24"
+          fill="none"
+          xmlns="http://www.w3.org/2000/svg"
+        >
+          <path
```
jpellizzari, line 211: FYI: the only downside of doing this inline (instead of a separate `.svg` file) is that the browser can't do fancy things to cache the image and our bundle size increases.

Not a big deal. I would keep as is.

ui/components/Logo.tsx
```diff
@@ -24,11 +26,19 @@ export default styled(Logo)`
     display: flex;
     align-items: center;
   }
+  img:first-child {
+    margin-right: ${(props) => props.theme.spacing.xs};
+  }
+  img:nth-child(2) {
+    opacity: ${(props) => (props.collapsed ? 0 : 1)};
+    transition: opacity 200ms cubic-bezier(0.4, 0, 0.2, 1) 0ms;
```
jpellizzari, line 34: :sparkles:

ui/components/Nav.tsx
```diff
@@ -1,87 +1,133 @@
...
+  flex-direction: column;
+  justify-content: space-between;
   width: 100%;
   height: 100%;
   border-radius: 10px;
+  border-bottom-right-radius: 0px;
   background-color: ${(props) => props.theme.colors.neutral00};
-  padding-top: ${(props) => props.theme.spacing.medium};
+  //28px bottom padding is medium + xxs - in the theme these are strings with px at the end so you can't add them together. This lines up with the footer.
+  padding: ${(props) => props.theme.spacing.medium} 0px 28px 0px;
```
jpellizzari, line 51: `calc()` might work but I like this better anyway.

jpellizzari (commented):
I thought I understood the inline SVGs but then I saw the files at the bottom. When do we inline vs use a file in this case? Is it just a styling thing? Or do we change the color based on props?

ui/components/Icon.tsx
joshri: Great will move to outside components

ui/components/ImageAutomationIcon.tsx
```diff
@@ -0,0 +1,21 @@
+import * as React from "react";
+
+function ImageAutomationIcon() {
+  return (
+    <svg
+      width="24"
+      height="24"
+      viewBox="0 0 24 24"
+      fill="none"
+      xmlns="http://www.w3.org/2000/svg"
```
jpellizzari, line 10: Doesn't your linter always complain about this prop? Mine does.

jpellizzari (commented):
@joshri My question still stands: when you we inline vs using a file import?

joshri:
> @joshri My question still stands: when you we inline vs using a file import?

File imports turn into an `img` tag with the source as a data-url, which means we can't change attributes of the svg anymore. Making the component allows for the className in the Icon component to still have an effect.

PR #3447, opened 2023-02-27
Take "storage" version into account for Flux CRDs

### LINKED ISSUE #3446: Flux CRD list does not show active storage version
**Describe the bug**

The current Flux CRD overview does not take the active storage version into account, but rather just the first Custom Resource Definition version in the list:

https://github.com/weaveworks/weave-gitops/blob/16f2faafada6fdc16efe61a1ad8e33af88e49ce0/core/server/fluxruntime.go#L165-L167

To make this more precise, and show the latest "active" version, it should take into account the version which has `storage: true` set.

**Environment**

- Weave-Gitops Version: `<=v0.17.0`
- Flux Version: `v0.40.1`
- Kubernetes version: N/A

**To Reproduce**

1. Install Flux version `v0.40.1`, which has a new `imagerepositories.image.toolkit.fluxcd.io` version (`v1beta2`).
2. Go to "Flux Runtime" page and observe `v1beta1` being reported as the "current" CRD version.

**Expected behavior**

It correctly reporting `v1beta2`, which is the active version in etcd.

image

hiddeco:
This ensures the active version stored in etcd is showed in the UI, instead of the first version on the list. Providing a more accurate view of the current version existing in the cluster.

Fixes #3446

### DIFF
--- a/core/server/fluxruntime.go
+++ b/core/server/fluxruntime.go
@@ -162,8 +162,13 @@ func (cs *coreServer) ListFluxCrds(ctx context.Context, msg *pb.ListFluxCrdsRequ
 			for _, d := range list.Items {
 				version := ""
 
-				if len(d.Spec.Versions) > 0 {
-					version = d.Spec.Versions[0].Name
+				for _, v := range d.Spec.Versions {
+					// This is the "active" version of the CRD in etcd, and a CRD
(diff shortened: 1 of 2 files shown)


stefanprodan (approved):
LGTM

makkes (approved):

PR #3460, opened 2023-02-28
Openshift: Allow removal of securityContext with Helm values

erikgb:
Closes

**What changed?**

This PR puts a conditional `with` block around the `securityContext` properties in the Deployment Helm template. This will ensure that the element is not rendered if the corresponding value is empty. This seems to be the pattern used in this, and other templates of the Helm chart.

**Why was this change made?**

This allows installation on Openshift without elevated SCC permissions. This is a common problem when provisioning community software to Openshift. The intentions of default security contexts are good, but sadly Openshift does not allow it. Here's a random goog explanation: https://github.com/bitnami/charts/issues/4884#issuecomment-754648273

n/a (trivial change)

n/a (trivial change)

**Release notes**

n/a

**Documentation Changes**

n/a

### DIFF
--- a/charts/gitops-server/templates/deployment.yaml
+++ b/charts/gitops-server/templates/deployment.yaml
@@ -32,12 +32,16 @@ spec:
         {{- toYaml . | nindent 8 }}
       {{- end }}
       serviceAccountName: {{ include "chart.serviceAccountName" . }}
+      {{- with .Values.podSecurityContext }}
       securityContext:
-        {{- toYaml .Values.podSecurityContext | nindent 8 }}
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
       containers:
         - name: {{ .Chart.Name }}
+          {{- with .Values.securityContext }}
           securityContext:
-            {{- toYaml .Values.securityContext | nindent 12 }}
+            {{- toYaml . | nindent 12 }}
+          {{- end }}
           image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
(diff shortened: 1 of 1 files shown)


erikgb:
@makkes The CI failures seem unrelated, or am I overlooking something?

makkes:
The change looks legit. I'll give it a spin later this week and approve.

As to failing CI checks: We're aware that some CI checks will fail on PRs from forks and we're working to improve the situation for community contributors. You don't need to worry about them right now. I'll let you know if something is legitimately failing.

erikgb:
@makkes Any chance to get this reviewed and merged? I think it is a reasonable non-breaking change.

LappleApple:
@chanwit Can you please share the current status of this issue?

chanwit:
@erikgb sorry for the late response. I'll be viewing this PR.

chanwit:
These changes are about improving the robustness of the Helm chart by avoiding potential errors if the `podSecurityContext` and `securityContext` are not defined. 

The line `{{- toYaml .Values.podSecurityContext | nindent 8 }}` has been replaced with a `with` statement to enclose the `securityContext` definition. The with clause is used to avoid having to check if a value is set before accessing it, so it provides null safety. This change indicates that the `podSecurityContext` will only be included if `.Values.podSecurityContext` is set. If `.Values.podSecurityContext` isn't defined, then the `securityContext` block won't be rendered at all, thereby reducing possible syntax errors.

Similar changes have been made for the `securityContext` in the containers section.

So I agreed that it would be a non-breaking change.

chanwit (approved):

PR #3462, opened 2023-02-28
Add DetailModal component, and convert pop up YamlViews to use it

### LINKED ISSUE #3439: Create right modal menu
As a user, I want to start to investigate the details of various k8s objects when issues arise.  We want to change the UX pattern where instead of details coming in as a modal pop-up in the middle of the screen, we want those to come sliding out from the right-side of the screen.

As a user, I can dismiss it by clicking outside of the modal.  Or I can use the "X" button per the design to dismiss it.  When I go to click on another item the right modal will pop-out.

joshri:
Closes #3439 

UPDATE:

As far as I understand this is actually only going to contain a more generic details screen. So it doesn't need to be as flexible as I made it before. We wanted a header thing in there as well which I've added in this screen: 
image

OLD PR DESCRIPTION BELOW:

The main purpose of this PR is to create a multi-purpose drawer component - the pop-up YamlViews have been converted to use it to start out, but the idea is that _anything_ will be able to go in there. Because we want to be able to trigger it from anywhere, I added it to AppContext. This makes the typing a little tricky since we're using components + props, when before it was just yaml so I always knew what to expect. 
Adding a new component would entail:  
- Adding the name to the `DetailOptions` enum
- Exporting the props and adding them to the combined `PropOptions` type
- Adding it to the switch statement in `DetailModal`

Looking for suggestions on if you would handle the typing any differently @jpellizzari 

https://user-images.githubusercontent.com/65822698/221993241-61650235-5e20-428f-81a8-be8229f57f7c.mov

### DIFF
--- a/ui/components/DetailModal.tsx
+++ b/ui/components/DetailModal.tsx
@@ -14,7 +14,7 @@ export type DetailViewProps = {
   props: PropOptions;
 };
 
-function DetailModal({ className, props, component }: DetailViewProps) {
+function DetailModal({ props, component }: DetailViewProps) {
   switch (component) {
     case DetailOptions.YamlView:
       return <DialogYamlView {...props} />;
--- a/ui/components/DetailModal.tsx
+++ b/ui/components/DetailModal.tsx
@@ -1,24 +1,47 @@
-import React from "react";
+import { IconButton } from "@material-ui/core";
+import { Close } from "@material-ui/icons";
+import React, { useContext } from "react";
 import styled from "styled-components";
-import { DialogYamlView, YamlViewProps } from "./YamlView";
-
-export enum DetailOptions {
-  YamlView = "YamlView",
-}
-
-export type PropOptions = YamlViewProps;
+import { AppContext } from "../contexts/AppContext";
+import { FluxObject, FluxObjectNode } from "../lib/objects";
+import Flex from "./Flex";
+import Text from "./Text";
+import { DialogYamlView } from "./YamlView";
 
 export type DetailViewProps = {
   className?: string;
-  component: DetailOptions;
-  props: PropOptions;
+  object: FluxObject | FluxObjectNode;
 };
 
-function DetailModal({ props, component }: DetailViewProps) {
-  switch (component) {
-    case DetailOptions.YamlView:
-      return <DialogYamlView {...props} />;
-  }
+const HeaderFlex = styled(Flex)`
+  margin-bottom: ${(props) => props.theme.spacing.xs};
+`;
+
+function DetailModal({ object, className }: DetailViewProps) {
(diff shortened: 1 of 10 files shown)


ui/lib/theme.ts
```diff
@@ -170,5 +170,11 @@ export const muiTheme = createTheme({
         overflowX: "hidden",
       },
     },
+    MuiDrawer: {
+      paper: {
+        width: "60%",
+        minWidth: 600,
```
jpellizzari, line 176: How does this look on a laptop screen vs a desktop screen? Mobile?

ui/components/DetailModal.tsx
```diff
@@ -0,0 +1,24 @@
...
+
+export type DetailViewProps = {
+  className?: string;
+  component: DetailOptions;
+  props: PropOptions;
+};
+
+function DetailModal({ className, props, component }: DetailViewProps) {
+  switch (component) {
+    case DetailOptions.YamlView:
```
jpellizzari, line 19: I think the `default` case here needs to return `null`. I am worried that when the modal gets closed, there is still something being rendered. If we don't return `null` when the `Drawer` is closed, old elements will appear there momentarily.

This also prevents weird state issues, because every time the `Drawer` gets opened, a new component will be instantiated.

jpellizzari: So I see below that the `Drawer` will get added/removed from the DOM. I wonder what happens when I go from one Yaml to another without closing it? Can I do that?

jpellizzari (approved):
Looking good! No blockers and and a nice use of `enum` here.

ui/lib/theme.ts
joshri: Here's a screen from a huge width - on small screens in OSS there'a min-width for the entire app that intentionally prevents any attempts at responsiveness - this doesn't exist in enterprise which is a much larger thing outside of this, but for now I've been told to not work on any views for mobile
image

ui/components/DetailModal.tsx
joshri: Totally fine returning null - the rest of the app is blocked when this modal is open - clicking anywhere outside of it will close the modal. Right now there's now way to navigate within the modal.

ui/components/Layout.tsx
```diff
@@ -104,7 +104,7 @@ function Layout({ className, children }: Props) {
   const { appState, setDetailModal } = useContext(AppContext);
 
   const detail = appState.detailModal;
-
+  console.log(appState);
```
jpellizzari, line 107: Leftover `console.log`

jpellizzari (approved):
Only note is what appears to be a leftover `console.log`. Then LGTM :+1:

PR #3465, opened 2023-03-01
docs: Add instruction for how to disable CAPI

Callisto13:
Even though it is tiny, I gave it its own page so that people could easily see it in the sidebar.

Closes https://github.com/weaveworks/weave-gitops-enterprise/issues/1339


foot (approved):
💯

PR #3467, opened 2023-03-01
docs: Tidy up the GitOpsTemplate docs

### LINKED ISSUE #3394: docs: Tidy up `Templates`

Callisto13:
This section was previously a giant wall of text which was quite hard to
read. I have broken it up into separate pages where it made sense, and
clarified some of the language. I have not changed much of the original
content itself.

Closes #3394


website/docs/gitops-templates/annotations.mdx
```diff
@@ -0,0 +1,34 @@
+---
```
makkes, line 1: For some reason I can't seem to find this page in the left nav of https://staging.docs.gitops.weave.works/revamp-template-docs/docs/next/gitops-templates/intro/

website/docs/gitops-templates/capd-example.mdx
```diff
@@ -0,0 +1,21 @@
+---
```
makkes, line 1: same here, it doesn't appear in the left nav.

website/docs/gitops-templates/annotations.mdx
Callisto13: huh 🤔 it's in the sidebars

 will take a look

Callisto13: definitely there

Screenshot 2023-03-03 at 15 06 24

website/docs/gitops-templates/capd-example.mdx
makkes: Ok, I think I know why I thought they weren't there: The file lives in ./gitops-templates/ but in the nav it's actually a level further down the hierarchy. Shouldn't the file locations represent the hierarchy?

Callisto13: They are required to represent the hierarchy if you are autogenerating the sidebars. But if you are not then it is optional.

I personally don't like mirroring it exactly, beyond grouping to a point, because it makes it easy to rearrange where they appear later on by just moving lines in the `sidebar.js`. BUT I will not be the one maintaining this, so will move if that is preferred.

makkes (approved):

website/docs/gitops-templates/capd-example.mdx
makkes: I'm fine either way. If it irritates someone in the future we can always change it.

PR #3468, opened 2023-03-01
Disconnect new nav from oss nav hooks and export it

### LINKED ISSUE #3479: Prepare new OSS nav for export
- Add all nav icons for enterprise nav to OSS IconType enum + svg code for hover and select states
- Make Nav and Logo "dumb" components so state can be supplied by enterprsie
- Celebrate

joshri:
Refs #3432 
Closes #3479 

What's happening in here? 
This is modifying and adding some components to OSS so they can be reused in Enterprise. 

Why?
Engineers have been doing double the work divided between two frontends for a while, and although sometimes they do need to be different, we should aim to reuse when we can. Duplicating styles not only adds more code to sift through, but has the potential to introduce a ton of hard to track bugs when a component works one way in OSS, and another in Enterprise. 

Summary of changes:
- Icons that are added as separate components are done this way so that the svg can be modified with a class (other svgs that have been imported into the Icon component as static images can not be modified). They have to be added to the IconType enum to be used, which means they must be added in OSS if they are to be used in a component that is expecting an IconType. 
- State (collapsed, route) has been moved out of Nav and Logo and into Layout, allowing Enterprise to pass down it's own separate values to the components. 
- the disabled prop in Nav allows for Enterprise to show/hide nav items based on feature flags. 
- a lot of additions to index.ts, and it's been alphabetized automatically by my linter which is why there are a lot of changes...I didn't choose these settings so...not pointing any fingers...but someone doesn't have their linter configured.
- I ordered theme.spacing since it's been bothering me for a looooong time.


opudrovs:
❓  Testing this PR. Noticed that in OSS only the icon part of the logo is clickable (but not the "weavegitops" text image), while in enteprise the whole logo (icon + text image) is clickable. What is the desired behavior which should be supported by the shared Logo component?

opudrovs:
💡 Would it be possible to make both the Collapse icon and "Collapse" text clickable to collapse the sidebar? Because not only the icon can be clicked for this, and it's small and hard to target.

opudrovs:
Ah, and probably in the logo only the icon is collapsible because after the sidebar is collapsed, the logo is collapsed to the icon only. But I still think that in case of the logo the whole image with text should be collapsible, I guess it's a de facto UI standard for similar logos.

opudrovs (approved):
LGTM! 🎉 Great job!

Tested it locally, works as expected.

The only minor questions are about clickable areas of the logo and collapse button.

PR #3469, opened 2023-03-02
Change refresh token to expire after the ID token.

bigkevmcd:
**What changed?**
This modifies the setting of the refresh token to expire after the ID token.

It also cleans up some code.

**Why was this change made?**
We are seeing users get logged out after the token cookie expiry time, and we think this is because the ID token is no longer available, and the refresh token can't be fetched.

Add an hour to the standard expiry time.

There's a separate discussion to be had around whether or not the cookie expiry should equal the expires time in the ID token...

Tests!

**Release notes**
Refresh token expiry should now allow refreshes to occur after the original cookie has expired.

**Documentation Changes**


foot (approved):
👌

PR #3481, opened 2023-03-03
Add clear docs link to navbar

Callisto13:
Also make it very clear that the current Chart ref is for OSS only (I would love to change the heading of the doc itself, but I cannot find which bit to configure in the autogen to make that happen).

image (6)

I have switched the `Getting Started` link to `Docs`. I don't want the nav bar getting too crowded, and the `Getting Started` is _right there_ when you click on the `Docs` link. Plus people will return to `Docs` and realistically only go to `Getting Started` once. That said, I can put both there if that is preferred, LMK.


makkes (commented):
✨

makkes (approved):

PR #3486, opened 2023-03-07
bump flux to v0.40.2

### LINKED ISSUE #3434: Bump Flux to the latest version
The Flux version used by Weave GitOps hasn't been updated for a while so we need to do that.

- Bump all flux dependencies in the go.mod file
- Update hard-coded versions in source code 
- Capture steps executed here to simplify the automation in #3277

chanwit:
Fixes #3434 

### For go.mod
1. Locate Flux version in `go.mod`
2. Change to a specific Flux version
3. Run `go mod tidy`

### For source controller version mapping to Flux:
1. Go to https://github.com/fluxcd/flux2/releases
4. Find the Flux version you want to support in the URL above.
5. Find the Source controller version which is used in the Flux version above. The Source version would normally be mentioned in the change log.
6. Don't forget to *remove the oldest mapping* from the map below

Idea to automate this part, we would probably use `docker` commands to pull Flux CLI, use it to generate manifest and locate the source-controller image.


opudrovs (approved):
LGTM! 🎉

Issue #2988, opened 2022-11-09
[Docs structure] Make Enterprise content easier to spot


dholbach:
Maybe it's sufficiently obvious to everyone else, but realised only when it was pointed out to me which parts of the content were about the Enterprise product and which about the open source project.

Right now we have these pieces of content that pertain to Enterprise
- everything in the Enterprise section
- complete sections like "Cluster Management"
- standalone docs like `deploying-capa`
- parts of docs like `installation/#installing-weave-gitops-enterprise`

Currently how you distinguish the two, you need to look out for paragraph headings like this:
![Bildschirmfoto vom 2022-11-09 14-09-54](https://user-images.githubusercontent.com/1346979/200839444-506aecd8-00c6-4b18-8405-fd411b4acaf0.png)

Maybe this could be made even clearer by using a different shade of background colour as well (ideally for both the content itself and the sidebar entries).

PR #3495, opened 2023-03-10
docs: POC to solve OSS vs EE versioning thing

Callisto13:
# DO NOT MERGE

## POC: Multi-Instance docs

This is a POC to demonstrate one way to solve the problem we have with
versioning the docs.

To recap:
- The docs are versioned to OSS
- But the majority of them refer to EE features
- It is not intuitive at all to know which version of the docs relates to which
  version of EE, which means it is very easy for users to select a version on
  the docs which has nothing to do with the version they are using.

The reasons why we are in this pickle are:
- OSS and EE are technically 2 entirely different products...
- But we want to document them all in the same place (because it makes sense for
  the product as a whole to do so)...
- So we have been copying many people who do this, notably Gitlab...
- A notable difference from Gitlab is that **they have all their code in one
  codebase**, meaning all tiers are versioned together.
- A non-docs solution to this has been floated: EE and OSS release at the same
  time. This will stop working the second EE needs to release an emergency patch
  on a Monday.

So basically until we find an engineering solution, we need to do something
about the docs before our customers notice that something doesn't line up.

So here we are.

### The solution

In this POC I am using a Docusaurus feature called [`multi-instance`](https://docusaurus.io/docs/next/docs-multi-instance).
This feature allows you to have multiple areas of the docs, each with **it's own
versioning system**.

This is without a doubt the **most underdocumented and pain in the ass thing to
configure** that I have yet come across in Docusaurus. I get the feeling like
they don't want people using it really? Or maybe the team were all part of the
Meta layoffs before they could write this up? idk.

Having said that, this is a feature which I imagine many people would love to
have work properly, so I think (hope) it will become less cumbersome in the
future.

See this video for how the multi-instance POC works.

https://user-images.githubusercontent.com/8898786/224337437-7791979c-9fbd-415f-9eb6-74ac83c13970.mov

The diff on this PR is large since I was reorganising everything. The main files
to care about are `sidebars.js`, `sidebarsEE.js`, `src/theme/NavbarItem/DocsVersionDropdownNavbarItem.js` and `docusaurus.config.js`.

 I have left OSS content in `docs` which is the default **and primary frame of
 reference for everything**, and moved all EE content under `enterprise`. You
 could go the other way and have EE be the default and put OSS in the new place.

 The Helm/CLI reference which I only recently added to the nav has had to be
 moved out in this solution because it only referenced the OSS chart/cli and
 there is no way to partially version a sidebar. I have tried to make the new
 reference location as clear as possible, but we may want to just have 2 things
 in the navbar for clarity.

 New versions can be created really easily in either OSS or EE's release process:

 ```bash
 # OSS
 yarn docusaurus docs:version 1.2.3
 # EE
 yarn docusaurus docs:version:enterprise 4.5.6
 ```

 The solution addresses some other common comments which I have heard a few times
 which are a) it's hard to tell which features are enterprise, and b) it would be
 nice if they had their on separate area or section in the docs.

 ### Things to know

 A couple of things which are either 'irritating' to 'unpleasant' when
 implementing multi-instance docs.

 #### Sidebars

 The navbar only recognises sidebars in `sidebars.js`, so any I tried to use from
 `sidebarsEE.js` were ignored. Sidebars are versioned with their instance so I
 could not get around this by combining them.

 #### Linking across instances

Linking across instances is weird, so far I have seen the following patterns:
- Link from `/docs/foo` to `/enterprise/bar` by markdown:
  `[link](@site/enterprise/bar)`
- Link from `/docs/foo` to `/enterprise/bar/index.mdx#heading` by markdown:
  `[link](../../enterprise/bar#heading)`
- Link from `/docs/foo` to `/enterprise/bar` at base of markdown `[link]:
  ../../enterprise/bar`
- Links from `<Link>` elements: `../../enterprise/bar`
- Links from sidebars have to be of type `html`:
    ```
    {
      type: 'link',
      label: 'Templates',
      href: '/enterprise/gitops-templates/intro'
    }
    ```
- Links from navbar have to be non-typed
    ```
    {
      position: 'left',
      label: 'Enterprise Docs',
      to: '/enterprise/intro-ee',
    }
    ```

The best thing about these rules are... they don't consistently apply! The links
which are failing in the build here are fine when I do `yarn build` locally 🙃 

#### Navbar version dropdowns

Obviously when you are on the EE instance you want to see the dropdown for EE
versions. The navbar is completely out of this system, however, so in order to
accomplish the desired UX, you have to wrap the Docusaurus navbar react with
some of your own. The navbar component is one of a few core components which
they advise you **not swizzling (wrapping) at all**. This is because there is a
risk the wrapped API could change in the future and upgrades could be a pain. In
our case we can _just about_ get away with it as it is literally 4 lines of code
so it will be pretty easy to spot breaks... but if the code grows to add more
customisations...

### Terraform controller

This is also a completely separate product with a totally different
version/release cycle. It would be sensible to do the same thing here if we want
to keep tf docs in with wg docs.


LappleApple:
Closing but adding a reference to this at #2988

PR #3498, opened 2023-03-10
Expand area that is clickable for logo and collapse in nav

### LINKED ISSUE #3497: Make logo text and collapse button text clickable in new nav
A `div` wrapping the logo text and collapse button text can fade these out without removing them from the DOM, while allowing them to still be clickable! 

AC:
- animation matches existing animation on main
- "Weave GitOps" and "collapse" are clickable

joshri:
Closes #3497 

New `Fade` styled component wraps collapse text and logo image - I didn't want to use the MUI version because it is super annoying about refs ala Popovers.

https://user-images.githubusercontent.com/65822698/224410517-16cfdba1-5a39-4eb6-ac3d-9d9f861a598b.mov

### DIFF
--- a/ui/components/Logo.tsx
+++ b/ui/components/Logo.tsx
@@ -3,6 +3,7 @@ import { Link } from "react-router-dom";
 import styled from "styled-components";
 import images from "../lib/images";
 import { V2Routes } from "../lib/types";
+import { Fade } from "../lib/utils";
 import Flex from "./Flex";
 
 type Props = {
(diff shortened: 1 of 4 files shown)


opudrovs (approved):
Look good to me! 🎉 

The only thing I would personally change is just wrapping the collapse icon and text with div to have a simple rectangular button clickable/with a hand cursor appearing over it instead of two separate areas of different height for the icon and the Collapse text. The Collapse element looks like a regular button to me. Please see the attached recording (no idea why it's not embedded properly).

The same is happening with the logo, but it's the same thing we have in enterprise now and logo elements are larger, so it's easier to target them with the mouse. The Collapse elements are smaller.

 But it is subjective.

https://user-images.githubusercontent.com/8824708/224421100-2e59de5f-ab69-4e7d-9cc8-cb086270c590.mov

opudrovs:
Oh, it got embedded successfully!

PR #3517, opened 2023-03-14
Speedup resource details

luizbafilho:
**What changed?**
- Adds a new `GetInventory` endpoint, that aims to increase the speed of the details page by fetching the inventory more efficiently by avoiding using List calls to k8s API and doing Gets as much as possible.

**Why was this change made?**

**Release notes**

**Documentation Changes**


Tiltfile
```diff
@@ -18,7 +18,7 @@ if advanced_go_dev_mode:
 
     local_resource(
         'gitops-server',
-        'GOOS=linux GOARCH=amd64 make gitops-server',
+        'GOOS=linux make gitops-server',
```
yitsushi, line 21: Why was it even there (GOARCH)? :O that would mean amd64 was used on newer mac laptops too (M1 chip)

yitsushi (approved):
Server side looks good to me, front-end... looks good too, but that doesn't mean to much from me :laughing:

PR #3524, opened 2023-03-15
Add detail page for Pods

### LINKED ISSUE #3514: Pod Detail View
AC:
- Add `Pod` to Kind enum and add fields to `FluxObject` class based on Lens CPU tab
- PodDetail component as an option inside DetailModal instead of YamlView
- YamlView still in a tab

joshri:
Closes #3514
Includes partial fix for #3561 

REMIX WITH CONDITIONS TABLE:

https://user-images.githubusercontent.com/65822698/227037059-2932968c-6e7e-43bc-a47a-b25d57d4ec5b.mov

HOT OFF THE PRESSES:
image

OLD:
image

Adds Pod to Kind enum and creates FluxObject class extension
Adds functionality for DetailModal to display other components
Adds PodPage with API request for object (Runs one time, no polling).
Adds PodDetail with basic display of info utilizing mostly existing components. 
Moves class overrides of MuiTabs into theme and Tab into a separate component to reduce repetition (can't reuse SubRouterTabs because these tabs are not using routing)
Adds functionality to PageStatus component to display a table of all conditions with a prop


JamWils:
Oh, why is there no status message then?

Ack on the table, maybe a separate table component so you don't have to make the other table be "one size fits all".

joshri:
I don't know why there's no message - this is a pod in the flux-system kustomization that's created automatically from the `reboot.sh` script in enterprise which we're all using. I can look into if it's normal to have undefined messages but maybe it's something we should handle regardless? Could be helpful for people starting out.

joshri:
Updated - I think the border around each container is a failed experiment but there has to be SOMETHING separating them.

opudrovs:
@joshri some checks are failing. Did you run `make proto`?

opudrovs (approved):
Tested it locally, everything is working as expected. Great job! 🎉 

Screenshots of the UI on regular laptop and narrow screens:

Screenshot 2023-03-17 at 17 15 07

Screenshot 2023-03-17 at 17 15 18

Screenshot 2023-03-17 at 17 15 34

Screenshot 2023-03-17 at 17 15 46

opudrovs:
@joshri Is the `Message not found` status message expected?

opudrovs:
@JamWils ^^^

joshri:
Re: Message not found - I added that to the PageStatus component to handle when no message comes back from `computeMessage` (i.e. Conditions objects have no message).

opudrovs:
@joshri is this message displayed for all pods or do some pods have the computable message in the status conditions? Because now it reads as an error even though it is marked with green. It might be a bit confusing to the user.

ui/components/PageStatus.tsx
```diff
@@ -15,7 +15,8 @@ type StatusProps = {
 };
 
 function PageStatus({ conditions, suspended, className }: StatusProps) {
-  const msg = suspended ? "Suspended" : computeMessage(conditions);
+  let msg = suspended ? "Suspended" : computeMessage(conditions);
+  if (!msg) msg = "Message not found";
```
opudrovs, line 19: 💡 This line can be replaced with a short-circuit assignment:
```
msg = msg || "Message not found";
```

joshri:
All of my pods in the `flux-system` kustomization have no message in any conditions objects. Maybe just "No message" instead of "No message found"? I think just having the icon there also looks like an error, so I'd like to put something there

joshri:
@opudrovs - 
James and I decided to try out a `Conditions` table that expands on click for this page specifically which handles the issue with no message on page status - this is new code in the `PageStatus` component since the last time you reviewed. 

To test it out, uncomment the case in `DetailModal` that gives access to the PodDetail page (also import `PodPage` into `DetailModal` again) 

We're planning to merge this code without giving access to the feature while product decides whether to continue with this new detail pages initiative.

opudrovs:
@joshri fine with me! I already approved this PR. Should I re-test it or will you do it yourself?

And some checks are failing. Probably another issue with the version of protobuf generator. Should I run and push `make proto`?

opudrovs:
The checks have passed, cool!

Issue #3530, opened 2023-03-16
Wrong kubectl command generated in the UI for non-namespaced resources


opudrovs:
<img width="1520" alt="Screenshot 2023-03-15 at 13 02 22" src="https://user-images.githubusercontent.com/8824708/225774744-9bbd7c0e-f817-4ce2-a06f-63ecc19d7222.png">

There is an error when using a copied command from the UI for non-namespaced resources (fro example, ClusterRoles):

```
kubectl get clusterrole flux-edit-flux-system -n  -o yaml


olga@macbook-pro ~ % kubectl get clusterrole flux-edit-flux-system -n  -o yaml 
NAME                    CREATED AT
flux-edit-flux-system   2023-03-15T11:54:28Z
Error from server (NotFound): clusterroles.rbac.authorization.k8s.io "yaml" not found
olga@macbook-pro ~ % kubectl get clusterrole flux-edit-flux-system -n  -o yaml 
NAME                    CREATED AT
flux-edit-flux-system   2023-03-15T11:54:28Z
Error from server (NotFound): clusterroles.rbac.authorization.k8s.io "yaml" not found
```

After fixing the command (removing `-n` without a namespace):

```
olga@macbook-pro ~ % kubectl get clusterrole flux-edit-flux-system  -o yaml
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  creationTimestamp: "2023-03-15T11:54:28Z"
  labels:
    app.kubernetes.io/instance: flux-system
    app.kubernetes.io/part-of: flux
    app.kubernetes.io/version: v0.41.1
    kustomize.toolkit.fluxcd.io/name: flux-system
    kustomize.toolkit.fluxcd.io/namespace: flux-system
    rbac.authorization.k8s.io/aggregate-to-admin: "true"
    rbac.authorization.k8s.io/aggregate-to-edit: "true"
  name: flux-edit-flux-system
  resourceVersion: "1719"
  uid: 5126ada9-dadb-4fad-b127-0409df0d6737
rules:
- apiGroups:
  - notification.toolkit.fluxcd.io
  - source.toolkit.fluxcd.io
  - helm.toolkit.fluxcd.io
  - image.toolkit.fluxcd.io
  - kustomize.toolkit.fluxcd.io
  resources:
  - '*'
  verbs:
  - create
  - delete
  - deletecollection
  - patch
  - update
olga@macbook-pro ~ % 
```

[...]

PR #3539, opened 2023-03-17
Support yaml commands for objects without a namespace

joshri:
Refs #3530 

Creates a utility function that is exportable for the yaml command. Will import in enterprise once this is merged.

### DIFF
--- a/ui/lib/__tests__/utils.test.ts
+++ b/ui/lib/__tests__/utils.test.ts
@@ -1,8 +1,10 @@
-import { Automation, HelmRelease, Kustomization } from "../objects";
 import { GetVersionResponse } from "../api/core/core.pb";
+import { Kind } from "../api/core/types.pb";
+import { Automation, HelmRelease, Kustomization } from "../objects";
 import {
   convertGitURLToGitProvider,
   convertImage,
+  createYamlCommand,
   formatLogTimestamp,
   formatMetadataKey,
   getAppVersion,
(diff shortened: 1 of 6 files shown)


ui/lib/__tests__/utils.test.ts
```diff
@@ -387,3 +389,21 @@ describe("utils lib", () => {
     });
   });
 });
+
+describe.only("createYamlCommand", () => {
```
opudrovs, line 393: ✨ Nice tests!

ui/lib/utils.ts
```diff
@@ -204,6 +204,18 @@ export function formatLogTimestamp(timestamp?: string, zone?: string): string {
...
+export const createYamlCommand = (
+  kind: string,
+  name: string,
+  namespace: string
+): string => {
+  if (kind && name) {
+    const namespaceString = namespace ? ` -n ${namespace}` : "";
+    return `kubectl get ${kind.toLowerCase()} ${name}${namespaceString} -o yaml`;
+  }
+  return null;
```
opudrovs, line 216: 💡 These lines can be simplified as:

```
if (!kind || !name) return null;

  const namespaceString = namespace ? ` -n ${namespace}` : "";
  return `kubectl get ${kind.toLowerCase()} ${name}${namespaceString} -o yaml`;
```

opudrovs (approved):

Issue #3499, opened 2023-03-10
Cannot Turn User Auth off on Sign In Page


scott-kausler:
**Describe the bug**

**Environment**
Kubernetes/OpenShift

- Weave-Gitops Version
0.18.0

- Flux Version
0.36.0

- Kubernetes version
1.24

**To Reproduce**
Set the following in the gitops-server helm chart and deploy
```
    envVars:
      - name: WEAVE_GITOPS_FEATURE_CLUSTER_USER_AUTH
        value: "false"
      - name: CLUSTER_USER_AUTH
        value: "false"
```

**Expected behavior**

The basic auth Username/Password fields do not appear on the sign in page.

**Actual Behavior**

The basic auth Username/Password fields DO appear on the sign in page.

foot:
Thanks for the report! Will see if I can repro.

Might be a bad comparison here https://github.com/weaveworks/weave-gitops/blob/f834a8a1b66cd724c62db61568bc757ee9d0dcdf/ui/pages/SignIn.tsx#L141 🤔

tun0:
> Thanks for the report! Will see if I can repro.
> 
> Might be a bad comparison here
> 
> https://github.com/weaveworks/weave-gitops/blob/f834a8a1b66cd724c62db61568bc757ee9d0dcdf/ui/pages/SignIn.tsx#L141

The three conditionals for the login stuff all resolve to `True`, always. Without configuring OIDC auth, both ways show up on the login screen as well.

foot:
Thanks for testing.

`flags.CLUSTER_USER_AUTH === "true"` might be the better comparison here, flags often come from env vars and are all strings.

Or perhaps adding a helper, to avoid this mistake in the future too.

```tsx
const { isFlagEnabled } = useFeatureFlags()
...
{isFlagEnabled("CLUSTER_USER_AUTH") ? (
...
```

PR #3551, opened 2023-03-21
Move feature flag status check to useFeatureFlags hook

AlinaGoaga:
Refactor following the conversation on https://github.com/weaveworks/weave-gitops/issues/3499

**What changed?**

**Why was this change made?**

**Release notes**

**Documentation Changes**

### DIFF
--- a/ui/hooks/featureflags.ts
+++ b/ui/hooks/featureflags.ts
@@ -5,5 +5,8 @@ export type FeatureFlags = { [key: string]: string };
 
 export function useFeatureFlags() {
   const { featureFlags } = useContext(CoreClientContext);
-  return { data: { flags: featureFlags || {} } };
+
+  const isFlagEnabled = (flag: string) => featureFlags?.[flag] === "true";
+
+  return { isFlagEnabled };
 }
--- a/ui/components/AlertsTable.tsx
+++ b/ui/components/AlertsTable.tsx
@@ -33,15 +33,15 @@ export const makeEventSourceLink = (obj: CrossNamespaceObjectRef) => {
 };
 
 function AlertsTable({ className, rows = [] }: Props) {
-  const { data: flagData } = useFeatureFlags();
-  const flags = flagData.flags;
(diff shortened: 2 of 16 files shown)


ui/hooks/featureflags.ts
```diff
@@ -5,5 +5,12 @@ export type FeatureFlags = { [key: string]: string };
...
   const { featureFlags } = useContext(CoreClientContext);
-  return { data: { flags: featureFlags || {} } };
+
+  const isFlagEnabled = (flag: string) => {
+    if (flag === "WEAVE_GITOPS_FEATURE_OIDC_BUTTON_LABEL") {
+      return featureFlags?.["WEAVE_GITOPS_FEATURE_OIDC_BUTTON_LABEL"] !== "";
+    } else return featureFlags?.[flag] === "true";
+  };
+
+  return { isFlagEnabled };
```
foot, line 15: WDYT about still exposing `flags` so we don't have to special case things in the `isFlagEnabled` helper?

- Most flag logic can use `isFlagEnabled` as its usually a string-bool
- Some flags will have other values like labels in them and they can still be accessed via `flags`?

```suggestion
  return { isFlagEnabled, flags };
```

ui/hooks/featureflags.ts
```diff
@@ -5,5 +5,8 @@ export type FeatureFlags = { [key: string]: string };
 
 export function useFeatureFlags() {
   const { featureFlags } = useContext(CoreClientContext);
-  return { data: { flags: featureFlags || {} } };
+
+  const isFlagEnabled = (flag: string) => featureFlags?.[flag] === "true";
+
+  return { isFlagEnabled, flags: featureFlags };
```
foot, line 11:
```suggestion
  return { isFlagEnabled, flags: featureFlags || {} };
```

To preserve old behaviour? I'm not sure what the loading pattern here looks like, if we might get undefined until flags are loaded etc?

foot (commented):
Code looks great! Will take it for a spin..

Issue #3519, opened 2023-03-14
Allow option to set custom root CA


padlar:
**Problem**
We have configured a dex openshift provider using custom CA in our setup and we get the following error when we run the weave gitops towards it -
```
Error: could not initialise authentication server: could not create auth server: could not create provider: Get "https://dex.apps.<domain>/.well-known/openid-configuration": x509: certificate signed by unknown authority
```

**Solution**
Solution would be to provide an option to trust custom CA in the gitops image. This can be achieved by reading in the CA bundle via ENV variable or using a trust-manager to support multiple root CAs (https://cert-manager.io/docs/projects/trust-manager/).

padlar:
As suggested by @bigkevmcd, custom CA can be used today with out any changes to the code by doing the following:

Create a secret with the custom CA and add env, extraVolumes and extraVolumeMounts to values.yaml as shown below -

```bash
apiVersion: v1
data:
  ca.crt: |
    RXdKT1R6RVJNQThHQTFVRUNoTUlVM1JoZEc1bGRIUXhHekFaQmdOVkJBTVRFbE4wWVhSdQ
    paWFIwSUZKdmIzUWdRMEVnTWpBZUZ3MHhNVEEwTVRNd056UXpNREZhRncwek5qQXlNVFl4
    TVRJd05URmFNRDB4CkN6QUpCZ05WQkFZVEFrNVBNUkV3RHdZRFZRUUtFd2hUZEdGMGJtVj
    BkREViTUJrR0ExVUVBeE1TVTNSaGRHNWwKZEhRZ1VtOXZkQ0JEUVNBeU1JSUJJakFOQmdr
    LS0tLQo=
kind: Secret
metadata:
  name: custom-root-ca
type: Opaque
```

```bash
    envVars:
      - name: SSL_CERT_FILE
        value: /etc/ssl/certs/ca.crt
    extraVolumes:
      - name: custom-ca-cert
        secret:
          secretName: custom-root-ca
    extraVolumeMounts:
      - name: custom-ca-cert
        mountPath: /etc/ssl/certs
        readOnly: true
```

PR #3563, opened 2023-03-27
add an option to set custom root CA (#3519)

padlar:
Closes 3519

**What changed?**

New option was added to specify custom Root CA path.

**Why was this change made?**

This change allows to communicate with OIDC provider with custom signed certificates.

New CLI option is added. Helm chart changes are not included in this change - will submit a follow-up PR.

All the unit tests passed.

**Release notes**

**Documentation Changes**

### DIFF
--- a/cmd/gitops-server/cmd/cmd.go
+++ b/cmd/gitops-server/cmd/cmd.go
@@ -62,6 +62,7 @@ type Options struct {
 	MTLS        bool
 	TLSCertFile string
 	TLSKeyFile  string
+	TLSRootCA   string
 	// Stuff for profiles apparently
 	HelmRepoName      string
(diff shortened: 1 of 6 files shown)


bigkevmcd:
There is a simpler way to do this, you can mount the CA and then set `SSL_CERT_FILE ` to point to it.

https://pkg.go.dev/crypto/x509@go1.20.2#SystemCertPool

This doesn't require _any_ changes to the code.

The Helm chart supports it already, via `extraVolumeMounts` and `extraVolumes` and `envVars`.

padlar:
Closing this as this is already supported in weave-gitops today. Thanks to @bigkevmcd for the tips.

PR #3565, opened 2023-03-27
Get inventory updates

luizbafilho:
Closes

**What changed?**
- adding info back that was removed previously. The UI need it.
- adding missing items to the root of the object so that in EE it didn't crash

**Why was this change made?**
- **[UI]** Convert **Entries** to FLuxObject not just spreading the output and this make things easy for UI to use what FluxObject provides .
- **[UI]** Adding `clusterName` to the response will avoid EE (multi-cluster) crash as now clusterName is defined 

**Release notes**

**Documentation Changes**

### DIFF
--- a/pkg/server/auth/jwt.go
+++ b/pkg/server/auth/jwt.go
@@ -74,7 +74,7 @@ func NewJWTAuthorizationHeaderPrincipalGetter(log logr.Logger, verifier tokenVer
 }
 
 func (pg *JWTAuthorizationHeaderPrincipalGetter) Principal(r *http.Request) (*UserPrincipal, error) {
-	pg.log.Info("attempt to read token from auth header")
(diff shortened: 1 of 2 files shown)


joshri (changes requested):
❌ I still have a blocker from the original PR: changing the `reconciledObjectAutomation` prop to individual fields breaks the usage of `ReconciledObjectsTable` in Enterprise. 

It's okay to change the prop (although we need an immediate followup PR in Enterprise ASAP), but I'm wondering if it's possible to just spread the automation as the prop `...automation` instead of listing out a billion props on the table AND the graph. What do we gain by changing it in this way?

Also from the original PR - `FluxObjectsTable` still cheats the objects prop using `as FluxObject` - we should know that those are FluxObjects and if they aren't that isn't good right?

PR #3583, opened 2023-03-29
Split up DataTable component and add column click handler (v2)

jpellizzari:
Part of https://github.com/weaveworks/weave-gitops-enterprise/issues/2609

**Save until after the release**

Breaks up the `DataTable` component without affecting the modules that import it. Also adds some sort skipping logic for when we want to sort elsewhere.

I had originally intended to break up `DataTable` further but it was too high-risk.

### DIFF
--- a/ui/components/DataTable/DataTable.tsx
+++ b/ui/components/DataTable/DataTable.tsx
@@ -12,41 +12,30 @@ import qs from "query-string";
 import * as React from "react";
 import { useHistory, useLocation } from "react-router-dom";
 import styled from "styled-components";
-import Button, { IconButton } from "./Button";
-import CheckboxActions from "./CheckboxActions";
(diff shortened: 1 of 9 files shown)


TheGostKasper (approved):
LGTM, 
What do you think of moving Props type to `type` and create a style file that contains all related styles to Datatable ?

I believe it's good to follow the same structure as it's in this PR 👏 .

jpellizzari:
> LGTM, What do you think of moving Props type to `type` and create a style file that contains all related styles to Datatable ?
> 
> I believe it's good to follow the same structure as it's in this PR clap .

It still made sense to keep `Props` with the `DataTable` I think. As far as styles go, I would need to see what that looks like with `styled-components`. Typically, we would want to couple styles with their components, so a centralized style file might be odd?

PR #3584, opened 2023-03-30
Add Text component to use in EE

AlinaGoaga:


ui/components/Flex.tsx
```diff
@@ -31,9 +32,11 @@ const Styled = (component) => styled(component)`
 
 class Flex extends React.PureComponent<Props> {
   render() {
-    const { className, children, onMouseEnter, onMouseLeave } = this.props;
+    const { className, children, onMouseEnter, onMouseLeave, testId } =
+      this.props;
```
foot, line 36: This is forwarding all props, can we do that and give Props the same type as `div`'s Props?

AlinaGoaga: I've adjusted it a bit :) , can't pass all props as that includes `tall, wide, wrap` etc which div can't use

foot: 🎉 .

Ah yeah, I seeeeee.

Hmm, can we collapse the whole thing down?

```tsx

const Flex = styled.div<Props>`
  display: flex;
  flex-direction: ${({ column }) => (column ? "column" : "row")};
  align-items: ${({ align }) => (align ? "center" : "start")};
  ${({ tall }) => tall && `height: 100%`};
  ${({ start }) => start && "justify-content: flex-start"};
  ${({ between }) => between && "justify-content: space-between"};
  ${({ center }) => center && "justify-content: center"};
  ${({ shadow }) => shadow && "box-shadow: 5px 10px 50px 3px #0000001a"};
`;

export default Flex;
```

kind of thing?

foot (approved):
💯

PR #3590, opened 2023-04-03
Resources health checks

luizbafilho:
Closes [#2596](https://github.com/weaveworks/weave-gitops-enterprise/issues/2596)

**What changed?**
- Adds a health checker for some resources.
- Initial support to:
  - Deployments
  - ReplicaSets
  - Pods

**Why was this change made?**

**Release notes**

**Documentation Changes**


joshri:
Front end looks okay but I can't test yet as OSS main appears to broken right now

yiannistri:
@joshri can you take another look now?

joshri:
What’s the difference between health check and what we do for status? I’m mainly asking bc the front end code ends up looking very similar to what we use for the status indicator, and I’m wondering if it can be reused - at least the JSX should wind up the same to reuse styles. 

Also - using the full message makes the column realllllly wide in the table. Is it correctly prioritized over status and message? Just expressing my overall desire to get rid of the horizontal scrolling

Are there plans to add this info to the detail pages?

TheGostKasper:
[UI] 
Both `KubeStatusIndicator` & `HealthCheckStatusIndicator` are pretty much the same in structure and naming Convension , different in dealing with props as the main diff is `healthChecks` doesn't deal with conditions to retrieve the text as it's in `KubeStatusIndicator`.
 Why not just change `KubeStatusIndicator` to cover `healthChecks` ?
 - _StatusIndicator_ is used everywhere in the system and doesn't worth the change .
 -  Creating new component that is responsible for `healthChecks` is more reasonable to cover specific job as in the near future we will update to it and use it in other pages .

@joshri

joshri:
Okay no worries - just trying to get some context as to what's going on. I was able to see the column and indicator for a Deployment in the UI, so consider the front end approved 👍

yiannistri (approved):
Looks good, I particularly liked how we go about testing this 👍

PR #3597, opened 2023-04-05
Fix concurrently refreshing the ID token

bigkevmcd:
Closes

**What changed?**
This serialises the ID token refresh using the refresh token, doing it concurrently was leading to errors which were returned as HTTP 40x.

The serialisation key is the refresh token.

**Why was this change made?**
To reduce conflicts when refreshing the token.

However, ideally we'd control this from the server-side.

Putting in a per-refresh-token sync mutex which is locked.

Tests and running the code.

**Release notes**

**Documentation Changes**


pkg/server/auth/auth.go
```diff
@@ -198,29 +198,58 @@ func WithAPIAuth(next http.Handler, srv *AuthServer, publicRoutes []string) http
...
+
+func (a *authenticatedMiddleware) ServeHTTP(rw http.ResponseWriter, r *http.Request) {
+	if IsPublicRoute(r.URL, a.publicRoutes) {
+		a.next.ServeHTTP(rw, r)
+		return
+	}
+
+	principal, err := a.principalGetter.Principal(r)
+
+	if refreshToken := getRefreshToken(r); err != nil || principal == nil && refreshToken != "" {
```
foot, line 235: I think we still wanna get into this block if refreshToken is empty, the `"Authentication required"`  inside this block is sort of the global 401

bigkevmcd: Slightly worrying that no tests fail in this case :-|

foot (approved):
Works a treat!

PR #3605, opened 2023-04-11
Allow group claim to be string, not only []string

asaf400:
Closes
No open issue exists for this PR.

**What changed?**
When using an OIDC Provider that filters the groups associated with the user principal,
for instance Jumpcloud, it is possible for the groups claim* to have a single string value, and not a list of strings or a list of one string, i.e:
claims[groupsKey] = ["group1"]
claims[groupsKey] = "group1"

This can occur when the user is only assigned to a single group in the OIDC Directory relevant to the Weave Gitops OIDC Application.

\* no matter the attribute name [groups | memberOf | other custom key]

**Why was this change made?**
I encountered this issue with the error message:
`failed to query user info endpoint: invalid groups claim "groups" in response Flux-Admins`

and so, in order to tix the issue, I have changed it so that when the type assertion for v.([]interface{}) fails, 
it is assumed that v might be a string, and is then type asserted that it is, and it is appended to groups as a single item.

it took me a while to get a hang on some of the golang idiom used in the code,
but I think that the end result is quite clean.

I was developing the change while debugging the tests,
and so I was able to:
- manufacture a new test for the new code
- locally test using a real actual JWT token from my OIDC (had to skip a few JWT  validations, but that's a given)
- make sure all everything works as expected with various inputs with my debugging sessions

**Release notes**
Adds support for OIDC providers that may send a string value for groups instead of a single element list.

**Documentation Changes**

Edits by maintainers are welcome

### DIFF
--- a/pkg/server/auth/claims.go
+++ b/pkg/server/auth/claims.go
@@ -45,7 +45,7 @@ func (c *ClaimsConfig) PrincipalFromClaims(token claimsToken) (*UserPrincipal, e
 	groups := []string{}
 
 	if v, ok := claims[groupsKey]; ok {
-		
+
 		gv, ok := v.([]interface{})
 
 		if ok {
@@ -57,7 +57,7 @@ func (c *ClaimsConfig) PrincipalFromClaims(token claimsToken) (*UserPrincipal, e
 				groups = append(groups, s)
 			}
 		} else {
-			if s, ok := v.(string); ok && len(s) > 0{
+			if s, ok := v.(string); ok && len(s) > 0 {
 				groups = append(groups, s)
 			} else {
 				return nil, fmt.Errorf("the groups claim %q is an empty value", groupsKey)
--- a/pkg/server/auth/claims.go
+++ b/pkg/server/auth/claims.go
@@ -45,16 +45,22 @@ func (c *ClaimsConfig) PrincipalFromClaims(token claimsToken) (*UserPrincipal, e
 	groups := []string{}
 
 	if v, ok := claims[groupsKey]; ok {
+		
 		gv, ok := v.([]interface{})
-		if !ok {
-			return nil, fmt.Errorf("invalid groups claim %q in response %v", groupsKey, v)
-		}
 
-		for _, v := range gv {
-			if s, ok := v.(string); !ok {
-				return nil, fmt.Errorf("invalid groups claim %q in response %v", groupsKey, v)
-			} else {
+		if ok {
+			for _, v := range gv {
+				if s, ok := v.(string); !ok {
+					return nil, fmt.Errorf("invalid groups claim %q in response %v", groupsKey, v)
+				} else {
+					groups = append(groups, s)
+				}
+			}
+		} else {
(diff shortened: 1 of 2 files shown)


pkg/server/auth/claims.go
```diff
@@ -45,16 +45,22 @@ func (c *ClaimsConfig) PrincipalFromClaims(token claimsToken) (*UserPrincipal, e
...
 
-		for _, v := range gv {
-			if s, ok := v.(string); !ok {
-				return nil, fmt.Errorf("invalid groups claim %q in response %v", groupsKey, v)
-			} else {
+		if ok {
+			for _, v := range gv {
+				if s, ok := v.(string); !ok {
+					return nil, fmt.Errorf("invalid groups claim %q in response %v", groupsKey, v)
+				} else {
```
bigkevmcd, line 55: You don't need an `else` here or below, as you're returning.

https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow

bigkevmcd (changes requested):
Looks good, thanks for the contribution.

Given the nature of claims, it's unfortunate that the parsing is the responsibility of the client, which leads to this sort of oddness.

I'd be happy to approve with a small tidyup.

pkg/server/auth/claims.go
asaf400: Yeah, 100% fair point, missed that :)

asaf400: I can do the same for the 2nd if, 
But I actually prefer that the condition is truthful with AND `ok && len(s) > 0`
as opposed to `!ok || len(s) == 0`.

asaf400:
updated with main branch via rebase

asaf400:
@bigkevmcd Do you know using why either:
`make local-docker-image` or `make docker-gitops-server` 
produced a docker image that failed to start the container when dropped in-place of the official image (using the helm chart in k8s) ?
I would guess that's something to do with the official docker build args, but didn't expect that

The error was `could not initialise authentication`
from https://github.com/weaveworks/weave-gitops/blob/main/cmd/gitops-server/cmd/cmd.go#L182:L183

bigkevmcd:
@asaf400 I think you were unlucky, there was a change that was reverted that _may_ have been causing your problem.

You might want to try again, and I see our pre-integration checks failing because of some minor formatting issue which feels fixable.

asaf400:
> @asaf400 I think you were unlucky, there was a change that was reverted that _may_ have been causing your problem.
> 
> You might want to try again, and I see our pre-integration checks failing because of some minor formatting issue which feels fixable.

Yeah, I just had to run `gofmt -w`
I was actually looking for the standard formatter for this project, but I missed the Makefile option for checking the format :)

asaf400:
updated with main branch once again

bigkevmcd (approved):
@asaf400 Thanks for the contribution, I'm happy with this.

asaf400:
@bigkevmcd I confirmed it's working by building the docker image, and launching it with the latest helm chart,
customizing the image used to our own ecr registry.

PR #3610, opened 2023-04-12
Wge 2693 - explorer documentation

enekofb:
Closes https://github.com/weaveworks/weave-gitops-enterprise/issues/2693

**What changed?**

User documentation for explorer

**Why was this change made?**

Before onboarding early adopters we should have user documentation ready. This PR takes cares of it. 

Locally 

Screenshot 2023-04-14 at 10 00 18


jpellizzari (approved):
LGTM. We can refine the larger story when we start adding more objects.

website/docs/explorer/configuration.mdx
```diff
@@ -0,0 +1,82 @@
...
+
+Explorer leverages existing authentication and authorization built-in the [application](https://docs.gitops.weave.works/docs/configuration/securing-access-to-the-dashboard/).
+It identifies for a user logged in the application: its identity and the access permissions via Kuberentes RBAC.
+Query results are filtered honouring the access determined via RBAC.
+
+## Authentication and Authorization for collecting
+
+As mentioned earlier, [GitopsClusters](https://docs.gitops.weave.works/docs/next/cluster-management/managing-existing-clusters/#connect-a-cluster)
+defines the connection and security context that explorer will use to collect data from the leaf clusters.
+Therefore, you should ensure that the user has access to watch the [Kinds](#kinds) that you would like explorer to serve.
```
JamWils, line 76: Is the collector, collecting these objects on a per user basis, or are we talking about the permissions on the `GitOpsClusters` object?  To me this reads that if Person A wants `Kustomizations`, and Person B wants `Kustomizations` and `HelmReleases` then they need to be set up with the "watch" permission?

enekofb: Agree that it is confusing: i have removed it. 

The collection is aync and out of the user context so there should not be reference to the user in this part. 
I do expect that out of this [issue](https://github.com/weaveworks/weave-gitops-enterprise/issues/2732) this section gets extended.

Issue #3585, opened 2023-03-31
Add OIDC Prefix support for impersonation


adberger:
**Problem**

We currently use GitLab as an OIDC provider for logging in to our Kubenetes clusters.
Our kube-apiserver is configured with `--oidc-username-prefix` & `--oidc-groups-prefix` (https://kubernetes.io/docs/reference/access-authn-authz/authentication/#configuring-the-api-server)

Now we set up the Weave Gitops UI to also use GitLab OIDC.
The problem now is, that Weave GitOps tries to impersonate without the prefix on the cluster and therefore fails.

Example:
GitLab group name: `sre-admins`
Kubernetes group claim: `idp_sre-admins`

We now see the `sre-admins` group in the Weave Gitops UI when we are logged in, and the impersonation tries to access resources as `sre-admins` instead of `idp_sre-admins`.

Does this description makes sense?

**Solution**

Possibly add another flag to set prefix for the OIDC?
Other solutions welcome

foot:
Hi @adberger, thanks for the issue!

Yes, this definitely makes sense and is something we want to add.

Probably into [OIDCConfig](https://github.com/weaveworks/weave-gitops/blob/15becb29f48663bf3e1ad70831d5d7b986b65910/pkg/server/auth/server.go#L109) secret alongside the other configuration.

foot:
- Read prefix values out of OIDCConfig secret
- When impersonating add the appropriate prefixes

PR #3611, opened 2023-04-12
Add OIDC prefix support for impersonation

sarataha:
Closes https://github.com/weaveworks/weave-gitops/issues/3585

**What changed?**
This PR adds support for configuring OIDC username and groups prefixes. Changes:
* Add two new flags, `oidc-username-prefix` and `oidc-groups-prefix`.
* Read prefix values out of OIDCConfig secret.
* Update the impersonating client to add the prefixes.

**Why was this change made?**
To allow for impersonation with OIDC prefixes.

Added a test case for `ConfigWithPrincipal` to ensure prefixes were added correctly.


pkg/kube/config_getter.go
```diff
@@ -7,6 +7,11 @@ import (
 	"k8s.io/client-go/rest"
 )
 
+type OIDCPrefixes struct {
```
bigkevmcd, line 10: This is exported, so should have a doc-comment.

But, it also feels like these are not `OIDCPrefixes` at this point, yes, we might use these to match against groups that come from OIDC, but these are not related to OIDC?

Is there a better name?

pkg/kube/config_getter.go
```diff
@@ -20,14 +25,15 @@ var _ ConfigGetter = &ImpersonatingConfigGetter{}
 // principal and if it finds one, it configures the *rest.Config to impersonate
 // that principal. Otherwise it returns a copy of the base config.
 type ImpersonatingConfigGetter struct {
-	insecure bool
-	cfg      *rest.Config
+	insecure     bool
+	cfg          *rest.Config
+	oIDCPrefixes OIDCPrefixes
```
bigkevmcd, line 30: Should this be `OIDCPrefixes` ?

pkg/kube/config_getter_test.go
```diff
@@ -78,3 +78,38 @@ func TestImpersonatingConfigGetterNoPrincipalInContext(t *testing.T) {
...
+	if cfg.Impersonate.UserName != expectedUser {
+		t.Fatalf("cfg username didn't match expected: %s", cfg.Impersonate.UserName)
+	}
+
+	expectedGroups := []string{"prefix-group1", "prefix-group2"}
+	if diff := cmp.Diff(expectedGroups, cfg.Impersonate.Groups); diff != "" {
+		t.Fatalf("cfg groups didn't match expected:\n%s", diff)
+	}
+
+	// Test with multiple calls.
```
bigkevmcd, line 108: You don't need to do this 10 times, you can just do it again :-)

bigkevmcd (changes requested):

pkg/kube/config_getter.go
sarataha: Makes sense to me 🙂 Is `UserPrefixes` suitable?

pkg/kube/config_getter_test.go
sarataha: Yes thank you for point this out 🙂 🙏

bigkevmcd (approved):
Looks good to me.

This will require changes to the enterprise product to pass these through too.

cmd/gitops-server/cmd/cmd.go
```diff
@@ -187,7 +191,12 @@ func runCmd(cmd *cobra.Command, args []string) error {
 
 	ctx := context.Background()
 
-	cl, err := cluster.NewSingleCluster(cluster.DefaultCluster, rest, scheme, cluster.DefaultKubeConfigOptions...)
+	oidcPrefixes := kube.UserPrefixes{
+		UsernamePrefix: options.OIDC.UsernamePrefix,
+		GroupsPrefix:   options.OIDC.GroupsPrefix,
+	}
```
foot, line 197: This should also incorportate authServer.AuthConfig.OIDCConfig.GroupsPrefix/UsernamePrefix I think?

Given https://github.com/weaveworks/weave-gitops/blob/cb8585c028ca6287220c49e9c3643d60f64a645d/pkg/server/auth/init.go#L40-L41

We should merge them w/ the secret taking higher precedence and show a similar logging message.

sarataha: Got it 👍 Thank you 🙂

cmd/gitops-server/cmd/cmd.go
```diff
@@ -187,7 +191,22 @@ func runCmd(cmd *cobra.Command, args []string) error {
...
+
+	// Incorporate values from authServer.AuthConfig.OIDCConfig
+	if authServer.AuthConfig.OIDCConfig.UsernamePrefix != "" {
+		log.V(logger.LogLevelWarn).Info("OIDC client configured by both CLI and secret. Secret values will take precedence.")
```
foot, line 201:
```suggestion
		log.V(logger.LogLevelWarn).Info("OIDC username prefix configured by both CLI and secret. Secret values will take precedence.")
```

foot (approved):

PR #3640, opened 2023-04-25
Fix the vulnerability in the `yaml` NPM package

opudrovs:
Closes https://github.com/weaveworks/weave-gitops/issues/3637

- Upgraded the `yaml` package to version 2.2.2 to resolve the vulnerability, reported by Snyk.

- Added an alias for the `yaml` package in `package.json` to resolve the ReferenceError happening in v2.2.2 with the default import. Parcel chokes on mixed types of exports (probably where export all/export with rename are mixed, as stated in the following issue https://github.com/parcel-bundler/parcel/issues/8792 ).

- Updated the `@docusaurus` package and its dependencies. Added a custom resolution to `yaml` v2.2.2 for website subdependencies.

Tested:
- Tested it in Tilt with regular mode and in the FAST_AND_FURIOUSER mode.
- Also ran `npm run build` to make sure it builds successfully.
- Ran the documentation with `yarn start`.

Testing:
To test it, run the app in Tilt and make sure that there are no errors displayed in the browser console in general or on object detail pages when displaying YAML views.

PR #3644, opened 2023-04-25
Remove the `yaml` package from direct website dependencies.

opudrovs:
Follow-up to https://github.com/weaveworks/weave-gitops/pull/3640

The `yaml` package was added to direct dependencies of the website because the `resolutions` example was found had the package in direct dependencies.

So, I tested removing it now and it works fine when keeping the `yaml` package only in resolutions (similar to the `ua-parser-js` package which was already there). 

The behavior of code stays the same. I tested running the documentation locally.


foot (approved):
👍

PR #3651, opened 2023-04-26
Add an additional page status for automations that summarizes their canaries

joshri:
Part of [2572](https://github.com/weaveworks/weave-gitops-enterprise/issues/2572) (Enterprise issue)

An additional `PageStatus` component has been added to the `AutomationDetail` component that displays a summarized condition created by looking at all of the automation's canaries. 
It displays one of four possible messages created by the `createCanaryCondition` function (tests in `ReconciledObjectsTable.test.tsx`):
- No canaries
- One or more canaries are in progress 
- All canaries have succeeded 
- One or more canaries failed

Assumptions made in implementation:
- Canaries will only have the `Promoted` condition in their conditions array.
- `Succeeded` and `Initialized` are the only indicators of success, `Failed` is the only one for failed, and all others should count as progressing.

This will not affect OSS as the new status only displays when the `AutomationDetail` component has custom tabs or actions passed down to it. No follow-up PR in enterprise is needed.

To test:
- run `export PUSH_PROGRESSIVE_DELIVERY_MANIFESTS_TO_GITOPS_DEV_REPO=1 ./tools/reboot.sh` in enterprise
- checkout this branch in OSS, then from the ui-cra directory in enterprise, run `make core-ui && make core-lib`
- check out the enterprise UI for the kustomization `progressive-delivery-canary`

Screenshot:
image


ui/components/ReconciledObjectsTable.tsx
```diff
@@ -25,6 +76,14 @@ function ReconciledObjectsTable({
     false
   );
 
+  //calculate canary status
+  useEffect(() => {
```
jpellizzari, line 80: This is effectively setting parent status from a child component, which we want to try and avoid. Instead, we should be lifting this logic up to where it is needed.

I realize this is a result of how the component tree is laid out.

jpellizzari (commented):
This is a bit un-react-y. A child component should not be calculating the state for a parent component.

The rest of the logic looks okay, but the status needs to be hoisted. Happy to help on this.

ui/components/ReconciledObjectsTable.tsx
joshri: Yeah the reconciled objects req was recently moved down into the individual tabs. Do we care if I move it back out or nah?

jpellizzari:
Upgraded parcel to get source maps working

jpellizzari (commented):
:+1: on the thorough tests. I would give a :heavy_check_mark: but I am now a contributor on this, so I will let others review it.

TheGostKasper (approved):
LGTM,

PR #3653, opened 2023-04-27
[WIP] upgrade go.mod to flux v2.0.0-rc.2

### LINKED ISSUE #3599: Upgrade and test WG OSS against Flux v2.0 RCs
Flux v2.0.0 RC-1 has released today. We have to catch up with it to make sure everything's working when the GA gets released.

chanwit:
Fixes #3599


foot:
Testing
- main + flux 2.0.0-rc.1
- upgrade-flux-v2-rc.1 + flux 2.0.0-rc.1
- upgrade-flux-v2-rc.1 + flux 0.41.2

### main + flux 2.0.0-rc.1

- Sources list / sync / suspend / resume
- GitRepo details / reconciled objects / events / yaml / sync / suspend / resume
- Applications list sync / suspend / resume
- Kustomization details / events / graph / yaml / sync / suspend / resume

Conclusion: main is "compatible" with flux 2.0.0-rc.1 and shows v1 CRs in the UI etc fine

### upgrade-flux-v2-rc.1 + flux 2.0.0-rc.1

In progress..

pkg/kube/kubehttp.go
```diff
@@ -28,6 +29,7 @@ func CreateScheme() (*apiruntime.Scheme, error) {
 	scheme := apiruntime.NewScheme()
 	builder := apiruntime.SchemeBuilder{
 		sourcev1.AddToScheme,
+		sourcev1b2.AddToScheme,
 		kustomizev2.AddToScheme,
```
foot, line 33: Should we add kustomize v1 here too?

foot:
> Notes: in the yaml view in the UI for gitrepo/kustomization the apiVersion is reported as v1beta2, so we might be querying that version still, rather than v1.


The code here seems to iterate over the available GVKs a bit randomly. https://github.com/weaveworks/weave-gitops/blob/2ab6561a1a42b08e8bd87bc67a54597c0d411da6/core/server/primarykinds.go#L21-L34

I get:

```
Added kind GitRepositoryList: source.toolkit.fluxcd.io/v1beta2, Kind=GitRepositoryList
Added kind GitRepository: source.toolkit.fluxcd.io/v1, Kind=GitRepository
Added kind GitRepository: source.toolkit.fluxcd.io/v1beta2, Kind=GitRepository
Added kind GitRepositoryList: source.toolkit.fluxcd.io/v1, Kind=GitRepositoryList
```

So `GitRepositoryList` would map to v1 and `GitRepository` would map to v1beta2 in the end. There is this idea of version priority but not sure if there is a helper around in client-go or something https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/#version-priority

foot:
Something like this seems to work better

```golang
	for _, gv := range scheme.PreferredVersionAllGroups() {
		for kind := range scheme.KnownTypes(gv) {
			gvk := schema.GroupVersionKind{
				Group:   gv.Group,
				Version: gv.Version,
				Kind:    kind,
			}
			kinds.kinds[kind] = gvk
			fmt.Printf("Added kind %+v: %+v\n", kind, gvk)
		}
	}
```

pkg/kube/kubehttp.go
chanwit: Thank you @foot 
Didn't realize that I totally miss it. I'll take a look.

chanwit:
So our conclusion would be the next release, OSS will drop support for Flux2 v0.41.x. Would that be ok?

foot:
> So our conclusion would be the next release, OSS will drop support for Flux2 v0.41.x. Would that be ok?

Great question!

So as it stands:
1. weave-gitops `v0.22.0` / `main` + flux `v0.41.2` works
2. weave-gitops `v0.22.0` / `main` + flux `v2.0.0-rc.1` works, it shows both v1 and v1beta2 GitRepositories and Kustomizations in the UI etc.
3. weave-gitops `upgrade-flux-v2-rc.1` + flux `v2.0.0-rc.1` works, it shows both v1 and v1beta2 GitRepositories and Kustomizations in the UI etc. 
4. weave-gitops `upgrade-flux-v2-rc.1` + flux `v0.41.2` will show _no matches for kind "GitRepository" in version "source.toolkit.fluxcd.io/v1_ everywhere in the UI, requiring you to upgrade (rerun `flux bootstrap`) to `v2.0.0-rc.1`

Benefits of dropping support for v0.41.2 right now:
- force users to upgrade their flux-runtime already.
- ?
- _Other benefits that come to mind?_

I only recently understood the full compatibility matrix that we have here. So just double checking that we want to drop support for `v0.41.2` now vs waiting for "GA final" etc, cc @JamWils

foot:
Note from testing we could include in migration guide:

You cannot get the "correct" `apiVersion` of a kustomization/gitrepo etc that is stored in git. k8s/our-code will dynamically replace it with whatever you request, so:
- weave-gitops `main`: will always say `apiVersion: v1beta1` in the yaml UI views, even if gitrepo.yaml in git says `apiVersion: v1`
- weave-gitops `upgrade-flux-v2-rc.1` will always say `apiVersion: v1` in the yaml UI views, even if gitrepo.yaml in git says `apiVersion: v1beta2`

chanwit:
We decided to skip Flux v2.0.0-rc.1 and go with Flux v2.0.0-rc.2 when it's released.

chanwit:
Now upgraded to Flux v2.0.0-rc.2.

LappleApple:
@chanwit can we close?

Issue #2704, opened 2022-09-05
SPIKE: adopt KYAML for YAML manipulation


chanwit:
We may need to manipulate YAML files in various ways for many tickets.
KYAML is a standard way to do so.
We as a team need to learn more about how to use this library.

chanwit:
@ozamosi request for refinement.

chanwit:
## Marshalling
1. We prepare an object to be marshaled, like `pod := corev1.Pod{ ... }`
2. We prepare a new scheme, `scheme := runtime.NewScheme()`
3. We add a scheme we want to the `scheme` variable, like `err := clientgoscheme.AddToScheme(scheme)`
4. We prepare an object serializer for the scheme, `objSerializer := serializer.NewCodecFactory(scheme).LegacyCodec(corev1.SchemeGroupVersion)`
5. After that we encode the object `pod` to the `jsonBytes` with `jsonBytes, err := runtime.Encode(objSerializer, &pod)`
6. Then we can use KYAML's `yaml.ConvertJSONToYamlNode(string(jsonBytes))` to convert JSON to KYAML's node.

chanwit:
## Adding fields

1. We prepare a KYAML node from the above step, or from `yaml.ReadFile(filename)`
2. To create a new field, we can use `yaml.LookupCreate()` function in `object.Pipe()`. For example, if we'd like to create `nameSuffix` field, we can do something like this:  `nameSuffixNode, err := obj.Pipe(yaml.LookupCreate(yaml.ScalarNode, "nameSuffix"))
3. We can use `yaml.FieldSetter` to set the value for a certain field.

Here's an example:

```golang
	obj, err := yaml.ReadFile("no-suffix.yaml")
	if err != nil {
		fmt.Println(err.Error())
		return
	}

	nameSuffixNode, err := obj.Pipe(yaml.LookupCreate(yaml.ScalarNode, "nameSuffix"))
	if err != nil {
		return
	}
	nameSuffix := yaml.GetValue(nameSuffixNode)
	obj, err2 := obj.Pipe(
		yaml.Tee(
			yaml.FieldSetter{Name: "nameSuffix", StringValue: nameSuffix + "-suffix"},
		),
	)
```

chanwit:
## Removing fields
To remove field, you can simply use the `yaml.Clear("field name")` function.
Here's a full example of how we use `yaml.Clear` to clean up multiple fields.

```golang
	scheme := runtime.NewScheme()
[...]

PR #3655, opened 2023-04-28
2704 redesign application details

TheGostKasper:
Part of [#2704](https://github.com/weaveworks/weave-gitops-enterprise/issues/2704)  (enterprise issue)

This is an enhancement for the information provided in the application data as is now, in order to highlight the most important data in the application details view

* Have the version and Last updated timestamp at the top of the page. 
    * For Helm Release, we will show the App Version
    * For Kustomization, we will show the App Revision
* Have a collapsable details for the application meta data 

**App Observability** - Include App Health Summary in App Status [#2701](https://github.com/weaveworks/weave-gitops-enterprise/issues/2701)

* The health summary should be located at the top of the application details page. 
* The health summary should be an aggregate of the health of all workloads in the application

https://user-images.githubusercontent.com/4614360/235336083-9c7be921-4feb-49c7-b6a9-ec5e7b65e7a9.mp4

**To discuss**
I went throw some test cases to find the best performance way to display the HealthCheck agg as follow 

* _Lifting GetInventory request up to Parent component to avoid unexpected behavior by passing data bottom-up from a child to parent which will cause a re-render and sometimes leaks_ .

**Doesn't that fetch getInventory even when Details tab is not selected** ? [Yes]
 To avoid that behavior, `refetchInterval` is turned off when we have the data (first call) & `Details` tab is not selected otherwise it turned on. One of it's cons is we won't get the latest data of the App when User is on other taps 

**Why not just keep the `refetchInterval` as it's ?** 
We will end up requesting `GetInventory` on every tab of the automation but in return HealthCheck status will be up to date .
 

**PR covers**
- lifting `GetInventory` up to parent 
- keep the refetchInterval to make sure HealthStatus is up to date
- Show HealthCheck Agg status
- Have a collapsable details for the application meta data 
- Have the version and Last updated timestamp at the top of the page


ui/components/AutomationDetail.tsx
```diff
@@ -57,6 +119,29 @@ function AutomationDetail({
...
+  const { path } = useRouteMatch();
+  const { data, isLoading, error } = useGetInventory(
+    type,
+    name,
+    clusterName,
+    namespace,
+    false,
+    {
+      retry: false,
+      refetchInterval: (data) =>
```
jpellizzari, line 140: There is an `enabled` flag that you can use to turn the hook on an off.

```suggestion
      enabled: (data) =>
```

jpellizzari (commented):
I don't think we should be turning this hook off. The top-level status component needs to keep polling to keep its state up to date. If it stops, it will not re-render when the state has changed.

One thing we can do to share state is to remove the `withChildren` parameter from the cache. Then the state will be shared no matter if `withChildren` has changed.

https://github.com/weaveworks/weave-gitops/blob/f85b49fed3a2259f11c1b848f75571de11639031/ui/hooks/inventory.ts#L31

Else, doing two requests is not a big deal. Both pieces of the UI need different data to work, so they can each request as they need to.

ui/components/AutomationDetail.tsx
TheGostKasper: Taking in consideration that user might navigate from/to different tabs and refresh the page or uses a link to specific tab other than Details Or navigate back and forward when Details is not the active tab  ... Than enabled won't work as expected

jpellizzari (approved):
Looks great :+1:

PR #3660, opened 2023-04-28
Update nav icons and colors

### LINKED ISSUE #3659: Nav icon polish
New icons:
image

joshri:
Closes #3659 

- adds new icon svg code
- modifies classes in icon component adding new svg tags (line, polygon, etc)
- swaps text and icons to gray and headers to black
- need to separately update explore icon IconType in enterprise

To see ALL the icons locally (not just OSS ones), swap to commit `a9f84cf0a18eeffe3436e5e7e03e0f52e6943db3`

https://user-images.githubusercontent.com/65822698/235235426-bd94b880-8399-44ed-b76f-5a7bf86772c2.mov

Screen with further changes suggested by design - new flux, notifications, and flux icons, search icon in data table changed, letter spacing added to headers
image


jpellizzari (approved):
LGTM. I assume the inline-ing of the svg icons is to avoid bundler problems. 

If we were able to use an `img` tag, the browser would cache them and they would load immediately, as well as ensure they are out of the `react` render thread. Not important at the moment, however.

Here was an effort to keep them as static files: https://github.com/weaveworks/weave-gitops/compare/jp-svg-inline

Maybe this could be adapted to use `img` tags instead of `fetch` -> html.

PR #3668, opened 2023-05-04
Fix to the `nsaccess` package 

### LINKED ISSUE #3652: Question: how does the namespace filtering feature work?
Hey, I have some questions about the correct behavior of the namespaces filtering feature.

I have a use case where the Weave GitOps UI v0.18.0 reports lots of errors for some of its users, coming from [this package](https://github.com/weaveworks/weave-gitops/blob/b05ebd8c090f923b4e99f3d1e43f55a47ec857b6/core/server/objects.go#L122), due to insufficient permissions to the Kubernetes Secrets the Helm releases are kept in. The error is correct because the group these users belong to lacks permissions to the secrets in some of the namespaces, they can though still see Flux CRs and related resources. This makes me confused however for I would expect server to filter out such namespaces, if access to the secrets counts as crucial for the namespace to be considered accessible. And if not, then I would expect the UI to not display any errors for the users and instead only display `HelmRelease` CRs without an inventory. 

To be more precise my questions are:
* how is the accessibility / inaccessibility determined, if determined at all?
* does partially accessible, when lacking certain permissions, equals to accessible, and in such case, is there a way to hide permission errors in the UI?
* is there a way to tweak how is the accessibility determined, in case it is not strictly tied to some set of permissions? That is, can I tell the server to skip certain namespaces for certain groups?

I also tried to answer these questions on my own, but got confused by what I have found. I eventually landed in here, in the [nsaccess package](https://github.com/weaveworks/weave-gitops/blob/v0.18.0/core/nsaccess/nsaccess.go#L67), for this seems involved in getting accessible namespaces upon impersonation. If that is the case, and if I get it right, it eventually translates what is returned on the [the SelfSubjectRulesReview request](https://github.com/weaveworks/weave-gitops/blob/4c048c23697ee4f3a463d61bebd15d7447fab8f6/core/nsaccess/nsaccess.go#L91) into [the derivedAccess map](https://github.com/weaveworks/weave-gitops/blob/4c048c23697ee4f3a463d61bebd15d7447fab8f6/core/nsaccess/nsaccess.go#L162). And then policy rules, which are being [these by default](https://github.com/weaveworks/weave-gitops/blob/4c048c23697ee4f3a463d61bebd15d7447fab8f6/core/nsaccess/nsaccess.go#L18), are being [checked against this map](https://github.com/weaveworks/weave-gitops/blob/4c048c23697ee4f3a463d61bebd15d7447fab8f6/core/nsaccess/nsaccess.go#L202) to determine value of [hasAccess boolean](https://github.com/weaveworks/weave-gitops/blob/4c048c23697ee4f3a463d61bebd15d7447fab8f6/core/nsaccess/nsaccess.go#L200), which is initially true. And this is the part that confuses me. 

The [DefautltWegoAppRules](https://github.com/weaveworks/weave-gitops/blob/v0.18.0/core/nsaccess/nsaccess.go#L18) starts with the `{"","secrets,pods,events","get,list"}`. In my case users have access to `pods` and `events`, so the `""` API group is available in the `derivedAccess` map, hence filtering passes the API group check. But then it does not find `secrets`, hence switches `hasAccess` to false, and continues to the [next rules](https://github.com/weaveworks/weave-gitops/blob/4c048c23697ee4f3a463d61bebd15d7447fab8f6/core/nsaccess/nsaccess.go#L217). Shouldn't it break here instead, for it knows the minimum permissions user needs are already incomplete? 

On continuing, it processes the next rules and it eventually gets to the [last item](https://github.com/weaveworks/weave-gitops/blob/4c048c23697ee4f3a463d61bebd15d7447fab8f6/core/nsaccess/nsaccess.go#L44) on the policy rules list, which is being the `{"","events","get,list,watch"}`, and in my case all users are granted this set of permissions, even those who don't have access to the secrets. In result this item passes the API group check, then the resource check, and it gets to the verbs check [here](https://github.com/weaveworks/weave-gitops/blob/4c048c23697ee4f3a463d61bebd15d7447fab8f6/core/nsaccess/nsaccess.go#L221). All the verbs are in the derived map, so the `hasAccess` boolean [is restored](https://github.com/weaveworks/weave-gitops/blob/v0.18.0/core/nsaccess/nsaccess.go#L230) to true, and since it was the last item on the list, the namespace is recognized as accessible as far as I understand it. It looks like user could lack permissions to the Secrets, Pods, and Flux CRs altogether, but having read-only access to `events` would still grant him permissions to the namespace, but maybe I've confused something.

Even for the future sake, is this the part of code that is actually responsible for determining the access? And if so, is my understanding of it correct?

ljakimczuk:
Closes #3652
Fixes #3652

**What changed?**

The PR changes the behaviour of the `hasAllRules` function. Up until now, all the items on the required rules were checked against the user derived permissions what resulted in the last item on that list often being decisive about the access. This PR changes this, and from now on the `hasAllRules` quits as soon as it realize user lacks permissions.

**Why was this change made?**

Users get permissions to the namespaces they don't have the right level of access to. This makes the UI to print errors.

It was implemented by first investigating the problem, then writing tests cases that were failing originally to make sure the problem is correctly recognized, and then the problem was solved by switching the `hasAllRules` from `continue` to `break` behaviour, fixing the tests.

With unit-tests, I may also build an image and test it in my environment if that's required.

Update: I've let it run in my environment, the error is gone, inaccessible namespaces are correctly filtered out.

**Release notes**

- Fixed the namespace filtering behaviour.

**Documentation Changes**


core/nsaccess/nsaccess_test.go
```diff
@@ -171,6 +171,84 @@ func TestFilterAccessibleNamespaces(t *testing.T) {
...
+		requiredRules := []rbacv1.PolicyRule{
+			{
+				APIGroups: []string{""},
+				Resources: []string{"secrets", "events", "pods", "namespaces"},
+				Verbs:     []string{"get", "list"},
+			},
+			{
+				APIGroups: []string{""},
+				Resources: []string{"pods"},
+				Verbs:     []string{"get"},
```
foot, line 239: Nice 👍

foot (approved):
Solid coverage, nicely done

chanwit:
@foot would this be labeled as a bug fix or an enhancement please?

foot:
> would this be labeled as a bug fix or an enhancement please?

Oops! Good one, bug.

PR #3680, opened 2023-05-09
UI application details followup

TheGostKasper:
Part of [EE issue](https://github.com/weaveworks/weave-gitops-enterprise/issues/2704)

image
image

### DIFF
--- a/website/docs/pipelines/promoting-applications.mdx
+++ b/website/docs/pipelines/promoting-applications.mdx
@@ -177,7 +177,7 @@ In the journey of creating a pull request, there are different secrets involved:
 
 1. Pipeline controller receives events via [webhook from leaf clusters](./#setup-notifications-from-leaf-clusters). Hmac is used for authN/authZ so an hmac key should be provided in this case.
 2. Pipeline controller clones and patches manifests to promote from the pipeline configuration repo. A set of [git credentials](https://fluxcd.io/flux/components/source/gitrepositories/#secret-reference) are required.
-3. Pipeline controller uses git provider api to create the pull request with the promoted manifests. A Personal Access Token (PAT) needs to be created to interact with pipelines git provider API.
+3. Pipeline controller uses git provider api to create the pull request with the promoted manifests. A Personal Access Token (PAT) needs to be created to interact with pipelines git provider API. This PAT is also used to list pull requests from the configured repository.
 
 Create a Kubernetes secret with the previous data.
 


ui/components/AutomationDetail.tsx
```diff
@@ -107,22 +104,18 @@ function AutomationDetail({
...
+  // const canaryStatus = createCanaryCondition(data?.objects);
   const health = computeAggHealthCheck(data?.objects || []);
+
   const defaultTabs: Array<routeTab> = [
     {
       name: "Details",
       path: `${path}/details`,
       component: () => {
         return (
           <>
```
joshri, line 118: i think we can remove the fragment tags here now

ui/components/AutomationDetail.tsx
```diff
@@ -190,56 +183,81 @@ function AutomationDetail({
...
+            disabled={automation.suspended}
+          />
+          <Spacer padding="xs" />
+          <Button
+            onClick={() => suspend.mutateAsync()}
+            loading={suspend.isLoading}
+          >
+            {automation.suspended ? "Resume" : "Suspend"}
+          </Button>
+          <CustomActions actions={customActions} />
```
joshri, line 201: I know that there're custom actions between the sync/suspend and applied revision/chart text, but imo it needs a little more space, as right now I only get ~12 characters. Maybe that's enough to determine what you need to out of it?

ui/components/Collapsible.tsx
```diff
@@ -1,47 +1,46 @@
 import { Collapse } from "@material-ui/core";
 import React from "react";
+import styled from "styled-components";
+import Flex from "./Flex";
 import Icon, { IconType } from "./Icon";
+import Text from "./Text";
 
 const Collapsible = ({ children }) => {
```
joshri, line 8: can we avoid the inline styles in this component?

ui/components/Flex.tsx
```diff
@@ -23,7 +24,8 @@ type Props = {
 const Styled = (component) => styled(component)`
   display: flex;
   flex-direction: ${({ column }) => (column ? "column" : "row")};
-  align-items: ${({ align }) => (align ? "center" : "start")};
+  align-items: ${({ align, alignItems }) =>
+    alignItems ? alignItems : align ? "center" : "start"};
```
joshri, line 28: this is kind of confusing for me to understand - is there a way to write it without the chained ternary?

joshri (approved):
Nothing's crashing, and I know we want to get it in before the release, but there are a couple small comments I have when you get the chance

ui/components/Flex.tsx
TheGostKasper: Yes, by using only the `alignItems` props instead of `align` but doing that will require to loop over the Components that uses Flex center and change it so that i didn't change it for now

ui/components/AutomationDetail.tsx
TheGostKasper: Valid concern and will mentioned in the next iteration to this page

ui/components/Collapsible.tsx
TheGostKasper: What concerns -as i don't really get it here :eyes:-  ?

TheGostKasper:
@yiannistri @joshri your concerns/suggestion will be addressed in the next iteration to Application details up next

chanwit:
Wow. The update looks really nice. Thank you @TheGostKasper :1st_place_medal: 

Note to self for some possible follow-up bits:
```tsx
const Styled = (component) => styled(component)`
  display: flex;
  flex-direction: ${({ column }) => (column ? "column" : "row")};
  align-items: ${({ align, alignItems }) => {
    if (alignItems) {
      return alignItems;
    } else if (align) {
      return "center";
    } else {
      return "start";
    }
  }};
  ${({ gap }) => gap && `gap: ${gap}px`};
  ${({ tall }) => tall && `height: 100%`};
  ${({ start }) => start && "justify-content: flex-start"};
  ${({ between }) => between && "justify-content: space-between"};
  ${({ center }) => center && "justify-content: center"};
  ${({ shadow }) => shadow && "box-shadow: 5px 10px 50px 3px #0000001a"};
`;
```

and `.trim-text`:

```css
 .trim-text {
    max-width: 150px; // <- maybe this, right?
    white-space: nowrap;
    overflow: hidden;
    text-overflow: ellipsis;
  }
```

PR #3689, opened 2023-05-10
Added a redirect for getting started page

Himangini:
Fixes https://github.com/weaveworks/weave-gitops-interlock/issues/414

**What changed?**
This pr adds a redirect for old page which no longer exists. Also removed redundant info from marketplace docs.

**Why was this change made?**
The page link https://docs.gitops.weave.works/docs/getting-started/ currently goes to older versions of docs with outdated information causing confusion. 

When the new getting started page was introduced https://docs.gitops.weave.works/docs/getting-started/intro we must have missed updating outdated links.

Adding a redirect plugin https://docusaurus.io/docs/using-plugins

Staging build and local testing

**Release notes**

**Documentation Changes**
This pr adds a redirect for the old page which no longer exists. Also removed redundant info from marketplace docs.
Added two plugins - 1 for redirect and 1 for generating a sitemap for our documentation site so crawlers have up to date info on our website


foot (approved):
Makes sense to me!

PR #3711, opened 2023-05-19
Add prop to hide dark mode switch

joshri:
Enterprise imports UserSettings from OSS, and needs to be able to disable the dark mode switch for now.

When running tests, I found a random console error where the `to` prop on Link was undefined in the Logo component, so I fixed that as well

### DIFF
--- a/ui/components/Logo.tsx
+++ b/ui/components/Logo.tsx
@@ -13,15 +13,15 @@ type Props = {
   link?: string;
 };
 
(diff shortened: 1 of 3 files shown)


jpellizzari (approved):
LGTM, but can't EE just hide this with CSS?

joshri:
> LGTM, but can't EE just hide this with CSS?

I guess it could be like `${Switch} {display: none}`? This is hopefully very temporary anyway

PR #3712, opened 2023-05-21
Show Application Violations in WeaveGitOps OSS

### LINKED ISSUE #3704: Show Application Violations in WeaveGitOps OSS
As a user using Weave Policy Agent, I need to see the policy violations in my application. 

#### Application Criteria 
* We need to replicate the violations tab in the application view in the Enterprise product.

waleedhammam:
Closes #3704

**What changed?**

- Added APIs for list application violation
- Added `Violations` tab to automationDetails page
- Created `ViolationDetails` page

**Why was this change made?**

- As a request to Show Application Violations in WeaveGitOps OSS

- By porting the feature from WGE and making the proper fixes & enhancements 
- Method `ListObjects` couldn't work as we rely on events to lists policies violations

- Manual Testing
- [UI] added unit tests

https://github.com/weaveworks/weave-gitops/assets/4614360/cb7312b8-530e-4831-acac-b94a7b367b2d

**Release notes**

**Documentation Changes**


ui/components/Policies/Utilis/PolicyUtilis.tsx
```diff
@@ -0,0 +1,50 @@
...
+  );
+};
+
+export const ParameterWrapper = styled(Flex)`
+  border: 1px solid ${(props) => props.theme.colors.neutral20};
+  box-sizing: border-box;
+  border-radius: ${(props) => props.theme.spacing.xxs};
+  padding: ${(props) => props.theme.spacing.base};
+  width: 100%;
+`;
```
joshri, line 45: I'd prefer to have these Parameter components in `Parameters` rather than here unless you anticipate them being used elsewhere - not a blocker though. I'd also go with the pattern of shortening utilities to "utils" rather than "utilis"

ui/components/Policies/Utilis/PolicyUtilis.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+export const ChipWrap = styled(Chip)`
+  color: ${(props) => props.theme.colors.black} !important;
```
joshri, line 48: I'd prefer to use the use the MUI class instead of `!important` for overrides - this also seems similar to what we did for Labels in the Metadata component. Let's use that here if it works.

api/core/core.proto
```diff
@@ -183,6 +202,62 @@ message GetInventoryResponse {
...
+    string created_at = 6;
+    string entity=7;
+    string namespace=8;
+    string violating_entity=9;
+    string description=10;
+    string how_to_solve=11;
+    string name=12;
+    string clusterName = 13;
+    repeated PolicyValidationOccurrence occurrences = 14;
+    string policy_id = 15;
```
jpellizzari, line 220: Mixing camel-case with snake case here. Can we use camel-case for proto fields?

ui/components/YamlView.tsx
```diff
@@ -53,18 +60,19 @@ function UnstyledYamlView({ yaml, object, className, theme }: YamlViewProps) {
...
             className="yaml-copy"
             size="small"
           />
-        )}
-      </YamlHeader>
+        </YamlHeader>
+      )}
+
       <SyntaxHighlighter
-        language="yaml"
```
joshri, line 67: These changes in YamlView don't seem to be used elsewhere? Are they necessary?

api/core/core.proto
```diff
@@ -183,6 +202,62 @@ message GetInventoryResponse {
     repeated InventoryEntry entries = 1;
 }
 
+message PolicyValidation{
```
jpellizzari, line 205: Nit: can we format these to aligned? In `vscode` you can use the `Align` command to get them all to line up.

joshri (commented):
Can you add a link to the design for this new page, and some steps to add objects to test locally?

core/server/policy_violations.go
```diff
@@ -0,0 +1,263 @@
...
+	opts := []sigsClient.ListOption{}
+	if m.Pagination != nil {
+		opts = append(opts, sigsClient.Limit(m.Pagination.PageSize))
+		opts = append(opts, sigsClient.Continue(m.Pagination.PageToken))
+	}
+	opts = append(opts, &sigsClient.ListOptions{
+		LabelSelector: labelSelector,
+		FieldSelector: fieldSelector,
+	})
+	opts = append(opts, sigsClient.InNamespace(v1.NamespaceAll))
```
jpellizzari, line 71: Do we want to query all namespaces here? It looks like the `msg` has a `.Namespace` field.

core/server/policy_violations.go
```diff
@@ -0,0 +1,263 @@
...
+		return nil, fmt.Errorf("error building selector for events query: %v", err)
+	}
+	opts := []sigsClient.ListOption{}
+
+	fields := k8sFields.OneTermEqualSelector("type", "Warning")
+	opts = append(opts, &sigsClient.ListOptions{
+		LabelSelector: selector,
+		FieldSelector: fields,
+	})
+	opts = append(opts, sigsClient.InNamespace(v1.NamespaceAll))
```
jpellizzari, line 107: Same as above regarding `NamespaceAll`

core/server/policy_violations.go
```diff
@@ -0,0 +1,263 @@
...
+		return nil, fmt.Errorf("error getting events: %v", err)
+	}
+	if len(validationsList.Errors) > 0 {
+		return nil, fmt.Errorf("error getting events: %s", validationsList.Errors[0].Message)
+	}
+	if len(validationsList.Validations) == 0 {
+		return nil, fmt.Errorf("no policy violation found with id %s and cluster: %s", m.ViolationId, m.ClusterName)
+	}
+	return &pb.GetPolicyValidationResponse{
+		Violation: validationsList.Validations[0],
```
jpellizzari, line 120: This is a little confusing. We only care about the first validation?

core/server/policy_violations.go
```diff
@@ -0,0 +1,263 @@
...
+
+	var validations []*pb.PolicyValidation
+	for listClusterName, lists := range clist.Lists() {
+		if clusterName != "" && listClusterName != clusterName {
+			continue
+		}
+		for _, l := range lists {
+			list, ok := l.(*v1.EventList)
+			if !ok {
+				continue
```
jpellizzari, line 149: Should we log or return an error here?

core/server/policy_violations.go
```diff
@@ -0,0 +1,263 @@
...
+func getAnnotation(annotations map[string]string, key string) string {
+	value, ok := annotations[key]
+	if !ok {
+		return ""
+	}
+	return value
+}
+
+func getPolicyValidationParam(raw []byte) ([]*pb.PolicyValidationParam, error) {
+	var paramsArr []map[string]interface{}
```
jpellizzari, line 212: Can this be described by a `struct` instead? I see that we have known fields below, which makes me thing we could do:

```go
type validationParam struct {
  Name string `json:"name"`
  ...
}
```

core/server/policy_violations.go
```diff
@@ -0,0 +1,263 @@
...
+		if err != nil {
+			return nil, err
+		}
+		param.Value = val
+		parameters = append(parameters, &param)
+	}
+	return parameters, nil
+}
+
+func getParamValue(in interface{}) (*any.Any, error) {
```
jpellizzari, line 240: Curious about this. Why are we doing these low-level type conversions?

ui/components/Policies/Utilis/HeaderRows.tsx
jpellizzari, line 1: Typo here?: `Utilis` -> `Utils`

ui/components/Policies/Utilis/Severity.tsx
```diff
@@ -0,0 +1,36 @@
+import React from "react";
+import Flex from "../../Flex";
+import Icon, { IconType } from "../../Icon";
+import Text from "../../Text";
+
+function Severity({ severity }: { severity: string }) {
+  return (
+    <Flex alignItems="center" gap="4" data-testid={severity}>
+      {(() => {
```
jpellizzari, line 9: Can this be lifted up? A `switch` inside escaped JSX seems unnecessarily complex. Maybe something like this:

```jsx
function Severity() {
        let icon;
        switch (severity.toLocaleLowerCase()) {
          case "low":
            icon = 
              <Icon
                type={IconType.CallReceived}
                color="primary20"
                size="base"
              />
                   ...
        }

    return (
      <Flex>{icon}</Flex>
   )


}
```

ui/components/Policies/Utilis/Parameters.tsx
```diff
@@ -0,0 +1,48 @@
+import React from "react";
+import { PolicyValidationParam } from "../../../lib/api/core/core.pb";
+import { ParameterCell, ParameterWrapper, parseValue } from "./PolicyUtilis";
+
+const Parameters = ({
```
jpellizzari, line 5: Withiin this PR, there are a couple different ways that components are being declared. Some via `const` and some via `function`.

Any thoughts on the tradeoffs and is it worth making them consistent?

ui/components/Link.tsx
```diff
@@ -82,4 +82,11 @@ function Link({
 
 export default styled(Link)`
   text-decoration: none;
+
+  &.ellipsis {
+    max-width: 300px;
```
jpellizzari, line 87: Why would it always be 300px? This seems like something we should just override with CSS where we need to.

ui/components/Policies/PolicyViolations/PolicyViolationDetails.tsx
```diff
@@ -0,0 +1,187 @@
...
+          onClick={() => setDetailModal({ object: entityObject })}
+        >
+          {entityObject.namespace}/{entityObject.name}
+        </Text>
+      ),
+    },
+  ];
+
+  return (
+    <Flex wide tall column gap="32">
```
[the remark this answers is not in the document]
TheGostKasper: It's perfect for the eye in that page :D

ui/components/AutomationDetail.tsx
```diff
@@ -179,6 +180,24 @@ function AutomationDetail({
...
+              application: name,
+              clusterName,
+              namespace,
+              kind: type,
+            }}
+            sourcePath="kustomization"
+          />
+        );
+      },
+      visible: !customTabs?.find((c) => c.name === "Violations"),
```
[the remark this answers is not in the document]
TheGostKasper: Because we want to prevent an dependencies for now as we won't use the same Violation component in the WGE so separate things is a good way to do so, and yes there will be a follow up PR to export and make use of it in WGE once it's ready. 

taking in consideration the validation check over the CanaryStatus depends on customTabs .

serboctor:
> Can you add a link to the design for this new page, and some steps to add objects to test locally?

@joshri you can follow [this](https://github.com/weaveworks/policy-agent/blob/open-source-policy-agent/docs/getting-started.md) guide to run the agent and add policies. There's also an example to generate violations. But in your case, you'll need to do it for a resource that belongs to an application to be able to see the violations under the applications page.

We didn't have a solid design :sweat_smile:, we are replicating a page that we already have in WGE.

ui/components/Breadcrumbs.tsx
```diff
@@ -25,21 +26,29 @@ export const Breadcrumbs = () => {
...
-            size="large"
-            color="neutral40"
-          />
-          <Link
-            to={formatURL(currentPage, parsed)}
-            textProps={{ size: "large", color: "neutral40" }}
-          >
-            {parsed.name}
-          </Link>
+          {label && (
```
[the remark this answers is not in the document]
TheGostKasper: Due to the navigation (breadcrumb) issue and it's all a static routes and only display the name from the queryParams as it's 
` Parent > child ` and there's no proper way to set the **Breadcrumb route** for this page as ` Application > (hr/kustomization) automation details > violation message ` beside it won't make any sense to display the violation details as `Application > violations` so that we went for displaying only the `name` and as there won't be a parent here then UI show a **>** before the name so that check was needed for better visualization .


I'm working on moving the breadcrumb to Page so that we can pass the routes easily for derived routes like the way we do in WGE

ui/components/Policies/PolicyViolations/PolicyViolationDetails.tsx
```diff
@@ -0,0 +1,187 @@
...
+      <HeaderRows headers={headers} />
```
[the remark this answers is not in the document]
TheGostKasper: The difference relies on 
```
interface Header {
  rowkey: string;
  value?: any;
  children?: any;
  visible?: boolean;
}
```

with much controlling on `HeaderRow` beside it's not globally used and is cover the needs here at Violations & Policies in the upcoming pr

ui/components/Policies/Utilis/PolicyUtilis.tsx
TheGostKasper: Yes, most of the Utilis and sharedComponent will be used in the other PR the @AsmaaNabilBakr is working on and it's kinda the same work as Listing Policies & policy details + (policy) Violations

ui/components/YamlView.tsx
TheGostKasper: This meant to be used in displaying the `violatingEntities` in Violation details and Policy code in PolicyDetails and currently we dropped `violatingEntities` and sowed as a modal .
beside this will give the ability to change it when we want .

ui/components/Breadcrumbs.tsx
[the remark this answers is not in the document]
joshri: Why not just leave the Breadcrumbs as they are? Isn't this just another tab on an automation?

ui/components/Policies/PolicyViolations/PolicyViolationDetails.tsx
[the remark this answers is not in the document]
joshri: I don't understand the difference - it just seems really complex for something we already have

ui/components/Policies/Utilis/Parameters.tsx
TheGostKasper: `Severity` is the only function and it's exported as default as there's no other fn in the file to export.

ui/components/Breadcrumbs.tsx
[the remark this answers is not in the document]
TheGostKasper: Yes because the violations is now a tab in the AutomationDetails for hr and kustomization so the breadcrumb should follow

`Applications > flux-system > Container Running As User in deployment helm-controller (1 occurrences) ` 

this not doable in current OSS as mentioned

ui/components/Policies/PolicyViolations/PolicyViolationDetails.tsx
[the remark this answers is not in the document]
TheGostKasper: The main difference is that when you create the listOf headers you set the the visible directly instead of doing the check and push/pop to a list also the display of the Flex not a table which is meant for this page

ui/components/Link.tsx
```diff
@@ -82,4 +82,11 @@ function Link({
...
+    white-space: nowrap;
+    overflow: hidden;
+    text-overflow: ellipsis;
+  }
```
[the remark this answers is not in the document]
TheGostKasper: It appears that we can't wrap a link in a tooltip so changing it to Text :

core/server/policy_violations.go
waleedhammam: Yeah indeed, the `Namespace` field is optional, we can filter by cluster scoped `PolicyId` or by `ClusterName` which will require to query on all namespaces

core/server/policy_violations.go
waleedhammam: Yeah indeed, Cause `Namespace` field is optional, we can filter by cluster scoped `PolicyId` or by `ClusterName` which will require to query on all namespaces

core/server/policy_violations.go
waleedhammam: The method purpose is to get the details of one violation, yet the method we query with `listEvents` (used in list&get)
returns a list, so we only need that one object

There also safety checks [here]( https://github.com/weaveworks/weave-gitops//blob/ff6ff54f6e376e916e8cb52390f93d22b6ee37d5/core/server/policy_violations.go#L113) if something went wrong and the query returned none or more than object we return some errors

core/server/policy_violations.go
waleedhammam: It's not an error here, we only care about a specific for of events we query on
logging that error can flood the logs and will make it hard to track the real errors (it can be log debug for skipping this object which doesn't meet the query critera)

core/server/policy_violations.go
waleedhammam: We already have `PolicyValidationParam` struct for the response but the `value` field could have different types depending on the policy (could be numeric or boolean or string) it's already implemented as `google.protobuf.Any` in the proto response

core/server/policy_violations.go
waleedhammam: that's actually because the `value` field could have different types depending on the policy, policy params could be different types as the code shows could be (string, numberic .. )

example params in policy `Containers Minimum Replica Count`

image

api/core/core.proto
waleedhammam: done

api/core/core.proto
waleedhammam: done

core/server/policy_violations.go
```diff
@@ -0,0 +1,267 @@
...
```
serboctor, line 244: I am not very fond of using `in` as an argument name, I would prefer something more descriptive. `in` gives me the feeling like it will be used in a channel or queue.

core/server/policy_violations.go
```diff
@@ -0,0 +1,267 @@
...
+
+type validationList struct {
+	Validations []*pb.PolicyValidation
+	Token       string
+	Errors      []*pb.ListError
+}
+
+func (cs *coreServer) ListPolicyValidations(ctx context.Context, m *pb.ListPolicyValidationsRequest) (*pb.ListPolicyValidationsResponse, error) {
+	var respErrors []*pb.ListError
+	clustersClient, err := cs.clustersManager.GetImpersonatedClient(ctx, auth.Principal(ctx))
```
serboctor, line 31: in the `ListObjects` they check on cluster name and accordingly use either `GetImpersonatedClientForCluster` or `GetImpersonatedClient`...do you know why it's being done like that there and is there a reason why you're not following the same approach?

core/server/policy_violations.go
```diff
@@ -0,0 +1,267 @@
...
+		if merr, ok := err.(*multierror.Error); ok {
+			for _, err := range merr.Errors {
+				if cerr, ok := err.(*clustersmngr.ClientError); ok {
+					respErrors = append(respErrors, &pb.ListError{ClusterName: cerr.ClusterName, Message: cerr.Error()})
+				}
+			}
+		}
+	}
+	labelSelector, err := k8sLabels.ValidatedSelectorFromSet(map[string]string{
+		"pac.weave.works/type": "Admission"})
```
serboctor, line 42: can we have the type as part of the request payload? that way to support Audit we wouldn't need to change the API again?

core/server/policy_violations.go
```diff
@@ -0,0 +1,267 @@
...
+	}
+	if len(validationsList.Validations) == 0 {
+		return nil, fmt.Errorf("no policy violation found with id %s and cluster: %s", m.ViolationId, m.ClusterName)
+	}
+	return &pb.GetPolicyValidationResponse{
+		Violation: validationsList.Validations[0],
+	}, nil
+}
+
+func (cs *coreServer) listEvents(ctx context.Context, clusterClient clustersmngr.Client, clusterName string, extraDetails bool, opts []sigsClient.ListOption) (*validationList, error) {
```
serboctor, line 128: this function eventually returns validation objects not events, I think the name should reflect that. Initially I thought it lists events, and the function that calls it does the mapping, which is what the name implied to me,

core/server/policy_violations.go
```diff
@@ -0,0 +1,267 @@
...
+	}
+
+	return &validationList{
+		Validations: validations,
+		Token:       clist.GetContinue(),
+		Errors:      respErrors,
+	}, nil
+}
+
+func toPolicyValidation(item v1.Event, clusterName string, extraDetails bool) (*pb.PolicyValidation, error) {
```
serboctor, line 172: How about we call it `eventToPolicyValidation` to make it more indicative?

serboctor (changes requested):

core/server/policy_violations.go
```diff
@@ -0,0 +1,267 @@
...
+}
+
+func (cs *coreServer) GetPolicyValidation(ctx context.Context, m *pb.GetPolicyValidationRequest) (*pb.GetPolicyValidationResponse, error) {
+	clusterClient, err := cs.clustersManager.GetImpersonatedClient(ctx, auth.Principal(ctx))
+	if err != nil {
+		return nil, fmt.Errorf("error getting impersonating client: %w", err)
+	}
+
+	selector, err := k8sLabels.ValidatedSelectorFromSet(map[string]string{
+		"pac.weave.works/type": "Admission",
```
serboctor, line 98: can we have this as part of the request payload so we can easily use the API for audit too

serboctor (changes requested):

api/core/core.swagger.json
```diff
@@ -19,7 +19,7 @@
   "paths": {
     "/v1/child_objects": {
       "post": {
-        "summary": "GetChildObjects returns the children of a given object, specified by a GroupVersionKind.\nNot all Kubernets objects have children. For example, a Deployment has a child ReplicaSet, but a Service has no child objects.",
+        "summary": "*\nGetChildObjects returns the children of a given object, specified by a GroupVersionKind.\nNot all Kubernets objects have children. For example, a Deployment has a child ReplicaSet, but a Service has no child objects.",
```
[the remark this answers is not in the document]
waleedhammam: I think it's version of `buf`, I'll check with Araby about the correct version we should use

core/server/policy_violations.go
waleedhammam: will follow the same approch

serboctor (approved):
BE LGTM

core/server/utils.go
```diff
@@ -18,3 +18,11 @@ func GetTenant(namespace, clusterName string, clusterUserNamespaces map[string][
 
 	return ""
 }
+
+func ExtractValueFromMap(mapName map[string]string, key string) string {
```
jpellizzari, line 22: This feels like a bit of an anti-pattern, although I don't know where it is used. This would hide a key not being defined, which is probably something that should be handled on a case by case basis.

jpellizzari (approved):

core/server/utils.go
waleedhammam: I've renamed it and added proper docstring for it

PR #3731, opened 2023-05-25
Fix navigation

### LINKED ISSUE #3730: Breadcrumb restructure
**Problem**

As part of #3704 which introduce policy violation details which will be represented as `Applications > helmRelease/kustomization > violation` , this approach is not doable using the current structure of the breadcrumb as it's a static Routes .
**Solution**

Passing `path` prop to Breadcrumb component which will be set in the page would solve this issue , this will require to restructure the Layout a little bit and changes will be targets all the parent component which defines `Page`

TheGostKasper:
Closes #3730 

**What changed?**
Breadcrumb component now display the `path` as a prop and this value is passed down from the `Page` component 

**Why was this change made?**
This PR provide the ability to create chained routes in the Breadcrumb not just a static ones 

https://github.com/weaveworks/weave-gitops/assets/4614360/a7cc5850-18c5-4d82-adf5-cd185cb11164


jpellizzari:
@TheGostKasper The video isn't working for me

jpellizzari (approved):
Pulled the branch down and LGTM, although if there are any UI changes, images would be helpful.

TheGostKasper:
@jpellizzari i'll record a video once i finish the unit test :ok_hand:

ui/components/Logo.tsx
```diff
@@ -41,8 +41,9 @@ export default styled(Logo)`
     width: auto;
     height: 32px;
   }
-  padding-left: 20px;
+  padding-left: 12px;
```
joshri, line 44: I'm not sure that this is what we want - check out the comparison:
main:
image
this branch:
image

ui/components/Logo.tsx
```diff
@@ -41,8 +41,9 @@ export default styled(Logo)`
...
+  height: 65px;
   //nav width: 200px - space btwn nav and content: 24px - content padding: 24px. All together 248px - 20 for left padding to line up with detail page titles.
```
joshri, line 46: I care less about the alignment of the breadcrumbs, more about the logo padding, but I think this comment should be removed since it's not relevant anymore

ui/components/UserSettings.tsx
```diff
@@ -108,7 +108,7 @@ function UserSettings({ className, darkModeEnabled = true }: Props) {
 }
 
 export default styled(UserSettings)`
-  padding-right: ${(props) => props.theme.spacing.small};
+  // padding-right: ${(props) => props.theme.spacing.small};
```
joshri, line 111: Should this be removed?

joshri (commented):
Looks good - adding the logo left padding back in would make me happy <|:^)

ui/components/Logo.tsx
TheGostKasper: ah i see, setting it to `12px` center the logo to the Nav container not align it to just icons and it's equivalent to `margin:auto`
but with `margin` it flinches due to the change in `width > margin ` when i collapse so i went for `12px` 
specially when it's collapsed
- with `20px`
image

- with `12px`
image

TheGostKasper:
@joshri we can continue the discussion later on .

Issue #2187, opened 2022-05-20
Add CI to progressive-delivery repo


jpellizzari:
Testing, linting, etc on PRs!

PR #3749, opened 2023-06-01
Disable Sync btn while loading and invalidate the react-query cache

ahussein3:
…equest

Closes# [#2187](https://github.com/weaveworks/weave-gitops-enterprise/issues/2817)

**What changed?**
- Disabled sync btn on loading.
- Sent the right react-query key so that the right cache get invalidated.

- Testing Locally.

**Release notes**

**Documentation Changes**


joshri (approved):
Also change the query key in the `CheckboxActions` suspend handler to `object` - good fix!!

ahussein3:
> Also change the query key in the `CheckboxActions` suspend handler to `object` - good fix!!

Hey @joshri we will face an issue here as This component is being displayed in two different views and each view has different reactQuery key: in sources view the key is "sources" while in application the key is "automation"
- The solution here is maybe we can sent the key from automation table to the checkbox component. tell me if that ok I will start work on it.

Issue #3745, opened 2023-06-01
Add a linter for Protobuf files


opudrovs:
As discussed with @yiannistri , the linter should enforce camelCase for field names to suit the current convention in the UI.

opudrovs:
@yiannistri cool, thanks! Will add it.

And I should also probably add the lint-proto CI step somewhere, correct?

opudrovs:
@yiannistri or should linting protobufs be included in the existing `lint` target and step?

From the OSS makefile:
```
lint: ## Run linters against code
	@go install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.52.0
	golangci-lint run --out-format=github-actions --timeout 600s --skip-files "tilt_modules"
```

yiannistri:
I would prefer a single make target for linting so it makes sense to include it there.

opudrovs:
@yiannistri I've fixed most of the linter warnings, so there are only two remaining. Are we sure we want to enable the rule to pluralize names of repeated fields?

This static check run
https://github.com/weaveworks/weave-gitops/actions/runs/5175099620/jobs/9322247561?pr=3752

fails with the following errors:
<img width="797" alt="Screenshot 2023-06-05 at 10 54 37" src="https://github.com/weaveworks/weave-gitops/assets/8824708/b9d6c9be-88d1-4712-8aa0-6dbdc3a50cf0">

The linter wants to pluralize the following `value` and `inventory` fields (to `values` and `inventories` correspondingly):
https://github.com/weaveworks/weave-gitops/blob/8e509d9c1f15dfe445be880dafd23b4573a41889/api/core/core.proto#L296
https://github.com/weaveworks/weave-gitops/blob/8e509d9c1f15dfe445be880dafd23b4573a41889/api/core/types.proto#L97

I am not sure about using `values` in `PolicyParamRepeatedString`, but `inventories` does not look right to me even if it is a repeated field.

yiannistri:
Let's disable that rule. Although I think pluralisation of collections is the right choice, our choice of field names is a bit odd (as you point out with `inventories`). Before enabling that rule we should come up with a better field name.

PR #3752, opened 2023-06-02
Add linter for Protobuf files

opudrovs:
Closes https://github.com/weaveworks/weave-gitops/issues/3745

- Added a config for the `protolint` Protobuf linter.
- Added linting Protobuf files to the `lint` Makefile target.
- Fixed Protobuf linter warnings (including splitting long comments to shorter lines).
- Ran `make proto`.

Notes:
- For linting, I am using the latest version of `protolint`, based on their readme. We can switch to using a fixed version.
- The directories to ignore were copied from the `.gitignore` file. We can extend the list as needed.
- I broke long lines with `returns` in .proto files right before `returns` (because of linter errors), because I have not found a convention for this. It seems to work fine.


.protolint.yaml
```diff
@@ -0,0 +1,58 @@
+---
+# Lint directives.
+lint:
+  # Linter directories to walk.
+  directories:
+    # The specific directories to exclude.
+    exclude:
```
yiannistri, line 7: There's no way to define the directories to _include_ instead of the ones to _exclude_ is there? From a quick look I didn't find anything in the example config.

yiannistri (approved):
Looks great 🥇

.protolint.yaml
opudrovs: I have not found an example for including dirs in the config yet. Protolint required a param passed for what to lint. So, I am passing a dot `protolint lint .`

opudrovs: Maybe it's considered the included dirs?

opudrovs: If I don't find how to add specifc paths to the config, I will add them as the positional argument.

yiannistri: > Maybe it's considered the included dirs?

Could be 🤔

opudrovs: Added the `/api` directory as a positional argument to the lint command (tested that it picks up errors in this dir recursively) and removed the directories to exclude from the config.

PR #3781, opened 2023-06-09
Add edit icon, modify Pipelines and Delivery icons for EE dark mode

joshri:
Adding edit icon so EE can supply non-Mui specific colors for dark mode. Adjust Pipelines and Delivery icons for EE dark mode

before:
image

after:
image

Unfortunately the Pipelines icon is in a tough spot - the fill on the middle section can not be hardcoded to white, but once it's transparent we get that extra filled-in section on the wings. Needs an adjustment in Figma.

Dark Mode Edit button:
image


TheGostKasper:
I believe we need to keep it as it's, there's a huge difference in Icons as 
image
image

but there's a chance to enhance the `PolicyMode` to mock the old styles `color fill or size` .

@AsmaaNabilBakr

joshri:
Cool - made an issue here - we can't have the same icon twice so the svg will need to support both styles https://github.com/weaveworks/weave-gitops/issues/3782

opudrovs:
Some tests are failing. The order of imports.

joshri:
oop thanks Olga - fixed.

opudrovs (approved):

Issue #3722, opened 2023-05-24
Add `break-glass` command for Terraform


chanwit:
Absorb this feature https://github.com/weaveworks/tf-controller/pull/618 into `gitops` cli.

Outcome: `gitops break-glass terraform <name>`

Tech details:
- Shim the break-glass implementation from `tfctl` as the command for the `gitops` cli.

Acceptance criteria
- [ ] Implement a PR

yiannistri:
We will create a new issue for the docs to keep things separate.

yiannistri:
We cannot continue with this right now because the dependencies in OSS are different from the dependencies of tf-controller.

From https://weaveworks.slack.com/archives/C0586V3N0BG/p1686833467740889

> I don't think we can use the tf-conrtoller in oss now, if we pull in the module, it requires `controller-runtime
 v0.15` while OSS is on v0.14. This update would require
update fluxcd/pkg/runtime
update changed function calls all over the codebase
change how we use client.NewDelegatingClient because it's not part of the controller runtime anymore
So that's a lot of core changes too

chanwit:
So it depends on and is being blocked by #3746

yiannistri:
Still blocked as controller-runtime in OSS is still on 0.14.6

LappleApple:
Closing due to change of plans.

PR #3791, opened 2023-06-14
Add break-glass command for Terraform

opudrovs:
Closes https://github.com/weaveworks/weave-gitops/issues/3722

WIP

### DIFF
--- a/cmd/gitops/breakglass/cmd.go
+++ b/cmd/gitops/breakglass/cmd.go
@@ -1,0 +1,24 @@
+package breakglass
+
+import (
+	"os"
+
+	"github.com/spf13/cobra"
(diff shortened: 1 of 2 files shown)


yitsushi:
Add my comment from Slack:

> I don't think we can use the `tf-conrtoller` in oss now, if we pull in the module, it requires `controller-runtime v0.15` while OSS is on `v0.14`. This update would require
> * update `fluxcd/pkg/runtime`
> * update changed function calls all over the codebase
> * change how we use client.NewDelegatingClient because it's not part of the controller runtime anymore
>
> So that's a lot of core changes too

LappleApple:
Closing due to change of plans for this effort.

PR #3724, opened 2023-05-24
Dark Mode UI bug follow ups

joshri:
Closes #3682 
Closes #3683

This PR fixes two requests made during the Dark Mode work:

https://github.com/weaveworks/weave-gitops/assets/65822698/e8906d6e-c1b7-495a-8618-3038eb533d9b

- Sync and ArrowDropDown buttons are now the same height, with their right and left borders removed respectively. 
- Border radius was being applied to each `th` cell - now it's only the first and last as intended.

PR #3796, opened 2023-06-20
Update theme

AlinaGoaga:
- Update theme to remove the dark border we have on buttons and update sync button style to fix some inconsistencies when disabled & hovered between "sync" and "sync options dropdown" (see below)

Before
Screenshot 2023-06-20 at 19 06 23

After
Screenshot 2023-06-20 at 19 09 17

- The changes to input will remain in EE since adding them to the theme would have meant having to overwrite the style on both SignIn page and in filters


joshri:
Manuela wanted the border separating `sync` and the dropdown removed so I did it here https://github.com/weaveworks/weave-gitops/pull/3724

Can we keep that or is that what broke things in light mode?

AlinaGoaga:
Yup it looked weird (it basically looks like in the "before" screenshot I attached with a different combination of colors 😅 ). Because it's not a button but a collection of buttons when you hover on top of one the other's state doesn't change. Without the border in between it looks a bit weird I think. 

We actually have another similar button on EE, with the border visible 
Screenshot 2023-06-21 at 17 50 28

joshri (approved):

PR #3819, opened 2023-07-06
Align flux-dependencies in go.mod

foot:
- Around the https://github.com/fluxcd/flux2/blob/v2.0.0-rc.3/go.mod
- v2.0.0-rc.3 is the last release that uses controller-runtime 0.14.x
- controller-runtime 0.15.0 introduces lots of breaking changes and a dependence on k8s golang 0.27.0

Follows https://github.com/weaveworks/weave-gitops/pull/3818 which reverts the migration to controller-runtime 0.15.0 for now.

Closes

**What changed?**

Align all the flux dep versions so they're more or less in sync.

**Why was this change made?**

Lots of diff deps in flux, want them to be aligned so no funny expected behaviour pops up

Looking up the v2.0.0-rc.3 go.mod and making weave-gitops' go.mod versions as similar as possible

- local tests
- pulled into weave-gitops-enterprise and ran smoke there too.

**Release notes**

**Documentation Changes**


go.mod
```diff
@@ -147,7 +147,7 @@ require (
 	github.com/exponent-io/jsonpath v0.0.0-20210407135951-1de76d718b3f // indirect
 	github.com/fluxcd/notification-controller/api v0.32.1
```
yiannistri, line 148: @foot can we also bump this one to `github.com/fluxcd/notification-controller/api v1.0.0-rc.3`? Not sure why this is not shown along with the other Flux imports, perhaps we can move it up too.

foot: Oops! Good catch

yiannistri (approved):

PR #3829, opened 2023-07-07
Allow for custom actions for data table checkboxes

joshri:
Part of: [#3034](https://github.com/weaveworks/weave-gitops-enterprise/issues/3034)

Draft because I want to make sure this does not conflict with @opudrovs current image automation work

PR #3834, opened 2023-07-10
Enhance types in checkbox actions for EE terraform sync/suspend all

joshri:
Follow up to: #3829 

Getting Typescript errors when actually trying to apply these changes in EE.

The props in `createElement` have to be typed otherwise whatever you're passing down is going to conflict - in this case since we want them to be custom I've given it the sad `any` type. But at least now we know that it's `any`.

### DIFF
--- a/ui/components/CheckboxActions.tsx
+++ b/ui/components/CheckboxActions.tsx
@@ -80,7 +80,7 @@ const DefaultSuspend: React.FC<{
 };
 
 export type Action = {
-  element: React.ReactElement;
+  element: React.FC;
   additionalProps?: { [key: string]: any };
 };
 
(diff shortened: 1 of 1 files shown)


ui/components/CheckboxActions.tsx
```diff
@@ -80,7 +80,7 @@ const DefaultSuspend: React.FC<{
 };
 
 export type Action = {
-  element: React.ReactElement;
+  element: React.FC;
```
jpellizzari, line 83: The prop name probably needs to be `component=` instead of `element` if we want a component instead of an instantiated element.

jpellizzari (approved):

Issue #3832, opened 2023-07-10
OIDC Refresh token renew twice and 401 Re-auth is occur after id_token expiration since 0.22


z15alall:
**Describe the bug**

Error With OIDC Authentification

Until 0.21 version i can login with oidc provider and stay logged after the id_token expiration (5 minutes for our case), but since 0.22 (i have also tested in 0.26 and 0.27) after the id_token expiration i have an authentification error 401 and go back to the login interface.

I supposed that error is introduce with this `[code](https://github.com/weaveworks/weave-gitops/pull/3597)` added in version 0.22

Indeed in 0.21 the id_token was refresh with the refresh token only once and i can stay logged more than 5 minutes (our id_token expiration time), now since 0.22 when the id_token expired we need we have a  401 error go back to the login interface.

In the log 0.27 below we saw at line _13_ a "_Successfully refreshed token_"  with good principal and line _15_ a "_refreshing token failed_" i supposed the same refresh_token is used twice to renew id_token and the second fail because the refresh_token was used by the first refresh token call that why we have an 400 error from our identity tool line 15.

**Environment**

- Weave-Gitops Version 0.27 
- Flux Version 2.0.0
- Kubernetes version: 1.26
 
**To Reproduce**

Reduce the tokenDuration to spend less time during the test to 30s for example :

```yaml
claimGroups: groups
claimUsername: email
clientID: clientid
customScopes: groups
issuerURL: https://xxxxxx
redirectURL: https://xxxx/oauth2/callback
tokenDuration: 0h0m30s
```

**Expected behavior**

Stay login until the access token/refresh token is availlable 

**Actual Behavior**

Unlogged when the id_token is expired because i think the refresh token is used twice en the second refresh token authentification failed

**Additional Context (screenshots, logs, etc)**

Log in 0.27 : 
```
  1 2023-07-10T08:47:14.438Z        DEBUG   gitops.auth-server      auth/jwt.go:107 parsed JWT token        {"expires": "2023-07-10T08:48:31.000Z"}
  2 2023-07-10T08:47:14.439Z        DEBUG   gitops.auth-server      auth/jwt.go:162 Found principal {"user": "xxxxxx", "groups": ["xxxx", "xxxx", "xxxx", "xxxx"], "tokenLength": 0, "method": "*auth.JWTCookiePrincipalGetter"}
[...]

PR #3842, opened 2023-07-13
UI token refreshing

foot:
Closes https://github.com/weaveworks/weave-gitops/issues/3832

**What changed?**

Token refreshing is now handled by the UI, not in-request by the server.

**Why was this change made?**

We cannot lock the refresh process on the server well, new requests will still come in and re-trigger it etc.

Add more logic the UI to deal with expired tokens
- Extend the UnAuthorizedInterceptor to try and refresh the token if we get a 401
- Keep track of inflight tokenRefreshes so we only have a single one at a time
- Wait for token refresh to complete before retrying requests

On the server we remove the inline token refresh logic

Manual testing mainly:

1. Setup up weave-gitops + OIDC w/ a provider that supports offline-access
2. Login
3. delete id_token cookie
4. Watch 401s return and a single refresh request made that restores the id_token

```
kind create cluster
  --namespace default \
  --from-literal=issuerURL=<your-issuer-url> \
  --from-literal=clientID=<your-client-id> \
  --from-literal=clientSecret=<your-oidc-client-secret> \
  --from-literal=redirectURL=http://localhost:4567/oauth2/callback
FAST_AND_FURIOUSER=1 SKIP_UI_BUILD=1 tilt up

npm install
npm start
open http://localhost:4567
```

Waterfall

Note 401s, single refresh, immediately followed by the 200s

image

Follow ups:
- [x] The refresh token navigates away, but resolves the promise, it perhaps should not do that, so no new requests start.
- [ ] /userinfo should also be wrapped by the unauth'dwrapper the UI too, its the first endpoint we hit when loading the UI to see if the user is logged in. We don't do this yet, so if you have a refresh-token and navigate to the UI fresh, we don't try and get you a new token right now.


bigkevmcd (approved):
This looks great!

AlinaGoaga (approved):
Looks great, tested locally and it's working as expected!

PR #3854, opened 2023-07-18
Remove 'with source' option from Sources Sync

### LINKED ISSUE #3853: Remove Sync dropdown from Sources and Source details pages
As discussed with @joshri , the Sync  button on the Sources page should not have the dropdown, it should be just Sync, similar to the Sync button on the Source Details page.

Screenshot 2023-07-18 at 14 21 47

@yiannistri WDYT?

joshri:
Closes #3853 

Adds an object where you can add any Routes that should have the `hideDropdown` prop enabled on their DataTable, in this case for the SourcesTable.

image


opudrovs (approved):
LGTM!

PR #3871, opened 2023-07-20
Tweaking dependabot to just raise PRs for security updates. 

enekofb:
Tweaking dependabot to just raise PRs for security updates not version updates. 

To have version updates managed by dependabot was not in scope so it is an unintended consequence. This PR tries to revert this behaviour as the documentation [suggests](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#open-pull-requests-limit)


Himangini (approved):
👍🏻

PR #3876, opened 2023-07-24
Remove Flux version from /v1/version response and page footers

### LINKED ISSUE #3665: Flux version not shown
**Describe the bug**

Whenever I access the Dashboard with a user that only has access to a single Namespace, this is what the footer shows:

The Flux version isn't shown, instead it reads `-`.

**Environment**

- Weave-Gitops Version: 0.22.0
- Flux Version: reproducible with 0.41.2 and 2.0.0-rc.1
- Kubernetes version: 1.27.1

**To Reproduce**

Access the Dashboard with a user that only has access to a single Namespace.

**Expected behavior**

The Flux version is shown. Alternatively show a message that tells me why I can't see the Flux version.

**Actual Behavior**

It just shows `-`.

**Additional Information**

This is logged by the weave gitops container when accessing the version endpoint:

```
2023-05-04T10:01:56.465Z        ERROR   gitops.core-server      server/version.go:104   error getting flux version      {"error": "error getting list of objects"}
```

makkes:
The Flux version has been displayed in two locations of the UI: In the footer of every page and on the "Flux Runtime" page. To display the Flux version in these locations, two different ways to retrieve said version have been implemented:

1. On the "Flux Runtime" page the version is retrieved from the Flux deployment objects for each cluster WeGO knows about.
2. In the footer the version is retrieved from the /v1/version endpoint which in turn tries to find the Flux namespace on the cluster that it runs on.

In order to properly display the Flux version properly, the UI user would have to be allowed to list (1) Namespaces on the management cluster and (2) Flux deployments on all clusters including the management cluster.

This leads to situations where one of these two permissions is not set and the user will see a Flux version properly rendered in one location but not the other which is potentially confusing.

In order to improve the user experience, this change completely drops the rendering of the Flux version in the footer so that there is only one consistent method to fetch the Flux version. Users that aren't allowed to list the Flux runtime information will not be able to find out the Flux version, anyway, so for those users rendering an "-" in the footer doesn't provide any information. All others can refer to the "Flux Runtime" page and will see all Flux versions on all clusters together with additional runtime information.

closes #3665


bigkevmcd (approved):
Looks good to me, thanks for this.

makkes:
@bigkevmcd @luizbafilho can one of you merge this for mw, please? I don't have permission.

PR #3885, opened 2023-07-25
Add sync and suspend for image update automations

joshri:
Allows sync and suspend for image update automations + fixes a bug where buttons were not disabled when un-checking the last checked button (having variable outside of useEffect prevented update -> see CheckboxActions component)

https://github.com/weaveworks/weave-gitops/assets/65822698/abdb7cc5-1670-45f7-a535-ae51955837a3

### DIFF
--- a/core/fluxsync/adapters.go
+++ b/core/fluxsync/adapters.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 
 	helmv2 "github.com/fluxcd/helm-controller/api/v2beta1"
+	imgautomationv1 "github.com/fluxcd/image-automation-controller/api/v1beta1"
 	reflectorv1 "github.com/fluxcd/image-reflector-controller/api/v1beta2"
 	kustomizev1 "github.com/fluxcd/kustomize-controller/api/v1beta2"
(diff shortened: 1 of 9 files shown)


yiannistri (commented):
I think we can remove the 'Sync without source' option, as for image update automation resources, there is no source to reconcile anyway.

Sidenote: It's somewhat confusing that the options we present in the UI are the opposite of what the Flux CLI uses. As an example, the following command is used to reconcile a resource **without also reconciling its source**:
```
# Trigger a Kustomization apply outside of the reconciliation interval
flux reconcile kustomization podinfo
```
The following command reconciles the resource **with its source**:
```
# Trigger a sync of the Kustomization's source and apply changes
flux reconcile kustomization podinfo --with-source
```
In my opinion, the UI should match the Flux CLI to avoid possible user confusion. By the way this is just a suggestion and can be addressed separately as it's unrelated to the changes of this PR.

joshri:
@yiannistri 

Removed source sync for image update automations:
image
image

Default sync behavior will be addressed in a separate PR as discussed in standup 🌞

yiannistri (approved):
LGTM 👍

PR #3894, opened 2023-07-28
Update docs with verified status

### LINKED ISSUE #3895: [Docs] Update docs with verified status information for sources

AlinaGoaga:
Closes #3895


AlinaGoaga:
@bigkevmcd @foot I've had to take new screenshots to show the verified status (I used `demo-01`). Some bits of the description are not as detailed as before so I'm wondering if we want to go back to that (I might need some help getting the sources/apps that we previously had in the screenshots on a local cluster).

foot (approved):
💯

PR #3915, opened 2023-08-04
Update video on main docs page

LappleApple:
Closes

**What changed?**
Remove GitOps Run demo video and replace with a newer video featuring Terraform Controller.

**Why was this change made?**

To provide fresher and more relevant content for users.


makkes (approved):
🎉

PR #3921, opened 2023-08-07
Add Azure-related pages to user docs

LappleApple:
**What changed?**
Adds documentation related to Azure Marketplace offering. Fixes an issue in the install OSS guide. Updates sidebars.js with new pages.


website/docs/enterprise/getting-started/install-enterprise-azure.mdx
```diff
@@ -0,0 +1,276 @@
...
+
+Click "Save". Then clicking "Next", which will give you a summary so you can review your input. Then click "Create". It will take about five minutes to deploy.
+
+You'll get to a new screen, which at the top-right shows "Notifications" and will display creation of the Flux configuration. When your deployment succeeds, go to the resource and pin to your dashboard. Then go to your terminal to see if it works in kubectl. In the terminal you'll get the GitRepository and Kustomizations. You should then get a green "succeeded" checkmark.
+
+The Kustomisations screen does not provide an option to inspect the path/target namespace—you have to supply the target Namespace in the Kustomization object.
+
+## 2. Install CAPZ, the CAPI Provider
+
+The installation wizard in the Azure Marketplace UI offers a checkbox where you can select to do this.
```
foot, line 51: This is not required anymore, could emphasise its optional if you want to provision or manage CAPI clusters. afaiui you can install it later and update the "capi support enabled" option in the WGE Azure marketplace installation wizard

website/docs/enterprise/getting-started/install-enterprise-azure.mdx
```diff
@@ -0,0 +1,276 @@
...
+
+## 3. Apply the Entitlements Secret
+
+Contact sales@weave.works for a valid entitlements secret. This will come in the form of a file “entitlements.yaml”. Apply it to the cluster:
+
+```
+kubectl apply -f entitlements.yaml
+```
+
+## 4. Configure Access for Writing to Git from the UI
```
foot, line 61: if we want to include all the options here (not just Azure) as we do in the other WGE installation instructions we could pull out the common instructions to a partial and import into both docs to keep it in sync https://docusaurus.io/docs/markdown-features/react#importing-markdown

website/docs/enterprise/getting-started/install-enterprise-azure.mdx
```diff
@@ -0,0 +1,276 @@
...
+- Namespace: flux-system
+- Scope: Cluster
+
+All of the displayed properties for the Flux objects screen are the same as what you'd supply to Flux bootstrap.
+
+Click "Next", then add:
+- Source Kind: Git repository
+- Repository URL: [your repository URL here]
+- Reference Type: Branch
+- Repository Type: Private
```
foot, line 29: This is still a useful guide but WGE _should_ auto-install flux now when you try and install it.

website/docs/enterprise/getting-started/install-enterprise-azure.mdx
foot: > The installation wizard in the Azure Marketplace UI offers a checkbox where you can select to do this. 

_If you are planning to manage or connect CAPI clusters to the WGE service make sure you first install the CAPI provider. Then during the WGE installation process be sure to select the "Enable CAPI support" checkbox._

website/docs/enterprise/getting-started/install-enterprise-azure.mdx
LappleApple: @foot Should we have the auto-install as the main/default doc text but reference these lines as "if for some reason WGE doesn't install Flux automatically ..."—like a fallback option?

foot: WGE won't install if the flux auto-installation fails, so could phrase it like that. Good backup option to work around some error there.

The guide to creating a source / kustomization is great though, the non-azure instructions ask you to `flux bootstrap` which does this initial git/ks creation for you. It should be done at some point before you use many WGE features like rendering templates out to git, which assumes that git repo will be being sync'd back into the cluster.

foot: What do you think about maybe moving it down the bottom to a troubleshooting section? e.g. The happy install path shouldn't require the user do this manually.

---

#### Troubleshooting

WGE will try and automatically install flux on a new cluster, if this fails for some reason or if you need a more custom flux installation you can manually install it before installing WGE.

foot (approved):
Otherwise I think we're looking pretty good!

website/docs/enterprise/getting-started/install-enterprise-azure.mdx
LappleApple: Good idea, I did that

PR #3926, opened 2023-08-07
Allow renaming the default cluster.

bigkevmcd:
Closes

**What changed?**
This provides functionality for renaming the default cluster in Weave GitOps from "Default" to a custom name.

Simple command-line rename, chart entry for setting it when required.

**Why was this change made?**
To allow us to pass in names from the Backstage plugin.

Tweak the helm values in `tools/helm-values-dev.yaml`:

```diff
+clusterName: testing
+
```

The new cluster name will be visible in the URL bar and queries to the backend will include the new name.

See above

**Release notes**
Nothing

**Documentation Changes**
This will be addressed in the Backstage docs.

### DIFF
--- a/charts/gitops-server/templates/deployment.yaml
+++ b/charts/gitops-server/templates/deployment.yaml
@@ -59,6 +59,9 @@ spec:
             - "--enable-metrics"
             - "--metrics-address=:{{ .Values.metrics.service.port }}"
             {{- end }}
+          {{- if .Values.clusterName }}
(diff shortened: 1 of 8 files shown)


foot:
Main flow works great.

The `GetVersion` endpoint seems to require the correct name too, it stops showing the k8s version in the status bar down the bottom rather than an error.

image

Other places its used:
- Policy: seems okay, uses the value coming from the UI
- GetSessionLogs: I can't actually see this being called from ts/golang.

Issue #3634, opened 2023-04-20
Notifications page cluttered with error messages


makkes:
**Describe the bug**

When navigating to the Notifications page with a user that doesn't have access to the notification resources in all Namespaces then an error is rendered for each and every Namespace:

![0057](https://user-images.githubusercontent.com/387444/233338152-33a9cebf-e101-4279-a440-88990e7314ef.png)

To actually view the resources I _**do**_ have acess to I need to scroll all the way down on the page:

![2023-04-20_10-36](https://user-images.githubusercontent.com/387444/233338194-fd116f63-ce35-476d-95a4-d02c2de1d3a4.png)

What's adding even more distraction is that the errors keep getting re-ordered, making the page really noisy.

**Environment**

- Weave-Gitops Version: 0.21.2
- Flux Version: 2.0.0-rc.1
- Kubernetes version: 1.27.1

**To Reproduce**

1. Install Weave GitOps
2. Log in as a user that only has access to a selected set of namespaces
3. Navigate to "Notifications"

**Expected behavior**

The page is not cluttered with error messages so that I'm not distracted by them and that I don't have to scroll down to see the resources I'm allowed to see.

**Actual Behavior**

see above

PR #3927, opened 2023-08-07
Filter duplicate error messages

joshri:
[Closes #3139](https://github.com/weaveworks/weave-gitops-enterprise/issues/3139)

Quick fix for OSS errors displaying the same error like seven or eight times. It would be useful for OSS to adopt Enterprise's cleaner error handling in the future


makkes:
@joshri is this related to #3634 as well?

joshri:
@makkes not quite, this only removes errors that have the exact same message from the list. Jordan's going to take on the permission errors situation

jpellizzari (approved):

PR #3932, opened 2023-08-09
Fix version in tf getting started guide

joshri:
the getting started guide currently leads to an error where the Terraform crd isn't found due to a version mismatch where v1alpha1 should really be v1alpha2 - FIXED IT


bigkevmcd (approved):
Awesome, thank you!

PR #3933, opened 2023-08-09
Add batch error handling to Notifications page

### LINKED ISSUE #3634: Notifications page cluttered with error messages

jpellizzari:
Closes #3634 

Adds an `ErrorList` component to Core, heavily inspired by the one used in EE. This also exports the `ErrorList` component so that it may be imported in EE and eventually subsume the `AlertListErrors` component (there is nothing wrong with that component currently, but this will enforce Consistency:tm:).

There are now two error "classes" on the page:

Error when we can't reach the backend:
Screenshot from 2023-08-09 12-40-04

Batched errors from various clusters (backend returns lots of errors):
Screenshot from 2023-08-09 12-40-22

Simulated the error state using this:

```diff
 core/server/objects.go | 10 ++++++++++
 1 file changed, 10 insertions(+)

diff --git a/core/server/objects.go b/core/server/objects.go
index 0bb6e26d5..f99290bbc 100644
--- a/core/server/objects.go
+++ b/core/server/objects.go
@@ -14,6 +14,7 @@ import (
 	pb "github.com/weaveworks/weave-gitops/pkg/api/core"
 	"github.com/weaveworks/weave-gitops/pkg/run/constants"
 	"github.com/weaveworks/weave-gitops/pkg/server/auth"
+	"github.com/weaveworks/weave-gitops/pkg/utils"
 	v1 "k8s.io/api/apps/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -148,6 +149,15 @@ func (cs *coreServer) ListObjects(ctx context.Context, msg *pb.ListObjectsReques
 		}
 	}
 
+	if msg.Kind == "Provider" {
+
+		for i := 0; i < 50; i++ {
+			msg, _ := utils.GenerateRandomString(5, 10)
+			respErrors = append(respErrors, &pb.ListError{ClusterName: "provider", Message: fmt.Sprintf("such and such is forbidden in namespace: %v", msg)})
+		}
+
+	}
+
 	return &pb.ListObjectsResponse{
 		Objects: results,
 		Errors:  respErrors,
```

### DIFF
--- a/ui/components/Alert.tsx
+++ b/ui/components/Alert.tsx
@@ -1,7 +1,8 @@
-import { AlertTitle, Alert as MaterialAlert } from "@material-ui/lab";
+import { Alert as MaterialAlert, AlertTitle } from "@material-ui/lab";
 import * as React from "react";
 import styled from "styled-components";
 import Flex from "./Flex";
+import Icon, { IconType } from "./Icon";
 import Text from "./Text";
 
 /** Alert Properties */
@@ -21,7 +22,12 @@ export interface Props {
 function UnstyledAlert({ center, title, message, severity, className }: Props) {
   return (
     <Flex wide start={!center} className={className}>
-      <MaterialAlert severity={severity}>
+      <MaterialAlert
+        icon={
+          <Icon type={IconType.ErrorIcon} size="medium" color="alertDark" />
+        }
+        severity={severity}
+      >
         <AlertTitle>{title}</AlertTitle>
         <Text color="black">{message}</Text>
       </MaterialAlert>
--- a/ui/components/ErrorList.tsx
+++ b/ui/components/ErrorList.tsx
@@ -1,0 +1,128 @@
+import { Box, Collapse } from "@material-ui/core";
+import { Alert } from "@material-ui/lab";
+import { sortBy, uniqBy } from "lodash";
+import * as React from "react";
+import styled from "styled-components";
+import Button from "./Button";
+import Flex from "./Flex";
+import Icon, { IconType } from "./Icon";
+import Text from "./Text";
+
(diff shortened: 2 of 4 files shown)


jpellizzari:
Authoring the PR made me realize the Alert icons are inconsistent, so I fixed that in a second commit:
Screenshot from 2023-08-09 12-54-49

joshri (approved):
LGTM - might be worth distinguishing between the request errors and permissions errors somehow...maybe permissions errors are more like warnings? Just a thought for the future

jpellizzari:
> LGTM - might be worth distinguishing between the request errors and permissions errors somehow...maybe permissions errors are more like warnings? Just a thought for the future

I think "errors is errors" as far as the backend is concerned. It doesn't differentiate between permissions and request errors.

PR #3944, opened 2023-08-15
Set static height on signin footer

### LINKED ISSUE #3210: Gitops run automation-kind annotation shows incorrect kind
Two screenshots here - one shows the annotation specifying the automation kind as `ks` in the dev console, second shows that it's actually a helm-release in the applications tab?! I didn't specify anything on creation which could be the problem. This is blocking linking run sessions to automations and sources [#1963](https://github.com/weaveworks/weave-gitops-enterprise/issues/1963) as the kinds are not lining up. Not sure which one is incorrect (run table or automation table).

image (1)
image

jpellizzari:
Closes #3210 

Sets a static height on the footer to ignore the "wheel" image height. I chose `300px` because it was roughly what both images were resolving to naturally.

### DIFF
--- a/ui/pages/SignIn.tsx
+++ b/ui/pages/SignIn.tsx
@@ -44,6 +44,7 @@ const OidcFlex = styled(Flex)`
 const Footer = styled(Flex)`
   & img {
     width: 500px;
+    height: 300px;
   }
 `;
 


joshri (approved):
LGTM 🌞

enekofb:
Btw, i rebased the PR to validate the behaviour of the release locking PR in case there was an issue to fix it with less impact possible to us.  It worked as you could see from the checks. Thanks you!

PR #3946, opened 2023-08-16
JWT cookie fetcher converted to session storage.

bigkevmcd:
Closes

**What changed?**
Add initial session storage

This pulls in the scs package for storing sessions.

This means that cookies are stored in-memory for now, with a short cookie being sent to the user's browser, we access the old ID and Refresh cookies from the session.

**NOTE**: This is an initial release of this code, we will provide support for configuring external storage.

**Why was this change made?**
Support longer cookies

Tests and running the code

**Release notes**
Switching to session storage for cookies.

**Documentation Changes**


pkg/server/auth/server.go
```diff
@@ -171,21 +175,36 @@ func claimsConfigFromSecret(secret corev1.Secret) *ClaimsConfig {
...
+func NewAuthServerConfig(log logr.Logger, oidcCfg OIDCConfig, kubernetesClient ctrlclient.Client, tsv TokenSignerVerifier, namespace string, authMethods map[AuthMethod]bool, noAuthUser string, sm SessionManager) (*AuthServerConfig, error) {
+	if authMethods[OIDC] {
+		if _, err := url.Parse(oidcCfg.IssuerURL); err != nil {
+			return nil, fmt.Errorf("invalid issuer URL: %w", err)
+		}
+
+		if _, err := url.Parse(oidcCfg.RedirectURL); err != nil {
+			return nil, fmt.Errorf("invalid redirect URL: %w", err)
+		}
+	}
```
foot, line 191: These two checks are in initAuthServer. They moved over there in #3958 as part of the the light refactoring to the very long arg lists. The thought was we could remove this `NewAuthServerConfig` function to and just create the AuthServerConfig literal.. or rename initAuthServer to become NewAuthServerConfig etc instead.

I think I stopped at that point as there were quite a few `NewAuthServerConfig`s around that would cause even more conflicts to resolve etc given the other inflight PRs..

pkg/server/auth/auth_test.go
```diff
@@ -407,3 +416,76 @@ func TestUserPrincipal_String(t *testing.T) {
...
+
+func contextWithSessionValues(values map[string]any) context.Context {
+	return contextWithValues(context.TODO(), values)
+}
+
+func contextWithValues(ctx context.Context, values map[string]any) context.Context {
+	return context.WithValue(ctx, sessionsCtxKey{}, values)
+}
+
+type fakeSessionManager struct {
```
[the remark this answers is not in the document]
foot: Can wait for someone to need it first.

cmd/gitops-server/cmd/cmd.go
```diff
@@ -186,12 +187,16 @@ func runCmd(cmd *cobra.Command, args []string) error {
 		return fmt.Errorf("couldn't get current namespace")
 	}
 
+	sessionManager := scs.New()
+	// TODO: Make this configurable
+	sessionManager.Lifetime = 24 * time.Hour
```
foot, line 192: Could make this a flag, but 24h is pretty good for now too.

foot (approved):
Tested w/ OIDC / no-auth / user-account all seems to be working great

Issue #2907, opened 2022-10-18
GitOps Run: session metadata


ozamosi:
For #2898, we need to be able to find our definition of vcluster, and extract a bunch of information about the session.

Already, a vcluster `StatefulSet` is tagged with the label `app: vcluster`. We should also set the following extra label:
 * `app.kubernetes.io/part-of: gitops-run`

We also needs to set these annotations:
 * `run.weave.works/cli-version` to the CLI version
 * `run.weave.works/port-forward` to a comma separated list of ports that are forwarded
 * `run.weave.works/command` to the full command that was run

We can now define a vcluster session as a `StatefulSet` with the two labels `app:vcluster` and `app.kubernetess.io/part-of: gitops-run`, and we can drive the UI with the additional annotations.

PR #3953, opened 2023-08-17
Add Flex with breakpoint prop

joshri:
Closes: [#2907](https://github.com/weaveworks/weave-gitops-enterprise/issues/2907)

Creates styled component extension of Flex with a breakpoint prop to swap `flex-direction` when below a certain screen width

laggy video here:

https://github.com/weaveworks/weave-gitops/assets/65822698/2fb955b0-feaf-4326-91bb-f0e61578ab06

### DIFF
--- a/ui/components/Flex.tsx
+++ b/ui/components/Flex.tsx
@@ -59,10 +59,4 @@ class Flex extends React.PureComponent<Props> {
   }
 }
 
-export const ColumnOnBreakpoint = styled(Styled(Flex))<{ breakpoint: number }>`
-  @media screen and (max-width: ${(props) => props.breakpoint}px) {
-    flex-direction: column;
-  }
-`;
-
 export default Styled(Flex);
--- a/ui/components/Flex.tsx
+++ b/ui/components/Flex.tsx
@@ -59,4 +59,10 @@ class Flex extends React.PureComponent<Props> {
   }
 }
 
+export const ColumnOnBreakpoint = styled(Styled(Flex))<{ breakpoint: number }>`
+  @media screen and (max-width: ${(props) => props.breakpoint}px) {
+    flex-direction: column;
+  }
+`;
+
 export default Styled(Flex);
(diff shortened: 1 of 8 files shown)


TheGostKasper:
I think we don't need to add the breakpoint as we can make use of the `flex-wrap` & `between` for `Flex`


```
   <Flex wide between wrap>
        <SyncActions
          name={name}
          ...
        />
        <Flex gap="14">
          {automation?.type === "HelmRelease" ? (
            ...
          <LargeInfo
           ... />}
          />
        </Flex>
      </Flex>
```

https://github.com/weaveworks/weave-gitops/assets/4614360/0969ea25-34d2-4ce6-b09e-3c24b6835a23

ui/components/Flex.tsx
```diff
@@ -59,4 +59,10 @@ class Flex extends React.PureComponent<Props> {
   }
 }
 
+export const ColumnOnBreakpoint = styled(Styled(Flex))<{ breakpoint: number }>`
```
jpellizzari, line 62: I dig this :+1:

jpellizzari (commented):
I like the creatrivity, but I would also like to see if @TheGostKasper 's suggestion would work here if possible.

joshri:
We'd lose the conditional top padding, which prevents this weird spacing:
image

But also...I might be losing my mind, but when I add the `wrap` prop nothing happens...I have my code in the screenshot to prove I added the prop haha
image

TheGostKasper:
Flex gap controls space in between, you can adjust the right gap. and by adding `wrap` sets whether flex items are forced onto one line or can wrap onto multiple lines.

 [This record](https://github.com/weaveworks/weave-gitops/pull/3953#issuecomment-1683700734)  without setting a gap value.

joshri:
@TheGostKasper Any ideas on why the wrap prop isn't working in my code in the right half of my screenshot?

TheGostKasper:
I believe it's because there's no overlap in width as the Flex Row still show both div without any problems.

joshri:
But the div on the right is clearly off the screen? What am I missing?

joshri:
There was an important diff between EE and OSS here - OSS added a special component for the sync and suspend buttons a couple weeks ago that is now set to 50% width by default, allowing for the large info flex to now wrap when it runs up against it.

TheGostKasper (approved):

PR #3967, opened 2023-08-27
Fix spaceing between buttons

### LINKED ISSUE #3891: Space in between buttons and weird black line
We have to define a space between the buttons and apply this rule to all. It doesn't matter what kind. I think 12px might be fine.

The line color on the sync arrow buttons should be gray, not black.

AsmaaNabilBakr:
Closes #3891 

**What changed?**
Set a fixed space between buttons to be `12px`

before:
image
after:
image

before:
image
after:
image

before:
image
after:
image

before:
image
after:
image


jpellizzari (commented):
Using `!important` in lots of places can make things hard to customize in the future. Are we sure we need these everywhere? I wonder if we could achieve it with regular CSS specificity logic.

Some docs on the subject: https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity

Love the thorough write up with pics :+1:

jpellizzari (approved):
Looking good! :+1:

PR #3974, opened 2023-08-30
Modify hook on verified status check

joshri:
I was asked about what the `verified` column meant on the automations table, and I noticed a potential improvement where we wouldn't have to filter through all the sources every time we display the status, and instead just request the object we need directly (and only running the request if it's a Git repo or OCI repo in the first place). Is this a thing?


ui/components/VerifiedStatus.tsx
```diff
@@ -23,17 +21,25 @@ const getVerifiedStatusColor = (status?: string) => {
...
 export const findVerificationCondition = (
-  a: VerifiableSource
+  a: VerifiableSource | undefined
 ): Condition | undefined =>
   a?.conditions?.find((condition) => condition.type === "SourceVerified");
 
-export const VerifiedStatus = ({
+const checkVerifiable = (sourceRef: ObjectRef): boolean => {
+  const { name, namespace, kind } = sourceRef;
+  //can sourceRef actually return undefined stuff?! Typescript says it can.
```
AlinaGoaga, line 30: Yup just double checking in `objects.ts` and it looks like it's defined as ` sourceRef(): ObjectRef | undefined`

ui/components/VerifiedStatus.tsx
```diff
@@ -46,27 +52,33 @@ export const VerifiedStatus = ({
...
-      </div>
-    </Tooltip>
+  if (source) return <VerifiedStatus source={source as VerifiableSource} />;
+  const {
+    name = "",
+    namespace = "",
+    kind = "",
+    clusterName = "",
+  } = sourceRef || {};
+  const { data: verifiable } = useGetObject<VerifiableSource>(
```
AlinaGoaga, line 76: Maybe I'd leave this as  `{ data: source }` and just send it as `source={source}` like you did above also for consistency?

AlinaGoaga:
Hey Josh, this looks good to me! It's code moved from Backstage, should probably look into something similar there but we'll leave that for later 🤔 I think the `VerifiableResource` interface was initially built like that to be more extensible though you are correct atm we're looking at `Git Repositories | OCI repositories` only as having the potential to be verified so we can leave it in the new format.

AlinaGoaga (approved):

Issue #3626, opened 2023-04-14
The YAML tab displays a wrong version of CRD


chanwit:
### Problem

According to this: https://github.com/weaveworks/weave-gitops/blob/5cd78d74f975b68a28572032d143a622058a0117/core/server/primarykinds.go#L21-L34

The DefaultPrimaryKinds() function in the provided code snippet is designed to loop over all known types in the Kubernetes scheme and create a map of GroupVersionKinds (GVKs). However, there is a hidden bug in this function, which makes it unable to obtain the storage version from the scheme.

The issue arises because the function relies on the scheme.AllKnownTypes() map, which contains types registered in the program during compile time. If the program uses a different version of an object (e.g., v1beta2 version of GitRepository) than the environment it's running in (e.g., an environment with the v1 version of GitRepository), the program does not have the knowledge of the environment's version (in this example, GitRepository/v1). This is because the environment's version only exists in the actual cluster, not in the program's registered types.

As a result of this bug, the YAML tab might display incorrect versions of objects, leading to potential issues and inconsistencies when working with Kubernetes objects.

### Proposed solution

To address the bug, the first approach would be to update the program to have matching versions of objects registered, ensuring consistency between the program and the environment it's running in. By doing this, the program will have knowledge of the correct versions and can display accurate information in the YAML tab.

If upgrading the program to have matched versions registered does not resolve the issue effectively, a follow-up ticket will be created to further investigate the problem and find an alternative solution. This ensures that the bug is addressed and the correct versions of objects are displayed, improving the overall user experience and preventing potential issues when working with Kubernetes objects.

joshri:
SO -

My podinfo GitRepo source has matching apiVersion and annotation:
![image.png](https://images.zenhubusercontent.com/617c4a20ede496b2fae35baf/1a503979-b243-413d-be74-30a89c9e6f62)

But my `helloworld` GitRepo source (which is from the tf-controller getting started guide) does not:
![image.png](https://images.zenhubusercontent.com/617c4a20ede496b2fae35baf/5cc5e450-89ec-48e9-91bb-059991c869cf)

When I get the yaml in a terminal, it also doesn't match:
![image.png](https://images.zenhubusercontent.com/617c4a20ede496b2fae35baf/13d0d51d-6d57-4eb4-9039-c0e85fadd0b8)

Which makes me think this isn't something that's happening on the frontend. 

[...]

PR #3980, opened 2023-09-01
predictable resolution for gitrepository gvk when v1 and v1beta2 exist

enekofb:
Closes https://github.com/weaveworks/weave-gitops/issues/3626

**What changed?**

Adjusted weights to avoid the ranking function computes v1 and v1beta2 with the same value.

**Why was this change made?**

We noticed that sometimes gvk resolution for gitrepositories was returning `v1beta2` instead of `v1` which was expected. The reason was that both `v1` and `v1beta2` level had the same ranking value. 

Therefore v1 or v1beta2 was returned depending on which one was first introduced in primaryKinds. Given  `AllKnownTypes ` was randomly returning either of them, we got to that unpredictable behavior.

Adjusted weights to avoid the ranking function computes v1 and v1beta2 with the same value.

Added test case for supporting the specific case and for the gitRepository from primaryKinds.

Also tested e2e 

Screenshot 2023-09-01 at 12 00 11

### DIFF
--- a/core/server/primarykinds.go
+++ b/core/server/primarykinds.go
@@ -25,7 +25,7 @@ const (
 	// It's multiplied with the digit following the "beta" suffix.
 	// Beta versions are more stable than alpha versions but might still contain bugs.
 	// They are given a higher weight than alpha versions.
-	betaWeight = 500
+	betaWeight = 50
 
 	// stableWeight is a weight assigned to stable versions (those without an "alpha" or "beta" suffix).
 	// Stable versions are the final, production-ready releases of software, hence they have the highest weight.
--- a/core/server/primarykinds_test.go
+++ b/core/server/primarykinds_test.go
@@ -1,7 +1,6 @@
 package server
 
 import (
-	"fmt"
 	"reflect"
 	"testing"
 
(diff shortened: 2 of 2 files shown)


core/server/primarykinds.go
```diff
@@ -25,7 +25,7 @@ const (
 	// It's multiplied with the digit following the "beta" suffix.
 	// Beta versions are more stable than alpha versions but might still contain bugs.
 	// They are given a higher weight than alpha versions.
-	betaWeight = 500
+	betaWeight = 50
```
jpellizzari, line 28: Just for clarification: we determine which kinds to use at runtime?

jpellizzari (approved):
A bit confused on the logic in `primarykinds.go`, but that is not the concern of the PR, so :+1:

core/server/primarykinds.go
enekofb: yes they are resolved at startup here https://github.com/weaveworks/weave-gitops/blob/17e51daa99e6f5f6d1732b8b6e1064d8d67b8fef/core/server/primarykinds.go#L126

chanwit (approved):

core/server/primarykinds.go
jpellizzari: Seems like a bit of an anti-pattern...

PR #3985, opened 2023-09-04
fix Space in between buttons

### LINKED ISSUE #3891: Space in between buttons and weird black line

AsmaaNabilBakr:
Closes #3891 

**What changed?**
Set a fixed space between buttons to be `12px`

before:
image
after:
image

before:
image
after:
image

before:
image
after:
image

before:
image
after:
image


AsmaaNabilBakr:
@jpellizzari FYI, @TheGostKasper  suggest another solution for #3891 so I create this PR to try and it works by:
- using `Flex` component and `gap` property in `EE`.
- remove `Spacer` and update `gap` in some components in `OSS`.
instead of over right on styles in `OSS`. 

`EE` PR : https://github.com/weaveworks/weave-gitops-enterprise/pull/3279

TheGostKasper (approved):

PR #3992, opened 2023-09-07
UI works on subpaths

### LINKED ISSUE #2529: How to configure the app and Ingress with different base path
**Describe the bug**

When installing WEGO with the Helm chart, I want to expose the WebUI for local development and testing via an ingress under a different path than `/` (e.g.: `/wego`).

Currently, as far as I can see, there is no parameter or ENV var exposed to configure the application to make use of it. The only thing that is possible, is to configure the ingress resource via the Helm chart, which does not function when set to something different than `/`.

Also figured out, that there is actually a path parameter for the application that seems to be dead code. One can start the app with, for example, `--path=/wego` with no effect at all (see: [cmd/gitops-server/cmd/cmd.go#L84](https://github.com/weaveworks/weave-gitops/blob/v0.9.1/cmd/gitops-server/cmd/cmd.go#L84)).

Some other Helm charts correctly configure the application and ingress by allowing the user to change the path and route prefix. This is something I did in my local cluster with, for example, the [kube-prometheus-stack](https://github.com/prometheus-community/helm-charts/tree/main/charts/kube-prometheus-stack) and it is working as expected (apps are reachable via the ingress). This can be achieved by setting `prometheus.prometheusSpec.routePrefix` and `prometheus.ingress.paths`.

**Environment**

- Operating System: MacOS 12.4
- Weave-Gitops Helm chart version: 2.2.5
- Weave-Gitops app version: v0.9.1
- Flux Version: 0.31.5
- Kubernetes: v1.21.12 (kind, running local)
- Kong (for ingress): 2.8

**To Reproduce**

1. have a local Kind cluster running with Kong as ingress controller
1. get the Helm charts
    ```
    helm repo add weave https://helm.gitops.weave.works
    helm repo update weave
    ```
1. deploy WEGO with the helm chart
    ```
    helm upgrade weaveworks-gitops weave/weave-gitops \
      --install \
      -n flux-system \
      -f values.yaml  # see the values that were used below in "Additional Context"
    ```
1. verify the app is running
    ```
    kubectl get deploy -n flux-system -l='app.kubernetes.io/name=weave-gitops'
    ```
1. try to open the website in a browser or use curl
    ```
    curl -I http://localhost/wego
    ```
1. wait for the backend timeout (with `HTTP/1.1 504 Gateway Time-out`). 
1. now try it with port-forwarding and do the curl again (expected response `HTTP/1.1 200 OK`)
    ```
    kubectl port-forward service/weaveworks-gitops-weave-gitops -n flux-system 9001:9001
    curl -I http://localhost:9001/wego
    ```

**Expected behavior**

When ingress path is set to `/wego` the website should be reachable. 

**Actual Behavior**

The website is not reachable and the requests time out.

Values used to deploy WEGO via the Helm chart:

```
adminUser:
  create: true
  username: admin
  # htpasswd -bnBC 10 "" 'CHANGEME' | tr -d ':\n' | sed 's/$2y/$2a/'
  passwordHash: <use the result from the command above>
ingress:
  enabled: true
  className: kong
  hosts:
  - host: localhost
    paths:
    - path: /wego
      pathType: ImplementationSpecific
```

Alternative values used (with annotations for Kong to modify the requests):

```
adminUser:
  create: true
  username: admin
  # htpasswd -bnBC 10 "" 'CHANGEME' | tr -d ':\n' | sed 's/$2y/$2a/'
  passwordHash: <use the result from the command above>
ingress:
  enabled: true
  annotations:
    konghq.com/preserve-host: "true"
    konghq.com/strip-path: "true"
  className: kong
  hosts:
  - host: localhost
    paths:
    - path: /wego
      pathType: ImplementationSpecific
```

foot:
Fixes #2529

**What changed?**

- Enables support for running under weave-gitops under a subpath e.g. example.com/wego/ in the UI router

**Why was this change made?**

- Common to expose weave-gitops dashboard under some subpath 

UI routing now checks for and respects a `<base href="/foo/" />` tag if present.

If its not present everything works as it did before.

### Understand

- The UI uses react-router with the `<BrowserRouter />`.
- The UI handles navigation etc itself.
- The golang server file router serves up /index.html on just about any path except those with extensions (for css/js/images) and the api endpoints
- This means that when you land at example.com/wego/applications/detail, index.html will be served.
  - The UI code is responsible for choosing what to show at this point.
  - From this route, we need to know where we can fetch index.js and index.css from. The correct url respectively would be /wego/index.js and /wego/index.css
- Setting `<base href="/wego/" />` tells the UI where to fetch relative resources from and where the api request can be made from (/wego/v1/api etc)

### On the backend

How to set the `<base />` tag in the served index.html

- Add a cli flag to set it and add it
- `<base />` tag injected from cli
- Mount the server handler under a subpath and add a redirect at / to point to it. e.g. `/` 302 -> `/weave-gitops`

#### Test

- uncomment the `additionalArgs` in `tools/helm-dev-values.yaml`
- `tilt up`
- open http://localhost:9001
- note you are redirected to http://localhost:9001/wego/ and the UI should all work

#### Test to check ingress works too..

Spin up a kind cluster w/ ingress support:

- Run `./tools/kind-with-ingress.sh`
- Uncomment `ingress` and `additionalArgs` sections in `tools/helm-dev-values.yaml`
- `tilt up`
- open http://localhost:30080 note the 404 from ingress controller
- open http://localhost:30080/wego/ everything should be working okay

**Release notes**

**Documentation Changes**


pkg/server/route_prefix.go
```diff
@@ -0,0 +1,36 @@
...
+		routePrefix = "/" + routePrefix
+	}
+	// ensure route prefix doesn't have a trainling slash
+	routePrefix = strings.TrimSuffix(routePrefix, "/")
+
+	routePrefixMux := http.NewServeMux()
+	routePrefixMux.Handle(routePrefix+"/", http.StripPrefix(routePrefix, mux))
+	// Redirect to the route prefix if the user hits the root of the server
+	// e.g. if they port-forward
+	routePrefixMux.Handle("/", http.RedirectHandler(routePrefix+"/", http.StatusFound))
```
bigkevmcd, line 21: Nice!

foot:
@LappleApple I'll write up some docs for this. "How to run weave-gitops on a sub path of your ingress controller", thoughts about where it should go?

pkg/server/route_prefix.go
```diff
@@ -0,0 +1,36 @@
...
+
+	return routePrefixMux
+}
+
+func InjectHTMLBaseTag(html []byte, routePrefix string) []byte {
+	baseHref := routePrefix
+	// ensure baseHref begins and ends with a slash
+	if !strings.HasPrefix(baseHref, "/") {
+		baseHref = "/" + baseHref
+	}
```
bigkevmcd, line 31: Should we calculate this once, and inject on every request?

Do we need to inject if if the RoutePrefix is "" ?

bigkevmcd (changes requested):
We should document how to do this?

bigkevmcd:
Looks good, but we probably need some Helm and documentation!

pkg/server/route_prefix.go
foot: Yeah, could do.

I don't think is very expensive to do on each request, but makes it clearer its not gonna change on each request etc.

> Do we need to inject if if the RoutePrefix is "" ?

Nope, the UI is designed to work without the `<base />` tag. We only add it if you're changing the base path etc. Will call that out in an improved comment here.

tools/kind-with-ingress.sh
jpellizzari, line 1: 👍

jpellizzari (approved):
Nice work! Love the UI tests and extra helper script.

LappleApple:
>

LappleApple:
@foot Let's chat next week about docs-y things :)

bigkevmcd (approved):
Tested it locally, and it works, let's get the exported functions either not-exported, if they don't need to be, or documented.

Otherwise, looks good.

pkg/server/route_prefix.go
foot: > > Do we need to inject if if the RoutePrefix is "" ?
>
> Nope, the UI is designed to work without the <base /> tag. We only add it if you're changing the base path etc. Will call that out in an improved comment here.

I did not think this through and you are right, we do always need to inject it.

heubeck (approved):
:pray:

gre9ory:
@foot 
I can confirm it works in my setup too after building an image of the repo with the `route-prefix` support and setting up the Helm chart to use that dev image. 

Still misses a new proper image and an updated Helm chart to reference that image to be usable by end users if not mistaken?

Anyhow, thank you for this - in my view - important feature!

bigkevmcd:
@gre9ory it should be in the next release

PR #3994, opened 2023-09-07
Add eslint rule to alphabetize imports

joshri:
Related: https://github.com/weaveworks/weave-gitops-enterprise/issues/3215

Took a dip into the `.eslintrc` to prepare for work on unifying our configs between OSS and EE. This PR adds a rule that alphabetizes imports in ascending order. You can easily configure this as part of your formatting on save if you're finding yourself having to run `npm run lint -- --fix` too much by setting 
```
"editor.codeActionsOnSave": {
    "source.fixAll.eslint": true
  },
```
in your settings.json


jpellizzari:
@joshri How does this compare to the `Organize Imports` command in VSCode?

joshri:
> @joshri How does this compare to the `Organize Imports` command in VSCode?
> 
> Reference: https://dev.to/shanesc/how-to-sort-and-cleanup-imports-on-save-in-vs-code-52p1

The main thing is that this will force everyone to do it since it will be tied to CI

jpellizzari (approved):

foot:
Cool! Is there some cli variant that we could invoke via make/yarn as a fallback for non-vscode users?

joshri:
> Cool! Is there some cli variant that we could invoke via make/yarn as a fallback for non-vscode users?

By cli variant do you mean `npm run lint -- --fix`? This setting is always able to be auto-fixed with that --fix flag

PR #4002, opened 2023-09-11
Add Artifact Metadata section to OCI Repository detail pages

joshri:
Closes: https://github.com/weaveworks/weave-gitops-enterprise/issues/3341

Adds a new section to the Metadata component for OCI Repositories for clarity - we were previously displaying revision and source in the info list, but folks were expecting to see it in Metadata (which only recognizes our metadata.weaveworks prefix).

image

### DIFF
--- a/ui/components/Metadata.tsx
+++ b/ui/components/Metadata.tsx
@@ -4,12 +4,12 @@ import { formatMetadataKey, isHTTP } from "../lib/utils";
 import Flex from "./Flex";
 import InfoList from "./InfoList";
 import Link from "./Link";
-import Spacer from "./Spacer";
 import Text from "./Text";
 
 type Props = {
   className?: string;
   metadata?: [string, string][];
+  artifactMetadata?: [string, string][];
   labels?: [string, string][];
 };
 
@@ -21,9 +21,8 @@ const Label = styled(Text)`
   background-color: ${(props) => props.theme.colors.neutralGray};
 `;
 
-function Metadata({ metadata = [], labels = [], className }: Props) {
+const makeMetadata = (metadata: [string, string][]): [string, any][] => {
   const metadataCopy = [];
-
   for (let i = 0; i < metadata.length; i++) {
     metadataCopy[i] = metadata[i].slice();
   }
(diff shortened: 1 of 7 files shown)


ui/components/Metadata.tsx
```diff
@@ -41,40 +40,62 @@ function Metadata({ metadata = [], labels = [], className }: Props) {
...
-          </Text>
-          <InfoList items={metadataCopy} />
-          <Spacer padding="small" />
+    <Flex column gap="8">
+      <Text size="large" color="neutral30">
+        {title}
+      </Text>
+      {title === "Labels" ? (
+        <Flex wide start wrap gap="4">
+          {items.map((label, index) => {
```
jpellizzari, line 57: Is it possible for `items` to be `null` or `undefined` here? If so, this will explode.

ui/components/Metadata.tsx
```diff
@@ -41,40 +40,62 @@ function Metadata({ metadata = [], labels = [], className }: Props) {
         </Link>
       );
   });
+  return metadataCopy;
+};
 
+const MetadataSection: React.FC<{ title: string; items: [string, any][] }> = ({
```
jpellizzari, line 46: Is `any` type really valid here? If I pass in `{ foo: { bar: 'baz' }}`, will the component render OK?

jpellizzari (approved):
:+1: No blockers, just nits.

yiannistri:
@joshri fyi https://github.com/weaveworks/weave-gitops-enterprise/issues/3341#issuecomment-1715496153

Maybe we should just show all annotations as key value pairs, wdyt?

ui/components/Metadata.tsx
joshri: The labels prop gets set to `[]` if undefined, and then it gets checked for length before rendering the section at all

ui/components/Metadata.tsx
joshri: It was `any` bc we're turning some of the values into links - I can change it into `string | JSX`

jpellizzari: For the sake of accuracy, I like `string | JSX`. That way the next person has an idea of what they can pass in.

PR #4006, opened 2023-09-12
Add EE exports

joshri:
Related: https://github.com/weaveworks/weave-gitops-enterprise/issues/3215

This PR handles exports for every OSS import that used a path other than `@weaveworks/weave-gitops`. EE PR will pull in this version of OSS once merged

### DIFF
--- a/ui/index.ts
+++ b/ui/index.ts
@@ -140,9 +140,9 @@ import { baseTheme, muiTheme, theme } from "./lib/theme";
 import { showInterval } from "./lib/time";
 import {
   V2Routes,
+  PageRoute,
   RequestError,
   ReactQueryOptions,
-  PageRoute,
 } from "./lib/types";
 import {
   createYamlCommand,
--- a/ui/index.ts
+++ b/ui/index.ts
@@ -1,5 +1,6 @@
 import { ReconciledObjectsAutomation } from "./components/AutomationDetail";
 import AutomationsTable from "./components/AutomationsTable";
+import { Breadcrumb } from "./components/Breadcrumbs";
 import BucketDetail from "./components/BucketDetail";
 import Button from "./components/Button";
 import ChipGroup from "./components/ChipGroup";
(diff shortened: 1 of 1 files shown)


ui/index.ts
```diff
@@ -91,7 +93,27 @@ import useNavigation from "./hooks/navigation";
...
+  ListPoliciesRequest,
+  ListPoliciesResponse,
+  GetSessionLogsResponse,
+  ListError,
+  LogEntry,
+  PolicyValidation,
+  PolicyObj,
+  PolicyParam,
+  Core as coreClient,
+} from "./lib/api/core/core.pb";
```
foot, line 116: All types are exported in the weave-gitops npm package underneath `/ui/` (would be neat to remove this prefix somehow / someday),  e.g. you can import these from a consuming app with:

```ts
import { GetObjectRequest } from '@weaveworks/weave-gitops/ui/lib/api/core/core.pb';
```

Note this is only for types, not for classes/object/functions etc.

ui/index.ts
```diff
@@ -100,6 +122,7 @@ import {
   Automation,
   Bucket,
   FluxObject,
+  FluxObjectNode,
```
foot, line 125: This one looks like a class so have to export it here 👍

ui/index.ts
joshri: This is for the eslint rule import/no-unresolved 
https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/no-unresolved.md

At the bottom of that doc under "When to not use this", it says: "If you're using a module bundler other than Node or Webpack, you may end up with a lot of false positive reports of missing dependencies." Maybe that's what we have here?

jpellizzari (approved):

joshri:
@jpellizzari and I discussed turning off the `no-unresolved` rule in both repos so this PR won't be necessary - closing

PR #4007, opened 2023-09-13
Expose OSS static asset handling to unify OSS/EE serving

foot:
- Asset handling is growing more logic like injecting <base> tags
- We should have the same code in OSS/EE

**What changed?**

Expose OSS static asset handling as exports

**Why was this change made?**

- Asset handling is growing more logic like injecting <base> tags
- We should have the same code in OSS/EE

Extracting some internal handlers etc

- [x] unit tests
- [ ] manual testing

**Release notes**

**Documentation Changes**

### DIFF
--- a/cmd/gitops-server/cmd/cmd.go
+++ b/cmd/gitops-server/cmd/cmd.go
@@ -11,7 +11,6 @@ import (
 	"os"
 	"os/signal"
 	"path"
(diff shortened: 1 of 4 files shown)


AlinaGoaga (approved):
LGTM 🥳

Issue #3318, opened 2023-01-30
Add logging page to the user-guide to cover what is logged


foot:
**Short description**
- As a user I want to understand what is logged out by weave-gitops to make better decisions about how to store the logs. For example some users may have stricter GDPR conformance setups and need to understand what may be logged

**Acceptance criteria**
- [ ] Documentation that describes what is logged

At the moment we log things like:
- OIDC `claimUsername` which is `email` by default. We output OIDC emails to the logs in certain `error` and `debug` logs

PR #4015, opened 2023-09-15
Move Docs link from sidebar to top of page

joshri:
Closes: [#3318](https://github.com/weaveworks/weave-gitops-enterprise/issues/3318)

light:
image

dark:
image

Also fixes a small console error in tests created by changes in the InfoList component using string values for `key` instead of index


jpellizzari (approved):

TheGostKasper (approved):

mmoulian:
@joshri Next week, if I have time, I would like to create icons for User and Document aligned with the sidebar style. We now have full style icons in the header and outlined in the sidebar. I think it's strange.
The header icons are normal Google font icons and I don't like how the outlined style looks. I prefer to create our own.

PR #4022, opened 2023-09-19
Fix the OIDC login button when using a RoutePrefix.

### LINKED ISSUE #4016: Sub-Paths not working with OAuth2
**Describe the bug**

In #3992 sub-path support for the UI has been introduced.

If I configure wego with oauth2 authentication I get redirected (e.g. from /wego sub-path) to /oauth2 and not $subpath/oauth2.

**Environment**

- Weave-Gitops Version : v0.32.0
- Flux Version: /n/a
- Kubernetes version: n/a

**To Reproduce**

1. Set up wego with OAuth2.
2. Configure OAuth2 / OIDC provider. ( via oidc-auth secret from helm chart).
3. Set clientID/clientSecret/issuerURL.
4. Set callbackURL to $wegoBaseUrl  + "/oauth2/callback" ( e.g. https://example.com/wego/oauth2/callback )
5. Open the UI.
6. Click on "Login with OIDC Provider".

**Expected behavior**

After clicking the "Login with OIDC Provider" button there should be a redirect to the provider and after successful authentication the callback should provide wego the required information (thus forwarding after login to the "real" UI).

**Actual Behavior**

1. Nothing (HTTP/404).
2. After adding the path /oauth2 to the ingress I just get redirected back to the login screen.

I think, as #3992 covered the path-prefix for the UI only, OAuth2 should made aware of the prefix as well.

bigkevmcd:
Closes #4016

**What changed?**
Make the OIDC login button relative so that it uses the route prefix from the HTML base element.

**Why was this change made?**
To address a bug when logging in with OIDC **and** a Route Prefix.

Tests.

**Release notes**
Allow use of the route prefix mechanism with OIDC authentication.

**Documentation Changes**


ui/pages/SignIn.tsx
```diff
@@ -92,7 +92,7 @@ function SignIn({ darkModeEnabled = true }: Props) {
 
   const handleOIDCSubmit = () => {
     const CURRENT_URL = window.origin;
-    return (window.location.href = `/oauth2?return_url=${encodeURIComponent(
+    return (window.location.href = `./oauth2?return_url=${encodeURIComponent(
```
foot, line 95: This seems to work. Which makes sense, but I thought I read something to the contrary recently -- that while `a`, `img`, `script` etc tags respect `base` tags, the JS world (e.g. setting `location.href`) does not.

However after doing some more testing it seems that it does, and this works fine, even if you end up at localhost:9001/wego/signin/ (with a trailing slash), which still routes correctly, and then logs into OIDC correctly too..

foot: JSDOM etc at least don't support navigation so this is a trickier case to test w/out more involved acceptance tests..

ui/pages/__tests__/Signin.test.tsx
```diff
@@ -35,6 +48,15 @@ describe("SignIn", () => {
     expect(screen.queryByText(defaultButtonLabel)).toBeTruthy();
   });
 
+  it("should redirect to the oauth2 endpoint with the correct return_url", async () => {
```
foot, line 51:
```suggestion
  it("should redirect to the oauth2 endpoint with a relative URL to support running under a subpath", async () => {
```

foot (approved):
💯

PR #4028, opened 2023-09-19
Fix 'cluster not found' error when requesting source from automation `sourceRef`

### LINKED ISSUE #4019: Cluster not found error
I see a "Cluster not found" error when viewing workloads or sources locally

Screenshot 2023-09-15 at 20 15 50

and in staging

Screenshot 2023-09-18 at 14 22 34

As discussed with @joshri , it must be a new issue. At least I have not found an existing issue for this yet. @joshri has a hypothesis that
> maybe it's expecting `management` to be the name and you don't have a cluster with that name.

Locally, I created two Kind clusters names `wego-dev` (the active one) and `wge-dev`.

joshri:
Closes #4019 

The request for an automation's source was using the `sourceRef` field, which doesn't have a namespace if the source is in the same one as an automation, and NEVER has a cluster name. This PR uses the automation's namespace and cluster name to actually get the source back on these requests. 

Also fixes not actually opening a new tab on the new docs button - careful when using the `as` prop for styles with a styled component, because it actually changed the `a` element into a button as well

### DIFF
--- a/ui/components/AutomationsTable.tsx
+++ b/ui/components/AutomationsTable.tsx
@@ -129,6 +129,7 @@ function AutomationsTable({
       label: "Verified",
       value: (a: Automation) => (
         <SourceIsVerifiedStatus
+          automation={a}
           sourceRef={getSourceRefForAutomation(a) || {}}
         />
       ),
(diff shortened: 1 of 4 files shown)


yiannistri (approved):
🥇

Issue #4001, opened 2023-09-11
No padding on Suspend button in Source Detail


joshri:
<img width="292" alt="image" src="https://github.com/weaveworks/weave-gitops/assets/65822698/4b7b9fa1-1d67-4056-a0c2-47d0d9e07855">

I can only see it on detail pages for Helm Charts for some reason

opudrovs:
@joshri I don't see any issues with Suspend button padding in the latest OSS main or in OSS or enterprise (demo 3) staging:

<img width="1526" alt="Screenshot 2023-09-15 at 18 06 29" src="https://github.com/weaveworks/weave-gitops/assets/8824708/dae66635-0c4a-482c-af72-c7896446f512">

<img width="1504" alt="Screenshot 2023-09-15 at 18 06 36" src="https://github.com/weaveworks/weave-gitops/assets/8824708/03fff91f-f5e1-4810-a737-22e316a53253">

<img width="1530" alt="Screenshot 2023-09-15 at 18 06 55" src="https://github.com/weaveworks/weave-gitops/assets/8824708/0078ed47-67dd-4b1b-a591-595fe9612dbc">

Are you running OSS in Tilt or via `yarn start`?

Could you please re-check the issue?

opudrovs:
Tried it with `yarn start`, the padding looks fine to me:

<img width="1522" alt="Screenshot 2023-09-15 at 18 43 46" src="https://github.com/weaveworks/weave-gitops/assets/8824708/3dc38c83-e253-4488-82d7-73d543ce300e">

<img width="1527" alt="Screenshot 2023-09-15 at 18 43 53" src="https://github.com/weaveworks/weave-gitops/assets/8824708/cb07dca3-a23a-4678-898a-96b0d865b8c7">

joshri:
@opudrovs - have you checked it out on a Helm Chart? I could only see this on Helm Charts for some odd reason

opudrovs:
OK, thanks. Checked HelmChart detail pages locally with Tilt and in staging: I can see that the button looks a bit different indeed and for HelmCharts only. Its padding stays the same, but its width is reduced:  54+ for me instead of 61+. Crazy.

What is weird, I can only see this issue locally
<img width="1531" alt="Screenshot 2023-09-15 at 20 06 30" src="https://github.com/weaveworks/weave-gitops/assets/8824708/ab28bf68-bdbc-4476-aa61-ebd9d1bc83d8">

[...]

PR #4030, opened 2023-09-19
Fix Suspend button width on Source details pages with short status message

opudrovs:
Closes https://github.com/weaveworks/weave-gitops/issues/4001

- Added `min-width: fit-content` to the `SyncActions` component to prevent Suspend button being to narrow on Source details pages in the status message is too short.

Notes:
- The padding on the Suspend button stays the same, but the button's with is forced to be too narrow because of the `width: 50%` specified in the `SyncActions` component. As discussed with @joshri, this width setting was added to make  the buttons play nicely with Applied Revision and Last Updated elements on Application details pages.

Testing:

- The original issue can be seen in staging on a HelmChart page with a short status message:
https://gitops.euw1.wego-gke.weave.works/helm_chart/details?clusterName=Default&name=dex-dex&namespace=flux-system

At least on my laptop's screen:
Screenshot 2023-09-20 at 11 40 49

To reproduce the original issue in the main branch, just go to a Source details page and change the object status message to smth. short (HelmChart have shorter message than other sources thats why this issue was visible on HelmChart pages):

Screenshot 2023-09-20 at 11 30 49

In the current branch with the fix, the Suspend button should look as expected:

Screenshot 2023-09-20 at 11 20 15

You can also check that the fix does not change the layout of Application details page (the issue with Suspend button was not happening on the Application details pages, but I would not like to introduce any changes to the elements' layout with the fix) and Applied Revision and Last Updated elements look like before on wide and narrow screens:

Screenshot 2023-09-20 at 11 21 11

Screenshot 2023-09-20 at 11 21 24

### DIFF
--- a/ui/components/SyncActions.tsx
+++ b/ui/components/SyncActions.tsx
@@ -75,4 +75,5 @@ const SyncActions = ({
 
 export default styled(SyncActions)`
   width: 50%;
+  min-width: fit-content;
 `;
--- a/ui/components/__tests__/__snapshots__/SyncActions.test.tsx.snap
+++ b/ui/components/__tests__/__snapshots__/SyncActions.test.tsx.snap
@@ -70,6 +70,9 @@ exports[`SyncActions snapshots hideDropdown 1`] = `
 
 .c1 {
   width: 50%;
+  min-width: -webkit-fit-content;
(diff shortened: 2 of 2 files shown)


joshri (approved):
Nice!! LGTM

PR #4038, opened 2023-09-21
Add text wrap to fit long term ex. progressive delivery

ahussein3:
Closes #[3301](https://github.com/weaveworks/weave-gitops-enterprise/issues/3301)

**What changed?**
- Add text wrap to break the long nav item into two rows.
- Update the Delivery icon which is used by progressive delivery to be a flagger icon.

**Why was this change made?**
- Add text wrap in nav items so that long term like "Progressive Delivery" would fit into two lines as it's requested [here](https://github.com/weaveworks/weave-gitops-enterprise/issues/3301#issuecomment-1729864228).


joshri:
Looks good - but does this close the issue? It doesn't actually change the nav text or page title to Progressive Delivery. Totally fine to do it separately if that was the plan

ahussein3:
@joshri Yes that's only updating the icon but the text pr will be in EE repo.

foot:
It feels a tiny bit cramped now when selected.

Maybe we could try
- removing the margin-bottom (~4px now)
- giving that space to the height of the links instead (32px -> 36px)

ahussein3:
> It feels a tiny bit cramped now when selected.
> 
> Maybe we could try
> 
> * removing the margin-bottom (~4px now)
> * giving that space to the height of the links instead (32px -> 36px)

here it how it looking after giving it space without removing the margin-bottom as they will stack up if we did
image

PR #4041, opened 2023-09-22
Fixes the OIDC return_url param when a subpath is configured

foot:
- We should return the configured subpath NOT the root which might be another app etc.
- Also fixes the redirect_url handling for OIDC, so if you land on /sources and are redirected to signin, logging in with OIDC will also redirect you to /sources on success (and /wego/sources if a subpath it configured)

Follow up to:
- #3992 
- #4022

### To test

Spin up a kind cluster w/ ingress support:

- Run `./tools/kind-with-ingress.sh`
- Configure OIDC w/ the correct redirectURL
   ```
     --namespace default \
     --from-literal=issuerURL=<issuerURL> \
     --from-literal=clientID=<clientID> \
     --from-literal=clientSecret=<clientSecret> \
     --from-literal=redirectURL=http://localhost:30080/wego/oauth2/callback
    ```
- Uncomment `ingress` and `additionalArgs` sections in `tools/helm-dev-values.yaml`
- `tilt up`
- open http://localhost:30080 note the 404 from ingress controller
- open http://localhost:30080/wego/ login w/ OIDC, note we are now redirected back to /wego/ and not / (which gives a 404).


AlinaGoaga (approved):
Looks good 🥳 , tested and it behaves as expected :)

PR #4049, opened 2023-09-29
Fix JSX/SVG warnings in DeliveryIcon

jpellizzari:
Dashes are not valid characters in JSX props.

Screenshot from 2023-09-29 08-27-33


opudrovs (approved):
LGTM.

ahussein3 (approved):

PR #4055, opened 2023-10-02
Standardize the column names on Events tabs

TheGostKasper:
Closes [EE-3376](https://github.com/weaveworks/weave-gitops-enterprise/issues/3376)

Currently some of our Events tabs use wording that is different from other Events tabs. But there's just one schema for Event objects. To prevent user confusion, let's standardize so that every Events tab includes these four column names:

- Reason
- Message
- From 
- Last Updated

This change will affect :
- [x] Applications (Reason, Message, Component, Timestamp) -- adjust
- [x] Sources (Reason, Message, Component, Timestamp) -- adjust
- [x] We also wish to **apply a tooltip** to the "Reason" column name to make its meaning clearer. This tooltip should say: " "Reason" refers to what triggered the event. It can be different according to each component."

### DIFF
--- a/ui/components/EventsTable.tsx
+++ b/ui/components/EventsTable.tsx
@@ -1,12 +1,10 @@
-import { CircularProgress } from "@material-ui/core";
 import * as React from "react";
 import styled from "styled-components";
 import { useListEvents } from "../hooks/events";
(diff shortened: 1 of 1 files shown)


opudrovs (approved):
LGTM!

Issue #4057, opened 2023-10-02
Message "spec.Kubeconfig is set on this HelmRelease" shows on Kustomizations


makkes:
**Describe the bug**

Looking at the details of a Kustomization shows the message "spec.Kubeconfig is set on this HelmRelease. Details about reconciled objects are not available.".

This happens on Kustomizations that don't even have `.spec.kubeConfig` set.

Also the field name is `kubeConfig`, not `Kubeconfig`.

**Environment**

- Weave-Gitops Version: 0.33.0
- Flux Version: 2.1.1
- Kubernetes version: 1.28.1

**To Reproduce**

Look at any Kustomization.

**Expected behavior**

The whole message shouldn't be there.

**Additional Context (screenshots, logs, etc)**

![screenshot_c582e](https://github.com/weaveworks/weave-gitops/assets/387444/01f121cc-5354-4fe9-b3e1-36987f0a2c96)

yiannistri:
This message was introduced in https://github.com/weaveworks/weave-gitops/pull/4021 cc @foot

enekofb:
Closing as fixed by the attached PR. Should be available in the next release.

PR #4066, opened 2023-10-04
Fix hr related message appearing on ks

AlinaGoaga:
Closes [#4057](https://app.zenhub.com/workspaces/team-pesto-6127680175876a0014fc5657/issues/gh/weaveworks/weave-gitops/4057)

### DIFF
--- a/ui/components/AutomationDetail.tsx
+++ b/ui/components/AutomationDetail.tsx
@@ -83,6 +83,10 @@ function AutomationDetail({
 
   const canaryStatus = createCanaryCondition(data?.objects);
   const health = computeAggHealthCheck(data?.objects || []);
+  // We cannot show reconciled objects for remote HelmReleases
+  const skipReconciledObjectsTable =
+    automation.type === "HelmRelease" &&
+    (automation as HelmRelease).kubeConfig !== "";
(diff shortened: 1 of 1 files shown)


ui/components/AutomationDetail.tsx
```diff
@@ -91,8 +91,9 @@ function AutomationDetail({
       component: () => {
         return (
           <RequestStateHandler loading={isLoading} error={error}>
-            {automation.type === "HelmRelease" &&
-            (automation as HelmRelease).kubeConfig === "" ? (
+            {automation.type === "Kustomization" ||
+            (automation.type === "HelmRelease" &&
+              (automation as HelmRelease).kubeConfig === "") ? (
```
foot, line 96: Maybe we flip the logic a bit so that we don't have to include other valid automation types in the conditional here (now only ks, but in the future maybe others)..

```ts
// We cannot show reconciled objects for remote HelmReleases
const skipReconciledObjectsTable = automation.type === "HelmRelease" && 
  (automation as HelmRelease).kubeConfig !== "";
```

Kind of thing?

foot (approved):
💯 !

PR #4069, opened 2023-10-06
Move header creation outside yaml view to allow for custom text

joshri:
We would like to use the DetailModal component for the PR Preview in the upcoming add source UI in EE - part of this means replacing the `kubectl get` command from the YamlView header with the path of your potential resource, so I moved the creation of the header up a level in the component tree, and added `path` as an option in the the createYamlHeader function.


opudrovs:
LGTM, but some UI tests are failing.

joshri:
yep - missed changing the prop in the test

opudrovs (approved):

PR #4070, opened 2023-10-09
Fixes staging-docs

foot:
- var interpolation wasn't working with single quotes
- also introduce env var into docusaurus so you can optionally disable versioning (`DISABLE_VERSIONING=true yarn build` to dramatically speed up builds when testing things..)

**What changed?**

Some of the github action code, switch from single `''` to double `""` quoted strings so we can do the var interpolation

Switch to env vars which relied on the variable expansion made in https://github.com/weaveworks/weave-gitops/pull/4046

**Why was this change made?**

So staging docs works

- Use heredoc for JSON payloads.. seems to work okay and avoids having to escape all the double quoted key names in JSON..

Checked that the staging docs now works

### DIFF
--- a/.github/workflows/docs.yaml
+++ b/.github/workflows/docs.yaml
@@ -45,9 +45,9 @@ jobs:
           npm i
           fi
           sed -i 's#url:.*$#url: "https://staging.docs.gitops.weave.works",#' docusaurus.config.js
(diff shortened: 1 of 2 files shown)


yiannistri (approved):
Well spotted 👍

PR #4072, opened 2023-10-09
Only build last 3 versions of user-guide for staging

foot:
- To speed it up a bit
- Also remove sed'ing of docusaurus.config.js, read env vars instead

**What changed?**

- Only build last 3 versions of user-guide for staging
- Hopefully 3 is good enough for previewing small tweaks to recent docs
- If need to preview more can tweak the number in the docu.config temporarily.

**Why was this change made?**

Staging docs build is very slow

Tweaks in the docusaurus config

Pushing to CI

image

3.5mins vs 13.5mins

Staging urls still work: https://staging.docs.gitops.weave.works/faster-docs-staging-build/

You can test locally with `STAGING_BUILD=true GA_KEY=fakekey yarn start` and see that only the last few versions appear:

image

**Documentation Changes**


bigkevmcd:
Could we also reduce the number we build for our main site?

Do we actually support all those versions?

Himangini (approved):
Cool ✨

foot:
Will merge for now, but understand versioning might change again soon!

PR #4074, opened 2023-10-11
Remove webkit-scrollbar css

TheGostKasper:
Closes [ee-2993](https://github.com/weaveworks/weave-gitops-enterprise/issues/2993)

**::-webkit-scrollbar**

Non-standard: This feature is non-standard and is not on a standards track. Do not use it on production sites facing the Web: it will not work for every user. There may also be large incompatibilities between implementations and the behavior may change in the future.


joshri (approved):
Looks good locally on Chrome

PR #4083, opened 2023-10-16
fix: Only run npm publish job on weave-gitops repo

yiannistri:
**What changed?**
There's a step in our CI process that pushes an npm package to the GitHub npm registry. This step fails when executed in forks of the repo. This change attempts to only run this step for PRs within the weave-gitops repo.

**Why was this change made?**
To prevent CI from failing when running in forks.


jpellizzari (approved):
Very nice!

PR #4084, opened 2023-10-16
Add colors and exports for new EE pipelines UI

joshri:
Adds new colors to the theme, and exports where needed, and makes one important modification:

Previously, we had both `black` and `neutral40` switching between `1a1a1a` and `fff` in light and dark mode respectively...so they were doing the exact same thing. In the new pipelines UI, we have a case where black needs to stay black in both light and dark mode, so I've made it so `black` and `white` in our theme will stay their respective colors no matter the theme, while the `neutral` variants will switch. This is nice regardless as it gives some extra consistency - so you'll see every use of `white` and `black` we currently have will be changed to their `neutral` variants in this PR.


ui/lib/theme.ts
```diff
@@ -117,6 +119,7 @@ export const theme = (mode: ThemeTypes = ThemeTypes.Light): DefaultTheme => {
         alertOriginal: "#BC3B1D",
         alertDark: "#9F3119",
         neutralGray: "#F6F7F9",
+        pipelineGray: "dde1e9",
```
opudrovs, line 122: The hashtag symbol is missing.

ui/lib/theme.ts
```diff
@@ -126,6 +129,7 @@ export const theme = (mode: ThemeTypes = ThemeTypes.Light): DefaultTheme => {
         grayToPrimary: "#737373",
         neutral20ToPrimary: "#d8d8d8",
         backGray: "#eef0f4",
+        pipelinesBackGray: "eef0f4",
```
opudrovs, line 132: Same here, the hashtag symbol is missing.

TheGostKasper (approved):

opudrovs (approved):

PR #4088, opened 2023-10-17
update explorer docs with granular configuration

jpellizzari:
Related to https://github.com/weaveworks/weave-gitops-enterprise/pull/3485

### DIFF
--- a/website/docs/explorer/getting-started.mdx
+++ b/website/docs/explorer/getting-started.mdx
@@ -21,7 +21,7 @@ Before using Explorer, please ensure that:
 
 ## Setup
 
-Explorer is enabled via configuration through the feature flag  `enableExplorer` that you could
+Explorer is enabled via configuration through the feature flag  `explorer.enabled` that you could
 configure in your Weave Gitops Enterprise HelmRelease values:
 
 
@@ -36,7 +36,7 @@ spec:
   # ... other spec components
   values:
   explorer:
-    enabled: true
+    enabled: true # global enable/disable flag
     collector:
       # ServiceAccount that explorer will use to watch clusters for resources
       serviceAccount:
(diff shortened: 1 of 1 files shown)


opudrovs (approved):

website/docs/explorer/getting-started.mdx
```diff
@@ -35,7 +35,20 @@ metadata:
 spec:
   # ... other spec components
   values:
-    enableExplorer: true
+  explorer:
```
enekofb, line 38: given its a getting started page, a minimal configuration would be better 

```
explorer:
  enabled: true # global enable/disable flag
```
and then leave other cases for the user to go onto the configuration page 

https://docs.gitops.weave.works/docs/explorer/configuration/

where to give a little more guidance and the configuration options that we provide 

https://docs.gitops.weave.works/docs/explorer/configuration/#setup

website/docs/explorer/getting-started.mdx
```diff
@@ -35,7 +35,20 @@ metadata:
...
+      serviceAccount:
+        name: "collector"
+        namespace: "flux-system"
+    cleaner:
+      disabled: false
+    enabledFor:
+      - applications
+      - sources
+      - gitopssets
+      - templates
```
enekofb, line 51: it feels that we should expand with comments to provide guidance to the user on how use these flags. for example: the user could think that is configuration for deciding which type of data to collect and not for the UI only

enekofb (commented):
Added comment with suggestion on how to split the configuration info.

enekofb (approved):
approving and merging to get it in today's wego release

followup around explorer docs and getting started vs configuration https://docs.gitops.weave.works/docs/explorer/configuration/

PR #4091, opened 2023-10-18
Cleanup datatable

TheGostKasper:
Closes [ee-3427](https://github.com/weaveworks/weave-gitops-enterprise/issues/3427)

**What changed?**

To summarize, `DataTable` should not have any state itself, but instead force the parent to maintain the state. so that i choose to not change DataTable itself as it's used allover the system so `TableVIew` is the clone `DataTable` which display **[header/body/emptyRow]** ...etc


ui/components/DataTable/TableView/modal.ts
jpellizzari, line 1: This is called `modal.ts` but seems to have a lot of different types. Is that file name intentional?

ui/components/DataTable/TableView/TableView.tsx
```diff
@@ -0,0 +1,57 @@
...
+      checkedFields = rows.map((r) => r.uid);
+    } else {
+      checkedFields = [];
+    }
+    onBatchCheck(checkedFields);
+  };
+  return (
+    <TableContainer id={id}>
+      <Table aria-label="simple table">
+        <TableHeader
```
jpellizzari, line 37: Really nice work on making this nice an readable :+1:

ui/components/DataTable/TableView/SearchedNamespacesModal.tsx
```diff
@@ -0,0 +1,33 @@
...
+import { SearchedNamespaces } from "../../../lib/types";
+import { IconButton } from "../../Button";
+import Icon, { IconType } from "../../Icon";
+import InfoModal from "../../InfoModal";
+
+interface Props {
+  searchedNamespaces: SearchedNamespaces;
+}
+
+const SearchedNamespacesModal = ({ searchedNamespaces }: Props) => {
```
jpellizzari, line 11: Not understanding where this is used. Where in the UI does this modal appear?

TheGostKasper: Used in `AutomationsTable` and `SourcesTable`

TheGostKasper:
Tests are covered in both `DataTable` and `DataTableFilters` :ok_hand:

jpellizzari (approved):

jpellizzari (approved):
Really nice work :+1:

ui/components/DataTable/TableView/TableView.tsx
```diff
@@ -0,0 +1,57 @@
...
+    if (checked) {
+      checkedFields = rows.map((r) => r.uid);
+    } else {
+      checkedFields = [];
+    }
+    onBatchCheck(checkedFields);
+  };
+  return (
+    <TableContainer id={id}>
+      <Table aria-label="simple table">
```
opudrovs, line 36: ❔  Is this `aria-label` informative to the user? It's a "simple table" for developers (comparing to the old all in one table), but for users it's just a table. And do we need an aria-label here at all?

[MDN](From https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-label):
> Note: While aria-label is allowed on any element that can have an accessible name, in practice, aria-label is only supported on interactive elements, widgets, landmarks, images and iframes.

Is the whole table considered an interactive element?

Maybe we could add some roles instead or in addition to `aria-label`?
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/table_role

ui/components/DataTable/TableView/TableBody.tsx
```diff
@@ -0,0 +1,87 @@
...
+    );
+  }
+
+  return (
+    <TableBody>
+      {rows?.map((r, i) => {
+        return (
+          <TableRow key={r.uid || i}>
+            {hasCheckboxes && (
+              <TableCell style={{ padding: "0px" }}>
```
opudrovs, line 51: ❔ Is "px" redundant here? We usually just use `padding: 0`, AFAIR.

ui/components/DataTable/TableView/SortableLabelView.tsx
```diff
@@ -0,0 +1,52 @@
...
+        </h2>
+      </TableButton>
+      {sortedField?.label === field.label ? (
+        <Icon
+          type={IconType.ArrowUpwardIcon}
+          size="base"
+          className={sortedField.reverseSort ? "upward" : "downward"}
+        />
+      ) : (
+        <div style={{ width: "16px" }} />
```
opudrovs, line 46: ❌  Can we use the standard `base` standard spacing constant from the theme here instead of the hardcoded "16px" value? We are already using `size="base"` for the icon.

ui/components/DataTable/TableView/TableView.tsx
```diff
@@ -0,0 +1,57 @@
...
+  fields,
+  id,
+  checkedFields,
+  defaultSortedField,
+  onSortChange,
+  hasCheckboxes,
+  onBatchCheck,
+}: TableViewProps) => {
+  const onCheckChange = (checked: boolean, id: string) => {
+    const chk = [...checkedFields];
```
opudrovs, line 18: 💡 I think we could use a more informative variable name here (preferably in plural) — to reflect that this variable holds an array of strings.

opudrovs (approved):
Tested it locally, works fine for me. Great job! 🎉

ui/components/DataTable/TableView/TableView.tsx
TheGostKasper: I'll remove the `aria-label`

ui/components/DataTable/TableView/SortableLabelView.tsx
TheGostKasper: I assume `hardcoded "16px"` is fine here as it doesn't worth the change of creating a styled div with `base` or to import the them to the file to use only the value

ui/components/DataTable/TableView/TableView.tsx
TheGostKasper: i'll rename it to `selectedFields` :ok:

PR #4098, opened 2023-10-24
Refactoring Status column 

luizbafilho:
Closes weaveworks/weave-gitops-enterprise#3529 

**What changed?**

Refactor to use HealthChecks instead to reduce confusion and properly represent the status of the object.

**Why was this change made?**

**Release notes**

**Documentation Changes**


joshri (approved):
Code works great - we are potentially losing some functionality where status for sources in the table won't be shown since they don't have a `health` attribute though. Probably not worth keeping the whole status column for this case imo just calling it out @yiannistri 
image

yiannistri:
@yiannistri to create follow-up issues for source objects with missing status.

Issue #4102, opened 2023-10-25
docsearch does not work for docs, does for static pages


enekofb:
**Describe the bug**

for static pages like security we find

<img width="1398" alt="Screenshot 2023-10-25 at 14 29 45" src="https://github.com/weaveworks/weave-gitops/assets/12957664/b272893b-5aa0-481f-8338-3ffda8ee141c">

we have no results for docs

<img width="1610" alt="Screenshot 2023-10-25 at 14 28 08" src="https://github.com/weaveworks/weave-gitops/assets/12957664/d2f02662-517c-43f9-9ae7-92416816cec0">

**Expected behavior**

i am able to search in both static and docs

**Additional Context (screenshots, logs, etc)**

enekofb:
looking in algolia https://crawler.algolia.com/admin/crawlers we could see that is marked as `blocked` 

<img width="543" alt="Screenshot 2023-10-25 at 14 33 42" src="https://github.com/weaveworks/weave-gitops/assets/12957664/aae32ea1-c5a9-47b3-bd58-921982b46765">

enekofb:
<img width="802" alt="Screenshot 2023-10-25 at 14 35 22" src="https://github.com/weaveworks/weave-gitops/assets/12957664/a7952528-0e43-4283-8bb1-05e6f5be067e">

enekofb:
https://github.com/typescript-eslint/typescript-eslint/issues/6856

enekofb:
given 

https://github.com/weaveworks/weave-gitops/issues/4102#issuecomment-1779292566

not sure whether replacing the production index is a solution given

enekofb:
sitemap in prod 

https://console.cloud.google.com/storage/browser/_details/production.docs.gitops.weave.works/sitemap.xml;tab=live_object?project=docs-gitops-ww-prod

is 

Created
Jul 12, 2023, 5:14:25 PM
Last modified
Jul 12, 2023, 5:14:25 PM

enekofb:
https://stackoverflow.com/questions/72838375/algolia-docsearch-in-docusaurus-doesnt-bring-any-results

enekofb:
https://github.com/weaveworks/weave-gitops/pull/3689/files

enekofb:
after pr merged, i could see sitemap now is generated with updated content 

[...]

PR #4103, opened 2023-10-25
generate sitemap

enekofb:
Part https://github.com/weaveworks/weave-gitops/issues/4102

**What changed?**
- removed metadata config to generate sitemap
- algolia key as env

**Why was this change made?**

suspect that the indexing issue in algolia would go away with this

```bash
➜  website git:(main) ✗ cat ./build/sitemap.xml|grep run-ui-subpath                                                                                                           <aws:sts>
<?xml version="1.0" encoding="UTF-8"?><urlset xmlns="http://www.sitemaps.org/schemas/sitemap/0.9" xmlns:news="http://www.google.com/schemas/sitemap-news/0.9" xmlns:xhtml="http://www.w3.
```

**Release notes**

**Documentation Changes**

### DIFF
--- a/website/docusaurus.config.js
+++ b/website/docusaurus.config.js
@@ -62,7 +62,6 @@ module.exports = {
     ],
   ],
   themeConfig: {
-    metadata: [{ name: "robots", content: "noindex, nofollow" }],
     navbar: {
       title: "Weave GitOps",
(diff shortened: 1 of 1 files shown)


foot (approved):
lgtm :zap

PR #4111, opened 2023-10-31
fix: Remove GitOps Run CLI commands

yiannistri:
Part of #weaveworks/weave-gitops-enterprise#3514

**What changed?**
Updated the gitops CLI to remove GitOps Run commands.

**Why was this change made?**
Because we are deprecating this feature.

By deleting the Go files that included GitOps Run-related commands.

```
make
./bin/gitops --help 
```

**Release notes**
TBD

**Documentation Changes**
We have updated docs already.


cmd/gitops/root/cmd.go
```diff
@@ -7,25 +7,20 @@ import (
...
 	cfg "github.com/weaveworks/weave-gitops/cmd/gitops/config"
 	"github.com/weaveworks/weave-gitops/cmd/gitops/create"
 	deletepkg "github.com/weaveworks/weave-gitops/cmd/gitops/delete"
 	"github.com/weaveworks/weave-gitops/cmd/gitops/docs"
 	"github.com/weaveworks/weave-gitops/cmd/gitops/get"
+	"github.com/weaveworks/weave-gitops/cmd/gitops/logs"
+	"github.com/weaveworks/weave-gitops/cmd/gitops/replan"
+	"github.com/weaveworks/weave-gitops/cmd/gitops/resume"
 	"github.com/weaveworks/weave-gitops/cmd/gitops/set"
+	"github.com/weaveworks/weave-gitops/cmd/gitops/suspend"
```
yitsushi, line 23: Do we want to keep these still? As I see it removes only `beta`, `remove`, and `run`.

If we want to keep replan, resume, suspend, etc, do we want to delete `remove`?

yiannistri: The reason for deleting `remove` is because the only command there was used to remove GitOps Run sessions.  `replan`, `resume` and `suspend` commands are interacting with Terraform resources which is why I kept them.

yitsushi (approved):

yitsushi (approved):

PR #4114, opened 2023-11-01
Add new svg icon as CLusterDiscovery icon

ahussein3:
Add New Icon for new navigation item ( Cluster Discovery Icon)


foot (approved):
Replaced Sources w/ ClusterDiscovery and it seemed to render okay 👍

jpellizzari (approved):
Approving, since we have other Icons like this, but we don't get the benefit of browser caching when we inline SVG into our JS bundles.

If this were an `` tag, the browser would cache the image and our bundle size would be reduced.

PR #4118, opened 2023-11-02
added adr with user docs navigation proposal

enekofb:
just draft to have further discussions


doc/adr/0016-release-ordering.md
```diff
@@ -2,7 +2,7 @@
 
 ## Status
 
-Proposed
+Accepted
```
enekofb, line 5: this is just taking the opportunity to update this ADR status

PR #4125, opened 2023-11-08
ci: Simplify docs creation

yiannistri:
**What changed?**
Simplified CI workflow that deploys user guide docs.

**Why was this change made?**
Some of the commands were not used/necessary.


foot (approved):
Nice one 👌

PR #4131, opened 2023-11-13
Add suspended annotations

ranatrk:
Closes BE part of https://github.com/weaveworks/weave-gitops-enterprise/issues/3546

**What changed?**
Add suspended-by and suspended-comment annotations to objects being suspended
Remove suspended-by and suspended-comment annotations from objects being resumed 

**Why was this change made?**
To log who/what performed the suspending action on the resource

Adding the annotations needed indicating `suspended-by` and `suspended-comment` that could possibly include extra details regarding the suspension 

- unit tests
- manually check annotations returned in objects suspended

**Release notes**

**Documentation Changes**


foot:
Lets add this little snipped to the makefile and run it to fix the ci lint job

https://github.com/weaveworks/weave-gitops/pull/4005/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R145-R148

core/server/suspend_test.go
```diff
@@ -137,58 +139,86 @@ func TestSuspend_Suspend(t *testing.T) {
...
+func checkSuspendAnnotations(t *testing.T, principalID string, annotations map[string]string, name types.NamespacedName, suspend bool) {
+	if suspend {
+		// suspended and annotations exist check
+		if suspendedBy, ok := annotations["weave.works/suspended-by"]; ok {
+			// principal check if suspended and annotations exist
+			if suspendedBy != principalID {
+				t.Errorf("expected annotation weave.works/suspended-by to be set to the principal %s", principalID)
+			}
+		} else {
+			t.Errorf("expected annotation weave.works/suspended-by not found for %s", name.String())
```
bigkevmcd, line 210: You don't need to call `.String()` because...
https://pkg.go.dev/fmt

> 5. If an operand implements method String() string, that method will be invoked to convert the object to a string, which will then be formatted as required by the verb (if any).

See https://go.dev/play/p/G1RCTkzZDoD

core/server/suspend_test.go
```diff
@@ -105,14 +106,15 @@ func TestSuspend_Suspend(t *testing.T) {
...
+		principalID := "anne"
+		md := metadata.Pairs(MetadataUserKey, principalID, MetadataGroupsKey, "system:masters")
 		outgoingCtx := metadata.NewOutgoingContext(ctx, md)
 		_, err = c.ToggleSuspendResource(outgoingCtx, req)
 		g.Expect(err).NotTo(HaveOccurred())
 
 		for _, tt := range tests {
 			name := types.NamespacedName{Name: tt.obj.GetName(), Namespace: ns.Name}
-			g.Expect(checkSpec(t, k, name, tt.obj)).To(BeFalse())
+			g.Expect(checkSpec(t, k, principalID, name, tt.obj, false)).To(BeFalse())
```
bigkevmcd, line 117: I'd probably have split this into to two expectations rather than piggy-backing on top of `checkSpec` ?

core/server/suspend.go
```diff
@@ -60,6 +60,8 @@ func (cs *coreServer) ToggleSuspendResource(ctx context.Context, msg *pb.ToggleS
 			return nil, err
 		}
 
+		changeSuspendAnnotations(obj, msg.Suspend, msg.Comment, principal)
+
 		if msg.Suspend {
```
bigkevmcd, line 65: A couple of things here...

https://github.com/weaveworks/weave-gitops/blob/35231b877e1af6f53fc5c04b9bb516f87ef8e759/core/server/suspend.go#L15-L19

We could reuse the principal from the first call in the second?

Also, we could include the "principalID" in the `log.Info` so that the user who does it is logged out?

We could also include the name and namespace, and maybe even Kind of the resource, so that the logs would be searchable.

"Who unlocked the GitRepository test/test"?

Bonus points for including the Cluster name :-)

bigkevmcd (changes requested):

core/server/suspend_test.go
bigkevmcd: I don't think you need to add a switch?

You've got a `client.Object` https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/client#Object  which embeds https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1#Object ?

core/server/suspend.go
ranatrk: the name,namespace, and kind are already included . I'll add the principal to the logs :+1: 
https://github.com/weaveworks/weave-gitops/blob/35231b877e1af6f53fc5c04b9bb516f87ef8e759/core/server/suspend.go#L44-L49

bigkevmcd (approved):
Looks good to me, 👍🏻

PR #4144, opened 2023-11-22
Add note about lack of persistent storage.

bigkevmcd:
Closes

**What changed?**
This just documents our current lack of persistent storage for sessions.

**Why was this change made?**
We are trying to figure out what we're going to support, but we don't currently support it, this just makes that clear in the documentation.

**Release notes**

**Documentation Changes**


sarataha (approved):
LGTM!

PR #4150, opened 2023-11-29
add weave gitops oss release notes links for v0.37 

enekofb:
to help out ee customers to have the info more easily accessible temp for this version. we should find a long-term solution to the problem.


MostafaMegahid (approved):

paulcarlton-ww (approved):
This adds a convenient link, we should consider reworking this in the future to distinguish between a dependancy you are required to satisfy and other software that is included in this release

PR #4158, opened 2023-12-06
Fix `gitops check`

### LINKED ISSUE #4157: gitops check failing after kubectl --short flag deprecation
**Describe the bug**

Running `gitops check` return exit status 1 instead of skipping/handling the error. 

**Environment**

- Weave-Gitops Version

```zsh
gitops version
Current Version: 0.37.0
GitCommit: 3f850ced58c9fe9d593807e299058e0e06c735eb
BuildTime: 2023-11-22T13:24:19Z
Branch: releases/v0.37.0
```

- Flux Version

```zsh
 flux -v
flux version 2.1.2
```

- Kubernetes version

```
kubectl version
Client Version: v1.28.4
Kustomize Version: v5.0.4-0.20230601165947-6ce0bf390ce3
Server Version: v1.28.3+k3s2
```

**To Reproduce**

1. Having a running flux and gitops installation

```zsh
kubectl get po -n flux-system
NAME                                           READY   STATUS    RESTARTS   AGE
helm-controller-5f964c6579-kz2vg               1/1     Running   0          2d16h
kustomize-controller-9c588946c-nhrvf           1/1     Running   0          2d16h
notification-controller-76dc5d768-l8hqh        1/1     Running   0          2d16h
source-controller-6c49485888-jz5t6             1/1     Running   0          2d16h
image-reflector-controller-84c449dc57-vpqqj    1/1     Running   0          2d12h
image-automation-controller-7764f8957c-l7xll   1/1     Running   0          2d12h
ww-gitops-weave-gitops-5fcfdb67b4-6x2fg        1/1     Running   0          22h
```

2. Run `gitops check`

```zsh
gitops check
Error: unable to get kubernetes version: exit status 1
```

**Expected behavior**

To run without errors, maybe checking kubectl version compatibility?

**Actual Behavior**

```zsh
gitops check
Error: unable to get kubernetes version: exit status 1
```

It seems this is related to kubectl `--short` flag used in this project, see [here](https://github.com/weaveworks/weave-gitops/blob/76ffb12731fb0b9554b16d72ab429dda1ea7f1b4/pkg/services/check/check.go#L26).

As it says in [kubectl's CHANGELOG.md](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.28.md?plain=1#L1718), `--short` is deprecated and changed `kubectl version` default output to be identical to what `kubectl version --short` printed

makkes:
closes #4157

**What changed?**

1. The `gitops check` command works again.
2. The oldest Kubernetes version yielding a positive result is bumped to 1.26.0.

**Why was this change made?**

1. The `--short` flag has been removed from `kubectl version` in 1.28 (https://github.com/kubernetes/kubernetes/blob/7fe31be11fbe9b44af262d5f5cffb1e73648aa96/CHANGELOG/CHANGELOG-1.28.md#deprecation) so the command obviously fails now.
2. We state in [our docs](https://docs.gitops.weave.works/docs/open-source/getting-started/install-OSS/#check-your-clusters-kubernetes-version) that we only support K8s releases that are supported upstream.

This commit changes the behaviour of the `gitops check` command to create a client-go DiscoveryClient and use that to retrieve the server version. That way we don't have to rely on forking a `kubectl` process and the output being consistent.

The code is now much cleaner, easier to read and properly tested.

Unit tests and manual runs.

**Release notes**

- Fix `gitops check` command with most recent version of `kubectl`.
- `gitops check` will only yield a positive result if the Kubernetes version is at least 1.26.0.

**Documentation Changes**

n/a

### DIFF
--- a/pkg/services/check/check.go
+++ b/pkg/services/check/check.go
@@ -8,7 +8,7 @@ import (
 )
 
 const (
-	kubernetesConstraints = ">=1.20.6-0"
+	kubernetesConstraints = ">=1.26"
 )
 
 // KubernetesVersion checks if the Kubernetes version of the client is recent enough and
@@ -30,5 +30,5 @@ func KubernetesVersion(c discovery.DiscoveryInterface) (string, error) {
 		return "", fmt.Errorf("✗ kubernetes version %s does not match %s", sv.Original(), kubernetesConstraints)
 	}
 
-	return fmt.Sprintf("✔ Kubernetes %s %s", sv.String(), kubernetesConstraints), nil
+	return fmt.Sprintf("✔ Kubernetes %s %s", sv, kubernetesConstraints), nil
 }
--- a/pkg/services/check/check_test.go
+++ b/pkg/services/check/check_test.go
@@ -39,12 +39,12 @@ func TestKubernetesVersion(t *testing.T) {
 		{
 			name:          "server version satisfies constraint",
 			serverVersion: "v1.28.4",
-			expectedRes:   `✔ Kubernetes 1.28.4 >=1.20.6-0`,
+			expectedRes:   `^✔ Kubernetes 1.28.4 >=1.`,
 		},
 		{
 			name:          "server version too low",
 			serverVersion: "v1.20.5",
-			expectedErr:   `✗ kubernetes version v1\.20\.5 does not match >=1\.20\.6-0`,
+			expectedErr:   `✗ kubernetes version v1\.20\.5 does not match >=1\.`,
 		},
 		{
 			name:          "server version not semver compliant",
@@ -72,7 +72,7 @@ func TestKubernetesVersion(t *testing.T) {
 				g.Expect(err).To(MatchError(MatchRegexp(tt.expectedErr)))
 			}
 
-			g.Expect(res).To(Equal(tt.expectedRes))
+			g.Expect(res).To(MatchRegexp(tt.expectedRes))
 		})
 	}
 }


pkg/services/check/check.go
```diff
@@ -2,71 +2,33 @@ package check
...
 
-	if _, err := semver.StrictNewVersion(version); err != nil {
-		return nil, err
+	cons, _ := semver.NewConstraint(kubernetesConstraints)
+	if !cons.Check(sv) {
+		return "", fmt.Errorf("✗ kubernetes version %s does not match %s", sv.Original(), kubernetesConstraints)
 	}
 
-	return semver.NewVersion(version)
+	return fmt.Sprintf("✔ Kubernetes %s %s", sv.String(), kubernetesConstraints), nil
```
bigkevmcd, line 33: Minor here, you don't need to use `sv.String()` just using `sv` will get `.String()` called on it by `%s`.

bigkevmcd: I appreciate that it's not in this PR, but the `kubernetesContraints` looks overly broad, I am not sure that Flux supports those versions (and they're not supported upstream)?

bigkevmcd (approved):
I'm ok with this, it works when the previous version doesn't 😄

pkg/services/check/check.go
```diff
@@ -2,71 +2,33 @@ package check
...
-	"os/exec"
-	"strings"
 
 	"github.com/Masterminds/semver/v3"
+	"k8s.io/client-go/discovery"
 )
 
 const (
-	kubernetesConstraints = ">=1.20.6-0"
+	kubernetesConstraints = ">=1.26"
```
bigkevmcd, line 11: 🙇🏻‍♂️

bigkevmcd (approved):
Thanks for fixing the K8s bits!

PR #4166, opened 2023-12-12
Add latest flux version for Flux discovery.

bigkevmcd:
Closes

**What changed?**
This adds the latest source-controller v1.2.2 as a marker for Flux v2.1.2.

Dependencies are updated, and code updated to reflect the updates to dependencies.

This requires upstream changes in tf-controller (because of updates to the Flux dependencies).

**Why was this change made?**
To support the latest version of Flux when the version isn't on the namespace.

**Release notes**

**Documentation Changes**


foot (approved):

core/server/inventory.go
```diff
@@ -446,3 +449,16 @@ func parseInventoryFromUnstructured(obj *unstructured.Unstructured) ([]*unstruct
 
 	return objects, nil
 }
+
+func storageVersionFromHelmRelease(helmRelease *helmv2.HelmRelease) int {
+	storageVersion := helmRelease.Status.LastReleaseRevision
+
+	if storageVersion == 0 && helmRelease.Status.History != nil {
```
foot, line 456: 🎉 

If a user is migrating from flux 2.1.2 -> 2.2.0, will they have any "residual status", e.g. should we flip this check to try for history first and then fallback to the old `LastReleaseRevision` which might still be present?

Or when the 2.2.0 controller starts it will clear `LastReleaseRevision` etc?

foot: https://github.com/fluxcd/helm-controller/blob/main/internal/controller/helmrelease_controller.go#L355

It explicitly clears `LastReleaseRevision`

core/server/inventory.go
```diff
@@ -145,16 +142,15 @@ func (cs *coreServer) getHelmReleaseInventory(ctx context.Context, k8sClient cli
...
-
-	storageVersion := helmRelease.Status.LastReleaseRevision
+	storageVersion := storageVersionFromHelmRelease(helmRelease)
 	if storageVersion < 1 {
 		// skip release if it failed to install
 		return nil, nil
 	}
 
+	storageNamespace := helmRelease.GetStorageNamespace()
+	storageName := helmRelease.GetReleaseName()
```
hiddeco, line 152: It would be better to collect this from the `Current` as well, which contains a `.Name`. Something like:

```go
func storageDataFromHelmRelease(helmRelease *helmv2.HelmRelease) (string, int) {
	if latest := helmRelease.Status.History.Latest(); latest != nil {
		return latest.Name, latest.Version
	}
	return helmRelease.GetReleaseName(), helmRelease.Status.LastReleaseRevision
}
```

hiddeco: Acually, if you do this. It would simply be better to also collect the `.Status.StorageNamespace`. So you would end up with something like `(ns string, name string, version int)`, and:

```
helmrelease.Status.StorageNamespace, latest.Name, latest.Version
```

or

```
helmRelease.GetStorageNamespace(), helmRelease.GetReleaseName(), helmRelease.Status.LastReleaseRevision
```

core/server/inventory.go
```diff
@@ -446,3 +436,26 @@ func parseInventoryFromUnstructured(obj *unstructured.Unstructured) ([]*unstruct
 
 	return objects, nil
 }
+
+func secretNameFromHelmRelease(helmRelease *helmv2.HelmRelease) *client.ObjectKey {
```
hiddeco, line 440: The logic below continues to be problematic, as one of the things `v2beta2` addressed is field mutations which change the target installation of a release.

When someone changes their `.spec.targetNamespace`, `.spec.storageNamespace` or `.spec.releaseName`. The controller will now first uninstall this release before installing the new one. Given this, you _have_ to take the `.status.storageNamespace` and data from `.status.history` into account to get an accurate view of the active release. As it may take some time for the controller to perform a successful uninstall, which means the data taken from `.spec` may not be an accurate representation of the active state, it's _desired_ state.

core/server/inventory.go
```diff
@@ -446,3 +436,26 @@ func parseInventoryFromUnstructured(obj *unstructured.Unstructured) ([]*unstruct
...
+	if helmRelease.Status.History == nil && helmRelease.Status.LastReleaseRevision == 0 {
+		return nil
+	}
+
+	releaseRevision := helmRelease.Status.LastReleaseRevision
+
+	if helmRelease.Status.History != nil {
```
hiddeco, line 447: This check is redundant, `Latest()` itself will check if `len(in) == 0` which will be `true` for `nil`.

core/server/inventory.go
```diff
@@ -258,18 +245,41 @@ func unstructuredToInventoryEntry(clusterName string, objWithChildren ObjectWith
...
 		go func(obj unstructured.Unstructured) {
 			defer wg.Done()
 
+			// Set the namespace of the object if it is not set.
+			if obj.GetNamespace() == "" {
+				// Manifest does not contain the namespace of the release.
+				// Figure out if the object is namespaced if the namespace is not
+				// explicitly set, and configure the namespace accordingly.
+				objGVK := obj.GetObjectKind().GroupVersionKind().String()
+				if _, ok := isNamespacedGVK[objGVK]; !ok {
```
hiddeco, line 269: You will need another mutex for the `map`.

core/server/inventory.go
```diff
@@ -258,18 +245,49 @@ func unstructuredToInventoryEntry(clusterName string, objWithChildren ObjectWith
...
+				// Manifest does not contain the namespace of the release.
+				// Figure out if the object is namespaced if the namespace is not
+				// explicitly set, and configure the namespace accordingly.
+				objGVK := obj.GetObjectKind().GroupVersionKind().String()
+				gvkMu.RLock()
+				namespaced, ok := isNamespacedGVK[objGVK]
+				gvkMu.RUnlock()
+
+				if !ok {
+					namespaced, err := apiutil.IsObjectNamespaced(&obj, k8sClient.Scheme(), k8sClient.RESTMapper())
```
hiddeco, line 275:
```suggestion
					var err error
					namespaced, err = apiutil.IsObjectNamespaced(&obj, k8sClient.Scheme(), k8sClient.RESTMapper())
```

hiddeco (approved):
From a Flux perspective, changes now look all good to me 💯

PR #4181, opened 2023-12-28
Add support for V2Routes.Runtime in navigation getParentNavRouteValue

enekofb:
Fixes the issue of the nav item menu not being highlighted:

Screenshot 2023-12-28 at 11 45 17


opudrovs (approved):
tested it locally and can see the Runtime navigation item highlighted. ✨

PR #4193, opened 2024-02-15
Fix log row which caused panic

marulkan:
Closes

**What changed?**
Log line which cause panic

**Why was this change made?**
Bugfix

Implemented change in cluster where we had a helmchart which triggered this panic (old PSP cr).

**Release notes**

**Documentation Changes**


casibbald:
@marulkan Please can you provide more details on this fix

erikgb:
@casibbald I am pretty sure this PR is fixing a real issue. If you look at the signature of logr Logger function:

````go
func (l Logger) Error(err error, msg string, keysAndValues ...any)
````

The last variadic parameter needs to have a pair number arguments (key + value).

File: core/server/inventory.go [hunk 1]
```diff
@@ -286,7 +286,7 @@ func getObjectsWithChildren(ctx context.Context, defaultNS string, objects []*un
 				if !ok {
 					namespaced, err = apiutil.IsObjectNamespaced(&obj, k8sClient.Scheme(), k8sClient.RESTMapper())
 					if err != nil {
-						logger.Error(err, "failed to determine if %s is namespace scoped", obj.GetObjectKind().GroupVersionKind().Kind)
+						logger.Error(err, "failed to determine if %s is namespace scoped", "entry", obj)
```
erikgb, line 289:
```suggestion
						logger.Error(err, "failed to determine if object is namespace scoped", "kind", obj.GetObjectKind().GroupVersionKind().Kind)
```

erikgb (changes requested):
@marulkan Do you want to fix/rebase/squash this PR, or should I just do it? Please let me know!

PR #4203, opened 2024-05-21
fix: remove rbac Secrets from app rules

DanielAntos90:
Be able to watch flux resources in UI without having rights to read secrets in k8s.


casibbald:
@DanielAntos90 what is the reason for this change? 
looking forward to understanding your objectives with the change

casibbald:
@DanielAntos90 Drop me a PM to connect up on this.

PR #4229, opened 2024-11-23
Bump Flux CRDs to Flux 2.4

erikgb:
It seems like most of the Go tests passes when Flux CRDs are bumped to Flux 2.4. I see that we already have moved into "Flux 2.4 land" in the go.mod file. What if we try to head directly for Flux 2.4? We use ww-gitops with Flux 2.4 in our clusters, so at least for the main parts, it works.

/cc @casibbald


casibbald:
Yeah heading directly to 2.4.0 is totally fine with me.

PR #4283, opened 2024-12-27
fix: upgrade/migrate buf to latest release (1.48.0)

erikgb:
Before starting to bump dependencies, ref. https://github.com/weaveworks/weave-gitops/pull/4282, it seems required to upgrade `buf` - at least on my workstation. I was unable to regenerate protocol buffers after bumping dependencies. Upgrading `buf` requires migrating config files. Some new linters are added, but I just excluded the linters failing for now.


casibbald:
Merging, if we encounter issues we can always revert and figure out what the issue is.

PR #4293, opened 2024-12-28
chore: remove direct dependency to github.com/pkg/errors

erikgb:
The https://github.com/pkg/errors project has been archived for some time in favor of the proposals that landed in the stdlib. We have very few references to this module so I took a stab at migrating it.


casibbald:
@erikgb Is this ready to merge? Are there any other references we need to update?

erikgb:
@casibbald This should be ready to merge. Some of the code changed here seems unused - at least in this project.

PR #4294, opened 2024-12-28
chore: remove dependency to gotest.tools

erikgb:
The `gotest.tools` dependency does not seem to give much over using the standard Go testing package. And when it comes to dependencies "less is more".


erikgb:
@casibbald This should also be ready to merge.

PR #4327, opened 2024-12-30
ci: renovate GHA versions manually

erikgb:
Closes

**What changed?**

Upgrading all GHA to their latest releases in an attempt to make Dependabot workflow great again.
It might have been the Flux action that made it fail, but let's see how it goes.

Some are upgraded to new major versions, so we might need to fix eventual issues in follow-up PRs.

**Why was this change made?**

To enable Dependabot upgrades.

**Release notes**

**Documentation Changes**


casibbald:
LGTM

PR #4368, opened 2025-01-02
fix: enable uploads to ghcr.io, upload binaries to github

casibbald:
Closes

**What changed?**

- [x] Container Images will be published via free ghcr.io  
- [ ] Containers will be signed with cosign
- [ ] Validate signed containers
- [x] Binaries will be uploaded to GitHub
- [x] Binaries will be signed with the new weaveworks.org GPG signature
- [ ] Validate signed binaries
- [ ] Generate and socialise release note draft

**Why was this change made?**

- Publishing via ghcr.io is necessary due to no commercial funding
- Images need to be signed with the new notary@weaveworks.org GPG key as necessary in the switch to a community project

**Release notes**

**Documentation Changes**


.github/workflows/pr.yaml
```diff
@@ -138,21 +138,15 @@ jobs:
...
-#        with:
-#          service_account: ${{ secrets.service_account }}
-#          workload_identity_provider: ${{ secrets.workload_identity_provider }}
-#      - name: Login to gcloud for docker
-#        run: gcloud --quiet auth configure-docker ${{ env.CI_CONTAINER_REGISTRY }}
-#      - name: Push images to gcloud
-#        run: |
-#          docker load --input /tmp/${{ matrix.docker-image }}.tar
-#          docker push "${{ env.CI_CONTAINER_REPOSITORY }}/${{ matrix.docker-image }}:${{ needs.ci-generate-tag.outputs.tag }}"
+      - name: Login to ghcr
```
erikgb, line 141:
```suggestion
      - name: Login to GHCR
```

.github/workflows/pr.yaml
```diff
@@ -138,21 +138,15 @@ jobs:
...
+        run: echo ${{ secrets.WEAVE_GITOPS_BOT_ACCESS_TOKEN }} | docker login ghcr.io -u weaveworks --password-stdin
+      - name: Push images to gcloud
```
erikgb, line 143:
```suggestion
      - name: Push images to GHCR
```

casibbald: @erikgb 
${{ secrets.WEAVE_GITOPS_BOT_ACCESS_TOKEN }}
I think we need to use this token

erikgb: I hope not. I think we should try to be as much key/secret less as possible. 😄

PR #4382, opened 2025-01-03
ci(deps): improve react Dependabot group

erikgb:
Closes

**What changed?**

The current react Dependabot group is too wide, making https://github.com/weaveworks/weave-gitops/pull/4363 worthless.
This PR improves the group definition by looking at the proposed versions in https://github.com/weaveworks/weave-gitops/pull/4363. Over time, I think that will be a better approach. I will try to follow up deps that might be missing from the group.

**Why was this change made?**

Create mergeable Dependabot PRs.

**Release notes**

**Documentation Changes**

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -7,10 +7,14 @@ updates:
       interval: "daily"
     groups:
       react:
-        patterns:
+        patterns: &react-deps
           - react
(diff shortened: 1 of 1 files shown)


casibbald (approved):
LGTM

PR #4392, opened 2025-01-04
ci(deps): yet another attempt to fix Dependabot grouping

erikgb:
Closes

**What changed?**

Yet another attempt to create working Dependabot groups.

**Why was this change made?**

I am pretty sure ChartGPT is wrong about the Dependabot matching on prefix. It seems like it matches exact and that we have to use `*` to get any wildcard functionality.

**Release notes**

**Documentation Changes**

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -9,14 +9,15 @@ updates:
       react:
         patterns:
           - react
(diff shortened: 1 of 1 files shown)


casibbald (approved):
Fingers crossed

PR #4456, opened 2025-01-06
Migrate eslint config

erikgb:
Closes

**What changed?**

**Why was this change made?**

**Release notes**

**Documentation Changes**


eslint.config.mjs
```diff
@@ -0,0 +1,41 @@
...
+                    order: "asc",
+                    caseInsensitive: true,
+                },
+            }
+        ],
+        "@typescript-eslint/explicit-module-boundary-types": 0,
+        "@typescript-eslint/no-explicit-any": 0,
+        "@typescript-eslint/ban-ts-comment": 0,
+        "import/no-named-as-default": 0,
+        "@typescript-eslint/switch-exhaustiveness-check": "error",
```
mproffitt, line 30: I am thinking this is too strict and will lead to some very ugly code as the enterprise code is merged in with the community edition code.

This is caused by a change in behaviour between `eslint < 9.0.0` and `eslint >=9.0.0` which detected more missing cases.

These missing cases should normally be handled by a `default` case which as far as I can see is set in all errors related to `switch-exhaustiveness-check`

I would be in favour of changing this line to become

```json
"@typescript-eslint/switch-exhaustiveness-check": 0,
```

The alternative seems to be to either encourage constructs such as the following, or disable the check on a per statement basis.

```typescript
  switch (paramType) {
    case 'boolean':
      return paramType ? 'True' : 'False';
    case 'object':
    case 'function':
    case 'symbol':
    case 'undefined':
    case 'bigint':
    case 'number':
    case 'string':
    default:
      return param;
  }
```

mproffitt:
Regarding the check errors. All of the `switch` check errors can be dealt with by the comment I left inline on `eslint.config.mjs` 

For the `try {} catch (error) {}` error `ui/lib/objects.ts` - this is a simple fix to convert the catch to simple 

```typescript
try {
...
} catch {
```

There is no need for the error check at this point as it's simply used to set an empty inventory

That would leave 2 errors

```plaintext
/home/mproffitt/src/github.com/weaveworks/weave-gitops/ui/components/AddKustomizationForm.tsx
  24:7  error  'defaultInitialState' is assigned a value but only used as a type  @typescript-eslint/no-unused-vars

/home/mproffitt/src/github.com/weaveworks/weave-gitops/ui/components/ReconciliationGraph.tsx
  17:1  error  Expected an assignment or function call and instead saw an expression  @typescript-eslint/no-unused-expressions
```

Dealing with these two in the order shown then for `AddKustomizationForm.tsx`:

```typescript
const defaultInitialState = () => ({
  name: "",
  namespace: "",
  source: "",
  path: "",
});

type FormState = ReturnType<typeof defaultInitialState>;
```

may be better written as simply

```typescript
type FormState = {
  name: string;
  namespace: string;
  source: string;
  path: string;
};
```

I'm still not sure how to proceed with the `d3` error in `ReconciliationGraph.tsx`. This becomes a linting error as we move across eslint major versions but I can replicate this error on the enterprise codebase now however I cannot tell the implications of simply removing the statement without extensive additional testing.

I would be inclined to add a per-line disable for this check with a `TODO` and reason, something like

```typescript
// TODO: Investigate if this can be removed safely via a component upgrade path
// eslint-disable-next-line @typescript-eslint/no-unused-expressions
d3;
```

erikgb:
Thanks for the solid review with all the explanations/suggestions for a non-frontend developer, @mproffitt! I will address your suggestions tomorrow!

eslint.config.mjs
erikgb: Thanks! Instead of disabling the rule completely, I override one of the rule options, ref. https://typescript-eslint.io/rules/switch-exhaustiveness-check/#options

mproffitt (approved):
Tests currently don't run locally for me but they [do run in CI](https://github.com/weaveworks/weave-gitops/actions/runs/12675478857/job/35326593075) and are successful which is the main thing :green_circle: 

UI still starts and executes in Tilt and a quick visual inspection of the UI shows no apparent changes or breakages have been introduced through fixing lint errors in the code.

image

Overall for a first F/E PR this is fairly solid work.

erikgb:
@casibbald or @tenstad, are one you able to approve this PR, so we can get it merged? Or add @mproffitt as an eligible approver? I can probably do the latter, but it feels wrong for me to adjust such settings as a newcomer to this project.

casibbald (approved):
Fingers crossed

PR #4473, opened 2025-01-09
Revert "ci: init new workflow to build images (#4468)"

casibbald:
Reverts weaveworks/weave-gitops#4468

We Need to resubmit this with potentially a different token


erikgb (approved):
But I don't think there is anything wrong with using the GitHub Action token for pushing packages to GHCR. Any remains from the Enterprise setup? We should investigate this more.

erikgb:
This issue is fixed now, so no need to revert: https://github.com/weaveworks/weave-gitops/actions/workflows/build-image.yaml?query=branch%3Amain. Roll-forward is always better than rolling back. 😉

PR #4489, opened 2025-01-11
ci: correct image sign verify command

erikgb:
Closes

**What changed?**

Now the signing seems to work, but there is an error in the sign verification command which is corrected in this PR.

**Why was this change made?**

Make the `main` branch CI green again.

**Release notes**

**Documentation Changes**

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -127,8 +127,8 @@ jobs:
       - name: Verify the image signing
         if: ${{ github.event_name != 'pull_request' && github.repository == 'weaveworks/weave-gitops' }}
         run: |
-          cosign verify ${{ env.IMAGE_NAME }}@${{ steps.build.outputs.digest }}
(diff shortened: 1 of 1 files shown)


.github/workflows/pr.yaml
```diff
@@ -127,8 +127,8 @@ jobs:
       - name: Verify the image signing
         if: ${{ github.event_name != 'pull_request' && github.repository == 'weaveworks/weave-gitops' }}
         run: |
```
tenstad, line 129: Change to > and remove indents instead, to avoid the backslashes?

erikgb: Fixed

tenstad: Think you have to remove indents as well, that's the downside of >

erikgb: Then I actually prefer the backslash solution.

tenstad (approved):

PR #4565, opened 2025-01-18
Replace yaml library with js-yaml

gusevda:
**What changed?**
In this PR:
- `yaml` library was replaced with `js-yaml` library;
- scope hoisting was enabled for `parcel build`;

**Why was this change made?**

There is a problem with `yaml` package that causes builds for production to fail. As a temporal solution, parcel scope hoisting was disabled in [the PR](https://github.com/weaveworks/weave-gitops/pull/4529). This PR fixes production builds.

Tested locally by running with `yarn start` and separately with `tilt up`.

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -8,6 +8,7 @@ const transformIgnorePatterns = [
   'install',
   'jest-canvas-mock',
   'js-sha3',
+  'js-yaml',
   'lodash',
   'luxon',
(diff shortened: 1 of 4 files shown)


erikgb (commented):
LGTM, but someone with more frontend experience should review this. 😉

mproffitt (approved):
LGTM

PR #4572, opened 2025-01-19
Replace deprecated testing library in unit tests

mproffitt:
**What changed?**

This commit upgrades the testing library from `react-test-renderer` to `@testing-library/react`.

This largely results in changes to `renderer.create().toJSON()` becoming `render().asFragment()`. To ensure minimal snapshot changes were recorded, I used `firstChild` from the fragment so the fragment wrapper was not added to the snapshots.

**Why was this change made?**

This change was made to replace the deprecated `react-test-renderer` with a more modern equivelent which is compatible with future versions of React framework

**Side effects**

During testing, a large number of failure messages such as:

```plaintext
An empty string ("") was passed to the src attribute
```

This message would cause tests to fail and is the result of `svg` images being passed to the fileMock in `ui/lib/fileMock.js`

To resolve this, I have replaced `svg` management with `jest-transformer-svg`  and introduced a helper method to convert imported `svg` images to a usable base64 encoded version for use in CSS styling

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -23,10 +23,12 @@ const transformIgnorePatterns = [
 const config = {
   preset: "ts-jest",
   moduleNameMapper: {
-    "\\.(jpg|ico|jpeg|png|gif|eot|otf|webp|svg|ttf|woff|woff2|mp4|webm|wav|mp3|m4a|aac|oga)$":
+    "\\.(jpg|ico|jpeg|png|gif|eot|otf|webp|ttf|woff|woff2|mp4|webm|wav|mp3|m4a|aac|oga)$":
       "<rootDir>/ui/lib/fileMock.js",
     "\\.(css|less)$": "<rootDir>/ui/lib/fileMock.js",
+    "^.+\\.svg$": "jest-svg-transformer",
   },
+
   transform: {
     "\\.tsx?$": "ts-jest",
     "\\.jsx?$": [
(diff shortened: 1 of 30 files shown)


erikgb:
Seems like the package intentions is to support newer versions of Jest, ref. https://github.com/Em-Ant/jest-transformer-svg/issues/5. Maybe we should file an issue and eventually help with fixing the Jest 29 support?

casibbald:
Ideal if we can.

mproffitt:
I've done some more digging on this and it's not a bug in the `jest-transformer-svg` library, rather it's related to these lines from `DarkModeSwitch.tsx`

https://github.com/weaveworks/weave-gitops/blob/22a1bd7bfaf37c1687a6a48e66588af326112791/ui/components/DarkModeSwitch.tsx#L28-L34

This seems to be the result of this issue in `jest-styled-components` https://github.com/styled-components/jest-styled-components/issues/97 which causes an SVG to be embedded somehow in a way that cannot be parsed correctly.

From what I understand, the correct behaviour to resolve this would be to encode the images as base64 instead of plain XML. This would allow both the browser, and the tests to interpret them properly instead of them only being interpretable inside the browser.

mproffitt:
Introduced a util method to convert from `image/svg+xml` to `image/svg+xml;base64` - this allows tests to pass without interfering with the front end

erikgb:
> Introduced a util method to convert from `image/svg+xml` to `image/svg+xml;base64` - this allows tests to pass without interfering with the front end

Sounds good! Does this mean the PR is reviewable/mergeable? Maybe update the PR description with the latest news? 😉

mproffitt:
It should be reviewable yes. I'd like to wait for @gusevda as he has way more experience with Typescript than I do and may know of a better way to solve the same issue.

mproffitt:
I think if you're happy for it to go, it's a `sed` followed by `yarn` command to drop them and I'd rather start as we mean to go on. Lets set a precedence with choosing the least confusing route for devs :smile_cat:

PR #4582, opened 2025-01-19
ci(lint): fix react/no-unknown-property eslint violations

erikgb:
Closes

**What changed?**

Fix and enable the react/no-unknown-property eslint rule.

**Why was this change made?**

Follow-up after introducing react eslint. The react/no-unknown-property rule is a recommended rule.

**Release notes**

**Documentation Changes**

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -45,7 +45,6 @@ export default tseslint.config([
 
             "react/display-name": 0,
             "react/no-children-prop": 0,
-            "react/no-unknown-property": 0,
             "react/prop-types": 0,
             "react-hooks/exhaustive-deps": 0,
             "react-hooks/rules-of-hooks": 0,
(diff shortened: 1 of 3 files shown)


eslint.config.mjs
```diff
@@ -45,7 +45,6 @@ export default tseslint.config([
 
             "react/display-name": 0,
             "react/no-children-prop": 0,
-            "react/no-unknown-property": 0,
```
mproffitt, line 48: did you mean to still remove this? Looks like you fixed the issues it related to so maybe then leave it in place?

erikgb: Yes, my intent was to remove it - since it currently turns the rule off. It is turned on by the recommended react eslint rules.

mproffitt (approved):

PR #4583, opened 2025-01-19
ci(release): remove publish to https Helm repository

erikgb:
Closes

**What changed?**

This removes the publishing of the Helm chart to the https Helm repository, leaving OCI as the only option for installing the Helm chart. Also updated the Helm install example in the docs.

**Why was this change made?**

We no longer have access to the `helm.gitops.weave.works` domain, and OCI is now the preferred mechanism for publishing Helm charts.

**Release notes**

**Documentation Changes**

### DIFF
--- a/.github/workflows/helm.yaml
+++ b/.github/workflows/helm.yaml
@@ -13,10 +13,6 @@ on:
       - 'charts/**'
   workflow_dispatch:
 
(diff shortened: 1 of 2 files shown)


casibbald (approved):
LGTM

PR #4686, opened 2025-02-04
ci: fix goreleaser arch template

casibbald:
Closes

**What changed?**

**Why was this change made?**

**Release notes**

**Documentation Changes**

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -15,7 +15,6 @@ archives:
       gitops-
       {{- title .Os }}-
       {{- if eq .Arch "amd64" }}x86_64
-      {{- else if eq .Arch "arm64" }}arm64
       {{- else }}{{ .Arch }}{{ end }}
 builds:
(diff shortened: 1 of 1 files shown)


.goreleaser.yml
```diff
@@ -14,7 +14,7 @@ archives:
       gitops-
       {{- title .Os }}-
       {{- if eq .Arch "amd64" }}x86_64
-      {{- if eq .Arch "arm64" }}arm64
+      {{- else if eq .Arch "arm64" }}arm64
```
erikgb, line 17:
```suggestion
```

It should be picked up by the else.

erikgb (changes requested):
I think this bug was introduced unintentionally in https://github.com/weaveworks/weave-gitops/pull/4675/files.

erikgb (approved):
The additional if is not needed IMO - as it will be pick up by the else below.

PR #4689, opened 2025-02-05
chore: fix version tag in UI

### LINKED ISSUE #4688: Footer should support pre-releases
I installed our [latest release candidate](https://github.com/weaveworks/weave-gitops/releases/tag/0.39.0-rc.1) today, and while most of the features we use seem to work, I noticed that the footer is broken. And I think it's because the UI doesn't support pre-release versions. It looks like this:

Image

I am not a frontend developer, but I think the problem might be somewhere around

https://github.com/weaveworks/weave-gitops/blob/2c7103534b94a67c4e6dc597ab8be2df7c9ac1b4/ui/lib/utils.ts#L207

casibbald:
Closes #4688 

**What changed?**
Semver display code

**Why was this change made?**
Version does not show correctly in RC's

image

Screenshot 2025-02-06 at 00 18 24

**Release notes**

**Documentation Changes**

### DIFF
--- a/ui/lib/utils.ts
+++ b/ui/lib/utils.ts
@@ -204,13 +204,14 @@ export function getAppVersion(
 ): AppVersion {
   const shouldDisplayApiVersion =
     !isLoading &&
-    (versionData?.semver || "").replace(/^v+/, "") !== defaultVersion &&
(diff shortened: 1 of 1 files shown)


erikgb (approved):
LGTM

PR #4749, opened 2025-02-21
ci: Make upgrade Flux CI great again

erikgb:
Closes

**What changed?**

This PR is a rework of the automation on upstream Flux releases. Instead of trying to mess with the Go dependencies, I suggest to update the CRDs used in testing instead. So I simplified this Makefile target and included it in the upgrade automation.

**Why was this change made?**

This proposed change was triggered by https://github.com/weaveworks/weave-gitops/pull/4740, which was raised automatically after the release of Flux 2.5.0. It is currently failing because of our conventional commits check, but what it tries to do makes little sense to me. We have no dependencies on the main Flux Go module in this project. We could introduce it, but it would probably be a "faked" dependency. I rather use Dependabot to keep up with the Flux Go modules.

**Release notes**

**Documentation Changes**


casibbald (changes requested):
LGTM, can we squash it down to one commit as per your previous requests.

erikgb:
> LGTM, can we squash it down to one commit as per your previous requests.

It will be squashed on merge - using the PR title as commit message.

casibbald (approved):

PR #4768, opened 2025-02-23
ci: Disable upgrade-flux workflow in forks

erikgb:
Closes

**What changed?**

Disable the upgrade-flux workflow in forks. It should only run in the main repo, as it will fail in "any" fork.

**Why was this change made?**

I was getting failing runs in my fork after Flux 2.5 was released.

**Release notes**

**Documentation Changes**

### DIFF
--- a/.github/workflows/upgrade-flux.yaml
+++ b/.github/workflows/upgrade-flux.yaml
@@ -11,6 +11,7 @@ permissions:
 jobs:
   has-new-flux:
     runs-on: ubuntu-latest
+    if: "${{ github.repository_owner == 'weaveworks' && github.ref_name == 'main'}}"
     outputs:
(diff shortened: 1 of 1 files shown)


.github/workflows/upgrade-flux.yaml
```diff
@@ -11,6 +11,7 @@ permissions:
 jobs:
   has-new-flux:
     runs-on: ubuntu-latest
+    if: "${{ github.repository_owner == 'weaveworks' && github.ref_name == 'main'}}"
```
tenstad, line 14: just minor fmt

```suggestion
    if: "${{ github.repository_owner == 'weaveworks' && github.ref_name == 'main' }}"
```

casibbald (approved):

PR #5312, opened 2025-12-18
fix(ui): replace deprecated innerRef prop with React.forwardRef

### LINKED ISSUE #5311: UI crashes after login due to deprecated innerRef prop in React 19
**Describe the bug**
After logging in, the UI crashes with the following React error:

Image

This occurs because the codebase uses the deprecated `innerRef` prop (from older styled-components/react-router patterns) in the Link, SubRouterTabs and Nav components

With the current dependencies (React 19, styled-components 6, react-router 7), the innerRef prop should be replaced with the standard ref prop using `React.forwardRef`

**Environment**

- Weave-Gitops Version: [main-160cb9aa0](https://github.com/weaveworks/weave-gitops/commit/160cb9aa0)
- Flux Version v2.5.1
- Kubernetes version v1.29.2

**To Reproduce**

* Start the UI server
* Complete the login flow
* See the React error crash screen a bit after login

**Expected behavior**

After login, the user should be redirected to the dashboard without any errors

umegbewe:
The innerRef prop is deprecated in styled-components v5+ and react-router v6+. This causes ui to crash a bit after login

Closes #5311

**What changed?**
- Refactored Link component to use React.forwardRef
- Updated SubRouterTabs and Nav to pass ref instead of innerRef


casibbald (approved):

casibbald:
@umegbewe thank you for the contribution

